PR #93, opened 2014-12-08
Make Rscript work for linux binary package, and use it to obtain the build variables

tkluck:
I *think* this will actually also make it possible to build against any other installed version of R, but I didn't test it so there's probably a bit of work to do for that. I just wanted to send an as-small-as-possible pull request with only "a view towards" linking against system R. Hopefully, you can review and merge this easily.


jonathon-love:
done!

PR #162, opened 2015-01-10
ANOVA (added name dependent variable in title )

raviselker:
For both the ANOVA and ANCOVA table

### DIFF
--- a/JASP-Engine/JASP/R/ancova.R
+++ b/JASP-Engine/JASP/R/ancova.R
@@ -334,12 +334,27 @@ Ancova <- function(dataset=NULL, options, perform="run", callback=function(...) 
 	
 	if (is.null(options$covariates)) {
 	
-	    anova[["title"]] <- "ANOVA"
-	
+	    if (options$dependent != "") {
+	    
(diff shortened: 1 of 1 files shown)


jonathon-love:
how would you feel about something a bit more minimalist?

rather than

`ANOVA (Dependent Variable: *Intelligence*)`

Perhaps just:

`ANOVA - Intelligence`

raviselker:
I think it should be clear that it refers to the dependent variable, but I could make it a little more mimimalist:

ANOVA (Dependent: *Intelligence*)

jonathon-love:
For the person doing the analysis, the very minimalist will be clear that it refers to the dependent (because they dragged it into the dependent box after all).
If you were working with some colleagues on a particular dataset, it would be clear if you sent them the very minimalist table that that was the dependent.
I think it would only be unclear to someone unfamiliar with the data, and perhaps unfamiliar with the design - but in this case, a person would provide a figure caption anyway.
My thinking with putting it in the title, is for if you open up some old results. you may have run two similar analyses with similar dependents - this lets you tell which is which at a glance, without having to open the options dialog.
I just think the really minimalist looks really good; whereas the less minimalist is a bit busy.
Granted, we're on shaky, not-so-object ground.

what do you think?

raviselker:
I agree

PR #186, opened 2015-01-17
ANCOVA (added header and fixed issue)

raviselker:


jonathon-love:
just keep an eye on the spaces/tabs things - a few spaces are creeping in here.

PR #187, opened 2015-01-19
Fix to bug in Bayesian Ancova

### LINKED ISSUE #101: Bayesian Ancova: Bug in error handling when interaction effects are specified without main effects.
There is still a problem with the error handling of interaction effects when there are no other variables in the Bayesian Ancova R-code:

screen shot 2014-12-17 at 16 22 39

mmarsman1:
This is to fix #101

### DIFF
--- a/JASP-Engine/JASP/R/ancovabayesian.R
+++ b/JASP-Engine/JASP/R/ancovabayesian.R
@@ -249,7 +249,8 @@ AncovaBayesian	 <- function(dataset=NULL, options, perform="run", callback=funct
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


jonathon-love:
Hi, looks good, but a number of your if statements contravene the [style guide](https://google-styleguide.googlecode.com/svn/trunk/Rguide.xml)

> Always begin the body of a block on a new line.
> 
> BAD:
> `if (is.null(ylim)) ylim <- c(0, 0.06)`
> `if (is.null(ylim)) {ylim <- c(0, 0.06)}`

PR #251, opened 2015-02-05
Update jasp-error-list.md

AlexanderLyNL:


jonathon-love:
how about:

1. Correlation co-efficient is undefined - not enough observations
2. Correlation co-efficient is undefined - one (or more) variables contain infinity

this will then be the same as the classical correlation.
and then dropping this one because we handle missing values automatically:

* Sample correlation co-efficient is undefined - one (or more) variables contain an NA, or do not vary

cheers

AlexanderLyNL:
Change the error messages of the classical correlation analysis. It makes sense to call it the sample correlation, or the r-statistic. 

Furthermore, it didn’t handle missing values automatically the last time I ran it.

jonathon-love:
hey @EJWagenmakers 

in the classical correlation, when a correlation can not be carried out we have an error message like:

`Correlation co-efficient is undefined - not enough observations`

@AlexanderLyNL would like this changed to

`Sample correlation co-efficient is undefined ... `

i prefer the former, but did you have a preference?

AlexanderLyNL:
There is always a difference between things we see (data and statistics) and parameters that are not observed. In the simple models we use, we make use of the observables to infer about the unobservables. One way to make this distinguish is by referring to the observed ones, as:

- sample mean 
- sample effect size 
- sample correlation co-efficient
- sample variance

and the unobserved ones, as:

- population mean  
- population effect size  
- population correlation co-efficient 
- population variance

Another way to do this is by referring to the observed ones as statistics. 

t-statistic is undefined -…. 

Though, you’re actually interested in the effect size \delta and the t-statistic is considered as a remnant of the population effect size. Similarly, you could  say 

r-statistic is undefined -….

However, as we don’t call it the r-statistic, we should call it Sample correlation co-efficient, hence, there’s something “wrong" with the data (model misspecification is the other alternative). 

I don’t care what spss does, but the difference between observed and unobserved quantities should always be clear. Furthermore, this separation helps the user to distinguish the frequentist from the Bayesian result. Frequentist statistics is always concerned about the observed quantities, the statistics, while Bayesian statistics is always concerned with what the observed quantities tell us about the unobserved quantities.

AlexanderLyNL:
In addition, I think that we should report the posterior in the Bayesian correlation matrix with a reference to \rho rather than the posterior mode that coincides with the observed r

EJWagenmakers:
I agree with Alexander. I don't want to use the term "population" (I prefer
latent or unobserved, or just nothing), but I agree that adding "sample" is
a good idea in this case.
EJ

jonathon-love:
right-o, so if i understand correctly, we can keep the classical one the way it is? or should i change that one too?

if you drop out the third error message @AlexanderLyNL i can merge this.

AlexanderLyNL:
I changed the third error message, which is useful when people try to calculate the correlation of a variable consisting of one number of all the time. Can you activate “remove NA” automatically? More specifically, I changed this into the following:

* Sample correlation co-efficient r is undefined - not enough observations
* Sample correlation co-efficient r is undefined - one (or more) variables contain infinity
* Sample correlation co-efficient r is undefined - one (or more) variables do not vary

You should definitely change this for the classical analysis as well, as that is only about the sample correlation co-efficient and not about the population/theoretical/latent/unobserved/“true” correlation co-efficient

jonathon-love:
> Can you activate “remove NA” automatically?

`.readDataSetToEnd()` has an option `exclude.na.listwise` which can be used to remove missing values... well.. listwise, but there's nothing available for doing it "analysis by analysis" (i can't think of a way to provide this that would make it less complicated than just doing it in the analysis). "analysis by analysis" is the default. i'm pretty sure this is implemented in the classical correlation if you want to copy/paste.

PR #272, opened 2015-02-11
Table heading

### LINKED ISSUE #246: Bayesian ANOVA and friends have no table heading
you should also add in an analysis heading; take a look at the classical ANOVA, or t-tests to see how this is done.

screen shot 2015-02-04 at 9 06 43 pm

mmarsman1:
This fixes #246

Does mean that in RM the heading is “dependent” to make things as
obscure as possible (another reason why we should split the functions).

### DIFF
--- a/JASP-Engine/JASP/R/ancovabayesian.R
+++ b/JASP-Engine/JASP/R/ancovabayesian.R
@@ -65,8 +65,7 @@ AncovaBayesian	 <- function(dataset=NULL, options, perform="run", callback=funct
 	posterior <- list()
 	posterior[["title"]] <- "Bayesian ANOVA: Model Comparison"
 	posterior <- list(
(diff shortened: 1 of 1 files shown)


jonathon-love:
in the RM ANOVA i would just leave the dependent variable out of the table title.

mmarsman1:
I can fix this at the RM code, I just realised, will do that.

On 11 Feb 2015, at 13:05, Jonathon Love <notifications@github.com<mailto:notifications@github.com>> wrote:


in the RM ANOVA i would just leave the dependent variable out of the table title.

jonathon-love:
how would you feel about simplifying the B ANOVA table headings? (and adding in an analysis heading?)

what about:

`Bayesian ANOVA - dependent`
`Analysis of Effects - dependent`

currently they're a bit busy.

PR #289, opened 2015-02-13
Development

quentingronau:


JASP-Engine/JASP/R/ttestbayesianpairedsamples.R
```diff
@@ -196,7 +196,13 @@ TTestBayesianPairedSamples <- function(dataset=NULL, options, perform="run", cal
 					r <- BayesFactor::ttestBF(c1, c2, paired = TRUE, r=options$priorWidth, nullInterval= nullInterval)
 					
 					bf.raw <- exp(as.numeric(r@bayesFactor$bf))[1]
-
+					
+					if (is.na(bf.raw)) {
+				
+						unplotable <- TRUE
+						unplotableMessage <- "Bayes factor is NaN"
```
jonathon-love: i'll merge this, but isn't a great error message. it's good if we can provide something which hints at why the error occurred.

Also, NaN, doesn't mean anything to the average user, which is why we put footnotes explaining it; so we can't put NaN in the footnotes :)

PR #369, opened 2015-02-23
Suggestions to error list

mmarsman1:
These are a couple of suggestions to make the list of errors more alike across the different analyses.

Please feel free to discuss @jonathon-love @AlexanderLyNL @EJWagenmakers @raviselker

### DIFF
--- a/Docs/development/jasp-error-list.md
+++ b/Docs/development/jasp-error-list.md
@@ -30,7 +30,7 @@ value errors for *t*
 
 - t-statistic is undefined -- the dependent variable contains infinity
 - t-statistic is undefined -- the grouping variable contains less than two levels once missing values in the dependent are excluded
-- t-statistic is undefined -- one or both levels of the dependent contains all the same value (the variance is zero)
+- t-statistic is undefined -- one or both levels of the dependent contains all the same value (zero variance)
 - t-statistic is undefined -- one or both levels of the dependent contain too few observations
 - *Note.* All tests, variances of groups assumed equal
 - *Note.* All tests, variances of groups not assumed equal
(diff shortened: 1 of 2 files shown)


EJWagenmakers:
All looks good except for one issue concerning "less" vs "fewer"
E.J.

********************************************
Eric-Jan Wagenmakers
Department of Psychological Methods, room 2.09
University of Amsterdam
Weesperplein 4
1018 XA Amsterdam
The Netherlands

Web: ejwagenmakers.com
Book: bayesmodels.com
Stats: jasp-stats.org
Email: EJ.Wagenmakers@gmail.com
Phone: (+31) 20 525 6420

“Man follows only phantoms.”
Pierre-Simon Laplace, last words
********************************************

jonathon-love:
hey @ravi,
@mmarsman1 has proposed some burdening changes to the ANOVA.

in other news, i've implemented these for the t-tests

mmarsman1:
I implemented these for the Bayesian Linear models.

@AlexanderLyNL I think this makes the reported errors more in line with very similar error messages in other analyses. For instance, "not enough observations" whereas other messages speak of "too few observations". We could also change the other messages to yours of course :)

@EJWagenmakers will change "less than" into "fewer than"

@jonathon-love we do need to keep this list in line with the actual errors :) Shall I check that once the dust settles on these burdening changes, but prior to the 0.6?

AlexanderLyNL:
Sample correlation coefficient r is undefined - one or more of the variables contain all the same value (zero variance)

of the???? contain all the same value???

mmarsman1:
What are you trying to say?

mmarsman1:
Would we want to change "Sample correlation coefficient r" to "Pearson correlation coefficient" as we may introduce other correlations in the future? 
Also as small *r* is used in the error messages, yet a capital *R* appears in the Table.

mmarsman1:
@jonathon-love could you merge these so that I can rebase?

PR #432, opened 2015-03-05
B t tests: implemented Richard's code one-sided BFs 

quentingronau:


jonathon-love:
there's some peculiar spaces in this.

PR #612, opened 2015-04-17
File loading error checking + bugs

dropmann:
- changed the way that an analysis status is saved. This will cause analyses saved prior to this update to load without any of the analyses that do not comply. (i.e. all of them)
- increased the robustness of loading jasp files with broader error
checking. 
- fixed analysis id management bug and the resulting null references.


dropmann:
i forgot to add the closes tag till after you merged. Is there a way i can close it?

jonathon-love:
which issue are you wanting to close?

Issue #606, opened 2015-04-16
Correlation matrix fixes


EJWagenmakers:
1. Pearson's r (not R)
2. sample r, not posterior median (my bad)

PR #615, opened 2015-04-18
Issue #606

AlexanderLyNL:


jonathon-love:
Looks good. One point of style, try not to call functions `myFunction`, `myX`, etc. but no biggie.

PR #617, opened 2015-04-18
C correlation: fixes #556

### LINKED ISSUE #556: C Correlation: Confidence intervals need to be implemented
The classical correlation does not present a button to compute CIs. The plot function does show them, and that's great, but it would be better to have the button as well (as we do for the other analyses).
E.J.

quentingronau:


JASP-Engine/JASP/R/correlation.R
```diff
@@ -370,22 +370,22 @@ Correlation <- function(dataset=NULL, options, perform="run", callback=function(
...
-					width <- 400
-					height <- 400
-					
-				} else {
-				
-					width <- 250 * l
-					height <- 250 * l
-					
-				}
+				 if (l <= 2 && (options$plotDensities || options$plotStatistics)) {
```
jonathon-love: hey, some spaces have snuck in here.

PR #618, opened 2015-04-18
variables to dot notation, though, I think it's fucking ugly.

AlexanderLyNL:


jonathon-love:
you can keep the camel-case if you prefer. it was just the `my...` thing

AlexanderLyNL:
It denotes the locality of the variables, hence, I don’t care about it that much, or it’s based on user input, though, it should be taken out of function names. Furthermore, referring to someMedian is better than calling a variable median, due to the fact that it overrides the median function.

PR #636, opened 2015-04-23
Saving error handling

dropmann:
- Added error handling for saving files


dropmann:
I fixed the issue with opening empty jasp files. It has been committed and just needs pulling across.

PR #649, opened 2015-04-28
Minor alterations

### LINKED ISSUE #637: File save: should take the user back to the common tab
i think the behaviour in office is to take the user back to the main tab after they've saved a file in the file tab.

### LINKED ISSUE #648: JASP Version reported as 0.07 rather than 0.7
We've been referring to our versions as:

0.4
0.5
0.6
etc.

and 0.07 seems a lot smaller. can we make this 0.7 ?

dropmann:
- dropped leading zero from the minor number of version strings. Closes
#648
- After saving the common tab is returned to. Closes #637

### DIFF
--- a/JASP-Common/version.cpp
+++ b/JASP-Common/version.cpp
@@ -25,13 +25,13 @@ Version::Version(std::string versionString)
 	char buildString[8];
 	unsigned char buildIndex = 0;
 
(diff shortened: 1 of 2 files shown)


jonathon-love:
the sscanf's, are they buffer overflows waiting to happen?

might be better of using the sstream

PR #657, opened 2015-05-06
Finished various issues

### LINKED ISSUE #650: Network access manager should disable caching of images from analyses
Sometimes the JASP engine reuses filenames for images. This can be a problem when the web results panel has cached the images, and uses the cache rather than getting the fresh version from disk.

the Network access manager should set the no-cache header in the responses, so the UI always loads the latest version.

### LINKED ISSUE #656: JASP file format: should labels be stored separately inside JASP files?
hey @dropmann 

thinking through our current strategy with storing the nominal labels in the dataset meta information, this may be problematic.

for example, a data set might be a database dump, with fields like *first name*, *last name*. every one of these would be unique, and the meta information would be massive.

this in itself wouldn't be so bad, but most JSON parsers load the whole JSON representation into memory; which means parsing a lot of data, when a previewer may only want access to the column names and a description.

so perhaps we should separate these out.

dropmann:
- Attempt to disable cache of the web browser. Closes #650
- Stored labels separately to other meta data. Closes #656
- Changed the default file name for saving from a csv base to the name
of the csv file with a JASP extension.

### DIFF
--- a/JASP-Desktop/lrnamreply.cpp
+++ b/JASP-Desktop/lrnamreply.cpp
@@ -28,6 +28,7 @@ LRNAMReply::LRNAMReply(const QString &path, QObject *parent)
 		{
 			setHeader(QNetworkRequest::ContentTypeHeader, "image/png");
 		}
+		setRawHeader(QString("Cache-Control").toLatin1(), QString("no-cache").toLatin1());
 	}
 
 	_file.setFileName(path);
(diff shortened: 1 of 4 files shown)


JASP-Desktop/lrnamreply.cpp
```diff
@@ -28,6 +28,7 @@ LRNAMReply::LRNAMReply(const QString &path, QObject *parent)
 		{
 			setHeader(QNetworkRequest::ContentTypeHeader, "image/png");
 		}
+		setRawHeader(QString("Cache-Control").toLatin1(), QString("no-cache").toLatin1());
```
jonathon-love: tell me about this toLatin1(), it shouldn't be toUtf8()?

jonathon-love:
hey, this all looks good. for future reference, it's good if you break fixes into separate commits. i.e. if there's a regression, it's easier to track down if it's one issue or feature per commit.

PR #766, opened 2015-07-06
Merging Backbone stuff

jonathon-love:


JASP-Desktop/asyncloader.cpp
```diff
@@ -72,6 +72,18 @@ void AsyncLoader::saveTask(const QString &filename, DataSetPackage *package)
...
+		int sleepTime = 100;
+		int delay = 0;
+		while (package->analysesHTML.empty())
+		{
+			if (delay > maxSleepTime)
+				break;
+
+			sleep(sleepTime);
+			delay += sleepTime;
+		}
```
jonathon-love: is this a race condition? what if what you're waiting for takes more than 2 seconds?

PR #773, opened 2015-07-07
Temp development

jonathon-love:


JASP-Desktop/html/index.html
```diff
@@ -27,7 +27,7 @@
 	<div id="templates" ></div>
 	<div id="intro">
 		<h1>Welcome to JASP!</h1>
-		<p style="text-align: right ;">Version 0.7 Beta 4</p>
+		<p style="text-align: right ;">Version <span class="app_version">Unknown</span></p>
```
jonathon-love: i'm pretty sure we don't use underscores in class names, or do we?
and perhaps this could be unpopulated/blank until setAppVersion() is called.

i'm invisaging people taking and using our result renderer in other projects in the future, and it would be a bit easier if they could ignore setAppVersion() if it wasn't important to them.

JASP-Desktop/html/js/jaspwidgets.js
```diff
@@ -82,22 +82,23 @@ JASPWidgets.Encodings = {
...
 		copy: 0,
-		save: 1,
+		save: 1
 	},
 
 	imageFormat: {
 		temporary: 0,
 		resource: 1,
-		embedded: 2,
+		embedded: 2
```
jonathon-love: you could use string literals here, or some "consts". as it is i'm not sure what 0, 1, or 2 correspond to. just a suggestion.

JASP-Desktop/mainwindow.h
```diff
@@ -104,6 +104,7 @@ private slots:
 	void showAnalysesMenuHandler(QString options);
 	void copySelected();
 	void citeSelected();
+	void menuHidding();
```
jonathon-love: spelling?

PR #825, opened 2015-07-21
Fixed JASP file version checking

### LINKED ISSUE #816: JASP shouldn't reject backward compatible files
the typical file format version convention is to have version of the form: X.Y.Z

increments to Z indicate backward compatible changes (software can just use file as is; an example of this would be the transition from JASP versions 0.7.0 to 0.7.1; only Z need be incremented)

increments to Y indicate forward compatible changes (software can just use the file as is, but might like to warn the user about a loss of functionality, i.e. when we add annotations; pre-annotation versions should just display the analyses, and leave the annotations out)

increments to X indicate incompatible changes (the software should refuse to open files with higher X values)

(we have an additional massive, which can just be zero)

JASP should follow these conventions (i.e. 0.7.0 shouldn't have rejected files from 0.7.1)

dropmann:
- Added warning message when using older version or pre-release version.
- closes #816

### DIFF
--- a/JASP-Common/datasetpackage.cpp
+++ b/JASP-Common/datasetpackage.cpp
@@ -13,6 +13,7 @@ void DataSetPackage::reset()
 	analysesHTML = std::string();
 	analysesData = Json::arrayValue;
 	hasAnalyses = false;
+	warningMessage = std::string();
 	_isLoaded = false;
 	setModified(false);
 }
--- a/JASP-Common/importers/jaspimporter.cpp
+++ b/JASP-Common/importers/jaspimporter.cpp
@@ -28,8 +28,14 @@ void JASPImporter::loadDataSet(DataSetPackage *packageData, const string &path, 
 
 	readManifest(packageData, path);
 
-	if ( ! isCompatible(packageData))
+	Compatibility compatibility = isCompatible(packageData);
+	if (compatibility == JASPImporter::NotCompatible)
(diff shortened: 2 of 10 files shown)


JASP-Common/datasetpackage.cpp
```diff
@@ -13,6 +13,7 @@ void DataSetPackage::reset()
 	analysesHTML = std::string();
 	analysesData = Json::arrayValue;
 	hasAnalyses = false;
+	warningMessage = std::string();
```
jonathon-love: just FYI, this isn't necessary; strings are always initialised as empty and don't need to be explicitly constructed.

JASP-Common/importers/jaspimporter.cpp
```diff
@@ -28,8 +28,14 @@ void JASPImporter::loadDataSet(DataSetPackage *packageData, const string &path,
 
 	readManifest(packageData, path);
 
-	if ( ! isCompatible(packageData))
+	Compatibility compatibility = isCompatible(packageData);
+	if (compatibility == JASPImporter::NotCompatible)
 		throw runtime_error("The file version is too new.\nPlease update to the latest version of JASP to view this file.");
+	else if (compatibility == JASPImporter::Limited)
+		packageData->warningMessage = "WARNING: This file was created by a newer version of JASP and may not have complete functionality.";
```
jonathon-love: you could probably tone this down a bit; i.e. WARNING in capitals is a bit full on.

PR #834, opened 2015-07-26
Correlation Matrix plots: Improved error messages

quentingronau:
Incorporated Alexander's suggestion to display "Correlation coefficient
undefined".


jonathon-love:
Cool. Are we able to provide an explanation for why it is undefined?

quentingronau:
Yes, the error messages should now be two lines long (for the scatterplots, the posterior plots and the statistics plots, but not for the density plots, since they do not display a correlation). The first line is "Correlation coefficient undefined:", the second line mentions the reason, which is at the moment either "Variable has zero variance" or "Variable contains infinity".

PR #888, opened 2015-08-23
C ANOVA's: removed colons UI

raviselker:


jonathon-love:
*chuckles*, such punctual service! :)

PR #993, opened 2015-09-19
refactored ttests and added non-parametric alternatives

dostodabsi:


JASP-Engine/JASP/R/ttestindependentsamples.R
```diff
@@ -1,698 +1,695 @@
...
+        list(name = "lev", type = "string", title = ""),
+        list(name = "W", title = "W", type = "number", format = "sf:4;dp:3"),
+        list(name = "p", title = "p", type = "number", format = "dp:3;p:.001")
+    )
+    
+    normalityTests[["schema"]] <- list(fields = fields)
+    
+    footnotes <- .newFootnotes()
+    .addFootnote(footnotes, symbol = "<em>Note.</em>",
+                 text = "Significant results indicate a deviation from normality")
```
jonathon-love: We prefer milder language for significance; we go for *suggest* rather than *indicate*
and i think language about *violating an assumption* is more consistent with what we've done so far (not completely sure).

`Significant results suggest a violation of the assumption of normality`

JASP-Engine/JASP/R/ttestindependentsamples.R
```diff
@@ -1,698 +1,695 @@
...
+    
+    ## for a independent t-test, we need to check both group vectors for normality
+    normalityTests.results <- list()
```
jonathon-love: try to be consistent in your naming convention; either use camel case or dot notation for variables, but certainly don't mix them :P

JASP-Engine/JASP/R/ttestindependentsamples.R
```diff
@@ -1,698 +1,695 @@
...
+                  
+                } else if (length(data) > 5000) {
+                    err <- "Too many datapoints (N > 5000) to compute statistic reliably"
+                    foot.index <- .addFootnote(footnotes, err)
+                    row.footnotes <- list(W = list(foot.index), p = list(foot.index))
+                    error <- TRUE
+                }
+                
+                ## if the user did everything correctly :)
+                if (!error) {
```
jonathon-love: good practice to put spaces around the exclamation mark

    if ( ! error) {

it is easy to miss the exclamation mark (depending on the editor, syntax highlighting, etc. this often isn't bad, but we like to accommodate the general case).

some people go further and do the:

    if (error == FALSE) {

to be extra explicit (but not me)

JASP-Engine/JASP/R/common_ttest.R
```diff
@@ -0,0 +1,59 @@
+#' The plan of this file is to successively abstract functionality
+#' from the individual t-tests into a common interface to reduce clutter
+#'
+#' returns the result object and the dataset
+.initialize_ttest <- function(dataset, options, perform, type = 'not-paired') {
```
jonathon-love: underscores are verboten!

JASP-Engine/JASP/R/common_ttest.R
```diff
@@ -0,0 +1,59 @@
...
+    
+    groups <- options$groupingVariable
+    depvars <- unlist(options$variables)
```
jonathon-love: should be camelCase or dot.separated

JASP-Engine/JASP/R/ttestindependentsamples.R
```diff
@@ -1,698 +1,695 @@
...
+
+                        df <- as.numeric(r$parameter)
+                        m <- as.numeric(r$estimate[1]) - as.numeric(r$estimate[2])
+                    }
+
+                    ## if the user doesn't want a Welch's t-test,
+                    ## give a footnote indicating if the equality of variance
+                    ## assumption is met; seems like in this setting there is no
+                    ## sampling plan, thus the p-value is not defined. haha!
+                    if (!wants_welchs && wants_students) {
```
jonathon-love: i would always place this warning beside the student's statistic if equality of variances is violated (perhaps not, if they're explicitly testing the assumption).

JASP-Engine/JASP/R/ttestindependentsamples.R
```diff
@@ -1,698 +1,695 @@
...
+            if (perform == "run" && length(variables) > 0 && !is.null(levels)) {
+
+                ## get the dependent variable at a certain factor level
+                data <- na.omit(dataset[[.v(variable)]][dataset[[.v(factor)]] == level])
+                
+                row.footnotes <- NULL
+                error <- FALSE
+                
+                if (length(data) < 3) {
+                    err <- "Too few datapoints (N < 3) to compute statistic reliably"
```
jonathon-love: maybe `too few samples` would be better than `too few datapoints`

dostodabsi:
Note that the comment for normality assumption (and Levene's) was already there -- I didn't change it. Also, the (!error) is your doing :P

I guess I won't change your mind in [favor of underscores](http://adv-r.had.co.nz/Style.html)? I wanted to make a scientific point that they are easier to read, but I misremembered the respective paper [paper](http://www.cs.loyola.edu/~binkley/papers/icpc09-clouds.pdf).

Anyways, I will make the appropriate changes (I guess there will be more), and make another pull request?

jonathon-love:
hey, that's a great study!

we also use tabs rather than spaces, and we don't use underscores in filenames either.

but these are just tiny things; it looks really good!

you don't need to do another pull request, whatever you commit to that branch will automatically get added to this pull request.

cheers

jonathon-love:
> you don't need to do another pull request, whatever you commit to that branch will automatically get added to this pull request.

but you'll probably need to ping me, because i won't get emails about your subsequent commits.

dostodabsi:
alright, sounds great!

dostodabsi:
I've removed the underscores, changed from spaces to tabs, and in general made things consistent. I've also changed the "indicate" to "suggest". I don't think we should say "too few samples"; "too few datapoints" seems much clearer.

A few things still:
The footnotes do not have -- and never have had -- a full stop at the sentence ending. Is this APA? It seems weird to me. Also, while we notify the user when the assumption of equality of variance is violated in the case of an independent t-test, we do not do so in case of normality. This seems inconsistent. In general, I don't think we should conduct tests without the user requesting them. In any case, Levene's will be underpowered most of the time; should we do [Welch's T-Test as a default](http://daniellakens.blogspot.nl/2015/01/always-use-welchs-t-test-instead-of.html) instead?

jonathon-love:
> I don't think we should say "too few samples"; "too few datapoints" seems much clearer.

my only objection to *datapoint* is that it's not really a word :P what about "too few observations"?

> The footnotes do not have -- and never have had -- a full stop at the sentence ending. Is this APA? It seems weird to me.

i've always thought of footnotes as being like bullet points, but i'm happy for us to adopt the full stop if that's APA.

> Also, while we notify the user when the assumption of equality of variance is violated in the case of an independent t-test, we do not do so in case of normality. This seems inconsistent.

yes! i didn't realise that. we should make it consistent. these tests have been written and contributed to by multiple people, and inconsistencies like this tend to creep in. good catch.

> In general, I don't think we should conduct tests without the user requesting them.

i agree we shouldn't display a table with the shapiro and homogeneity tests unless the user explicit asks for them, but putting a footnote in the case of violations is a good idea. it's small, unobtrusive, and i think will greatly reduce the number of people [accidentally/forgetfully/unknowingly] overlooking these things.

> In any case, Levene's will be underpowered most of the time; should we do Welch's T-Test as a default instead?

it's a good idea, but i think it's at odds with people's expectations. if i ask for a t-test, i expect to get a student's t-test. and if i compare the result with SPSS, i'll assume JASP is broken. the JASP philosophy is to encourage people to best practice (i.e. with giving them helpful footnotes, etc.) but not being heavy handed about it. giving them a welch's instead isn't *that* heavy handed, but i try not to make decisions for people.

it's a fine line, but i think footnotes falls on one side, and giving them a different test statistic falls on the other.

cheers

dostodabsi:
given what we discussed, I think this is alright now :)

PR #1306, opened 2016-05-05
Bugfix/tab change

akashrajkn:
Fixes the tab-switch bug

Changes made:
1) MainWindow::tabChanged() function - gets  the current active tab from TabBar::getCurrentActiveTab() function and switches the tab accordingly


akashrajkn:
This was tested on Linux ( Ubuntu 15.04 )

PR #1311, opened 2016-05-11
changed x-axis range according to prior width

### LINKED ISSUE #1310: Bayes factor robustness plot - user prior point plotted outside the graph
The code for creating Bayes factor robustness plot uses the x-axis (Cauchy prior width) range from 0 to 1.5. Suppose the user sets the "prior width" in the range 1.5 to 2, the Bayes factor point corresponding to the user prior is plotted outside the graph [See the picture attached].
bfrobustness

akashrajkn:
Fixes #1310 
By default, the range of 0 to 1.5 is used. 
If the user chooses a prior width more than 1.5, then the graph is plotted from 0 to 2


boutinb:
@quentingronau Can you check this fix?

PR #1314, opened 2016-05-12
Update regressionlinear.R

borisma:
Hi! There is problem with the R squared change option in Linear Regression. 
The overall problem is that the R squared change option does not make really sense as long as one cannot add blocks of parameters (hierarchical regression). I am really waiting for this feature because it is so basic.
Thus, for the time being the R squared change option just repeats R squared of the model (r.squared.old <- r.squared, line 673), and the F.change statistic should repeat the model F-Test. It doesn't, however, in case of more than one predictor, because the df1 always  equals 1. This is because abs(length(lm.model[[ m ]]$variables) - length(lm.model[[ m-1 ]]$variables)) always equals 1. As soon as you have the "real" F-change function you need: df1 <- abs(length(lm.model[[ m ]]$variables) - length(lm.model[[ m-(NUMBER OF ADDED PARAMETERS)  ]]$variables)).
But as long as you can't report a real F-Change-Statistic, you need to set df1 = m (number of predictors). Otherwise you get the correct model F-value * m in the F-Change outcome.
Thanks!
Boris

### DIFF
--- a/JASP-Engine/JASP/R/regressionlinear.R
+++ b/JASP-Engine/JASP/R/regressionlinear.R
@@ -656,7 +656,7 @@ RegressionLinear <- function(dataset=NULL, options, perform="run", callback=func
 					df1 <- 1
 					if (m > 1){
 					
-						df1 <- abs(length(lm.model[[ m ]]$variables) - length(lm.model[[ m-1 ]]$variables) )
+						df1 <- abs(length(lm.model[[ m ]]$variables))
 					}
 					
 					df2 <- length( dataset[[ dependent.base64 ]]) - length(lm.model[[ m ]]$variables) - 1


boutinb:
@quentingronau Can you check this fix?

quentingronau:
Hi Boris,

Thank you very much for pointing out this issue.
You are absolutely right, in the case that you described (i.e., multiple predictors, method = "Enter"), the
F-change and the df are incorrect. However, your commit does not solve the issue. The reason is that "m" corresponds to the number of models and not the number of predictors. As you correctly mentionted, since, at the moment, it is not possible to conduct a hierarchical regression, there is only one model that is tested when selecting the method "Enter". This means that, no matter how many predictor variables are used, m will always be one and not larger than one (when the method is "Enter"). There is one case when the R squared change option is already useful which is when you select a stepwise method, e.g., "Forward". In this case m can be larger than one and the F-change values are correctly displayed in the current version of JASP:

jasp_forward_method

For comparison, here are the SPSS results:

spss_forward_method

I made a commit which fixes the issue with the R squared change option for the method = "Enter" case (#1332).
Again, thank you very much for reporting the issue!

Cheers,
Quentin

borisma:
Thank you! Do you already know when hierarchical regression will be implemented? Without it it is difficult to use JASP for teaching. JASP is so great and I really wanna use it... 
Best, Boris

boutinb:
This is fixed will pull request #1332

PR #1355, opened 2016-06-25
call of overloaded function ambiguous-fixed

akashrajkn:
currently build fails with the error - " call of overloaded 'isnan(double&)' is ambiguous ". Explicitly adding the namespace std, solves it

### DIFF
--- a/JASP-Desktop/importers/spss/spssrecinter.cpp
+++ b/JASP-Desktop/importers/spss/spssrecinter.cpp
@@ -216,7 +216,7 @@ const
 string SPSSColumn::format(double value)
 {
     QString result;
-    if (!isnan(value))
+    if (!std::isnan(value))
     {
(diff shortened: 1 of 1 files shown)


williamstome:
When I clone the repo on Ubuntu 14.04, I get this error as well. I checked what branch I was on, and by default I was on development. If I switch to master I no longer have this issue (but the build fails for other reasons)

akashrajkn:
@williamstome thanks for reporting it. Some of the recent code added on development uses isnan() function without specifying the namespace explicitly. I will fix it.

Can you please give more information just for reference? - gcc version your system is using (and/or version of libstdc++ library installed in your linux machine). Cheers and Thank you.

williamstome:
Sure thing: gcc (Ubuntu 4.8.4-2ubuntu1~14.04.3) 4.8.4

Issue #1457, opened 2016-09-14
SumStats correlation bug


EJWagenmakers:
@akashrajkn 

When I run correlation pairs and switch to BF_01, the dot remains in the old (wrong) spot. See screenshot!
I haven't checked the other analyses but that would be worth doing.

![image](https://cloud.githubusercontent.com/assets/8603015/18515453/b3684880-7a95-11e6-9d6a-f455b7689c4c.png)

akashrajkn:
Fixed this in pull request #1462

PR #1462, opened 2016-09-18
solved bug-BF01 plot point shown correctly

akashrajkn:
1) solves issue #1457 
2) added comments to functions


boutinb:
@akashrajkn Your modified file has a different code layout, which makes really difficult to see your changes. Could you set it back to the previous code layout? Or did you do that on purpose?

akashrajkn:
@boutinb I changed tabs into spaces - which caused this. I will make changes in the original layout and make a different pull request.

Issue #1470, opened 2016-09-21
[Debian] compiled tests fail (reason unknown)


mih:
Log snippet:

```
make[2]: Entering directory '/home/mih/debian/jasp/git/JASP-Tests'
( test -e Makefile.JASP-Desktop-staticlib || /usr/lib/x86_64-linux-gnu/qt5/bin/qmake /home/mih/debian/jasp/git/JASP-Tests/JASP-Desktop-staticlib.pro -nocache 'QMAKE_CFLAGS_RELEASE=-g -O2 -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2' 'QMAKE_CFLAGS_DEBUG=-g -O2 -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2' 'QMAKE_CXXFLAGS_RELEASE=-g -O2 -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2' 'QMAKE_CXXFLAGS_DEBUG=-g -O2 -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2' QMAKE_LFLAGS_RELEASE=-Wl,-z,relro QMAKE_LFLAGS_DEBUG=-Wl,-z,relro QMAKE_STRIP=: PREFIX=/usr -o Makefile.JASP-Desktop-staticlib ) && make -f Makefile.JASP-Desktop-staticlib 
make[3]: Entering directory '/home/mih/debian/jasp/git/JASP-Tests'
mkdir -p /home/mih/debian/jasp/git/JASP-Tests/../Resources/ ; cp -R /home/mih/debian/jasp/git/JASP-Desktop/../Resources/* /home/mih/debian/jasp/git/JASP-Tests/../Resources/ ;
cp: '/home/mih/debian/jasp/git/JASP-Desktop/../Resources/Data Sets' and '/home/mih/debian/jasp/git/JASP-Tests/../Resources/Data Sets' are the same file
cp: '/home/mih/debian/jasp/git/JASP-Desktop/../Resources/Help' and '/home/mih/debian/jasp/git/JASP-Tests/../Resources/Help' are the same file
[...]

PR #1471, opened 2016-09-21
WiP: Debian packaging

mih:
This is not asking for a merge. The purpose of this PR is to document the state of the effort, allow for comments and track related issues.

Related issues:
- [x] #1470
- [ ] #1472
- [ ] #1473
- [ ] #1476
- [x] #1477
- [x] #1478
- [ ] #1480
- [ ] #1481
- [ ] #1511
- [ ] #1689 
- [ ] #1690
- [ ] #1691


debian/control
```diff
@@ -21,8 +21,6 @@ Package: jasp
 Architecture: any
 Depends: ${shlibs:Depends},
          ${misc:Depends},
-         libqt5webkit5,
-         libqt5svg5,
```
akashrajkn: libqt5webkit5 and libqt5svg5 are required, am I right?

mih: Yes, but they are discovered automatically.

mih:
FTR: Binary packages for Debian sid and stretch, as well as Ubuntu xenial are now available from NeuroDebian.

Backports to earlier releases require backports of rinside, which in turn require backports of R 3.3. Doable but requires some effort.

boutinb:
Hi @mih: thanks really much for your work!
If I well understood your work, you have made a package under NeuroDebian of JASP Beta 0.8.2. And Debian & Ubuntu can use this package to install JASP.
2 questions:
Is there a way to update this package when we have a new release?
Should we merge this pull request now?

mih:
@boutinb Yes that is right.

You could ping me for an update when you make a release. I'll build binary packages and update the NeuroDebian repo.However, there is no secret sauce. We could have a (virtual) sit-down to demonstrate how you could update and build packages on your own.

You could consider merging this into a separate branch, e.g. 'debian'. Merging it into master will cause more trouble than it help in the long run. Often the distributions will require things that make less sense or are needlessly complicated to achieve in the general case -- when it is straightforward for a particular environment.

mih:
@boutinb One more thing:

The folder `debian/patches` contains a few patches that I would very much appreciate to appear in master, though!

boutinb:
@mih Cannot you use a define NEURODEBIAN or something like that, and use it directly in the code? In this way, the patches files are not necessary anymore.
We can make also a folder debian/neurodebian and debian/ubuntu, so that we can keep both versions in the same repository. Or do you think this is a bad idea?
To have a (virtual) sit-down would be really much appreciated! Have you any plan to come to Amsterdam? Eric-Jan Wagenmaker asked you once, but we don't know if you answered him. If not, we can arrange a Skype session.
We would like really much to have a sustainable port of JASP on the linux platforms!

boutinb:
@mih I have just discussed with Eric-Jan, you are more than welcome in Amsterdam if you would like. We are ready to pay you back travel and hotel costs. If not, would you prefer a Skype session?

mih:
@boutinb Sorry for the delay, missed the github notification. I would love to come, but the semester just started and experience tell that a little later will result in less penalty for leisure activities ;-)

boutinb:
@mih When you have time, you would more than welcome!

boutinb:
@mih What should we do with this request? Can I close it?

mih:
Pushed binary packages of 0.8.1.0 (amd64) for Debian stretch/sid and Ubuntu 16.04 to NeuroDebian.

a-0-p:
Hi @mih, if you have time could you spend a couple of words on how to properly install jasp on Debian 9? I see that it's on the neurodebian repositories, however when trying to install it I have multiple missing or broken dependencies...for example r-cran-afex which is needed to run jasp is not available on Debian repositories. If you could guide less experienced users with the installation, that'll be nice.

Thanks!

a-0-p:
A quick update: it turned out that once I have installed the package r-cran-afex from the unstable repositories, there are no more problems in running jasp on Debian. Generally, I would avoid installing packages from unstable into stable; however, r-cran-afex is very unlikely to give any problems.

So that's what worked for me...but in case anyone has any suggestion or any other workaround to install jasp on Debian 9, I think it's worth sharing them here.

Thanks!

boutinb:
@mih I think we can close this pull request now. We use FlatPack now, and it works on Debian (and on most Linux versions).

Issue #1444, opened 2016-09-03
Values of the credible intervals descriptive plots


tressoldi:
Is it possible to obtain the values of the credible intervals descriptive plots for example in the Bayesian paired t-test?

Issue #1359, opened 2016-06-28
Credible intervals in descriptives


EJWagenmakers:
There was a forum question about the x% credible intervals shown in the t-test "descriptives" plot. The interval bounds are not provided anywhere in their numerical form. I suggest we add them to the output of the descriptives table (Note: if the user sets the plot to produce a z% interval, the table should also show the bounds for that z% interval).
E.J.

PR #1494, opened 2016-09-29
Added credible intervals descriptives

TimKDJ:
Issues #1359 #1444


akashrajkn:
@TimKDJ it is difficult to view the changes because of different formatting. Probably the editor you are using has 'Remove trailing Whitespace' enabled by default.

TimKDJ:
@akashrajkn you were right about the trailing white spaces; I've changed the settings in my editor. It should be more legible now

PR #1548, opened 2016-11-11
Calculate BayesFactor when Kendall Tau is selected as the correlation coefficient (Summary Stats)

akashrajkn:
Added BF calculation when Kendall Tau is used as the correlation coefficient


akashrajkn:
@JohnnyDoorn can you please review this?

PR #1551, opened 2016-11-13
Runtime requirements incomplete

rphlypo:
Documentation of build process does not mention that `r-cran-rjson` is a runtime requirement, now added to the list of runtime requirement for Linux.
Without rjson an error caused the JASP interface to close and crash.


rphlypo:
Cleaning with a rebase against development and will resubmit

Issue #1507, opened 2016-10-07
Fedora 24: Jasp immediately crashes


silberzwiebel:
After finally [managing to compile Jasp under Fedora 24](https://github.com/jasp-stats/jasp-desktop/issues/1430), I started it by running `./jasp`
In the terminal I get these messages:

```
QObject::connect: No such signal FSBModel::hideAuthentication()
QObject::connect: No such signal FSBModel::hideAuthentication()
QObject::connect:  (receiver name: 'browser')
QObject::connect: No such signal FSBMExamples::hideAuthentication()
QLayout: Attempting to add QLayout "" to BackStageWidget "backStage", which already has a layout
QLayout: Attempting to add QLayout "" to VariablesWidget "variablesPage", which already has a layout
```

And as graphical output I see the main Jasp window for a short time, followed by an error dialog that has a non-conclusive error message:

```
JASP has experienced an unexpected internal error.

""

If you could report your experiences to the JASP team that would be appreciated.

JASP cannot continue and will now close.
```

Could you help me to run Jasp?

akashrajkn:
@silberzwiebel, need more details to find what is causing the problem. What version of Qt are you using? [on linux it should be 5.4]

akashrajkn:
Do you have the required R packages at /usr/lib64/R ?

silberzwiebel:
I'm installing them at the moment (and face different R related issues). Apparently, I was to quick to run the program. 

It seems that I'm using Qt 5.6.1.

akashrajkn:
@silberzwiebel , Qt 5.6 does not support webkit module any more. You will have to separately compile this module if you want to use 5.6 - this is a bit of a hassle. It's more convenient to use Qt 5.4

silberzwiebel:
Hmm, okay. I've installed the required R packages but Jasp still won't start.
In the fedora repos there is a QtWebkit package, though: 
```
qt5-qtwebkit.x86_64                     5.6.1-3.b889f46git.fc24
qt5-qtwebkit-devel.x86_64           5.6.1-3.b889f46git.fc24
```
I do have installed both pacakges. There are no Qt 5.4 packages available in the fedora repos. Could you give me a hint for compiling qtwebkit on my own?

akashrajkn:
@silberzwiebel, look at [this link](https://trac.webkit.org/wiki/BuildingQtOnLinux), they tell how to compile webkit

[...]

PR #1554, opened 2016-11-16
show application-crash message when rjson package is not installed

akashrajkn:
JASP currently does not show an error message when it crashes (when rjson package is not installed). This is an **attempt** to solve this. 

(see #1507)

### DIFF
--- a/JASP-Engine/JASP/R/common.R
+++ b/JASP-Engine/JASP/R/common.R
@@ -137,7 +137,13 @@ run <- function(name, options.as.json.string, perform="run") {
 
 checkPackages <- function() {
 
-	rjson::toJSON(.checkPackages())
+	p <- try(silent = FALSE, expr = {
+		return(rjson::toJSON(.checkPackages()))
+	})
+
+	if (class(p) == "try-error") {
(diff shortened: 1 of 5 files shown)


akashrajkn:
@boutinb this fix is kind of a hack, is there a better way to fix this? 

I resorted to explicitly checking the standardError message because, the _fatalError variable (in MainWindow.cpp) is not set when QProcess:ProcessError signal is emitted.

JASP-Engine/JASP/R/common.R
```diff
@@ -109,7 +109,13 @@ run <- function(name, options.as.json.string, perform="run") {
 
 checkPackages <- function() {
 
-	rjson::toJSON(.checkPackages())
```
akashrajkn: This causes the app to crash - exception: when package rjson is not installed

boutinb:
@akashrajkn Hello Akash. Let's wait for this fix. We are ging to release a 8.0.1 version first. We have to think of a beter solution, but I don't know yet how.

akashrajkn:
This pull request is obsolete.

PR #1587, opened 2017-01-17
Feature errorcheck

TimKDJ:
Major changes with this PR:
- created the distinction between exceptions (unexpected, needs fixing) and errors (expected, user needs to make changes to data/options)
- added a stack trace to exceptions of at most the last 10 calls
- improved the error check for use with grouping vars
- applied error checking to the independent samples t-test


boutinb:
Nog Niet merge!!!!

PR #1594, opened 2017-01-23
Modules - BAS

akashrajkn:
Only the basic functionality available.
- output table contains top 5 models, and their Bayes factor values
- a graph showing the log posterior odds for the computed models


akashrajkn:
I'll resolve the merge conflicts and create another pull request.

PR #1595, opened 2017-01-23
Modules/BAS - TESTING

akashrajkn:
Only basic functionality available
- show best 5 models and their Bayes factor values in the table
- graph - log posterior odds for the computed models
- graph - posterior distribution of the coefficient (intercept only)


akashrajkn:
This PR is obsolete.

PR #1601, opened 2017-01-27
Activate Vovk-Sellke Maximum p-Ratio functionality

vankesteren:


JASP-Desktop/analysisforms/binomialtestform.cpp
```diff
@@ -39,9 +39,9 @@ BinomialTestForm::BinomialTestForm(QWidget *parent) :
 	ui->testValue->setLabel("Test value");
```
akashrajkn: @vankesteren can you remove the #ifdef ... #endif lines also

PR #1603, opened 2017-01-31
Descriptives Bayesian ANOVA

TimKDJ:
Initial implementation of descriptive tables and plots for the Bayesian ANOVA.
Also added status$error (inline with other analyses) and fixed the spacing to match the Google style guide.


TimKDJ:
I have finalized adding the descriptives table and descriptives plots to the Bayesian ANOVA/ANCOVA/RM ANOVA.

PR #1649, opened 2017-02-21
Separated the module files from JASP-Desktop pri file

akashrajkn:
The idea is to keep the modules independent, so that addition of a new module does not involve modification of JASP-Desktop files.


akashrajkn:
If this is fine, I will add a commit on this with all common analysis as a module

akashrajkn:
This PR is obsolete.
(Needs discussion before going ahead with it)

Issue #1546, opened 2016-11-10
The first t-test takes pretty long


EJWagenmakers:
When I compute a Bayesian t-test, the first analysis takes pretty long. After that's done, the robustness check and sensitivity analyses are pretty quick. @quentingronau, any clue as to why this happens? 
E.J.

quentingronau:
Hi @EJWagenmakers ,

I am not completely sure. It could be the case that this is due to the state system which saves results and prevents them from being recomputed unless this is necessary, but this depends on what exactly changes from one analysis to the next. If you are sure that, e.g., the second analysis cannot be retrieved from the state object and is still taking less time than the first, this might be a question for the programmers.

Cheers,
Quentin

PR #1659, opened 2017-03-01
Solution for the slow Bayesian init

TimKDJ:
This solves the problem raised in issue #1546.
It is a bit hacky.

### DIFF
--- a/JASP-Engine/JASP/R/packagecheck.R
+++ b/JASP-Engine/JASP/R/packagecheck.R
@@ -121,6 +121,8 @@
 	
 	}
 	
+	.initPackages(installed.package.names)
+	
 	if (length(messages) == 0) {
 	
 		list(official=TRUE)
@@ -132,4 +134,12 @@
 }
 
 
+.initPackages <- function(installed) {
+	
+	# First call to BayesFactor initializes its functions and speeds up subsequent calls.
+	if ('BayesFactor' %in% installed) {
+		try(BayesFactor::BFInfo(print=FALSE), silent=TRUE)
+	}
+	
+}
 


akashrajkn:
@TimKDJ I used microbenchmark package to compare the existing code with your changes, but for the few datasets (of different sizes), I _could not_ find much difference in the execution times - first run is always much slower than the subsequent runs.

To understand the variability, I considered only lower quartile, median and upper quartile values in the microbenchmark results. I could be wrong in interpreting the results, could you please carry out the same test?

**Details**
Analysis - TTestBayesianIndpendentSamples 
Code tested - [line 940](https://github.com/jasp-stats/jasp-desktop/blob/development/JASP-Engine/JASP/R/ttestbayesianindependentsamples.R#L940)

`microbenchmark::microbenchmark(BayesFactor::ttestBF(data=subDataSet, formula=f, r=r.size)[1])`

TimKDJ:
Hi @akashrajkn 

In your microbenchmark call you don't specify the times argument.
If this is not set to 1, the impact of namespace loading is hard to discriminate, because it averages the time of the first call -- which needs to load the namespace -- with the 99 times it does not need to.

## R
You can compare the differences in R:
```
library(microbenchmark)

### run microbenchmark 100 times
unloadNamespace('BayesFactor')
print(paste('Before running, is namespace loaded:', isNamespaceLoaded('BayesFactor')))
time <- microbenchmark(BayesFactor::ttestBF(data=sleep, formula=extra ~ group)[1])
print(paste('After running, is namespace loaded:', isNamespaceLoaded('BayesFactor')))
print(time)

### run microbenchmark twice separately
unloadNamespace('BayesFactor')
print(paste('Before first run, is namespace loaded:', isNamespaceLoaded('BayesFactor')))
time <- microbenchmark(BayesFactor::ttestBF(data=sleep, formula=extra ~ group)[1], times=1)
print(time)
print(paste('Before second run, is namespace loaded:', isNamespaceLoaded('BayesFactor')))
time <- microbenchmark(BayesFactor::ttestBF(data=sleep, formula=extra ~ group)[1], times=1)
print(time)
```

## JASP
If I perform a timing experiment in JASP itself I get the following result:

**Without fix**
"[1]" 
" "After first func call. Perform is set to: init Time needed: 3.138179063797"

"[1]" 
" "After first func call. Perform is set to: run Time needed: 0.00399994850158691"

**With fix**
"[1]" 
" "After first func call. Perform is set to: init Time needed: 0.0110001564025879"

[1] "After first func call. Perform is set to: run Time needed: 0"

**Details**
Analysis - Bayesian ANOVA - [line 306](https://github.com/jasp-stats/jasp-desktop/blob/development/JASP-Engine/JASP/R/commonbayesianlinearmodels.R#L306)

```
begin <- Sys.time()
model.list <- try(BayesFactor::enumerateGeneralModels (model.formula, 
		whichModels = "withmain", neverExclude = paste ("^", neverExclude, "$", sep = "")), 
		silent = TRUE)
end <- Sys.time()
print(paste('After first func call. Perform is set to:', perform, 'Time needed:', end-begin))
```
## Note
It would be much nicer to use:
```
if ('BayesFactor' %in% installed && base::isNamespaceLoaded('BayesFactor') == FALSE) {
	try(loadNamespaces('bayesFactor'), silent=TRUE)
}
```
But the R version we use in JASP does not support this.

akashrajkn:
Ah, I hadn't considered 'times' parameter.

Issue #1677, opened 2017-03-14
Error with only within-subjects variables in an ANOVA


lcoome:
Hi,

I noticed another error in JASP - When you have only within-subjects variables in an ANOVA, the calculation of the between-subjects SS is wrong.
It should be 27.333 in the attached 3X3 example from one of the tests we gave to our students (attached).

Best,

Lindsay

[2WithinTest2.xlsx](https://github.com/jasp-stats/jasp-desktop/files/841996/2WithinTest2.xlsx)

JohnnyDoorn:
Hi Lindsay,
Thanks again for taking the effort to point out a bug. I downloaded your dataset and converted it to csv (see attached file, as .txt since github does not support .csv for some reason). When I load it into JASP, specify your 3x3 model, and specify the first column with the subject numbers to indicate individual cases, I do get 27.333 in the between-subjects table. 
When you ran this, did you specify the participant column? If not, then maybe it would be more logical for JASP to automatically take each row as an individual case, rather then the user having to specify it themselves. Also, are you using the latest JASP version? Because in the latest version, there is no between subjects table shown if there are no between subjects variables specified. Furthermore, it might have something to do with which type of SS you specify - I just tested this with a build of JASP that shows the between subjects table regardless of whether such variables are specified, and I do get the value 27.333 when I specify the SS as type 1. In that case I did not need to specify the first column as a between subjects variable. However, I could be overlooking something, so I look forward to your reply! 
Kind regards,
[2WithinTest2.txt](https://github.com/jasp-stats/jasp-desktop/files/844672/2WithinTest2.txt)

Johnny

PR #1697, opened 2017-03-22
Added Tim Error Handling

JohnnyDoorn:
Also Fixed issue #1677 with ANOVA table


boutinb:
@TimKDJ Kun je de code reviewen?

JASP-Engine/JASP/R/correlation.R
```diff
@@ -698,6 +691,13 @@ Correlation <- function(dataset=NULL, options, perform="run", callback=function(
 					 } else {
 
 						if (perform == "run") {
+						  errors <- .hasErrors(dataset, perform = "run", message = 'short', type = c('variance', 'infinity'),
```
TimKDJ: I think it's better to generally stick to `perform = perform` rather than a hard-coded value. If the if-statement above it is ever changed this might lead to unexpected behavior (same for other files).

JASP-Engine/JASP/R/correlationbayesianpairs.R
```diff
@@ -483,41 +483,21 @@ CorrelationBayesianPairs <- function(dataset=NULL, options, perform="run", callb
 					}
 					
 					errorMessage <- NULL
-					
+					errors <- .hasErrors(dataset, perform = "run", message = 'short', type = c('observations','variance', 'infinity'),
```
TimKDJ: The statement right above checks for 1 or -1 correlation, this could possibly be added as a built-in check of .hasErrors(). 
Is this sort of check used in other correlation code?

PR #1710, opened 2017-04-01
Add Tim error handling in EFA and PCA

hummyhummy:
But when I use argument type('infinity','variance')  in function .hasErrors() without indicating target it can not detect variables whose variance is 0 or contain infinity cases. What should I do for this issue?


TimKDJ:
@hummyhummy please come see me sometime so we can look at it together.

PR #1735, opened 2017-04-19
(1) credible intervals Bayesian correlation matrix and Bayesian corre…

AlexanderLyNL:
…lation pairs, (2) p-values for sumstat Kendall's tau

R changes:
- give sensible p-values and BF for Kendall’s tau when n <=2
- Only partially cleaned up correlation.R
- added isTRUE in code for robustness reason
- Removed additional layer in the bfObject when it comes to ci. Check
for ci goes via bfObject$ciValue, say, 0.95, and the intervals are in
bfObject$ci$twoSided, bfObject$ci$plusSided, bfObject$ci$minSided.
- Example: bfObject$ciValue <- 0.95 and bfObject$ci$twoSided <- c(-0.3,
-0.2, 0.1). Tells me that the 95% CI is from -0.3 to 0.1. The point
-0.2 is the median.

Style:
- reordered the input in kendall tau functions, first data, then
parameters then hyperparameters: Data: n, tauObs. Parameters: tauPop,
Hyperparameters: kappa, var=1
- .bfKendallTau is a wrapper around .bfCorrieKernelKendallTau.
Basically, it switches between various methods and various variances
- Removed double function definitions in .correlationbayesianpairs.R
- renamed functions: (1) .credibleInterval to
.computePearsonCredibleInterval

UI son and ui, cpp:
- Bayesian Correlation matrix and pairs now have button for credible
intervals <- slot is called “credibleInterval” which is true or false,
percentage is now referred to as “ciValue”


TimKDJ:
Hi @AlexanderLyNL 

Some things I've noticed:
- There is a .Rhistory file included in your commit
- I noticed a number of ){ in if-statements (just ctrl+f)
- You removed the spacing around the log( BF10 ) option, but that would make it inconsistent with other analyses

On a personal note:
The mark up comments might be a little excessive...
For example:
```
+	# MarkUp: General: Make container for all the results
  	results <- list()
  	
 +	# MarkUp: General: Make container for the structure 
  	meta <- list()
```
These lists are created in every analysis and people know what they are.
The comments don't add very much in my view.

JASP-Engine/JASP/R/.Rhistory
```diff
@@ -0,0 +1,512 @@
+.hJeffreysApprox(n=someNRep, r=someRRep, someDomain)*.hJeffreysApprox(n=someNOri, r=someROri, someDomain)*
```
akashrajkn: @AlexanderLyNL can you please remove this file from the pull request

AlexanderLyNL:
Thanks for the comments:
- Removed .RHistory
- I removed spacing log( BF10 ) option in every Bayesian analysis form

The mark up comments are indeed a bit excessive, but it gives me an idea of how to make a generic createTable function.

PR #1751, opened 2017-04-26
Reapply earlier commit and add R v3.3 functionality

TimKDJ:
This reapplies PR #1527 and incorporates features from the newest R version.

### DIFF
--- a/JASP-Engine/JASP/R/packagecheck.R
+++ b/JASP-Engine/JASP/R/packagecheck.R
@@ -350,11 +350,14 @@
 }
 
 
-.initPackages <- function(installed) {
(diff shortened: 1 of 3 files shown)


JASP-Engine/JASP/R/packagecheck.R
```diff
@@ -350,11 +350,14 @@
 }
 
 
-.initPackages <- function(installed) {
+.initPackages <- function(installed.packages) {
```
vankesteren: The variable name might be a bit inconvenient: it's kind of clashy with utils::installed.packages. Perhaps a better name would be `installedPackages`?

Issue #1761, opened 2017-05-07
Typo in reliability analyses


EJWagenmakers:
@vandenman (Don, I assume this is you)
Hi guys,

Under Reliability analysis we have the "McDonalds' \omega". However, the name of the progenitor is McDonald, without an s at the end. So it should be "McDonald's \omega" -- the apostrophe should immediately follow the "d". Can we fix this before we have our new release please? NOTE: it should also be fixed in the output.

Cheers,
E.J.

vandenman:
Yes that's me! Good catch, will fix this tonight.

PR #1762, opened 2017-05-07
Reliability updates

vandenman:
Added the following things:

-Missing value handling (delete listwise or pairwise).
-Greatest lower bound (another reliability statistic).
-Fixed typo in McDonald's omega (issue #1761).
-Fixed bug for <=2 nonmissing observations.

Some details:

.cpp file adjusted so the missing values handling under advanced options is hidden by default.
.ui & .json files adjusted to include glb, missing values and typo.
.R file adjusted to fix bug, typo, glb and missings.

The glb implementation in psych can be rather error prone and can crash with seemingly valid data, hence it is wrapped in a try-catch statement (a footnote will appear if it crashes). Also, the advanced options box is a simple copy paste with minor changes from a different file. It does what it's supposed to do but please let me know if it is (not) the intended way to do this :)


JASP-Engine/JASP/R/reliabilityanalysis.R
```diff
@@ -73,7 +74,7 @@ ReliabilityAnalysis <- function(dataset = NULL, options, perform = "run",
 
 		# check for errors
 		anyErrors <- .hasErrors(dataset = dataset, perform = perform,
-								type = c("infinity", "variance"))
+								type = c("infinity", "variance", "observations"), observations.amount = " <= 2")
```
TimKDJ: It's more intuitive (and consistent with other analyses) to use observations.amount = "< 3"

JASP-Engine/JASP/R/reliabilityanalysis.R
```diff
@@ -73,7 +74,7 @@ ReliabilityAnalysis <- function(dataset = NULL, options, perform = "run",
...
 
 		doUpdate <- base::identical(anyErrors, FALSE)
```
TimKDJ: .hasErrors has the argument exitAnalysisIfErrors; when this is set to TRUE it exits the analysis immediately. Consequently you would not need this sort of if-statement. 
See also: [this doc](https://github.com/jasp-stats/jasp-desktop/blob/development/Docs/development/r-analyses-guide.md#calling-the-error-check-function)

(I don't think you need the doUpdate <- TRUE in the else-statement below this either, as I cannot find another mention of doUpdate elsewhere in your code)

JASP-Engine/JASP/R/reliabilityanalysis.R
```diff
@@ -136,16 +145,27 @@ ReliabilityAnalysis <- function(dataset = NULL, options, perform = "run",
...
+			
+		} else { # try since the glb is rather error prone...
+			
+			r[["glb"]] <- try(psych::glb(r = d, key = key)[["glb.max"]], silent = TRUE)
+			
+		}
 
 		# calculate McDonalds omega
-		omega <- psych::omega(d, 1, flip = FALSE)[["omega.tot"]]
+		omega <- psych::omega(d, 1, flip = FALSE, plot = FALSE)[["omega.tot"]]
```
TimKDJ: Just a suggestion: it might be good to always wrap calls to external packages in a try. But you know the behavior of these packages and the intended behaviour when an error occurs in your output table better than me.

JASP-Engine/JASP/R/reliabilityanalysis.R
```diff
@@ -194,20 +217,37 @@ ReliabilityAnalysis <- function(dataset = NULL, options, perform = "run",
...
+		
+		if (options[["glbScale"]]) {
+			
+			if (length(variables) <= 2) {
+				
+				message <- paste(message, "Warning: Greatest lower bound can only be calculated for three or more variables.")
+				
+			} else if (isTryError(r[["glb"]])) {
+				
+				message <- paste(message, "Warning: Greatest lower bound could not be calculated.")
```
TimKDJ: Maybe it's also useful to include (a part of) the message the Psych package returns to the trycatch, to be as informative as possible.

JASP-Engine/JASP/R/reliabilityanalysis.R
```diff
@@ -124,8 +125,16 @@ ReliabilityAnalysis <- function(dataset = NULL, options, perform = "run",
 
 	if (perform == "run" && !is.null(variables) && length(variables) > 1) {
 
-		d <- as.matrix(dataset[complete.cases(dataset), ])
-
+		d <- as.matrix(dataset)
```
AlexanderLyNL: d is not very descriptive. Perhaps change this into dataAsMatrix or something

JASP-Engine/JASP/R/reliabilityanalysis.R
```diff
@@ -136,16 +145,27 @@ ReliabilityAnalysis <- function(dataset = NULL, options, perform = "run",
 
 		# calculate chronbach alpha and gutmanns lambda6
 		r <- psych::alpha(d, key = key)
```
AlexanderLyNL: r is not that descriptive, perhaps call it relyFit or something

JASP-Engine/JASP/R/reliabilityanalysis.R
AlexanderLyNL: you can check by using

if (isTryError(omega)) {
    ........
}

or for multiple objects

if ( isTryError(list(omega1, omega2)) ) {
    ........
}

JASP-Engine/JASP/R/reliabilityanalysis.R
```diff
@@ -136,16 +145,27 @@ ReliabilityAnalysis <- function(dataset = NULL, options, perform = "run",
...
+		}
 
 		# calculate McDonalds omega
-		omega <- psych::omega(d, 1, flip = FALSE)[["omega.tot"]]
+		omega <- psych::omega(d, 1, flip = FALSE, plot = FALSE)[["omega.tot"]]
 
 		# calculate McDonalds omega if item dropped
 		omegaDropped <- NULL
 		if (ncol(d) > 2) {
 			omegaDropped = numeric(length = ncol(d))
```
AlexanderLyNL: <-

JASP-Engine/JASP/R/reliabilityanalysis.R
AlexanderLyNL: Isn't it possible to centralise these messages in the commonmessages.R ?

vandenman:
Hopefully squashed the commits (whole process seemed a bit wonky to me). Updated `.haserrors` to use `< 3` and `exitAnalysisIfErrors = TRUE` (Tim's comments). Also changed some variable naming to be more informative (Alexanders comments). Generally, I prefer not to use abundant "try-catch" constructions because I want to catch bugs explicitly. However, the errors in glb are likely caused by a bug in the psych package. The error occurs inside the glb function in a construction like `colnames(object) <- c(...)` but in *some* situations there are more columnnames than obj has columns.

TimKDJ:
> Generally, I prefer not to use abundant "try-catch" constructions because I want to catch bugs explicitly.

The problem with this is that any error not caught in trycatch, will crash the entire analysis. If there are still some useful computations to be made on other parts of the data/columns/etc, a trycatch would allow you to to do so. I suppose it depends on how much the different calculations depend on one another (and so if an error somewhere will surely be propagated or is pretty self-contained).

The squash did not work. Although I believe @boutinb can also do this when he merges this PR.

vandenman:
> The problem with this is that any error not caught in trycatch, will crash the entire analysis.

I agree with that line of reasoning but I feel by default try-catch wrapping of every function call to non-base packages is a bit overdoing it. While the reliability statistics seem independent, their computations are all pretty similar (similar matrix multiplications, similar requirements for correlations matrices, they always use the all data provided). If one reliability statistic errors, it is likely that they all will. Also, almost all statistics are calculated in the call `psych::alpha`. If that fails all but 2 output options will disappear. I think it's better to just let the analysis crash at that point, but other ideas for intended user experience are welcome. The case with the glb is an exception here that (I think) needs to be solved in the psych package.

PR #1772, opened 2017-05-11
temporary fix state plot issue B. independent samples t-test

quentingronau:


TimKDJ:
Let me see if I can find a more permanent solution

PR #1773, opened 2017-05-12
Prevent figures from being added to an empty state

TimKDJ:
This fixes the crashing problem of the Bayesian Independent Samples T-Test. 
I closed PR #1772 as this is a more permanent fix.

### DIFF
--- a/JASP-Engine/JASP/R/common.R
+++ b/JASP-Engine/JASP/R/common.R
@@ -1021,7 +1021,7 @@ saveImage <- function(plotName, format, height, width){
 .imgToState <- function(imgobj) {
 
 	result <- list()
-	if (!is.list(imgobj))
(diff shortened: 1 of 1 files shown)


JASP-Engine/JASP/R/common.R
```diff
@@ -1021,7 +1021,7 @@ saveImage <- function(plotName, format, height, width){
 .imgToState <- function(imgobj) {
 
 	result <- list()
-	if (!is.list(imgobj))
+	if (!is.list(imgobj) || is.null(names(imgobj)))
```
akashrajkn: @TimKDJ can you please add a comment before the ```if``` statement (for future reference)?

TimKDJ:
Sure!

PR #1780, opened 2017-05-18
 Corrected file extension when writing to OSF and read jasp files  without analyses

FransMeerhoff:
When files are saved to the OSF the file extension in the OSF file
input window was not updated to .jasp. This has been fixed.
It is now also possible to read a data file and save it directly to a
jasp file and read the jasp file back with errors.

### DIFF
--- a/JASP-Desktop/backstage/opensavewidget.cpp
+++ b/JASP-Desktop/backstage/opensavewidget.cpp
@@ -37,22 +37,6 @@ OpenSaveWidget::OpenSaveWidget(QWidget *parent) : QWidget(parent)
 	_tabWidget->setMaximumWidth(800);
 
 	QWidget *webWidget = new QWidget(this);
-	/*QGridLayout *webWidgetLayout = new QGridLayout(webWidget);
-	webWidgetLayout->setMargin(36);
-
-	QFrame *webFrame = new QFrame(webWidget);
-	QGridLayout *webLayout = new QGridLayout(webFrame);
-	webFrame->setLayout(webLayout);
-	webLayout->setMargin(0);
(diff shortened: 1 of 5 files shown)


JASP-Desktop/backstage/opensavewidget.cpp
```diff
@@ -391,6 +376,42 @@ Utils::FileType OpenSaveWidget::getCurrentFileType()
...
+QString OpenSaveWidget::getDefaultOutFileName()
+{
+	QString path = getCurrentFilePath();
+	QString DefaultOutFileName="";
+	
+	if (path != "")
+	{
+		QString name =  QFileInfo(path).baseName();
+		QString ext = QFileInfo(path).suffix();
+		Utils::FileType filetype = Utils::getTypeFromFileName(path.toStdString());
```
boutinb: fileType is not used

JASP-Desktop/backstage/opensavewidget.cpp
```diff
@@ -391,6 +376,42 @@ Utils::FileType OpenSaveWidget::getCurrentFileType()
...
+		switch (_mode)
+		{
+			case FileEvent::FileSave:
+				ext="jasp";
+				break;
+			case FileEvent::FileExportResults:
+				ext="html";
+				break;
+			case FileEvent::FileExportData:
+				if (ext=="") ext = "csv";
```
boutinb: Why here is there a 'if (ext=="")'. Probably because the file type can be .dat or .txt. But it should never be .html or .jasp. So I would be more explicit:
if (ext != 'dat' && ext != 'txt') ext = 'csv'

Issue #1184, opened 2016-02-20
"p<.001" issue


EJWagenmakers:
@raviselker 
I noticed that in the classical binomial test, p-values that are lower than .001 are summarized by "p<.001". Is this the case for other classical analyses as well? I would think that --similar to how we report extreme values for the Bayes factor-- we might want to switch to scientific notation for those cases. The way we are doing it now just hides information which some people might find relevant (I personally care whether p=.0009 or p=.000000000000000000000000000000000000000000000000000000000001.

raviselker:
It's best if we make the p-value in the table clickable so you can access the actual p-value if you would like to. Summarizing this to p<.001 by default is nice behavior as it makes the table very clear. Btw, all the classical analyses do this.

EJWagenmakers:
I agree it makes the table look nice, but the information should be
available somehow. An alternative is to make this an option somewhere.

E.J.

********************************************
Eric-Jan Wagenmakers
Department of Psychological Methods, room G 0.29
University of Amsterdam
Nieuwe Prinsengracht 130
1018 VZ Amsterdam, The Netherlands

Web: ejwagenmakers.com
Book: bayesmodels.com
Stats: jasp-stats.org
Email: EJ.Wagenmakers@gmail.com
Phone: (+31) 20 525 6420

“Man follows only phantoms.”
Pierre-Simon Laplace, last words
********************************************

hvidaldossantos:
Ciao all,
As some journals require exact p values for publication, not having the option to get exact p values in tables or at least access this information otherwise, turns out to be a deal breaker.
I would be happy to contribute to this, but I'm new to JASP and I don't know how the program is structured. In particular I don't know where to find the code that creates tables.
Thanks

EJWagenmakers:
@AlexanderLyNL @TimKDJ  @vankesteren 
Thanks for the feedback. We hope to address this soon.

AlexanderLyNL:
Just to clarify to the people following this thread. We now have implemented a global option that allows JASP to report the exact p-value. By default we report p<.001 (in accordance to APA guidelines), whenever the exact p-value is indeed smaller than .001. When the exact p-value option is turned on and the p-value is small, we now report its exact value in scientific notation. This feature will be implemented in our next release.

vankesteren:
Hi,

We just created a pull request for this. Look for the "display exact p-values" preference setting in the next release of JASP!

[gif](http://i.imgur.com/uQLNhCO.gifv)

[...]

PR #1802, opened 2017-06-06
Add exact p-value option

vankesteren:
Added an exact p-value global preferences option. This option is persistent across sessions. Exact p-values are displayed with scientific notation in the form 1.234e-56. This fixes issue #1184.


hvidaldossantos:
Hi,
Thanks for this needed feature.
I cannot find how to enable this. How do I set global preferences?
I found no mention if FAQ, User guide, or plain google search.
Thanks

boutinb:
Hi @hvidaldossantos, this feature has just been implemented in our current build, but this will be released only in the next version of JASP (hopefully begin of juli).

hvidaldossantos:
Great thx,
I thought it was already implemented in 0.8.1.1
I'll wait.
Cheers

PR #1806, opened 2017-06-08
Inclusion probabilities for matched models

TimKDJ:
Adds an option to perform a targeted inclusion analysis for Bayesian (RM) AN(C)OVA. See also https://www.cogsci.nl/blog/interpreting-bayesian-repeated-measures-in-jasp

Additionally I
- removed stepwise code (no longer used)
- made style changes


quentingronau:
I checked this for an example by hand and the numbers matched quite well (only the third decimal was different but this may be due to rounding), so I think this is fine.

Cheers,
Quentin

TimKDJ:
@quentingronau also add the if > 1 = 1 and < 0 = 0?

quentingronau:
Yes, that is probably a good idea.

TimKDJ:
Alright, guess it's good then

quentingronau:
Yes, great.

PR #1809, opened 2017-06-12
add error handling and state system to PCA

hummyhummy:
If it works for PCA I will change EFA in the same way. But there are some potential issues: (1) not sure how to use stateKey to retrieve unchanged objects and (2) not sure how to write image to file when using ggplot2 (in .screePlot the original command is print)


TimKDJ:
I'm not completely sure what you mean by "unchanged objects" - they are unchanged by nature, the question is only whether or not we can re-use them. This depends on the UI options that changed.
If you added a key then the next time the analysis is run it will be provided with all state items that are okay to be used again.
So if (! is.null(state[["model"]])) it means you can use the model object. Does that make sense?

For the plots, instead of printing you return it. The object is then provided to writeimage.
(from the top of my head, but there is documentation available)
```
p <- .screePlot(args)
content <- .writeImage(width = someX, height = someY, plot = p, obj = TRUE)

screeplot[["convertible"]] <- TRUE
screeplot[["obj"]] <- content[["obj"]]
screeplot[["data"]] <- content[["png"]]
screeplot[["status"]] <- "complete"
```

I can also help in person if needed. Just come see me :)

PR #1843, opened 2017-07-04
Machine learning common

hummyhummy:
add boosting regression and classification, add editable files for random forest regression and classification.

### DIFF
--- a/JASP-Engine/JASP/R/graphPackage.R
+++ b/JASP-Engine/JASP/R/graphPackage.R
@@ -1,0 +1,1088 @@
+.graphOptions <- list2env(list(
+        fontsize = 18,
+        legend.cex = 1.25,
+        legend.position = "auto",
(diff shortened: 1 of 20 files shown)


boutinb:
@koenderks Can you review these changes?

JASP-Engine/JASP/R/graphPackage.R
```diff
@@ -0,0 +1,1088 @@
+.graphOptions <- list2env(list(
```
koenderks: You should not include this file in the commit, since the R package should be put in the JASP folder on your computer.

hummyhummy: The JASPGraph package can not be added as a package in the JASP framework at the moment, and Mac users can not add package/library in JASP by themselves so Frans suggests this is the easiest way to call the package right now.

Issue #1853, opened 2017-07-16
Cronbach alpha confidence interval


Marian-Popa:
## Environment (for bugs)

* JASP version:
* OS name and version:
* Analysis:
## Environment (for enhancement requests)
 
* Enhancement:
* Purpose:

Marian-Popa:
Hi to the JASP Team! 
You did and do a great job, and I know what I say, beeing a long time SPSS teacher with psychology students.
I would like to suggest to add confidence interval for Cronbach alpha coeficient, which is insistently urged by some authors în the recent years.
(http://www.psyctc.org/stats/R/Feldt1.html)
Best wishes
Marian

EJWagenmakers:
Yes, this is an excellent suggestion, also for the other measures. 
@vandenman?

vandenman:
It's on my todo list!

AlexanderLyNL:
Implemented in 8.2.0

Issue #1693, opened 2017-03-20
Error in Reliability test


niteshkhilwani:
## Environment (for bugs)

* JASP version: Mac 0.8.1.0
* OS name and version: Mac OS
* Analysis: While performing reliability analysis on the dataset with 

- Mean: 0.009804
- Std. Dev: 0.099
- Min: 0
- Max: 1

Getting following error: 

This analysis terminated unexpectedly.
Error in min(eigens$values): invalid 'type' (complex) of argument
Stack trace
analysis(dataset = NULL, options = options, perform = perform, callback = the.callback)
.reliabilityResults(dataset, options, variables, perform)
psych::alpha(d, key = key)
alpha.1(C, R)
smc(R)
cor.smooth(R)

boutinb:
Can you send us the csv and jasp files you use (in a zip format)?

phehks:
I have an issue exactly same as this. I noticed that the problem only occurs when I have more than 25 items for reliability analysis.

Please advise. Thank you.

![image](https://cloud.githubusercontent.com/assets/28689487/26035868/5ebfcdaa-3906-11e7-872b-e5f724746045.png)

vandenman:
Hi, @phehks could you perhaps send us the .csv and .jasp files you use (in a zip format)? That way we can reproduce the error and see what exactly goes wrong. An initial guess is that this error is caused by missing values.  You could try to rerun the analysis without cases that contain missing values.

akashrajkn:
#1860 mostly fixes this issue. Please reopen the issue if an error occurs.

phehks:
Thank you so much.

PR #1860, opened 2017-07-21
Reliability updates

### LINKED ISSUE #1854: Reliability analysis using Cronbach alpha is not completed
## Environment (for bugs)

* JASP version: 0.8.1.2
* Analysis: Reliability analysis
[jasp cronbach.docx](https://github.com/jasp-stats/jasp-desktop/files/1155389/jasp.cronbach.docx)

The analysis has been running for more than 5 minutes and has not completed.  There is no result for the analysis.  I have tried with less variables and it is able to complete the analysis.  But it seems to remain stuck when I include all 27 variables.  

## Environment (for enhancement requests)
 
* Enhancement:
* Purpose:

vandenman:
This PR adds:

- Confidence intervals for Cronbach's alpha (feature request #1853).
- Changes in internal structure: Correlation matrix is now computed manually instead of by `psych::alpha`. This avoids a bug where `psych` functions treat square datasets as correlation matrices (it fixes #1854 and perhaps also #1693, I verified the first, for the second no data was supplied). 
- Bug fix where analysis would chrash when all variables were reverse-scaled.
- Implements state system with Tim's method.


JASP-Engine/JASP/R/reliabilityanalysis.R
```diff
@@ -69,19 +60,24 @@ ReliabilityAnalysis <- function(dataset = NULL, options, perform = "run",
...
-								observations.amount = " < 3",
-								exitAnalysisIfErrors = TRUE)
-
-		resultsAlpha <- .reliabilityResults(dataset, options, variables, perform)
-
-	} # otherwise results retrieved from state
-
+	
+	if (!is.null(variables) && length(variables) > 0) {
+		if (is.null(resultsAlpha)) { # was the main analysis retrieved from state?
```
TimKDJ: Is it possible for `state[["resultsAlpha"]]` to be null and `state[["confAlpha"]]` to be non-null?
In that specific case `.reliabilityResults()` would not be called, because you add both of these state items to `resultsAlpha` earlier.

vandenman: That would break the analysis, but it should not be possible because:


```
        # lines 43-49
	defaults = c("variables", "reverseScaledItems", "missingValues")
	stateKey <- list(
		resultsAlpha = defaults,
		confAlpha = c(defaults, "confAlphaLevel")
	)
	resultsAlpha <- state[["resultsAlpha"]]
	resultsAlpha[["ciAlpha"]] <- state[["confAlpha"]]
```

should ensure that if `resultsAlpha ` is not retrieved from the state then neither is `confAlpha`.

PR #1872, opened 2017-07-29
Map application with JASP window

akashrajkn:
Problem - In unity desktop, sometimes, the application window and the launcher do not connect. This leads to a generic icon to be shown on the launcher.

Fix - `StartupWMClass` will map jasp window with its WM class.


akashrajkn:
Tested on ubuntu 16.04.

PR #1918, opened 2017-08-16
Freq anova factors

vandenman:
This should avoid legends taking up too much space in the descriptive plots of the frequentist ANOVA. The error message shown is: `"Cannot plot separate lines for factors with more than 15 levels (variable: {{variable name}}; levels: {{number of levels of variable}})."` Perhaps it should be mentioned in the helpfile of this analysis that a workaround to the error is to put the variable with the most levels on "horizontal axis".


TimKDJ:
It appears that we can have more than just 15 (see attachment).
How did you decide on this maximum?

[factorLevelsPlot.jasp.zip](https://github.com/jasp-stats/jasp-desktop/files/1228542/factorLevelsPlot.jasp.zip)

vandenman:
@TimKDJ My choice for 15 is arbitrary. For your example, we can use more than 15. In the attachement, I changed the factor names from `0`, `1`, `2`, ... to `aLongerName0`, `aLongerName1`, `aLongerName2`, ... and the result is a huge legend dominating the plot. So plot visibility depends on multiple factors:
- Variable name length.
- If multiple column legend, the distribution of variable name lengths across columns.
- The custom width/ height used by a user (larger = better, increase the size in my attachment and suddenly the plot is readable).

Perhaps, we should find some maximum number of levels for which anova is suited in the literature and use that? 
[factorLevelsPlot2.zip](https://github.com/jasp-stats/jasp-desktop/files/1228643/factorLevelsPlot2.zip)
.

TimKDJ:
Ah, that is unfortunate. I guess we need to think this one through a little more.
Perhaps your second solution is the better one. If we can get the size of the legend we can figure out if the plotable area remains big enough or not.

Luckily this is an exotic scenario which won't really occur very often.
Let's figure this one out for the release after the current.

vandenman:
@TimKDJ the fix I implemented opens up a temporary device, calculates the ratio of legendsize / total plot size and gives an error if the legend is too big. Perhaps the error message can be formatted more clearly, and perhaps the location of this error checking can be more general. Thoughts?

Issue #1906, opened 2017-08-09
Uninformative error message when repeated-measures factor name matches a between-subjects factor name


tvickery:
## Environment (for bugs)

* JASP version: 0.8.1.2
* OS name and version: Platform-independent (test Windows 10 + macOS Sierra)
* Analysis: If you name a repeated-measures factor the same name as a between-subjects factor in a mixed-effects anova, you receive an uninformative error message (pasted below). I eventually discovered that I had done this, but the error greatly puzzled me and it took some time for me to figure out what I had done. The code currently doesn't appear to check for this and warn the user.

ERROR MESSAGE THIS YIELDS:

This analysis terminated unexpectedly.

Error in t.test.default(unlist(postHocData[listVarNamesToLevel[[k]]]), : 'y' is missing for paired test

Stack trace
analysis(dataset = NULL, options = options, perform = perform, callback = the.callback, state = state)

.resultsPostHoc(referenceGrid, options, dataset, fullModel)

t.test(unlist(postHocData[listVarNamesToLevel[[k]]]), unlist(postHocData[listVarNamesToLevel[[i]]]), paired = T, var.equal = F)

t.test.default(unlist(postHocData[listVarNamesToLevel[[k]]]), unlist(postHocData[listVarNamesToLevel[[i]]]), paired = T, var.equal = F)

stop(''y' is missing for paired test')

To receive assistance with this problem, please report the message above at: https://jasp-stats.org/bug-reports

TimKDJ:
You are correct, we do not check for this. 
I can imagine that the error message left you rather confused ;)

It shouldn't be a very hard thing to add. Thanks for notifying us!

JohnnyDoorn:
This issue will be fixed in the next release, I just updated the code!

Issue #1805, opened 2017-06-07
Include effect size in Post Hoc Tests for ANOVAs


YamilVidal:
Include effect size in Post Hoc Tests for ANOVAs.
The request is quite self explanatory.

Reporting effect sizes is (should be) more important than reporting p-values.

Thanks

boutinb:
@JohnnyDoorn Could you answer this request?

JohnnyDoorn:
Hi,
I looked into this, but did not find any cohesive way of doing this - do you have a suggestion yourself? It would be bad practice to simply report cohen's d for each post hoc comparison, as this does not correct for multiple testing. Maybe there's a way to correct cohen's d that I am not aware of.
Cheers,
Johnny

AlexanderLyNL:
Included in 0.8.2.0

https://jasp-stats.org/download/

PR #1923, opened 2017-08-22
Fixed bug #1805, #1893 and #1906

JohnnyDoorn:
- Fixed bug #1805 and #1893 - added effect sizes for posthoc tests RM ANOVA

- Fixed bug #1906 - give informative error message when same name for RM factor and between subjects factor


boutinb:
@vankesteren Hey Erik-Jan, heb je dit request gereviewed?

PR #1937, opened 2017-08-27
Minimal viable meta-analysis module

raoelg:
Added a meta-analysis module that is just complete enough to be usable. But needs a lot of work still. Many UX options not connected to any output behaviour yet. Variable names handling is wrong (quick hack to make it usable, but doesn’t work with non latin characters).

### DIFF
--- a/JASP-Desktop/JASP-Desktop.pri
+++ b/JASP-Desktop/JASP-Desktop.pri
@@ -189,7 +189,8 @@ SOURCES += $$PWD/main.cpp \
 		$$PWD/importers/importcolumn.cpp \
 		$$PWD/importers/csvimportcolumn.cpp \
 		$$PWD/preferencesdialog.cpp \
-    $$PWD/ribbons/ribbonmeta_analysis.cpp
+    $$PWD/ribbons/ribbonmeta_analysis.cpp \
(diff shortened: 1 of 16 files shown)


boutinb:
@raoelg Can you set this module available only in DEBUG mode (with #ifdef QT_DEBUG in mainwindow.cpp and tabbar.cpp)?

lindeloev:
Will this  be available in JASP and when? I had to use OpenMetaAnalyst for teaching this year but would really like to see it integrated in JASP so that students can use just one piece of software.

AlexanderLyNL:
When do you need it? :)

EJWagenmakers:
Hopefully in a week or two. No promises!
EJ

lindeloev:
Awesome! I will add MA to the growing list of ways JASP is superior to SPSS.

Since you ask: I'll teach MA with exercises on Monday but I was just asking so that I could let students know when the could expect this to arrive in JASP. Task solved, thanks!

AlexanderLyNL:
It's almost finished! Raoul and Tim are working very hard on it and only a couple of things need to be adjusted followed by some testing. Too bad we can't finish it before you discuss MA in your class, as that would have also provided a great additional test opportunity.

PR #1962, opened 2017-09-06
Tiny setup - Clean up R.Framework

FransMeerhoff:
With the requiredPackages.R script (Tim de Jong) de total R-code is by
means of namespaces checked for needed packages


Tools/requiredPackages.R
```diff
@@ -0,0 +1,65 @@
...
+if (! "stringr" %in% installed.packages()) {
+  install.packages("stringr")
+}
+library(stringr)
+
+basePkgs <- installed.packages(priority="high")
+basePkgs <- basePkgs[basePkgs[, "Priority"] == "base", 1]
+
+reqPkgs <- NULL
+expr <- '([a-zA-Z0-9.]{2,}(?<![.]))(?:::|:::)[a-zA-Z0-9._]+'
```
akashrajkn: @TimKDJ can you please add a comment for this regex.

TimKDJ: # matches namespace in [valid namespace -> triple or double colons -> valid function name]
@FransMeerhoff

TimKDJ:
GPArotation will have to be manually included in the list.
The psych package lists it under 'Suggests', but actually imports it.
I have contacted the author, but in the meantime we must add it ourselves; on line 31:
```
reqPkgs <- 'GPArotation'
```
instead of 
```
reqPkgs <- NULL
```

Tools/requiredPackages.R
FransMeerhoff: @TimKDJ @akashrajkn 

- Added GPArotation to the list
- Added comment to RegExp
- Updated the reduced Framework in https://static.jasp-stats.org/development/R3.3%20OSX%20JASP%200.8.3.zip 
(Window follows)

TimKDJ: I think it's best to add GPArotation in the script (as per my post), so we do not forget it at some point.

Issue #1959, opened 2017-09-05
Vovk-Sellke option makes correlation crash


TimKDJ:
* JASP version: 0.8.2
* OS name and version: macOS Sierra
* Analysis: Correlation Matrix
* Steps to reproduce:
1) From debug.csv add "debCollin1", "debCollin2" and "contNormal"
2) Tick "Flag significant correlations"
3) Tick "Vovk-Sellke maximum p-ratio"

I think the cause lies in the footnote the Vovk-Sellke option tries to add.

Edit: somewhat related, if you add "debInf" and then remove some other variable, the infinity footnote disappears.

JohnnyDoorn:
I just talked to Quentin (macOS Sierra 10.12.6) and Don (windows 10), who also could not replicate the issue, just like myself (ubuntu 16.04). I'll discuss it with the developers, maybe they can zoom in on when/where this happens..

Issue #1874, opened 2017-07-30
Spearman and Kendall's Tau confidence intervals


tressoldi:
## Environment (for bugs)

* JASP version: 8.1.2
* OS name and version: Windows 10
* Analysis: Correlation
## Environment (for enhancement requests)
 
* Enhancement: Contrary tot the Pearson correlation, I don't see the confidence intervals of the Spearman and Kendall's Tau
* Purpose: Is it possible to add CIs to these stats?

EJWagenmakers:
You are right! We ought to fix that. Let me ask @JohnnyDoorn or @vandenman to take a look at this. 
Cheers,
E.J.

lindeloev:
When you do, make the settings- and results layout consistent with the Bayesian analysis. 

Settings: I have a slight preference for the radio button (select one) over the tick boxes (select multiple); mostly because tick boxes allow for selecting zero. IF multiple tick boxes, then perhaps one table per correlation type rather than everything in one table (?) - or the choice.

AlexanderLyNL:
Added this on the 7th of September

PR #1963, opened 2017-09-07
Added confidence intervals for rank correlations

JohnnyDoorn:
Fix issue #1959
Fix issue #1874


TimKDJ:
Johnny was unable to reproduce the main problem in #1959. Can anyone else?

vandenman:
@TimKDJ After numerous attempts, I could replicate on windows 10.
- load Debug.csv
- preferences -> ensure exact p-values is ticked.
- correlation matrix
- Pick Debcollin1, Debcollin2, contNormal (in that order!)
- Now click Vovk-Sellke.
- JASP should be unresponsive and Correlation matrix title should dissapear.

For some reason, following the exact same steps but selecting the variables in the order contNormal, Debcollin1, Debcollin2 doesn't replicate the error for me.

TimKDJ:
Thanks for confirming this Don.
I rewrote the table function and it seems to have fixed the issue for me.
Once I do a pull request we can see if it solves the problem for you as well.

PR #1967, opened 2017-09-12
Prevent recursion of custom objects

TimKDJ:
Custom objects returned by analyses do not (should not) contain JASP images, prevent recursion of these objects.

@raoelg this fixes the problem when rma.fit is added to the state.


TimKDJ:
@vankesteren I'm guessing that in addition to this we could move
```
results$state   <- NULL # remove the state object
```

from line 119 to 98?
Why do we recurse over the state with `.ImgToResults` in any case (line 117)?

PR #2000, opened 2017-09-26
Hierarchical regression

### LINKED ISSUE #1418: Block Hierarchical Regression
would like to request that there be an option for Block hierarchical regression.  Social Sciences people rarely use stepwise regression.  We need some control on which block of variables would come first, and which block would come next. Thank you.

JohnnyDoorn:
Implemented hierarchical regression, allowing the user to check predictors to include them in the null model.
Fix #1418

### DIFF
--- a/JASP-Engine/JASP/R/regressionlinear.R
+++ b/JASP-Engine/JASP/R/regressionlinear.R
@@ -190,34 +190,53 @@ RegressionLinear <- function(dataset=NULL, options, perform="run", callback=func
 	# Fit Linear Model
 	lm.model <- list()
 	empty.model <- list(lm.fit = NULL, variables = NULL)
-
+	lm.fit.index.one.model <- 1
+	includes.nuisance <- FALSE
+	
 	if (length(options$modelTerms) > 0) {
 
 		variables.in.model <- NULL
 		variables.in.model.base64 <- NULL
-
+    variables.in.null.model <- NULL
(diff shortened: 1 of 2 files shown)


diogenesbido:
In fact, this is not hierarchical. 
It was implemented stepwise, backward, forward and enter methods, but the idea was to enter a block of independent variables (IV), and after this, another set of (IV) and so on. The order should be chosen by the researcher. Usually, we enter with control variables first and the IV in a second set. The results should show the R² change.

JohnnyDoorn:
Hi Diogenes,
The stepwise/forward/backward regression was already present in JASP. However, what was added in this merge was the ability to check your predictors 'as nuisance', to have these included in your null model. This enables you to have control over what your specified model is compared to (as opposed to the model with just the intercept), which is reflected in the relevant statistics (r^2 change, f change, etc).
JASP version 8.3, which is planned to be released next week, will have this functionality (JASP 8.2 does not).
Kind regards
Johnny

diogenesbido:
ok, I misunderstood.

I was expecting something like this (a weird example in R):

attach(infert)
mod1 <- lm(stratum ~ age) ## control variable
mod2 <- lm(stratum ~ age + induced) ## complete model

summary(mod1)
summary(mod2)
**anova(mod1, mod2)**

JohnnyDoorn:
Yes, this will be possible in the next release!
A little sneak peek:
hierarchicalregressionexample

Issue #1921, opened 2017-08-22
PostHoc output disappears


kevinvschie:
## Environment (for bugs)

* JASP version: 0.8.2.0
* OS name and version: Windows 7 Enterprise
* Analysis: posthoc test after BANOVA

Running posthocs seems to work fine. JASP produces an output table. However after a couple of sections the output table is removed. It's possible to rerun the posthoc tests and the output table appears again, but after that it disappears too.

AlexanderLyNL:
Hi Kevin,

Could you please send us the JASP file in a zip file? Then we can have a look at it.

Cheers,
Alexander

kevinvschie:
Sure, where can I send it too?

AlexanderLyNL:
a.ly@uva.nl please.

TimKDJ:
Thanks for bringing this to our attention @kevinvschie!
I can reproduce the problem and will work on a fix.
This appears to be an older bug (I can also do reproduce it in the frequentist RM ANOVA of version 0.8.1.2).

PR #2012, opened 2017-09-28
Anova Repeated Measures Post Hoc factors error

boutinb:
When setting a Post Hoc factor, hide & show the parameter panel removes
the factor


TimKDJ:
Fixes #1921

TimKDJ:
There is still an issue with the Bayesian RM plot.
Also, changing the name of the RM factor cancels elements the factor is involved in, it would be better if it just reset them.

PR #2025, opened 2017-10-03
Made view robust to empty results

TimKDJ:


JASP-Desktop/html/js/analysis.js
```diff
@@ -481,11 +481,13 @@ JASPWidgets.AnalysisView = JASPWidgets.View.extend({
 	render: function () {
 
 		var results = this.model.get("results");
-		if (results == "") { // called to update progressbar
-			var $progressbar = this.progressbar.init(this.model.get("progress"), this.model.get("id"), this.model.get("status"));
-			this.$el.find(".jasp-progressbar-container").replaceWith($progressbar);
-			this.handleVisibilityProgressbar(this.progressbar.status());
-			
+		if (results == "" || results == null) {
```
akashrajkn: Can you please tell me the case when results is null?

TimKDJ: When someone performs an empty return whilst the status is initing.

Issue #931, opened 2015-09-07
Pairwise Mean Comparison for RM ANOVA and a few more things


renenob:
Good day Jonathon!

I am not sure if these have been previously requested.
1. Pairwise mean comparison for RM ANOVA (this would greatly improve the quality of information that can be inferred from the results of 1way RM, 2way RM, and split plot ANOVAS)
2. In the context of 2way RM, 2way BS, and split plot ANOVAs - It would be nice to have significance test and post hoc (or pairwise comparison) within each level of one of the factors or independent variable.  For example, the levels Factor A is  Male/Female, while Factor B is High / Average / LOW.  I would like to be able to test is there is a significant differences in the DV among High / Average / and Low, but among Males only (plus post hoc). Then, I would like to  do the same significance and post hoc test again, but his time only for females.  This would be great for moderation analysis when the IVs are categorical. 

I hope that made sense.  Thank you very much!

borisma:
Hi @renenob and @jonathon-love 
This is what I also suggested in my recent email with suggestions! Would be really important.

renenob:
Thank you for hearing us on Hierarchical Regression!  Will we see this feature in the near future, too?

AlexanderLyNL:
Implemented

PR #2050, opened 2017-10-11
Included effect sizes for all anova's

JohnnyDoorn:
Now also for regular anova and ancova


AlexanderLyNL:
Does this solves issue #931?

JohnnyDoorn:
No, this simply adds effect size measures (Cohen's d) to the post hoc tests in the non-RM ANOVA's - before this was only included in the RM ANOVA. The issue you refer to pertains to having a split by option for post hoc tests, or to be able to include an interaction term in the post hoc tests. It is a nice suggestion though and I will look into implementing it.

Issue #1719, opened 2017-04-06
JASP cannot handle non-English systems


baltzis:
## Environment (for bugs)

* JASP version: 0.8.1.1
* OS name and version: Windows 10 64-bit (Greek)
* Analysis: Probably JASP cannot work in OS that is not in English not only because it cannot read path names with non-ASCII characters, but also because even if the files are stored in paths with names in plain ASCII, then there is always the incapability of handling paths to user profiles that may not contain plain ASCII characters. The following message was received after a failure to carry out an analysis:

This analysis terminated unexpectedly.
Error in grid.newpage(): could not open file 'D:/Users/������ & ����/AppData/Local/JASP/temp/7280/resources/5/_18.png'
Stack trace
analysis(dataset = NULL, options = options, perform = perform, callback = the.callback)
.independentSamplesTTestDescriptivesPlot(dataset, options, perform)
print(p)
print.ggplot(p)
grid.newpage()
To receive assistance with this problem, please report the message above at: https://jasp-stats.org/bug-reports

This may also answer the issue about .jasp files that although they have been saved in paths with names containing only ASCII characters, for some obscure reason after a while or some time later, they can't be opened by JASP which then crashes.

## Environment (for enhancement requests)
 
* Enhancement:
* Purpose:

aknight1-uva:
Duplicate #1695

boutinb:
Solved by #2064

Issue #1704, opened 2017-03-30
Please add Chinese support in your next version


Sciurus365:
## Environment (for bugs)

* JASP version: 0.8.1.1
* OS name and version: Windows 7 
* Analysis:
I found that JASP can not open any files with name or path including Chinese words.
The figures generated by JASP can not show Chinese words, either.
I really like this software, but some of my works are required to be reported in Chinese, so I can't use JASP to analysis these data. Could you please add Chinese support in your next vision? Thank you!

SCgeeker:
Hi, be sure your Chinese words are encoded in utf-8. You could check my files for teaching.

https://osf.io/x4h7d/

Sciurus365:
I've checked my file and it is encoded in uft-8. And I tried again with your file. The words are displayed correctly in the **charts**, but Chinese words are still not able to be shown in the **plots** (see the picture attached. Could you please check it?

aknight1-uva:
Duplicate issue #1695

boutinb:
Solved by #2064

PR #2064, opened 2017-10-19
Read File with path with strange Characters

### LINKED ISSUE #1695: Opening .sav file, graphics output
## Environment (for bugs)

* JASP version: 0.8.1.1
* OS name and version: Windows 10 (64bit)
* Analysis: When I try to open a .sav file in a directory other than route, I get the message:
"Unable to open file. No header found in .SAV file".
I suspect that this is because my system is in Greek and several directories have Greek names. The same file, however, can be opened from the route directory of "My Documents".

* Second problem: While labels in Greek appear OK in tables (e.g. frequencies), in plots they appear in gibberish (see attached image).
test-jasp

boutinb:
Problem on Windows 10
Fixes #1695, #1704, #1719

### DIFF
--- a/JASP-Engine/JASP/R/common.R
+++ b/JASP-Engine/JASP/R/common.R
@@ -101,16 +101,7 @@ run <- function(name, options.as.json.string, perform="run") {
 				state[["figures"]] <- c(state[["figures"]], .imgToState(results$results))
 			}
 
-			location <- .fromRCPP(".requestStateFileNameNative")
-
-			relativePath <- location$relativePath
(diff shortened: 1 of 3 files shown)


JASP-Engine/JASP/R/common.R
```diff
@@ -377,21 +368,27 @@ isTryError <- function(obj){
 }
 
 .saveState <- function(state) {
-
+	result <- list()
+	relativePath <- NULL
+	fullPath <- NULL
```
TimKDJ: fullPath should be root.

PR #2079, opened 2017-10-26
Meta analysis v1

raoelg:
model without predictors and no intercept handled more gracefully;
heterogeneity parameters and confint not computed when method = FE;
changed methods names in UI to full names; call to rma() wrapped in
tryCatch() for graceful failure; removed option ‘GENQ’ from UI; made
images convertible; fixed influential cases; fixed issues with
as.jaspTable.data.frame; added d64.rma method; added and altered
previous versions of b64/d64 methods; other fixes


JASP-Engine/JASP/R/classicalmetaanalysis.R
```diff
@@ -0,0 +1,784 @@
...
+      if( identical(formula.rhs, ~ 1 - 1) )
+        .quitAnalysis("The model should contain at least one predictor or an intercept.")
+
+      rma.fit <- tryCatch(
+        metafor::rma(
+          yi = get(b64(effsizeName)), sei = get(b64(stderrName)), data = dataset,
+          method=method, mods = b64(formula.rhs), test = options$test,
+          slab = if(options$studyLabel != "") get(b64(options$studyLabels)),
+          level = options$regressionCoefficientsConfidenceIntervalsInterval
+        ), error = .quitAnalysis)
```
TimKDJ: In general .quitAnalysis is used when we expect the analysis to abort.
In these cases we can give a clear, informative error message.

I don't see the added benefit of this tryCatch, because it will still be an exception, not an expected error.

JASP-Engine/JASP/R/common.R
```diff
@@ -501,7 +501,8 @@ isTryError <- function(obj){
 			
 		} else {
 		
-			stop(paste("bad call to .unv() : v is \"", v, "\"", sep=""))
+			#stop(paste("bad call to .unv() : v is \"", v, "\"", sep=""))
+		  vs[length(vs)+1] <- v
```
TimKDJ: What happens here? Is this for cases when it is a character like ":"?

JASP-Engine/JASP/R/common.R
```diff
@@ -754,6 +755,28 @@ callback <- function(results=NULL, progress=NULL) {
 	stop("could not clean value")
 }
 
+# .clean <- function(value) {
```
TimKDJ: Can you remove these commented out lines? We can update .clean at some point, but this seems unnecessary :)

JASP-Engine/JASP/R/common.R
```diff
@@ -1107,11 +1130,121 @@ saveImage <- function(plotName, format, height, width){
...
+  as.name(d64(as.character(x)))
+}
+b64.list <- function(x, ...) {
+  rapply(x, b64, ..., how = "replace")
+}
+d64.list <- function(x, ...) {
+  rapply(x, d64, ..., how = "replace")
+}
+	
+.newProgressbar <- function(ticks, callback, skim=5, response=FALSE) {
```
TimKDJ: Looks like a merge conflict. Please remove the upper .newProgressbar()

JASP-Engine/JASP/R/common.R
raoelg: Actually: .unv intends to throw an input error but instead crashes the execution. It makes more sense to do no transformation if the condition is not satisfied. Also with the b64 and d64 functions it crashes if (some of) the variable names are not in base 64 encoding

TimKDJ: Seems sensible to me too. Just remove the commented out line.

Issue #2065, opened 2017-10-20
Logistig Regression in Ubuntu


WillemVlakveld:
When conducting a Logistic Regression applying JASP Version 0.8.3.1 running in the Linux operating system Ubuntu version 16.04 LTS, I get an error messages as soon as I select a dependent variable (with the values 0 and 1). This error does not occur when I do the same analysis with JASP running in Windows. The Error message is:

This analysis terminated unexpectedly.

Error: could not find ‘startsWith’

Stack trace
analysis(dataset = NULL, options = options, perform, callback = the.callback, state = state)
.hasErrors(dataset, perform, type = ‘factorLevels’, factorLevels.target = options[[‘dependent’]], factorLevels.amount = ‘≠ 2’, exitAnalysisifErrors = TRUE)

vandenman:
Hi,

This is a known issue with the R version in Ubuntu repositories being a outdated (as was similarly observed  in #2063 ). A workaround for updating R is also given in #2063 .

Issue #2063, opened 2017-10-19
Ubuntu 16.04 R version dependency


luketudge:
## Error: could not find function startsWith
* JASP version: 0.8.3.1
* OS name and version: Ubuntu 16.04.3 (xenial)
* Analysis: any
* Steps to reproduce: start Jasp, open example data, try any analysis

I had updated Jasp a couple of times with the last updates (via the Ubuntu ppa), but not used it for a while. Then I noticed that all analyses had started returning the error 'could not find function startsWith'.

I have resolved this problem on my system, but I submit this issue in case anybody else has encountered it.

The startsWith function is available only in more recent versions of R (3.3). Since Jasp now seems to need this function a lot, Ubuntu users will need to upgrade their R installation in order to be able to use Jasp. The official Ubuntu repositories tend to be a bit slow in providing new versions, so for the moment, R 3.3 is not available there. The problem can be resolved by getting R from a CRAN repository instead of from Ubuntu. One method is described here:

[https://www.digitalocean.com/community/tutorials/how-to-install-r-on-ubuntu-16-04-2]

The CRAN versions are more up to date than Ubuntu's. So once I added a CRAN repository and reinstalled / upgraded R, Jasp worked as normal again.

If this is a common issue at the moment for Ubuntu users, then it might be worth adding a note to the installation guide. Hopefully it will resolve itself eventually once Ubuntu catches up to R 3.3.

Issue #1842, opened 2017-06-30
Various t-test errors


lindeloev:
## Environment (for bugs)

* JASP version: 8.1.2
* OS name and version: Ubuntu 16.04
* Analysis: t-tests

I just got the latest JASP from your Debian repo and took it for a spin out of pure procrastination. I'm considering JASP for a stats course in the fall, so I'm following you guys - excellent work! :-) I discovered a few bugs in the t-test analyses. While I'm at it, I also propose a few improvements.

# Frequentist independent t-test error:
The first thing I noticed: the frequentist independent samples t-test fails with an error on the Bugs dataset:

```
Error: could not find function 'startsWith'

Stack trace
analysis(dataset = NULL, options = options, perform = perform, callback = the.callback, state = state)

.hasErrors(dataset, perform, type = 'factorLevels', factorLevels.target = options$groupingVariable, factorLevels.amount = '≠ 2', exitAnalysisIfErrors = TRUE)
```

I haven't look into your project's development setup, but this should've been caught by a unit test?

# Bayesian t-test credible interval error
1. When selecting an informed prior, the plot does not update it's intervals. E.g. set a Normal(0, 0.001) prior. The effect of that should be visible on the plot, but I see no change.
2. It's not a "descriptive plot" if there are credible intervals since the latter are inferential. Maybe a title like "Plot of parameters" or the like?
3. It would be nice to get the credible interval(s) in a table as well. Perhaps under "additional statistics."

# Inconsistent t-test layout
When moving between the different types of t-test, the layout changes for fields where the content is the same. E.g. the box for entering the credible interval is indented more in paired samples than independent samples. This is true of Bayesian and frequentist analyses as well.

JohnnyDoorn:
Hi Jonas,
Thanks for your interest and following! I can respond to your first and third inquiry - the second one is maybe for @quentingronau , as he implemented the subjective Bayesian t-test.
1. I have had this issue as well (I'm on Ubuntu 16.04), when I was still on an older version of R (I think I was at 3.2, when 3.4 was already out). Updating my R version fixed this error message - this startsWith function is probably a function that was included in R in the new version.
3. Thank you for catching this - I had already fixed the indentations and spacings for the frequentist t-tests a couple of days ago, so it should be in the next release! I will also take a look at the Bayesian tests and fix the indentations accordingly.
Kind regards,
Johnny

AlexanderLyNL:
[...]

PR #2081, opened 2017-10-26
Add startsWith, endsWith for R < 3.3

TimKDJ:
These base functions were recently added and break linux installs.

Issues #1842, #2063 and #2065


vandenman:
I like this as a fix for R's backwards compatibility issues, but it does feel kinda hacky. Also, on windows/ mac, will those use these functions instead of the new R > 3.3 `startsWith()`? We could avoid loading these functions in R > 3.3 by wrapping the function definitions in something like:

```
if (as.numeric(version$minor) < 3.3) { 
       # load startsWith and endsWith
        source("compatabilityR3_3.R")
}
```
Perhaps we should also use something similar to create a message (pop-up) on jasp startup to tell people they should update R, and warn them that if they don't errors might appear. I bet there are more compatibility issues than just the ones we encountered thus far.

TimKDJ:
Good ideas.
I added an if-statement around the two functions.
The warning would probably be nice too, I'll give this some thought.

PR #2098, opened 2017-10-31
Make hasErrors more robust to bad checks

TimKDJ:


JASP-Engine/JASP/R/commonerrorcheck.R
```diff
@@ -256,9 +256,14 @@
...
-      checkResult <- base::do.call(check[['callback']], callingArgs)
+      checkResult <- try(base::do.call(check[['callback']], callingArgs))
     } else {
-      checkResult <- check[['callback']]()
+      checkResult <- try(check[['callback']]())
+    }
+    
+    # See if the check itself terminated with an exception (oh dear).
+    if (isTryError(checkResult)) {
+      next
```
AlexanderLyNL: only next, isn't that informative. Perhaps return a fail error check message

AlexanderLyNL: Say no more fam!

PR #2101, opened 2017-10-31
Network module fixes

vandenman:
The network module!

And some changes to error handling:
- `.checkVarCovMatrix` check now tests if matrices are positive definite instead of positive semi-definite.
- Implemented an error check to test if a correlation or covariance matrix is positive-definite when a user supplies raw data. The test accepts a grouping variable.


JASP-Engine/JASP/R/commonerrorcheck.R
```diff
@@ -552,3 +553,39 @@
...
+  
+  result <- list(error=FALSE, errorVars=NULL)
+  
+  if (is.null(corFun))
+    corFun <- stats::cor
+
+  if (is.null(grouping)) {
+    dataset <- list(dataset)
+  } else {
+    groupingData <- dataset[[.v(grouping)]]
```
TimKDJ: Would .applyOnGroups() work for this?

JASP-Engine/JASP/R/networkanalysis.R
```diff
@@ -0,0 +1,2132 @@
...
+		if (is.null(state[["network"]])) {
+
+		  # default error checks
+		  checks <- c("infinity", "variance", "observations")
+
+			groupingVariable <- NULL
+			if (options[["groupingVariable"]] != "") {
+				groupingVariable <- options[["groupingVariable"]]
+				# these cannot be chained unfortunately
+				.hasErrors(dataset = dataset[.v(groupingVariable)], perform = perform,
```
TimKDJ: Is this still needed with the more robust error handling? If so, changes must be made.

JASP-Engine/JASP/R/commonerrorcheck.R
vandenman: I tried some things, but `.applyOnGroups` doesn't work for me. The problem is with the line `func(data[[.v(target)]])`. In my case, `data[[.v(target)]]` must return something that can be interpreted as a matrix. However, due to the double `[[` target cannot have length larger than 1. We could make an `.applyOnGroups` that allows for matrix returns, but so far this is the only use case.

JASP-Engine/JASP/R/networkanalysis.R
vandenman: Sure, but I'd like to test first if that that PR fixes the issues, before making the changes.

TimKDJ:
Finished, @vandenman?

vandenman:
Finished!

PR #2128, opened 2017-11-07
Fixes logistic regression bug

TimKDJ:


JASP-Engine/JASP/R/commonglm.R
```diff
@@ -442,7 +442,7 @@
     }
 
     # Add footnote of predicted level
-    if (options[["dependent"]] != "") {
```
vankesteren: Thanks, looks good

Issue #2122, opened 2017-11-06
spaces in Log(BF10)


EJWagenmakers:
This is obviously a minor point, but it is also easy to fix:
When selecting "Log(BF10)" (in all of the Bayesian analyses), the output in the table refers to this as "Log( BF10 )". I am not sure why we need the extra spaces. @TimKDJ?

AlexanderLyNL:
3 t-tests,

correlation pairs,
 
binomial,
contingency

and the equivalent tests in sumStat. Easy fix.

Issue #2083, opened 2017-10-26
Additional info for prior-posterior plot and robustness check on by default


AlexanderLyNL:
## Template for enhancement requests
 
* Enhancement: Have the additional on by default
* Purpose: Change default behaviour

Easy to get started

akashrajkn:
![screensht](https://user-images.githubusercontent.com/6959308/32067874-e652d302-ba84-11e7-9b38-c8d9ffa7cab8.jpg)

The selection of option `prior and posterior plots` should select `Additional info` by default.

Details - The default state of options are mentioned in the Resources (json) file of each analysis.

PR #2139, opened 2017-11-08
Fix/minor

### LINKED ISSUE #2121: SumStats p-value
In the SumStats module, go to Bayesian binomial, add number of successes > 1 and number of failures > 1, and set test value to 0 or to 1. Then the p-value returned is "false". I am not sure what the correct value is; the observed data are impossible under H0, so I would be tempted to say it should be 0. @AlexanderLyNL, what do you think? @akashrajkn 
See screenshot for proof.
pfalse

akashrajkn:
Fixes #2121, #2122 and #2083


akashrajkn:
Fixing #2122 involved too much leg-work. We should probably write a common BF-title function that should be used in all analyses

AlexanderLyNL:
Agree, same for starring them. I was thinking about a report bf function, that selects:
- BF+0, BF-0, BF10, 
- 1/BF or Log
- How many stars
- column label

Your SumStats functions can be used as a template. This also requires some naming convention across the analyses.

PR #2144, opened 2017-11-09
Fixes to all testing error reports

raoelg:
- wrapped call to rma() in tryCatch so that informative messages
generated by rma() are directly communicated via .quitAnalysis
- coerced variable values used as labels to strings to deal with NA’s
- added automatic adjustment of figure height for forest plot to suite
the number of studies better (needs to be changed to allow for user
adjustment)
- increased maxiter to avoid non-convergence of the Fisher scoring
algorithm (but this is still not guaranteed; should be parameter in the
interface, and error message has to be modified to indicate this)
- included a check on positiveness of the values of the Effect Size
Standard Errors, sending an error msg to use through .quitAnalysis
- resolved an issue with inconsistent interpretation of the value of
‘level’ in metafor package: if level=1 it is interpreted as a fraction,
if level > 1 it is interpreted as a percentage

### DIFF
--- a/JASP-Engine/JASP/R/classicalmetaanalysis.R
+++ b/JASP-Engine/JASP/R/classicalmetaanalysis.R
@@ -124,6 +124,8 @@ ClassicalMetaAnalysis <- function(dataset=NULL, options, perform="run", callback
     )
     
     .hasErrors(dataset, perform, type=c("infinity", "observations"),
+               custom = error_if(any(dataset[[b64(stderrName)]] <= 0), "Non-positive standard errors encountered in variable '", stderrName, 
+                                 "'. Standard errors should be positive. Please check your input."),
                all.target=numeric.variables, observations.amount="< 2",
                exitAnalysisIfErrors=TRUE)
   }
@@ -154,10 +156,6 @@ ClassicalMetaAnalysis <- function(dataset=NULL, options, perform="run", callback
       if (identical(formula.rhs, ~ 1 - 1))
         .quitAnalysis("The model should contain at least one predictor or an intercept.")
       
-      # check input data
-      if (any(dataset[[b64(stderrName)]] <= 0))
-        .quitAnalysis(paste0("Non-positive standard errors encountered in variable '", stderrName, "'. Standard errors should be positive. Please check your input."))
-      
       # analysis
       rma.fit <- tryCatch( # rma generates informative error messages; use them!
         metafor::rma(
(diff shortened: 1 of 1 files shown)


JASP-Engine/JASP/R/classicalmetaanalysis.R
```diff
@@ -124,39 +124,49 @@ ClassicalMetaAnalysis <- function(dataset=NULL, options, perform="run", callback
...
         formula.rhs = update(formula.rhs, ~ . + 0)
-      if( identical(formula.rhs, ~ 1 - 1) )
+      
+      if (identical(formula.rhs, ~ 1 - 1))
         .quitAnalysis("The model should contain at least one predictor or an intercept.")
-
-      rma.fit <- metafor::rma(
+      
+      # check input data
+      if (any(dataset[[b64(stderrName)]] <= 0))
```
TimKDJ: You can add this to the .hasErrors call. This has the added benefit of grouping all error checking and the ability to show multiple failures in one message.

```
    # check input data
    checkSE <- function() {
      if (any(dataset[[b64(stderrName)]] <= 0))
        paste0("Non-positive standard errors encountered in variable '", stderrName, "'. Standard errors should be positive. Please check your input.")
    }
    
    .hasErrors(dataset, perform, type=c("infinity", "observations"), custom=checkSE,
               all.target=numeric.variables, observations.amount="< 2",
               exitAnalysisIfErrors=TRUE)

```

JASP-Engine/JASP/R/classicalmetaanalysis.R
```diff
@@ -124,39 +124,49 @@ ClassicalMetaAnalysis <- function(dataset=NULL, options, perform="run", callback
...
+        metafor::rma(
           yi = get(b64(effsizeName)), sei = get(b64(stderrName)), data = dataset,
           method=method, mods = b64(formula.rhs), test = options$test,
-          slab = if(options$studyLabel != "") get(b64(options$studyLabels)),
-          level = options$regressionCoefficientsConfidenceIntervalsInterval
-        )
+          slab = if(options$studyLabel != "") paste0(get(b64(options$studyLabels))),
+          level = options$regressionCoefficientsConfidenceIntervalsInterval + 1e-9, # add tiny amount because 1 is treated by rma() as 100% whereas values > 1 as percentages
+          control = list(maxiter = 500)
+        ), error = .quitAnalysis)
```
TimKDJ: Still have some doubts about the blanket tryCatch(), e.g.:

`Error in metafor::rma(yi = get(b64(effsizeName)), sei = get(b64(stderrName)), : Fisher scoring algorithm did not converge. See 'help(rma)' for possible remedies.`
Gives a lot of useless information. Now if we change the error condition a little:
```
), error = function(e) .quitAnalysis(e$message))
```
We obtain a slightly better message:
`Fisher scoring algorithm did not converge. See 'help(rma)' for possible remedies.`
Although the 'help(rma)' bit is still a bit evil, because this does not help JASP users.

We'll keep it because in general the messages are quite good, but please add the code suggestion.

TimKDJ:
Problems from the .jasp file seem to be solved. Nice job!
Please incorporate the two code changes and we're good to go.

PR #2197, opened 2017-11-30
Changes after feedback on BAS

TimKDJ:
Main things:
- fully replaced our old Bayesian Linear Regression with the new BAS implementation
- added posterior estimates with different model selectors (median, most complex, best, BMA)
- added BFM column to the main table
- added scale parameter together with JZS prior
- removed marginal likelihood column (was literally log(BF10))
- removed alpha parameter from g-prior (no suitable default)
- removed combination of BAS+MCMC


akashrajkn:
Can you update the help file? (Options and References)

TimKDJ:
Could you make a draft? I need to make some more progress with the 'easy R' code

AlexanderLyNL:
Once the draft is completed, I'll look over it, so Tim can work on the other project.

PR #2203, opened 2017-12-04
jasptools update to v0.5.1

TimKDJ:
changes:
- changed name from JASPTools to jasptools (easier to type and more consistent)
- future proved jasptools by letting it call ‘run’ (in common.R) instead of calling analyses directly; this way any change to ‘run’ is automatically incorporated
- added ability to find pkg root if jasp-desktop is not called jasp-desktop
- improved pkg finding on Windows
- show the full stack-trace by default in browser windows


TimKDJ:
FYI: I've moved jasptools to jasp-stats, so maintenance can be performed by any team member.

Issue #2196, opened 2017-11-30
Bug during independent sample Bayesian t-test


philippberens:
* JASP version: 0.8.4.0
* OS name and version: Win 7
* Analysis: Independent sample Bayesian t-test
* Steps to reproduce: 
Load [ex6_geneexp.zip](https://github.com/jasp-stats/jasp-desktop/files/1517722/ex6_geneexp.zip)
Run Bayesian t-test
with Group as Factor and Expression as Variable

Output by JASP:
This analysis terminated unexpectedly.

Error in ttest.rows[[rowNo]]$BF: $ operator is invalid for atomic vectors

Stack trace 
analysis(dataset = NULL, options = options, perform = perform, callback = the.callback, state = state)

.ttestBayesianIndependentSamplesTTest(dataset, options, perform, state = state, diff = diff)

AlexanderLyNL:
In case you'd like to see the result sooner rather than later, here it is:

![image](https://user-images.githubusercontent.com/10271675/33517819-24e59afa-d78b-11e7-9673-8a82abaaf508.png)

philippberens:
Thanks!

PR #2204, opened 2017-12-05
handle integer overflow in Bayesian t-tests

ASarafoglou:
Fixes to #2196

### DIFF
--- a/JASP-Engine/JASP/R/ttestbayesianonesample.R
+++ b/JASP-Engine/JASP/R/ttestbayesianonesample.R
@@ -4517,9 +4517,8 @@ TTestBayesianOneSample <- function(dataset=NULL, options, perform="run", callbac
                             oneSided = FALSE, options) {
 
   tValue <- unname(t.test(x, y, paired = paired, var.equal = TRUE)$statistic)
-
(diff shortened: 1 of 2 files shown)


TimKDJ:
You need to set a seed for the unit test to work. Currently rnorm will draw random numbers every time.

JASP-Engine/JASP/R/ttestbayesianonesample.R
```diff
@@ -4517,9 +4517,8 @@ TTestBayesianOneSample <- function(dataset=NULL, options, perform="run", callbac
                             oneSided = FALSE, options) {
 
   tValue <- unname(t.test(x, y, paired = paired, var.equal = TRUE)$statistic)
-
-  n1 <- length(x)
-  n2 <- ifelse(paired, 0, length(y))
+  n1 <- as.numeric(length(x))
```
AlexanderLyNL: Let's add a comment to this

# length(x) is an integer and integer times integer leads can lead to an overflow
# numeric multiplication is more robust in R

AlexanderLyNL:
Fixes issue #2196

PR #2216, opened 2017-12-11
Fix to wrong residuals in between subjects table 

### LINKED ISSUE #1677: Error with only within-subjects variables in an ANOVA

TimKDJ:
The RM ANOVA between subjects table is wrong when there is no between subjects effect.

Fixes #1677.


TimKDJ:
Just noticed I apparently completely forgot to create a set of unit tests for the RM ANOVA.

Issue #2201, opened 2017-12-02
Weirdness in BUGS data set


EJWagenmakers:
Something has gone wrong on line 40 (38). I think we ought to delete that row.
![image](https://user-images.githubusercontent.com/8603015/33516387-8c40ffcc-d737-11e7-9b26-ed7e056c4abb.png)

EJWagenmakers:
And while you are at it, row 90 can also go

AlexanderLyNL:
That's strange. Everything is okay for me. 

Case 40 has a missing value under gender. 

<img width="1312" alt="screen shot 2017-12-02 at 17 45 40" src="https://user-images.githubusercontent.com/10271675/33517679-be89483a-d788-11e7-9577-f4dd75c71892.png">

Case 97 has a missing value under education.

<img width="1312" alt="screen shot 2017-12-02 at 17 46 23" src="https://user-images.githubusercontent.com/10271675/33517678-be6da5e4-d788-11e7-94df-74914d9179df.png">

AlexanderLyNL:
@EJWagenmakers What's the status of this? This data set works well for me.

boutinb:
Johnny solved this

PR #2239, opened 2017-12-19
Removed blank cells from Bugs example dataset

JohnnyDoorn:


AlexanderLyNL:
This wasn't needed for me, because JASP was able to detect the missing correctly, see my comments in issue #2201

PR #2254, opened 2018-01-04
Fixed regression linear issue

Janto91:


JASP-Tests/R/tests/testthat/test-regressionlinear.R
```diff
@@ -82,6 +116,28 @@ test_that("Coefficients Covariance table results match", {
   )
 })
 
+test_that("Coefficients Covariance table results match if one predictor is redundant", {
```
vandenman: This unit test (lines 120 - 131) is basically identical to the previous one. You can reuse the results from the previous one to do another one (as long as the objects are not overwritten, the testing works as a standard R script). Is it possible to add `options$regressionCoefficientsCovarianceMatrix <- TRUE`  to the previous test and then only make one call to `jasptools::run`?

Janto91:
I updated the unit-test :)

PR #2278, opened 2018-01-29
Multinomial Table widget

akashrajkn:
Tested only on Ubuntu 16.04

Current functionality:
 - Add column (all counts are initialized with 1)
 - Delete column
 - reset (This will remove all hypotheses and load the default one - H1)


akashrajkn:
@vankesteren can you check it out and tell me if you need any more functionality?

vankesteren:
This pull request is now in #2285 Multinomial Test. I needed to make a few more adjustments!

PR #2298, opened 2018-02-07
RM ANOVA Simple Effects

### LINKED ISSUE #1957: Feature request: Simple effects in Anova
* Enhancement: add simple effects analyses in Anova
* Purpose: would it be possible to include in JASP the possibility to test simple effects in Anova, please ? traditionnaly, when an interaction is significant (between two factors for example), it is interesting to describe this interaction by using simple effects (testing the effect of factor 1 for each condition of factor 2).

JohnnyDoorn:
Fix  #1957

### DIFF
--- a/JASP-Engine/JASP/R/anovarepeatedmeasures.R
+++ b/JASP-Engine/JASP/R/anovarepeatedmeasures.R
@@ -50,6 +50,7 @@ AnovaRepeatedMeasures <- function(dataset=NULL, options, perform="run", callback
 	stateLevene <- NULL
 	statePostHoc <- NULL
 	stateContrasts <- NULL
+	stateSimpleEffects <- NULL
 	stateSphericity <- NULL
 
 	if ( ! is.null(state)) {  # is there state?
@@ -95,6 +96,16 @@ AnovaRepeatedMeasures <- function(dataset=NULL, options, perform="run", callback
 			# old levene's table can be used
 
 			stateLevene <- state$stateLevene
+		}
+		
+		if (is.list(diff) && diff[['withinModelTerms']] == FALSE && diff[['betweenModelTerms']] == FALSE && diff[['repeatedMeasuresCells']] == FALSE &&
+		    diff[['repeatedMeasuresFactors']] == FALSE && diff[['simpleFactor']] == FALSE && diff[['moderatorFactorOne']] == FALSE &&
+		    diff[['moderatorFactorTwo']] == FALSE) {
+		  
+		  # old simple effects tables can be used
+		  
+		  stateSimpleEffects <- state$stateSimpleEffects
+		  
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


JASP-Engine/JASP/R/anovarepeatedmeasures.R
```diff
@@ -2501,6 +2527,273 @@ AnovaRepeatedMeasures <- function(dataset=NULL, options, perform="run", callback
...
+    orderOfTerms <- unlist(options$withinModelTerms[[length(options$withinModelTerms)]]$components)
+    indexofOrderFactors <- match(allNames,orderOfTerms)
+
+    for (level in lvls[[1]]) {
+      # For each level of the first moderator factor, take a subset of the dataset, and adjust the options object
+      # Suboptions is the same as options, except that the first moderator factor has been removed as a predictor 
+      # (because each subset only has one level of that factor). The same procedure is applied to the second moderator, if specified.
+      
+      subOptions <- options
+      if (isModeratorOneWithin) {
```
vankesteren: Yeah alright this if/else block is very long. Do you think it would be shorter with a [`switch()`](https://stackoverflow.com/questions/10393508/how-to-use-the-switch-statement-in-r-functions) statement? If not, leave it like this but maybe add some comments within each condition to show which case we're in.

JASP-Engine/JASP/R/anovarepeatedmeasures.R
```diff
@@ -2501,6 +2527,273 @@ AnovaRepeatedMeasures <- function(dataset=NULL, options, perform="run", callback
 	list(result=descriptivesPlotList, status=status, stateDescriptivesPlot=stateDescriptivesPlot)
 }
 
+.rmAnovaSimpleEffects <- function(dataset, options, perform, fullModel, fullAnovaTableWithin, fullAnovaTableBetween, status, singular, stateSimpleEffects) {
+  
+  if (identical(options$simpleFactor, "") | identical(options$moderatorFactorOne, ""))
```
AlexanderLyNL: Please call a items within a list using "[[]]" rather than with $, for instance, 

options[["simpleFactor"]] 

This is more robust, which is necessary due to the fact that you have items that are named similar such as "moderatorFactorOne" and "moderatorFactorTwo".

JASP-Engine/JASP/R/anovarepeatedmeasures.R
```diff
@@ -2501,6 +2527,273 @@ AnovaRepeatedMeasures <- function(dataset=NULL, options, perform="run", callback
...
+      # Suboptions is the same as options, except that the first moderator factor has been removed as a predictor 
+      # (because each subset only has one level of that factor). The same procedure is applied to the second moderator, if specified.
+      
+      subOptions <- options
+      if (isModeratorOneWithin) {
+        rmFactorIndex <- which(lapply(options$repeatedMeasuresFactors, 
+                                      FUN = function(x){x$name == options$moderatorFactorOne}) == TRUE)
+        splitNames <- unlist(lapply(strsplit(factorNamesV,  split = "_"), 
+                                      FUN = function(x) x[indexofOrderFactors[rmFactorIndex]]))
+        splitWithinNames <- unlist(lapply(strsplit(withinFactorNamesV,  split = "_"),
```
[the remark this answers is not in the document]
AlexanderLyNL: Perhaps we should make standard methods for this operation in common

JohnnyDoorn:
I have addressed the comments in the latest commits. Also added functionality to either pool the error terms or not - the literature seems to disagree on this, so I figured I put it in as an option. The default is set to non-pooled, as this is more robust to violations of homogeneity of variances.

Issue #2049, opened 2017-10-11
HiDPI issues on Windows


vankesteren:
# HiDPI support in JASP
Due to their high resolution at a small screen size, some modern Windows laptops have issues with the JASP user interface. Common screen resolutions for such laptops are 3200x1800 ("Quad HD") or 3840x2160 ("4K"). A number of issues occurs in all versions of JASP. Examples of issues on such problems are #2047, #2046, #2045,  #2027, #1234 and #1212. 

# Why is this an issue?
In order to make things legible, windows allows you to "scale" text and other elements on-screen. A common scaling factor for the resolutions mentioned above at 13.3 inch and 15.6 inch screens, respectively, is 200%. This happens in the settings:

![image](https://user-images.githubusercontent.com/11596858/31428576-7dfa8dec-ae6b-11e7-88d4-135c07b46e0c.png)

The problem is that the `Qt` framework we are currently using is not able to "access" this value and adjust its UI drawing based on this. This means that if we fix a UI-drawing issue for  HiDPI screens, by definition we are creating a UI-drawing issue for "normal" screens on Windows and all screens on any other platform. Currently, other `Qt` applications have similar issues, such as [Mendeley](https://www.mendeley.com/):

![image](https://user-images.githubusercontent.com/11596858/31428752-336a80a6-ae6c-11e7-91ab-a1e5cd53d173.png)

# When will this finally be fixed?
We plan to move to a newer version of `Qt` in November. This brings additional support for HiDPI displays: [blog](http://blog.qt.io/blog/2016/01/26/high-dpi-support-in-qt-5-6/), [reference manual](http://doc.qt.io/qt-5/highdpi.html). We will then start to investigate these tools to see if we can create an elegant global solution for this problem. We hope to fix this in 2018.

# What can I do? / Workaround
There are two options to work around these issues:
1. __Deal with it__. You'll have to resize windows and accept some funky-looking user interfaces. We strive to make sure that at least all the functionality is in a _useable_ state. That means that after resizing, at least all the labels are legible / understandable.
[...]

Issue #2047, opened 2017-10-10
OSF login box


EJWagenmakers:
On my (infamous) laptop with high-res screen, the OSF login box looks (much) worse than it did before. See screenshot.
E.J.
![osfloginscreen](https://user-images.githubusercontent.com/8603015/31414105-5170b52a-adea-11e7-8da0-078d2486a55d.png)

boutinb:
Duplicate can be closed

Issue #2046, opened 2017-10-10
About Box on High Res Screen


EJWagenmakers:
See attach. On my infamous laptop, the About box is a bit small, and I cannot resize to make the text more readable.
E.J.
![aboutboxabitsmallonhighresscreen](https://user-images.githubusercontent.com/8603015/31413937-9e045820-ade9-11e7-8da0-f1e72decd57f.png)

boutinb:
Duplicate can be closed

Issue #2045, opened 2017-10-10
Preference Box on High Res Screen


EJWagenmakers:
On my infamous laptop screen, the preference box looks suboptimal (see screenshot). However, I can fix that by simply resizing the box. 
1. It would be nice if it looked good without having to resize;
2. The boxes for "OK" and "Cancel" are slightly too small, and I can't resize those.
E.J.
![prefboxonhighresscreen](https://user-images.githubusercontent.com/8603015/31413883-6e254952-ade9-11e7-8608-c870e577ba3c.png)

boutinb:
Duplicate can be closed

PR #2308, opened 2018-02-14
Support hi-dpi on windows

vankesteren:
__What is this?__
This PR finally grants support for hi-dpi (usually high resolution laptop displays) on windows. 🎈 🎈 🎈 

❗️ __NB: This branch needs to be built using Qt 5.10.1 or higher__ ❗️ 
Qt 5.10 results in a bug where the webengine windows will have enormous scroll bars.
Testing still needed on: 
- [x] mac
- [ ] linux

__Why is this even necessary?__
 :+1: This branch fixes the long-standing and often-documented issues with drawing JASP on a high-dpi screen in windows: #2049, #2047, #2046, #2045, #2027, #1234 and #1212.

__What's next?__
Unfortunately, drawing issues are not completely solved yet: dragging the window from a HIDPI display to a lower dpi display will result in a black bar on the left and top of the main window. The reverse is true for the opposite operation. This holds for both Qt 5.10.0 and Qt 5.10.1. Issue #2322.

### DIFF
--- a/JASP-Desktop/main.cpp
+++ b/JASP-Desktop/main.cpp
@@ -27,6 +27,7 @@
 
 int main(int argc, char *argv[])
 {
+	QCoreApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
 	QCoreApplication::setOrganizationName("JASP");
 	QCoreApplication::setOrganizationDomain("jasp-stats.org");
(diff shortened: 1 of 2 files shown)


AlexanderLyNL:
I responded with a heart, because today is the holiest of all days

PR #2342, opened 2018-02-15
Fix loop bug

boutinb:


boutinb:
@akashrajkn Thanks, I have changed this function also.

PR #2358, opened 2018-02-20
JASPTools run does not work with Qt5.10

Janto91:


vankesteren:
Oh great! I saw this problem too, nice that you managed to fix this 👍

PR #2374, opened 2018-02-26
Added functionality to pool across RM factors for descriptive plots

### LINKED ISSUE #2368: To narrow confidence and credible intervals for main effects in repeated measures anova
OS name and version: Windows 10 Education
Analysis: Repeated measures ANOVA/ Baysian repreated measures ANOVA with between subject factor

Hi,

I suspect that Jasp is producing to narrow confidence and credible intervals for main effects of a between variable in repeated measures ANOVA and BANOVA. I have a design with 35 subjects (16 in one between subject condition 19 in the other). These subjects have each made 28 different judgements within-participants. A repeated measures ANOVA shows that the main effect of condition (between subject variable) is non-significant, however when I produce a graph over the main effect the confidence intervals are not overlapping, showing a clear significant effect. When I compute the same analysis in another software (STATISTICA) I get the same F and P-value, but the confidence intervals are clearly overlapping. This is expected with non-significant results. Similarily the credible intervals for the main effect of condition when computing a Baysian repeated measure anova look similar to the confidence intervals produced in Jasp, which makes me suspect that they are to narrow as well. 

I couldn't find a post on any bug regarding confidence and credible intervals. Is this a known issue? Are there any quick fixes?

Best,
August

Edit, here are the stats and the graph over the mean of each between condition with 95% confidence interval in JASP and STATISTICA. I have edited the STATISTICA graph so that the Y-axis is the same. The conditions are "Kvant" and "Kval" and they are placed on different sides (sorry for this). The mean values are the same, the only difference are the confidence intervals.
image
image

JohnnyDoorn:
Fix #2368
Improved indexing for omega calculation for interactions

### DIFF
--- a/JASP-Engine/JASP/R/anovarepeatedmeasures.R
+++ b/JASP-Engine/JASP/R/anovarepeatedmeasures.R
@@ -99,8 +99,8 @@ AnovaRepeatedMeasures <- function(dataset=NULL, options, perform="run", callback
 		}
 		
 		if (is.list(diff) && diff[['withinModelTerms']] == FALSE && diff[['betweenModelTerms']] == FALSE && diff[['repeatedMeasuresCells']] == FALSE &&
-		    diff[['repeatedMeasuresFactors']] == FALSE && diff[['simpleFactor']] == FALSE && diff[['moderatorFactorOne']] == FALSE &&
-		    diff[['moderatorFactorTwo']] == FALSE && diff[['poolErrorTermSimpleEffects']] == FALSE ) {
+		    diff[['repeatedMeasuresFactors']] == FALSE && diff[['sumOfSquares']] == FALSE && diff[['simpleFactor']] == FALSE && 
+		    diff[['moderatorFactorOne']] == FALSE && diff[['moderatorFactorTwo']] == FALSE && diff[['poolErrorTermSimpleEffects']] == FALSE ) {
 		  
 		  # old simple effects tables can be used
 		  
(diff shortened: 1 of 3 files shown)


JASP-Engine/JASP/R/anovarepeatedmeasures.R
```diff
@@ -2580,7 +2586,7 @@ AnovaRepeatedMeasures <- function(dataset=NULL, options, perform="run", callback
   
   simpleEffectsTable[["schema"]] <- list(fields=fields)
   
-  if (perform == "run" && status$ready && status$error == FALSE)  {
+  if (perform == "run" && status$ready && status$error == FALSE && class(fullModel) != "try-error")  {
```
AlexanderLyNL: I've made a funciton isTryError in common.R, so you might be interested to implement this as: 

!isTryError(fullModel)

JASP-Engine/JASP/R/anovarepeatedmeasures.R
```diff
@@ -2537,10 +2540,13 @@ AnovaRepeatedMeasures <- function(dataset=NULL, options, perform="run", callback
...
 
-.rmAnovaSimpleEffects <- function(dataset, options, perform, fullModel, fullAnovaTableWithin, fullAnovaTableBetween, status, singular, stateSimpleEffects) {
+.rmAnovaSimpleEffects <- function(dataset, options, perform, fullModel, fullAnovaTableWithin, 
+                                  fullAnovaTableBetween, status, singular, stateSimpleEffects) {
   
-  if (identical(options$simpleFactor, "") | identical(options$moderatorFactorOne, ""))
+  if (identical(options$simpleFactor, "") | identical(options$moderatorFactorOne, "")) {
     return (list(result=NULL, status=status))
+  }
+
```
AlexanderLyNL: Perhaps here is a good point to add the check for "Type 1" and return

JohnnyDoorn: I tried that, but it causes the whole ANOVA to crash - all the more reason for a rewrite!

PR #2380, opened 2018-03-01
Resize plot with JASP file

boutinb:
Cannot Resize plot (or saveAs action) when loading a JASP file without
updating it.


JASP-Engine/JASP/R/common.R
```diff
@@ -172,6 +172,27 @@ run <- function(name, title, dataKey, options, resultsMeta, stateKey, requiresIn
...
+		}
+	}
+	
+	if (base::exists(".requestTempRootNameNative")) {
+		paths <- .fromRCPP(".requestTempRootNameNative")
+		root = paths$root
+		base::Encoding(root) <- "UTF-8"
+		setwd(root)
+	} else {
+		print("Could not set the working directory!")
```
vandenman: When does this occur? This means the working directory cannot be set. That would be fatal for every subsequent analysis/ action involving saving to or reading from the state. Shouldn't this else statement return something to make jasp crash, or give a warning message?

boutinb: This "else" is actually not needed. If the function cannot be found, JASP would crash anyway.

Issue #2381, opened 2018-03-01
One Sample T-Test: Possible error in the estimation of the 95% CI for Mean Difference


rhileno:
## Template for bug reports

* JASP version: v. 0.8.6
* OS name and version: Windows 10 Home
* Analysis: One Sample T-Test - 95% CI for Mean Difference
* Steps to reproduce: We have detected a possible error in the One Sample T-Test. We believe that the estimation of the 95% CI for Mean Difference is incorrect (see picture). The CI would have to give -10.96 to -8.14. However, it gives 23.592 to 26.408. We believe that this confidence interval is not the 95% IC for Mean Difference, but that it is the 95% IC for Mean.

![one sample t-test - error ci for mean difference](https://user-images.githubusercontent.com/36966655/36861242-aa229a22-1d83-11e8-82a2-8a396ac512cc.png)

Janto91:
Thank you, @rhileno!

You are totally right! I fixed this issue and the correct version of the one sample t-test will be available in the next release of JASP!

AlexanderLyNL:
There's a difference between a confidence interval and a prediction interval. The confidence interval is an interval use to estimate and not necessarily to test, see 

https://arxiv.org/abs/1705.01064

Section 2.2 and Section 2.3

PR #2383, opened 2018-03-02
Fix incorrect confidence intervals for mean difference

Janto91:
Before, the confidence intervals for the mean difference did not take the test value intou account.

This fixes issue [#2381](https://github.com/jasp-stats/jasp-desktop/issues/2381).


JASP-Engine/JASP/R/ttestonesample.R
```diff
@@ -389,8 +389,8 @@ TTestOneSample <- function(dataset = NULL, options, perform = "run",
 					p <- as.numeric(r$p.value)
 					stat <- as.numeric(r$statistic)
 					m <- as.numeric(r$estimate - r$null.value)
-					ciLow <- .clean(as.numeric(r$conf.int[1]))
-					ciUp <- .clean(as.numeric(r$conf.int[2]))
+					ciLow <- .clean(as.numeric(r$conf.int[1] - r$null.value))
```
AlexanderLyNL: We have to discuss this properly first

Issue #2412, opened 2018-03-13
Frequentist RM ANOVA: Can't handle an object of class  “afex_aov” 


maciekszul:
Trying to do basic repeated measures ANOVA (frequentist), I get persistent problem, nonexistent when doing bayesian version. I use 2x2 design. Problem occurs even with different datafile, with the same dimensions.
datafile if required: https://mfr.osf.io/render?url=https%3A%2F%2Fosf.io%2Fp54d3%2Fdownload

> Error in ref.grid(object = structure(list(anova_table = structure(list(: Can't handle an object of class  “afex_aov” 
> Use help('models', package = 'lsmeans') for information on supported models.
> 
> Stack trace
> .referenceGrid(options, fullModel)
> 
> lsmeans::lsmeans(fullModel, formula)
> 
> lsmeans.formula(fullModel, formula)
> 
> lsmeans(object, .all.vars(specs), by = by, ...)
> 
> lsmeans.character(object, .all.vars(specs), by = by, ...)
> 
> lsmeans.character.default(object, .all.vars(specs), by = by, ...)
> 
> lsmeans.default(object, specs, ...)
> 
> do.call('ref.grid', rgargs)
> 
> ref.grid(object = list(anova_table = list(`num Df` = c(1, 1, 1), `den Df` = c(20, 20, 20), MSE = c(0.0108026413700208, 0.00347051711256845, 0.000732690559138553), F = c(0.0833405077745035, 292.708562291153, 1.54468907258066), ges = c(0.000514463193833267, 0.367406072632858, 0.000646654695436145), `Pr(>F)` = c(0.775791372315324, 2.07928260353455e-13, 0.228295544969146)), aov = list(`(Intercept)` = list(coefficients = 0.643998015870239, residuals = 0, effects = -5.90233931025348, rank = 1, fitted.values = -5.90233931025348, 
[...]

PR #2413, opened 2018-03-13
Created Unit Test for AnovaRepeatedMeasures

Kucharssim:


vankesteren:
Your new unit tests actually fail! Any idea why?

You can see the result [here](https://travis-ci.org/jasp-stats/jasp-desktop/builds/353090252#L1323)

Kucharssim:
The unit test does work on my computer.  
After consulting Bruno, it seems that the jasptools::run() cannot load other dataset than 'test.csv'. I'll try a workaround by creating the test datasets directly in the script - if it passes, at least we know it's because of this.

vandenman:
I tried the tests myself and they pass on windows but fail on Linux. After closer inspection of the error message, I got:


This analysis terminated unexpectedly.<br><br>Error in ref.grid(object = structure(list(anova_table = structure(list(: Can't handle an object of class  “afex_aov” <br> Use help('models', package = 'lsmeans') for information on supported models.<br><br>

This appears related to #2412, perhaps @JohnnyDoorn knows more about it? Anyway, the tests themselves look fine but I suggest putting this PR on hold until the other bug has been fixed.

JohnnyDoorn:
yay =)

Kucharssim:
Thanks, Johnny!

Let me revert to the version with loading the datasets - now it creates the data in the script, which makes the test slower.

Kucharssim:
Huh, now the RM ANOVA unit test works, but it somehow broke the principal component analysis? Everything is fine on Windows though, again...

JohnnyDoorn:
Hmm the PCA check also succeeds for me on Ubuntu 16.04. 
Here is the Travis output for the failed test:
test-principalcomponentanalysis.R:58: failure: Factor correlation table matches
`test` not equal to `ref`.
Length mismatch: comparison on first 2 components

JohnnyDoorn:
Try 2: yay??

PR #2418, opened 2018-03-14
Improved error checks for redundant variables

Janto91:


JASP-Engine/JASP/R/regressionlinear.R
```diff
@@ -455,6 +455,11 @@ RegressionLinear <- function(dataset=NULL, options, perform="run", callback=func
 
 	}
 
+	### check for errors
+	if (! is.null(independent.variables) && ! is.null(dependent.variable)) {
+	  errors <- .hasErrors(dataset, options = options, perform = perform, message = 'short', type = c('varCovData'), exitAnalysisIfErrors = TRUE, )
```
vandenman: here you could add:

`varCovData.weights = dataset[[options[["wlsWeights"]]`

also, the comma is not needed.

JASP-Engine/JASP/R/commonerrorcheck.R
```diff
@@ -542,10 +577,24 @@
     dataset <- split(dataset, groupingData)
   }
   for (d in seq_along(dataset)) {
-    cormat <- do.call(corFun, c(list(dataset[[d]]), corArgs))
+    
+    if (is.null(options$wlsWeights)) {
+      options$wlsWeights <- ""
+    }
+    
+    if (! options$wlsWeights == "") {
```
vandenman: by directly supplying weights in the argument, this can become:

    if (!is.null(weights)) {
      cov <- cov.wt(dataset[[d]], weights)
      ...
    } else {
      cormat <- do.call(corFun, c(list(dataset[[d]]), corArgs))
    }
    err <- .checkVarCovMatrix(cormat, nrow = FALSE, symm = FALSE)

that way, the check can still be used even if people don't store the weights in `options$wlsWeights`.

JASP-Engine/JASP/R/commonerrorcheck.R
```diff
@@ -489,12 +507,29 @@
...
+    vars <- .checkForPerfectCorrelations(dataset)
+    return(list(error=TRUE,reason="Matrix is not positive-definite", vars=vars))
   }
   
   return(list(error=FALSE, reason = ""))
 }
 
+.checkForPerfectCorrelations <- function(dataset) {
+
+  vars <- list()
```
vandenman: This function looks way to much like a C like function. Loops should be avoided over vectorized operations if possible. In addition, the code only checks for correlation of 1 whereas correlations of -1 should also be picked up.
Consider:

    # some data
    p <- 3
    dat <- matrix(rnorm(20*p), 20, p)
    dat <- cbind(dat, dat[, 1])
    cr <- cor(dat)

    # a method for finding the perfectly correlated pairs
    idx <- which(lower.tri(cr, diag = FALSE), arr.ind = TRUE) # the lower triangle of the matrix
    bad <- (1 - abs(cr[idx])) <= sqrt(.Machine$double.eps)    # check if correlations are awfully close to -1 or 1
    idx[bad, ]                                                # extract perfectly correlated pairs


Whenever dealing with not strictly integer values and testing for equivalence, it's safer to test against `sqrt(.Machine$double.eps)` than against exact values. (Remember `.1 + .2 != .3`!)

JASP-Engine/JASP/R/commonerrorcheck.R
```diff
@@ -91,12 +91,25 @@
       message <- paste0(openingMsg, '<ul>', message)
     }
   }
+
+  # See if we should add the variables causing the error (so far only for perfect correlations).
+  if (type == "varCovData" && ! is.null(args$variables) && length(args$variables) > 0) {
+    message <- paste0(message, "<ul><li> Note: The following pairs of variables: ")
+    for (i in 1:length(args$variables)) {
```
vandenman: maybe:

    nPair <- length(args$variables)
    if (nPairs > 10) {
      msg<- sprintf("<ul><li> Note: There were %d pairs of variables perfectly correlated. The first 10 are: ", nPairs)
    } else {
      msg<- "<ul><li> Note: The following pairs of variables are perfectly correlated: "
    }
    message <- paste0(message, msg)
    for (i in seq_len(nPairs)) {
      ...
    }

that way, if somebody drags in many, many bad variables, jasp won't accidentally crash because of it.

JASP-Engine/JASP/R/commonerrorcheck.R
```diff
@@ -91,6 +91,27 @@
       message <- paste0(openingMsg, '<ul>', message)
     }
   }
+
+  # See if we should add the variables causing the error (so far only for perfect correlations).
+  if (type == "varCovData" && ! is.null(args$variables)) {
+    nPairs <- length(args$variables$var1) # Could take either var1 or var2 as length is identical
+    if (nPairs > 0) {
```
vandenman: Question: can `nPairs` even have length 0? Because you already check if `args$variables` is`NULL`.

JASP-Engine/JASP/R/commonerrorcheck.R
```diff
@@ -489,12 +515,26 @@
...
+    return(list(error=TRUE,reason="Matrix is not positive-definite", vars=vars))
   }
   
   return(list(error=FALSE, reason = ""))
 }
 
+.checkForPerfectCorrelations <- function(dataset) {
+  
+  # if not already a correlation matrix
+  if (! any(diag(dataset) == 1)) {
```
vandenman: shouldn't `any` be `all`? Or `any(diag(dataset) != 1)`

JASP-Engine/JASP/R/commonerrorcheck.R
Janto91: In the way it was programmed previously it could. But I can see how this is confusing. Therefore, I changed the code in the .checkForPerfectCorrelations function.

Issue #1927, opened 2017-08-24
Hide observed cell counts in contingency tables when checkbox is cleared


HorwoodP:
* JASP version: 0.8.2
* OS name and version:  Windows 10, version 1607

I've noticed that Observed cell counts are displayed in contingency table regardless of whether the Observed checkbox is ticked or not.

Would it be possible to suppress the display of Observed cell counts when another of the cell display options are ticked (e.g. Column Percentages) and the Observed box is not ticked?

TimKDJ:
Yeah, that makes sense.
What good is a checkbox otherwise.

boutinb:
@ASarafoglou Can you have a look to this problem?

HorwoodP:
Sorry - didn't mean to close off this issue!

Just wondering if this bug will get looked at soon, since I'd like start generating contingency tables without the observed counts being displayed.

Thanks very much.

EJWagenmakers:
Good point. Can we fix this please?

PR #2500, opened 2018-04-30
Conttablefix

ASarafoglou:
I removed the button for observed counts from UI and .json file. 
Contingency tables should always display them.

### DIFF
--- a/JASP-Desktop/analysisforms/Common/contingencytablesform.ui
+++ b/JASP-Desktop/analysisforms/Common/contingencytablesform.ui
@@ -758,7 +758,7 @@
             <string>Counts</string>
            </property>
            <layout class="QGridLayout" name="gridLayout_5">
-            <item row="3" column="0">
+            <item row="2" column="0">
              <widget class="QWidget" name="widget_12" native="true">
               <property name="maximumSize">
                <size>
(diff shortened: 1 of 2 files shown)


HorwoodP:
Will this fix mean that observed counts are always displayed in all contingency tables?
If so, that's the opposite of what I was asking for in issue #1927 where I wanted the option to clear the observed counts tick box so that the contingency table could just display percentages (as we do when reporting survey results). Thanks.

ASarafoglou:
Yes, now the contingency table always displays the observed counts —
displaying e.g. only expected counts don’t make sense. Put you have a
point. Maybe we could add the option to display observed proportions
instead of counts like we do in the multinomial test.

Cheers,
Alexandra

Op di 1 mei 2018 om 03:00 schreef HorwoodP <notifications@github.com>

HorwoodP:
Yes - an observed proportions option would be useful.
Should I log this as a separate issue?

Thanks
Phil

ASarafoglou:
Yes, that would be great!
Thanks.

Op di 1 mei 2018 om 06:35 schreef HorwoodP <notifications@github.com>

akashrajkn:
Looks good.

Issue #2389, opened 2018-03-05
Linear Regression - Stepwise Procedure for Redundant Covariates


Janto91:
* JASP version: 0.8.6
* OS name and version: macOS High Sierra 10.13.3
* Analysis: Linear Regression
* Steps to reproduce: 
1. DV = debCollin2
2. Covariates = debCollin1, debCollin3
3. Method = Stepwise or Forward

Bug: Error in if (minimumPvalue < options$steppingMethodCriteriaPEntry)

JASP File: [JASP-Bug.jasp.zip](https://github.com/jasp-stats/jasp-desktop/files/1780757/JASP-Bug.jasp.zip)

PR #2508, opened 2018-05-04
Deal with na for regression coefficients

Janto91:
For example, this can occur if two variables are perfectly correlated.


Janto91:
This should fix this [issue](https://github.com/jasp-stats/jasp-desktop/issues/2389).

PR #2513, opened 2018-05-08
updates to unit-tests

vandenman:
Fixes the failing BAS unit tests. This also means that the r package .zip should now contain BAS 1.5.0 (from CRAN).


vandenman:
to summarize the changes:

- now uses jsonlite instead of rjson
- `fromJSON` and `toJSON` are now separate functions, so that switching between json packages should be very easy in the future.
- jsonlite reads one element of json for the multinomial test differently (it doesn't simplify a list to a vector but instead keeps it as a list). @ASarafoglou can you confirm that this won't break anything?
- added some `if (file.exists)` so that the network analysis doesn't give any more warnings.

PR #2597, opened 2018-06-05
Fix Meta-Analysis

Janto91:
Fixes the issues that after refreshing the JASP Meta-Analysis file from the data library, JASP kept loading but nothing ever happened


Janto91:
It adds a new default options in case the usual assignment does not work. But yeah it is not a very nice fix.

PR #2603, opened 2018-06-06
Independent Bayesian T-test new state system

vandenman:
(For the next release)
Implemented Tim's state system for the independent samples bayesian t-test. The main goal is to improve the legibility of this analysis and easy the transition to jaspResults. For now I focused on the state system, mainly to remove the giant if statements. The remainder should be cleaned when jaspResults is implemented. If this analysis goes well I plan to rewrite the Bayesian one sample and paired samples as well.

Changes to the state system:
- Elements of the statekey can now be lists containing expressions and options names.
- State can now be specified for collections (sublists), by adding a list to the attribute "collectionKey" of the specific element of the stateKey.

Together,  these changes allow for specifying "state$allPlots depends option1, option2", while also specifying "state$allPlots$contNormal depends on `'contNormal' %in% options[['variables']]`". This is particularly relevant for the T-tests, where removing one variable means most of the state is still useful. @vankesteren can you check whether the state changes make sense? And if the documentation about these changes is understandable by itself?

Changes to independent samples t-test:

- Removed legacy code (old checks for no. factor levels that are now handled by .hasErrors).
- Instead of using many different numeric indices, now uses variable names for indices.

@JohnnyDoorn since you added stuff to this analysis last, can you check if I didn't break anything?

All suggestions/ feedback are welcome!

### DIFF
--- a/JASP-Engine/JASP/R/ttestbayesianindependentsamples.R
+++ b/JASP-Engine/JASP/R/ttestbayesianindependentsamples.R
@@ -15,8 +15,13 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 
-TTestBayesianIndependentSamples <- function(dataset=NULL, options, perform="run", callback=function(...) 0, ...) {
+TTestBayesianIndependentSamples <- function(dataset=NULL, options, perform="run", callback=function(...) 0, state = NULL, ...) {
   # 
+  print(perform)
+  print('str(state$descriptivesPlots, max.level = 3)')
+  print(str(state$descriptivesPlots, max.level = 3))
+  print('str(state, max.level = 3)')
+  print(str(state, max.level = 3))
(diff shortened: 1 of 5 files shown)


JASP-Engine/JASP/R/ttestbayesianindependentsamples.R
```diff
@@ -1535,3 +1131,29 @@ TTestBayesianIndependentSamples <- function(dataset=NULL, options, perform="run"
                (priorDensZeroPoint / corFactorPrior) / (densZeroPoint / corFactorPosterior))
   return(bf)
 }
+
+.recodeBFtype <- function(bfOld, newBFtype, oldBFtype) {
```
akashrajkn: Can you move this in `common`? It can be used by all other analyses.

vankesteren:
will be jaspresults

PR #2616, opened 2018-06-11
bugs fix for one sample t-test

koenderks:
Bugs fixed:

- There is a weird note under the table reading: All tests, hypothesis is population mean is different from [number]. 
- "to" must be a finite number

### DIFF
--- a/JASP-Engine/JASP/R/ttestbayesianonesample.R
+++ b/JASP-Engine/JASP/R/ttestbayesianonesample.R
@@ -164,7 +164,7 @@ TTestBayesianOneSample <- function(dataset=NULL, options, perform="run", callbac
 
     if (options$testValue != 0) {
 
-      message <- paste("All tests, hypothesis is population mean is different from ", options$testValue, sep="")
+      message <- paste("For all tests, the alternative hypothesis specifies that the population mean is different from ", options$testValue,".", sep="")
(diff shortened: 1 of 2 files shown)


JASP-Engine/JASP/R/ttestbayesianonesample.R
```diff
@@ -164,7 +164,7 @@ TTestBayesianOneSample <- function(dataset=NULL, options, perform="run", callbac
 
     if (options$testValue != 0) {
 
-      message <- paste("All tests, hypothesis is population mean is different from ", options$testValue, sep="")
+      message <- paste("For all tests, the alternative hypothesis specifies that the population mean is different from ", options$testValue,".", sep="")
```
AlexanderLyNL: Just for parsimony, use paste0, then you don't need sep=""

Resources/Library/TTestBayesianOneSample.json
```diff
@@ -38,7 +38,9 @@
 			"format": "",
 			"name": "testValue",
 			"type": "Number",
-			"value": 0.0
+			"value": 0.0,
+			"max": 10000,
+			"min": -10000
```
AlexanderLyNL: Did you also checked this for other data sets? The variables in debug?

koenderks: yes I did

PR #2617, opened 2018-06-11
String Factors

Janto91:
Fix such that string factors can be added to factors fields


Janto91:
Good call! I made the same fix for these analyses.

Janto91:
Done

PR #2619, opened 2018-06-11
bugfix: base64 capitals in output table

vankesteren:
thanks @Kucharssim


JASP-Engine/JASP/R/semsimple.R
```diff
@@ -612,6 +602,18 @@ SEMSimple <- function(dataset=NULL, options, perform="run", callback=function(..
         sem_modind <- sem_modind[sem_modind$mi > options$outputModificationIndicesHideLowIndicesThreshold, , drop=FALSE]
       }
 
+      # Converting base 64 to normal
```
AlexanderLyNL: We typically remove the print statements, unless you think that this is a case that it's necessary.

vankesteren: Oopsie. Fixed

Issue #2557, opened 2018-05-22
anova he analysis terminated using post hoc 


dagiovanni:
## Template for bug reports

* JASP version: 0.8.6.
* OS name and version: Windows 7
* Analysis: Performing a post hoc with tukey method using Anova. The analysis terminated. 
![jaspbuganovaposthoctukey](https://user-images.githubusercontent.com/34913912/40356031-449184fa-5db8-11e8-9dc7-eee6cd071ae7.JPG)

i used this file.
ist is a spss file. Attached an zipped.
[Bugreport.zip](https://github.com/jasp-stats/jasp-desktop/files/2026065/Bugreport.zip)

* Steps to reproduce: 
load data, use anova, dv is chen, av is einkommenneu. use post hoc. take einkommenneu in the left window. 

## Template for enhancement requests
 
* Enhancement:
* Purpose:

JohnnyDoorn:
Hi,
I just ran your analysis in the latest version of JASP (to be released within 2 weeks) and it works (see attached screenshot)
![posthocfail](https://user-images.githubusercontent.com/15704203/40667563-ca4b51ac-6362-11e8-9cb3-72bb540a7c5f.png)
, so along the way your bug must have been fixed. Thanks for reporting it though, and please let me know if this issue persists in the latest JASP version for you (the version will be 0.9).
Kind regards
Johnny

JohnnyDoorn:
Hi, 
We looked into this, and the error is because of an umlaut in your level names. If you remove the umlaut, your post-hoc test will run! For me it worked at first because I first converted your data file to a csv, which also gets rid of the level names. 
Of course this is not ideal behavior, and we will look into better handling of umlauts in variable level names!
Kind regards,
Johnny

boutinb:
Hi,
I have checked a bit further. The problem is not due to the umlaut, but because the Einkommenneu column has labels (Neun and Zehn) for values (9 and 10) that does not occur in the dataset. Apparently the Post Hoc test does not accepts labels (factors) that have no data.
@JohnnyDoorn Can you have a look to this problem?

JohnnyDoorn:
Yes, I fixed it - will push the fix after the next release (together with other, still to be tested, ANOVA features).
For the moment, converting the data to csv will fix the issue, because that removes labels that are not used.

PR #2620, opened 2018-06-12
Fix issue #2557

AlexanderLyNL:
The problem was that the post hoc test cannot handle factors without
counts.


AlexanderLyNL:
Just to clarify: Kudos to @JohnnyDoorn, because this is not mine fix.

PR #2628, opened 2018-06-15
FixSuggestedVariableTypes

Janto91:
Add that all variable types are suggested for descriptives
Add that scale is also a suggested variable type for reliability analyses

### DIFF
--- a/JASP-Desktop/analysisforms/Common/descriptivesform.cpp
+++ b/JASP-Desktop/analysisforms/Common/descriptivesform.cpp
@@ -34,6 +34,8 @@ DescriptivesForm::DescriptivesForm(QWidget *parent) :
 
 	TableModelVariablesAssigned *model = new TableModelVariablesAssigned(this);
 	model->setSource(&_availableVariablesModel);
+    model->setVariableTypesSuggested(Column::ColumnTypeNominal | Column::ColumnTypeOrdinal | Column::ColumnTypeScale);
+    model->setVariableTypesAllowed(Column::ColumnTypeNominal | Column::ColumnTypeOrdinal | Column::ColumnTypeScale | Column::ColumnTypeNominalText);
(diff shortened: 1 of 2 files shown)


JASP-Desktop/analysisforms/Common/descriptivesform.cpp
```diff
@@ -34,6 +34,8 @@ DescriptivesForm::DescriptivesForm(QWidget *parent) :
 
 	TableModelVariablesAssigned *model = new TableModelVariablesAssigned(this);
 	model->setSource(&_availableVariablesModel);
+    model->setVariableTypesSuggested(Column::ColumnTypeNominal | Column::ColumnTypeOrdinal | Column::ColumnTypeScale);
+    model->setVariableTypesAllowed(Column::ColumnTypeNominal | Column::ColumnTypeOrdinal | Column::ColumnTypeScale | Column::ColumnTypeNominalText);
```
AlexanderLyNL: We should allow any type to be added in descriptives. This doesn't change that, does it?

Janto91: There is one other type not covered by this syntax, namely type "unknown". Do you want this type to be allowed as well?

AlexanderLyNL: Can you make a type that's unknown and see how JASP respond to it? I think it's a safeguard, but I'm curious and would like to see how it reacts. Otherwise, this pull request is great.

Janto91: Bruno explained to me that it is not possible to have data type unknown. The type is there for sth else.

I simplified the code by deleting the allowed line for descriptives because the default is that everything is allowed anyway.

PR #2654, opened 2018-06-23
New ANOVA features

JohnnyDoorn:
Several features:
- Brown-Forsythe and Welch AN(C)OVA
- Show df for contrast analysis
- Option to not assume equal variances in contrast analysis (AN(C)OVA)
- Option to add confidence intervals to post-hoc tests
- Dunn, Dunnett, Games-Howell post-hoc tests\
- New state system for ANOVA
- Rewrote standard post-hoc function to play better with the new state system
- Added unit tests for Friedman, Kruskal-Wallis tests, updated unit tests for posthoc/contrast analyses


JohnnyDoorn:
Hey @Janto91 
Thanks for taking a look. About the differing results - I checked this with R-packages, and the results do line up:
```{r}
library(userfriendlyscience)
library(DescTools)
mydat <- read.csv("/home/johnny/Downloads/fieldFeaturesCheck/comp1.csv")
oneway(mydat$Dose, y = mydat$Happiness, posthoc = 'games-howell')
DunnettTest(mydat$Happiness ~ mydat$Dose)
```
So maybe SPSS uses some different correction? Note that the p-values here are already correcting for multiple comparisons.
I still need to look into the post-hoc confidence intervals - the way I've done it is to use the confint function on the glht object from the multcomp-package, but here might be some stuff going on under the hood that is different from SPSS.
As for the other points, I incorporated them and will be in my next commit!
Cheers,
Johnny

JohnnyDoorn:
Hi @Janto91 ,
I updated the code:
- Added confidence intervals for Dunnett post hoc and contrast analysis
- Improved df for unequal variances contrasts
- Improved SE for Games-Howell
- The Dunnett discrepancy is because Field does it one-sided (todo: add functionality to do one-sided posthoc tests)
- Fixed titles in table
- Confidence intervals for post hoc tests corresponds to Fields R book - we use the multcomp package, which slightly deviates from what SPSS does.
Cheers,
Johnny

Issue #2669, opened 2018-07-02
Rendering on windows 7


oliats:
- JASP 0.9.XX
- WINDOWS 7 64bit
- Intel G45/G43 Express Chipset

When using Jasp 0.9 on a desktop running windows 7 /64bit I have a problem in rendering such as the one described here: https://forum.qt.io/topic/65435/qt-qml-renderig-bad-quality-windows/3
The bad rendering is only in the data window, the tools window and the reports window work good.
The video card drivers are upgraded, so I think that the reason of the problem is similar to that one described in the forum.qt.io.
With Jasp 0.8.6 everything works good.
I think it is not a bug, because on a laptop also running windows 7/64 bit everything works good.
I would like to try to solve the problem, as suggested in the forum qt: "This is most likely a problem with too old GPU / OpenGL driver. You can try to force your application to use DirectX by adding the following to your main.cpp: "// Options on Windows; set before instantiating QApplication QCoreApplication::setAttribute( Qt::AA_UseSoftwareOpenGL );" , but I don't know in witch file in the Jasp script I should add the string (in the Jasp script I cannot find any 'main.cpp' file).
I'm not a programmer and I don't know much about it.
Could anyone help me please?
Thank You very much
Oliviero

vankesteren:
Hi @oliats, thank you for the report! Especially the qt forum link is helpful. You cannot do anything about this from your end, but we might be able to fix this.

@JorisGoosen @boutinb this seems a relatively quick fix. The problem would be in identifying for which platforms this would need to be activated?

oliats:
Thank You very much.
I have the problem on Windows 7 64bit
JASP 0.9.0.1 for windows (both 32 and 64 bit)
Video card: Intel G45/G43 Express Chipset

oliats:
Thank you very much! Version 0.9.1.0 works good!

PR #2692, opened 2018-07-10
Fixes for syncing data, options/anova-crash and #2669

JorisGoosen:


JorisGoosen:
Fix for #2669 is conjecture

PR #2703, opened 2018-07-17
Rewrote BinomialTest with jaspResults

Janto91:


vankesteren:
I will use your edits to write the binomial test as an example analysis for the new style!

PR #2714, opened 2018-07-24
Rewrote TTestIndependentSamples with jaspResults

Janto91:
I also added more effect sizes and changed the computation of them via packages (after consultation with E.J. and Simon).
I also added an option for the user to determine which standard deviation should be used for computing Cohen's d and Hedges' g.

### DIFF
--- a/JASP-Engine/JASP/R/ttestindependentsamples.R
+++ b/JASP-Engine/JASP/R/ttestindependentsamples.R
@@ -15,822 +15,842 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 
-TTestIndependentSamples <- function(dataset = NULL, options, perform = "run",
-									callback = function(...) 0, ...) {
-	
-	state <- .retrieveState()
-	
-	## call the common initialization function
-	init <- .initializeTTest(dataset, options, perform, type = "independent-samples")
+TTestIndependentSamples <- function(jaspResults, dataset, options, state = NULL) {
+  
+  # Update options
+  if (options$hypothesis == "groupOneGreater") {
+    options$hypothesis <- "greater"
+  } else if (options$hypothesis == "groupTwoGreater") {
+    options$hypothesis <- "less"
+  } else {
+    options$hypothesis <- "two.sided"
+  }
+  
+  # Define state if empty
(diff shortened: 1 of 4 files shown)


Janto91:
@Kucharssim: Could you take a look at unit tests for the effect sizes? I compared the results for Cohen's d to the ones reported in Field (Chapter 4) but, so far, we do not have unit tests for Hedges' g and Cliff's delta and the corresponding confidence intervals.

JASP-Engine/JASP/R/ttestindependentsamples.R
```diff
@@ -15,822 +15,857 @@
...
+                                                           type = "number",          format = "sf:4;dp:3",
+                                                           overtitle = paste0(100*options$effectSizeConfidenceIntervalPercent, "% CI for Hedges' g"))
+    }
+  }
+  
+  # Fill up table with results
+  .fillUpIndependentSamplesTTestTable(independentSamplesTTestTable = independentSamplesTTestTableParametric, 
+                                      dataset = dataset, options = options, parametric = TRUE, ready = ready)
+  
+  return(NULL)
```
koenderks: I don't think you have to return anything in these jaspResult functions

JASP-Engine/JASP/R/ttestindependentsamples.R
```diff
@@ -15,822 +15,857 @@
...
+                                                              type = "number",             format = "sf:4;dp:3",
+                                                              overtitle = paste0(100*options$effectSizeConfidenceIntervalPercent, "% CI for Cliff's Delta"))
+    }
+  }
+  
+  # Fill up table with results
+  .fillUpIndependentSamplesTTestTable(independentSamplesTTestTable = independentSamplesTTestTableNonParametric, 
+                                      dataset = dataset, options = options, parametric = FALSE, ready = ready)
+  
+  return(NULL)
```
koenderks: same here (and for all other functions)

JohnnyDoorn:
What's the status on this? I would like to add some stuff to the t-test but want to wait for this rewrite.

Janto91:
I have rewritten this with jaspResults but not following the newest template. I can try to do this on Sunday.

PR #2730, opened 2018-08-14
updated jasptools for jaspResults

vandenman:
This PR updates jasptools to be compatible with jaspResults. Unit tests should work again,  they did locally, so hopefully travis is happy.

` resultsMeta <- rjson::fromJSON(resultsMeta)` is deleted from `runJaspResults()` because it's not an argument of the function (I discussed this with @JorisGoosen).

This doesn't mean the unit-tests automagically work though. Some restructuring might be necessary because the json outputted by jaspResults can look slightly different than before (e.g., different ordering) and because different internal names may have been used by programmers rewriting the R analyses. 

The binary files are compiled on Windows 10 with R 3.4.4, if this needs to be different, let me know. To see the source changes, see [this PR](https://github.com/jasp-stats/jasptools/pull/1).


vandenman:
Yeah, travis still isn't happy. Right now because I didn't adjust the travis script to install jaspResults. I suspect this will require some more tinkering. I'll drop by once I think it works properly again.

PR #2745, opened 2018-08-23
Make Travis more intelligent

vandenman:
installs car using `devtools::install_version` because the maintainers of car incorrectly specified that the package depends on R (> 3.2.0). In reality, car depends on R (> 3.5.0).

I'd still like to do some other testing on travis, so please do not merge yet.

### DIFF
--- a/Tools/requiredPackages.R
+++ b/Tools/requiredPackages.R
@@ -29,34 +29,46 @@ if (length(args) == 0) {
 }
 
 installed <- installed.packages()
+INSTALL_opts <- c("--no-docs", "--no-html", "--no-multiarch")
 options("repos" = "https://cloud.r-project.org")
 
 if (travis) {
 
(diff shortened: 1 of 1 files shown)


vandenman:
Travis can now distinguish between pull requests and pushes. In addition, it tracks which files have been modified. For example, see lines 1205 and 1207 of the Travis build. It then saves which .R files were modified (using `saveRDS(diff[endsWith(diff, ".R")], file = "modifiedRfiles.rds")`) such that these can be subject to static code analysis (which will be implemented later).

In addition, travis now prints a list of all installed packages and wether they were obtained from cache or installed (see line 1211 and onward).

Finally, `Tools/requiredPackages.R` creates the file `~/.R/Makevars` on Travis. This allows for specifing general compilation flags for packages. Since installing all the packages takes longer than running the tests, compilation flags are set to `-O0` for both C and C++ files.

PR #2765, opened 2018-08-31
add methods through .onAttach()

vandenman:
- adds missing methods to `jaspResults` through the `.onAttach()` function.
 - Added a function `destroyAllAllocatedRObjects` that should be called after `destroyAllAllocatedObjects()`. The latter destroys the c++ objects, making the pointers the R objects have invalid. The new function looks through the environment for any jasp objects with invalid pointers and deletes them.
 - modified the wrapper for `createJaspState` to use `create_cpp_jaspState` instead of `jaspResultsModule$create_cpp_jaspState`.


JorisGoosen:
Looks good, I'm merging it.

AlexanderLyNL:
Loving it

PR #2789, opened 2018-09-21
Fix distplot issues

vankesteren:
This pull request fixes (hopefully) 2 things:
1. Factors now get a barplot, not a distribution plot - as they should
2. Strange resolution with odd-size displays has been fixed

The resolution error still needs to be tested on mac @boutinb

### DIFF
--- a/JASP-Engine/JASP/R/common.R
+++ b/JASP-Engine/JASP/R/common.R
@@ -2247,9 +2247,10 @@ as.list.footnotes <- function(footnotes) {
   if (ggplot2::is.ggplot(plot)) {
 		ppi <- .fromRCPP(".ppi")
     ggplot2::ggsave(relativePathpng, plot, "png",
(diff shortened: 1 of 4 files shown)


AlexanderLyNL:
I've got some problems with the resolution. I tried both 72/96 and 52/96. Neither works on my mac 10.12.6

screen shot 2018-09-21 at 15 41 11

AlexanderLyNL:
The error only occurs for scale variables and not for factors.

vankesteren:
Let's talk about this wednesday?

AlexanderLyNL:
We can even do it today, if you have the time, because I'm at the UU now.

AlexanderLyNL:
Awesome works well

vankesteren:
Don't know yet why the PR checks fail; the error seems to be in the table but I did not change anything in the table 😒

PR #2811, opened 2018-10-01
fixed descriptives matrix plot

vandenman:
@JorisGoosen This should fix the issues with the plots.

Note that the matrix plots are a special kind of plot objects (a composition of ggplot objects).
They need their own print function, hence the changes in `writeImage.R` and `common.R`

### DIFF
--- a/JASP-R-Interface/jaspResults/R/writeImage.R
+++ b/JASP-R-Interface/jaspResults/R/writeImage.R
@@ -29,7 +29,7 @@ writeImageJaspResults <- function(width=320, height=320, plot, obj = TRUE, relat
   setwd(root)
   on.exit(setwd(oldwd))
 
-  if (ggplot2::is.ggplot(plot)) {
+  if (ggplot2::is.ggplot(plot) || inherits(plot, c("gtable", "ggMatrixplot", "JASPgraphs"))) {
     ppi <- .fromRCPP(".ppi")
 
     # fix for mac
@@ -63,8 +63,6 @@ writeImageJaspResults <- function(width=320, height=320, plot, obj = TRUE, relat
       .redrawPlot(plot) #(see below)
     } else if (inherits(plot, "qgraph")) {
       qgraph::plot.qgraph(plot)
-    } else if (inherits(plot, c("gtable", "ggMatrixplot", "JASPgraphs"))) {
-			gridExtra::grid.arrange(plot)
     } else {
       plot(plot)
     }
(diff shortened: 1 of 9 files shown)


AlexanderLyNL:
The fix is to convert the matrix plot to a ggplot, isn't it? If the background is a problem, would setEPS() in R fix this then?

vandenman:
Well since jaspResults cannot serialize any base plot, the current fix is to convert all of the base plots in descriptives to ggplots (basically implement Koen's work for descriptives only). The matrix plot is somewhat of a special case (it needs its own plot function, `gridExtra::grid.arrange()`. The current PR should fix both the issue Joris had as well as the one you showed me today.

AlexanderLyNL:
On my mac, the binomial as factor 1. doesn't show up and 2 crashes whenever I save as eps. Let's have a look at it tomorrow.

vandenman:
@AlexanderLyNL @JorisGoosen I think I solved the plotting, let's discuss this somewhere today.

JASP-R-Interface/jaspResults/R/writeImage.R
```diff
@@ -28,7 +28,7 @@ writeImageJaspResults <- function(width=320, height=320, plot, obj = TRUE, relat
   oldwd <- getwd()
   setwd(root)
   on.exit(setwd(oldwd))
-
+print("now in writeImageJaspResults")
```
AlexanderLyNL: Please remove the print statement.

vandenman: good catch!

JorisGoosen:
Well, the correlationplot shows up nicely now, even as eps. So when Alexander agrees I think we can merge this in?

I am working from home today, but we could discuss any open points here or through email? Or tomorrow.

vandenman:
There are actually three additional fixes aside from the correlationplot:

- Saving the density plot crashed jasp (this was still a base plot). I changed it to Koen's ggplot code.
- Both the Bayesian anova and ancova didn't have a keep specified. In other words, JASP would immediately delete the .png files of the descriptives plots before they could be shown.

I'm happy discussing the code/ open points anywehere.

JorisGoosen:
Well, this all looks fine to me so I've merged it in. 
The rest of the ggplots that @koenderks  will come in after the release of 0.9.1 right?

koenderks:
Yes, I think that is what is supposed to happen.

PR #2821, opened 2018-10-02
Shows descriptive plots Bayesian RM ANOVA

JohnnyDoorn:
Before you had to resize to show the plots.


AlexanderLyNL:
This was fixed by Don in https://github.com/jasp-stats/jasp-desktop/pull/2811  @vandenman a different implementation?

vandenman:
Nope I fixed the anova and ancova. Somehow this plot also needs the change. I vaguely remember that Tim made the keep redundant at one point, perhaps something changed in common and now it needs to be specified again? We should check other plots to make sure.

AlexanderLyNL:
I checked all the plots on my branch. The only one that doesn't work is Bayesian RM ANOVA, thus, once this is done we'll have all the plots covered.

AlexanderLyNL:
@vandenman Travis says that it's done, but github says Travis is still doing it's magic.

PR #2831, opened 2018-10-04
jaspHtml upgrade

vankesteren:
Copy-paste now works well in jasp html output.

_NB: also bumps engine version_


vankesteren:
This PR is not _necessary_ before the release

PR #2855, opened 2018-10-16
JASPgraphs updates

vandenman:
- now specifies `ggsave` in pixels rather than inches.
- changes to descriptives plots to improve matrix plot.
- some internal updates to JASPgraphs.


JASP-Engine/JASP/R/descriptives.R
```diff
@@ -533,19 +533,20 @@ Descriptives <- function(jaspResults, dataset, options, state=NULL)
 
 
   plotMat <- matrix(list(), l, l)
-
+  # scy <- ggplot2::scale_x_continuous(labels = scales::number)
```
vankesteren: These commented lines can be removed now right?

JASP-Engine/JASPgraphs/R/jaspScales.R
```diff
@@ -0,0 +1,66 @@
+#' @export
+scale_x_continuous <- function(name = waiver(), breaks = axesBreaks, minor_breaks = waiver(),
```
vankesteren: Ah so basically we overload the default scale functions

JASP-R-Interface/jaspResults/R/writeImage.R
```diff
@@ -29,23 +29,29 @@ writeImageJaspResults <- function(width=320, height=320, plot, obj = TRUE, relat
...
-    # fix for mac
-    if (Sys.info()["sysname"] == "Darwin") ppi <- ppi / 2
-
-    ggplot2::ggsave(relativePathpng, plot, "png",
-                    width  = 1.5*width/ppi,
-                    height = 1.5*height/ppi,
-                    dpi    = 2*ppi,
-                    bg     = "transparent")
+    pngMultip <- .fromRCPP(".ppi") / 96
+    ggplot2::ggsave(
```
vankesteren: I think this is a good idea, although it does (re)introduce the following 2 problems:
1. The density plot now has different line thickness than the bar plot in descriptives (should be easy to fix for this plot)
2. y axis label letters are slightly cut off on windows with 100% (default) scaling (look at the `n` and the `u` in `counts`; it's less prominent in this anti-aliased png screenshot than in JASP)

afbeelding

AlexanderLyNL:
Great work Don, it seems that the ppi issues are gone, even if I use JASP on a higher resolution monitor.

vankesteren:
@AlexanderLyNL fixed my issue: change the dpi to 300 in the preferences. Maybe it's an idea to change this by default to a higher resolution @boutinb?

PR #2874, opened 2018-10-24
Implement themeJASP in ggplots + convert non-ggplots to ggplots

koenderks:


vandenman:
everything looks good, I just need to check if the descriptives plots still work.

koenderks:
An update for the plot branch which solves all the before mentioned issues. Note that the Bayesian correlation matrix was not working because it was calling ggplot functions, I addded the base functions to keep it working.

AlexanderLyNL:
I noticed that the correlation matrix plot doesn't use Don's magical correlation plot from Descriptives

koenderks:
What do you mean exactly? The function .descriptivesMatrixPlot() can only be used in descriptives right?

AlexanderLyNL:
screen shot 2018-11-16 at 14 10 29

The matrix plot on the left is the one from your branch in "Correlation Matrix". On the right is Don's Correlation plot from "Descriptives". The formatting on the left would not please EJ I reckon.

koenderks:
image
Just like this then? We just transpose the matrix?

koenderks:
Also I don't know why your axes keep disappearing...I don't have this issue.

koenderks:
I see what you mean, the facFive plot was not good
image

AlexanderLyNL:
Sorry I wasn't clear, you don't need to transpose it, but it's more the @vandenman matrix structure that I was worried about, see 

https://github.com/jasp-stats/jasp-desktop/blob/9dee192a996fb84cfdb01ac4cdffce0b06d9708a/JASP-Engine/JASP/R/descriptives.R#L534

koenderks:
Okay I get what you mean, can you check if this is in line with the other structure?

AlexanderLyNL:
By the way, we still need "Counts" on the y-axis for factors when "Distribution plots" is chosen.

koenderks:
all adjusted.

AlexanderLyNL:
Great stuff!

PR #2898, opened 2018-11-07
Refactor binomial test for jaspResults

vankesteren:
This analysis can serve as an example for the new `jaspResults` analysis interface.

Note that the plot does not make use of the `JASPgraphs` package

makes #2703 obsolete


JASP-Engine/JASP/R/binomialtest.R
```diff
@@ -15,224 +15,315 @@
...
+.binomComputeResults <- function(jaspResults, dataset, options) {
+
+  # Take results from state if possible
+  if (!is.null(jaspResults[["stateBinomResults"]])) return(jaspResults[["stateBinomResults"]]$object)
+
+  # This will be the object that we fill with results
+  results <- list()
+
+  # First, we perform precalculation of variables we use throughout the analysis
+  results$spec <- .binomCalcSpecs(dataset, options)
```
vandenman: doesn't this mean that if a variable is called `spec` this will cause an error?

JASP-Engine/JASP/R/binomialtest.R
```diff
@@ -15,224 +15,315 @@
...
 
+  # Return results object
+  return(results)
+}
+
+.binomCalcSpecs <- function(dataset, options) {
+  specs <- list()
+
+  # Recode the hypothesis for binom.test()
+  if (options$hypothesis == "notEqualToTestValue") {
```
vandenman: probably something to deal with later, but I see this particular if statement in almost every analysis..

JASP-Engine/JASP/R/binomialtest.R
```diff
@@ -15,224 +15,315 @@
...
+  binomialTable$addColumnInfo(name = "total",      title = "Total",      type = "integer")
+  binomialTable$addColumnInfo(name = "proportion", title = "Proportion", type = "number", format = "sf:4")
+  binomialTable$addColumnInfo(name = "p",          title = "p",          type = "number", format = "dp:3;p:.001")
+
+  if (options$VovkSellkeMPR) {
+    binomialTable$addColumnInfo(name = "VovkSellkeMPR", title = "VS-MPR\u002A", type = "number", format = "sf:4")
+  }
+
+  if (options$confidenceInterval) {
+    binomialTable$addColumnInfo(name = "lowerCI", title = "Lower", type = "number", format = "sf:4",
```
vandenman: Perhaps
```suggestion
overtitle <- paste0(100 * options$confidenceIntervalInterval, "% CI for Proportion")
    binomialTable$addColumnInfo(name = "lowerCI", title = "Lower", type = "number", format = "sf:4", overtitle = overtitle)
        binomialTable$addColumnInfo(name = "upperCI", title = "Upper", type = "number", format = "sf:4", overtitle = overtitle)
```

JASP-Engine/JASP/R/binomialtest.R
```diff
@@ -15,224 +15,315 @@
...
+  }
 
+  # Add footnote: VovkSellkeMPR
+  if (options$VovkSellkeMPR) {
+    binomialTable$addFootnote(message = .messages("footnote", "VovkSellkeMPR"), symbol = "\u002A")
   }
 
-  return(binomResults)
+  # Add footnote: Alternative hypothesis
+  if (binomResults$spec$hypothesisRec == "two.sided") {
```
vandenman: Typically, I find a switch statement more readable:
```suggestion
  message <- switch(binomResults$spec$hypothesisRec,
    "two.sided" = .messages("footnote", "binomNeq",     value = options$testValue),
    "greater"   = .messages("footnote", "binomGreater", value = options$testValue),
    "less"      = .messages("footnote", "binomLess",    value = options$testValue)
  )
  binomialTable$addFootnote(message = message, symbol = "<em>Note.</em>")
```
When to use this is pretty much up to the developer, but since this should be an example analysis I figured I'd bring it up.

JASP-Engine/JASP/R/binomialtest.R
```diff
@@ -15,224 +15,315 @@
...
-	)
+}
 
-  return(list(results=results, status="complete", state=state))
+# Preprocessing functions ----
+.binomReadData <- function(dataset, options) {
+  if (!is.null(dataset)) {
+    return(dataset)
+  } else {
+    return(.readDataSetToEnd(columns.as.factor = options$variables))
```
vandenman: I'm a bit confused. You read the data in as factor, but have a lot of `as.factor` down the road. Does this mean that the reading of the data with `columns.as.factor` doesn't guarantee the data are factors? If yes, the `as.factor` calls down the road are superfluous.

JASP-Engine/JASP/R/binomialtest.R
```diff
@@ -15,224 +15,315 @@
...
-    } else {
-      hyp <- "less"
-    }
+  # Precompute the levels of the different variables
+  specs[["levels"]] <- NULL
+  for (variable in options$variables) {
+    column <- dataset[[.v(variable)]]
+    data <- column[!is.na(column)]
+    levels <- levels(as.factor(data))
+    specs[["levels"]] <- append(specs$levels, list(levels))
```
vandenman:
```suggestion
    specs[["levels"]] <- c(specs$levels, list(levels))
```

JASP-Engine/JASP/R/binomialtest.R
```diff
@@ -15,224 +15,315 @@
...
+  } else {
+    return(.readDataSetToEnd(columns.as.factor = options$variables))
+  }
 }
 
-.binomialTest <- function(dataset, options, perform, variables) {
+.binomCheckErrors <- function(dataset, options) {
+
+  # Check if results can be computed
+  if (length(options$variables) == 0) .quitAnalysis("Input at least one variable")
```
vandenman: This causes the default behavior to return an error if the user does not load any variables. We could make this the default, all analyses start with an error message, but perhaps we should discuss this with EJ. I don't think right now this currently very user friendly. The least we could do is create a separate function for cases like this such that the user experience looks better (i.e., a jasp-green box telling them the analysis won't run until they enter a), b) c).

JASP-Engine/JASP/R/binomialtest.R
```diff
@@ -15,224 +15,315 @@
...
-	results[["binomial"]] <- .BinomialTable(binomResults, options)
+  # Compute the results
+  binomResults <- .binomComputeResults(jaspResults, dataset, options)
 
-	if (! is.null(descriptPlots)) {
-		results[["descriptives"]] <- descriptPlots
-		results[["descriptives"]][["title"]] <- "Descriptive Plots"
-	}
+  # Output tables and plots
+  .binomTableMain(jaspResults, dataset, options, binomResults)
```
vandenman: you don't want to align the arguments of these three functions? :P

vandenman:
Overall I'd say this looks really good! I have some suggestions here and there, but most of it is perfectionistic since this analysis is supposed to be an example for others.

vankesteren:
Thanks for the review don, I'll look into it now

JASP-Engine/JASP/R/binomialtest.R
vankesteren: Nope, results here is simply a list. Do you forsee any issues with this naming? of course, we can just as well change this.

JASP-Engine/JASP/R/binomialtest.R
vankesteren: yes, ideally this would just be passed to the options directly as-is. I had a short conversation about this with @JorisGoosen. I think the argument @Janto91 gave is that [the `json` does cannot handle names with periods in them as in `two.sided`](https://github.com/jasp-stats/jasp-desktop/blob/development/Docs/development/r-analyses-guide.md#step-12-updating-the-options). This seems strange and fixable though!

JASP-Engine/JASP/R/binomialtest.R
vankesteren: great suggestion

JASP-Engine/JASP/R/binomialtest.R
vankesteren: this is probably true and an artefact of developing in RStudio using the original dataset without it having gone through the data loading pipeline of JASP. I'll look into it.

JASP-Engine/JASP/R/binomialtest.R
vankesteren: I was unsure how to do it. If you have a suggestion I'll be happy to take it ;)

JASP-Engine/JASP/R/binomialtest.R
vankesteren: yes, good point, I thought I had already changed this. I wanted to change it back to displaying an empty main table. I'll have to look how I can do this while maintaining as much separation as possible between checking for errors and computing results.

JorisGoosen: Maybe just put a jaspHtml with a message stating that one should add some variables? And then return from the (sub)analysis?

JASP-Engine/JASP/R/binomialtest.R
JorisGoosen: It seems that this isn't actually a problem in the json, but in the UI files (for the analysis widgets) where you give the name of an option as name of a widget.

And you cannot put a dot in the name because it is converted to c++. 
We could make this better in QML though, maybe. Although @boutinb would be able to answer this more authoritatively.

JASP-Engine/JASP/R/binomialtest.R
vankesteren: I have solved it similarly to how it's done now: an empty main table.

JASP-Engine/JASP/R/binomialtest.R
vankesteren: Did this.

JASP-Engine/JASP/R/binomialtest.R
vankesteren: I have removed as.factor inside the functions

JASP-Engine/JASP/R/binomialtest.R
vandenman: Use compute columns to create a variable called `spec`. Run the analysis, observe that it crashes: 

```
Error in switch(binomResults$spec$hypothesisRec, two.sided = .messages('footnote', : EXPR must be a length 1 vector
```

The problem is that a bit lower you call ` results[[variable]] <- list()` which in this scenario overwrites the information in `results[["spec"]]`.

vankesteren: Thanks for noticing this! I've fixed it in the most recent push. I added an extra level to the binomResults list so now it looks somewhat like so:

```
binomResults
├─ spec
|  ├─ hypothesisRec
|  └─ levels
|      └─ ...
└─ binom
   ├─ variable1
   |   └─ ...
   └─ variable2
       └─ ...

```

JASP-Engine/JASP/R/binomialtest.R
```diff
@@ -15,199 +15,327 @@
...
+  )
 
-    d <- dataset[[.v(var)]]
-    d <- d[!is.na(d)]
-		nObs <- length(d)
+  binomialTable$addFootnote(message = message, symbol = "<em>Note.</em>")
+}
 
-		descriptivesPlotList <- list()
+.binomEmptyMainTable <- function(tab) {
```
vandenman: Found by @ASarafoglou, 

Should this be a separate function? It's only used once, and the 9 is hardcoded.

JorisGoosen: How about: https://github.com/jasp-stats/INTERNAL-jasp/issues/80

JorisGoosen:
Can this be merged?

PR #2923, opened 2018-11-14
Compute column functions

AlexanderLyNL:


AlexanderLyNL:
I only have one issue with character strings though.

vandenman:
I'm not exactly sure what the use is of `findDominatingClass()`. Directly using `useMethod()` didn't work? Also, we would now have `ifelse()` and `ifElse()` available? That might be confusing.

AlexanderLyNL:
The problem with normal ifelse is that it doesn't work on labels, but on the underlying values instead, see for instance 

https://www.theanalysisfactor.com/r-tutorial-recoding-values/

This will confuse users, which is why I made ifElse. I didn't know how to deal with the multiple types, which is why I made findDominatingClass to deal with this. This function is more generic, so I thought why not keep it separate. 

The issue with character strings I mention above is concerned with users entering character strings in ifElse for a compute column in JASP, but JASP doesn't allow a text factor to be computed anymore. @JorisGoosen

vandenman:
I feel I'm missing something. In the example below `ifelse` and `ifElse` give me the exact same output:

```
a = letters[1:5]
ifElse(a == "a", 1, 2)
# [1] 1 2 2 2 2
ifelse(a == "a", 1, 2)
# [1] 1 2 2 2 2

a = factor(letters[1:5])
ifElse(a == "a", 1, 2)
# [1] 1 2 2 2 2
ifelse(a == "a", 1, 2)
# [1] 1 2 2 2 2
```
The link you send also uses `ifelse` directly on factor levels, without any issues.
Do you have minimal working example where they don't produce the same output?

AlexanderLyNL:
`kaas <- as.factor(5:9)

ifelse(kaas==5, 1, kaas)
ifElse(kaas==5, 1, kaas)`

and 

'bob <- as.factor(LETTERS[5:9])

ifelse(bob=="F", 1, bob)
ifElse(bob=="F", 1, bob)
'

JorisGoosen:
@AlexanderLyNL actually, jasp DOES allow a text factor to be computed in the current version.

PR #2932, opened 2018-11-19
Fixed base64 error in Meta Analysis

akashrajkn:
Fixes the jasp-stats/jasp-issues#248 and jasp-stats/jasp-issues#193

### DIFF
--- a/JASP-Engine/JASP/R/classicalmetaanalysis.R
+++ b/JASP-Engine/JASP/R/classicalmetaanalysis.R
@@ -127,12 +127,12 @@ ClassicalMetaAnalysis <- function(dataset=NULL, options, perform="run", callback
                exitAnalysisIfErrors=TRUE)
   }
 
-  method = switch(options$method, `Fixed Effects` = "FE", `Maximum Likelihood` = "ML",
+  method <- switch(options$method, `Fixed Effects` = "FE", `Maximum Likelihood` = "ML",
                   `Restricted ML` = "REML", `DerSimonian-Laird` = "DL", `Hedges` = "HE",
                   `Hunter-Schmidt` = "HS", `Sidik-Jonkman` = "SJ", `Empirical Bayes` = "EB",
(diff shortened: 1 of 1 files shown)


akashrajkn:
@vankesteren @AlexanderLyNL can you review and merge?

AlexanderLyNL:
A very quick check shows that = is used for assignment in a couple of case in the old code. Could you please lintr/style this? Secondly, the covariates and factors are displayed in the table as base 64 text. Lastly, the interaction doesn't seem to be taken into account. 

The screenshot is taken from the analysis of 
screen shot 2018-11-19 at 11 48 33

https://github.com/jasp-stats/jasp-issues/issues/248#event-1974274408

followed by a refresh and the additional input shown in the screenshot.

akashrajkn:
@AlexanderLyNL 

1. Fixed the base64 output in Coefficients table.
2. `<-` assignment is used instead of `=`. 

With respect to interaction: `options$modelTerms` gives the interactions, but the R-code does not seem to do anything with it. (I don't know how to handle the interactions for meta analysis.)

ma-fix
)

AlexanderLyNL:
Does the following help:

http://www.metafor-project.org/doku.php/tips:multiple_factors_interactions#model_with_interaction

Could you add that example as a unit test? If not, I'll look into it later today.

akashrajkn:
@AlexanderLyNL  Yes! That helps. I'll try to do it tomorrow. Cheers

akashrajkn:
@AlexanderLyNL In this dataset, the interaction between *Measurement* and *Timing* is basically equal to *Measurement* (when not NA). `rma` package automatically drops this with a warning message ` Redundant predictors dropped from the model.`

For other predictors, we can see the interaction, 
ma-fix

AlexanderLyNL:
Awesome, have you also tested the example that was in the link? And added that as a unit test?

PR #2933, opened 2018-11-19
Fix ANOVA issue with multiple covariates and no fixed factors

JohnnyDoorn:


JASP-Engine/JASP/R/ancova.R
```diff
@@ -856,7 +856,10 @@ Ancova <- function(dataset=NULL, options, perform="run", callback=function(...)
 				F <- if (is.na(result[term,"F value"])) {""} else { result[term, "F value"] }
 				p <- if (is.na(result[term,"Pr(>F)"] )) {""} else { result[term, "Pr(>F)"] }
 
-				if(i == 1 || term == "Residuals" || (!is.null(unlist(options$covariates)) && (terms.normal[i] == options$fixedFactors[[1]] || terms.normal[i] == options$covariates[[1]]) && 
+				firstFixedFactor <- ifelse(length(options$fixedFactors) == 0, "", options$fixedFactors[[1]])
```
AlexanderLyNL: Can you write this in an if-else statement. The ifelse construction can lead to some weird behaviour. That won't occur here though, but it's more a matter of style

AlexanderLyNL: In addition, could you add a unit test?

AlexanderLyNL:
Other than that the fix works like a charm.

JASP-Engine/JASP/R/ancova.R
JohnnyDoorn: What unit test? There is already one for ANCOVA right?

PR #2943, opened 2018-11-22
Strip comments from compute column code and filter code.

JorisGoosen:
Fixes jasp-stats/INTERNAL-jasp/issues/#72

Also disables the GUI for JASP when running unit-tests on a jasp-file recursively


JorisGoosen:
Actually fixes  #INTERNAL-jasp/issues/72

PR #2944, opened 2018-11-22
computeColumn call now executed within local environment

vandenman:
Compute column is now executed within an empty environment that is immediately deleted afterward. Functions/ objects that are not found are looked for one level higher, inside the global environment. I had to add the function `local()` to the whitelist.


JorisGoosen:
Not to be too nitpicky but you messed up the indentation in JASP-Engine/r_functionwhitelist.cpp
Would this also work for the filter? 
Because I think I can now still mess it up if I do the same thing there right?

vandenman:
Not nitpicky at all! The indentation uses tabs rather than spaces? if so, I'll change that. This should also work for the filter, I'll add that as well.

JorisGoosen:
Yeah, in cpp indents are 4-space tabs.

PR #2947, opened 2018-11-23
Stabilized JASP!

JorisGoosen:
Can now change datasize from small -> 1000000+ -> small ad infinitum and will not crash.
Although it CAN make your computer grind to a halt when you try to work with 100 of millions of rows because it doesn't play nice with many gigs of memory yet. But it wont crash.

Fixes issue https://github.com/jasp-stats/INTERNAL-jasp/issues/74


AlexanderLyNL:
w0000ttt

PR #3008, opened 2019-01-18
updates to jaspgraphs

vandenman:
As we discussed, this PR contains only the changes to JASPgraphs in #3003. I also changed order of the classes for vdiffr. I will update the other PR once this one has been merged into development.


JASP-Engine/JASPgraphs/R/parseThis.R
```diff
@@ -0,0 +1,40 @@
...
+#' @rdname parseThis
+#' @export
+parseThis <- function(x) {
+  UseMethod("parseThis", x)
+}
+
+#' @export
+parseThis.character <- function(x) {attr(x, "parse") <- TRUE; x}
+#' @export
+parseThis.factor    <- function(x) {attr(x, "parse") <- TRUE; x}
```
TimKDJ: Do you really need parseThis? It only seems to do something with types it doesn't actually mind.
E.g., for characters and factors it adds "parse" but then doesn't need to parse them, however for data.frames it doesn't add "parse" and then it does need to parse them.
Maybe I'm overlooking something

JASP-Engine/JASPgraphs/R/printJASPgraphs.R
```diff
@@ -0,0 +1,41 @@
...
+#' @method print JASPgraphs
+#' @export
+print.JASPgraphs <- function(x, ...) {
+
+  if (ggplot2::is.ggplot(x)) {
+    ggplot2:::print.ggplot(x, ...)
+  } else if (inherits(x, c("gtable", "gTree", "grob", "gDesc"))) {
+    gridExtra::grid.arrange(x, ...)
+  } else {
+    stop(sprintf(
```
TimKDJ: Does R support a way to call the next S3 method? Because this doesn't fail very gracefully, eventhough a different class might have an appropriate print method defined.

vandenman: what do you mean? We could do something like

```
class(x) <- class(x)[-1]
e <- try(
	print(x) # hope for the best
)
if (inherits(e, "try-error")) {
  stop("...")
} else {
  return(invisible(e))
}
```
instead? 

This function is only called whenever people use `print(jaspGraphsObject)` we (or I) better make sure it works for objects we've created.

vandenman: the function `NextMethod()`, immediately calls the next S3 method btw.

TimKDJ: I would say use `NextMethod()` given that `length(class(x)) > 1`. The error seems a bit overkill. Mainly thinking about future cases where 'random' people might work with our objects and  wish to extend them.
Up to you though, it's not that big of a deal.

PR #3010, opened 2019-01-18
Fixes to unit tests and jasptools

TimKDJ:
- Allow jasptools to parse qml files to easily generate an options list in R (`jasptools::analysisOptions("Anova")`)
- Visual regression testing the same way ggplot2 does it (with vdiffr, it should be much more stable now with the dependencies being statically linked)
- Allows jaspResults to create plots outside of the JASP environment


TimKDJ:
Okay so I'm finished here for now, but jaspResults causes a segment fault on Travis.
I deleted all caches and rebuild, so it definitely appears to be caused by the later modifications to jaspResults.

```
> library(jaspResults)
Loading required package: Rcpp
jaspResults version: 1.2
 *** caught segfault ***
address 0x7f1058a1a548, cause 'invalid permissions'
/home/travis/.travis/functions: line 104: 25617 Segmentation fault  
```

vandenman:
this could be caused by `create_cpp_jaspResults("Analysis Test")` which should be `create_cpp_jaspResults("Analysis Test", NULL)` (line 13 of zzzWrappers.R) and which is called in the `.onAttach()`.

PR #3012, opened 2019-01-22
added aliases

vandenman:
In Rstudio, only `?jaspContainer` shows the helpfile. Since people usually call `createJaspContainer`, they might get confused that `?createJaspContainer` does not exists. This PR redirects 
`?createJaspContainer` to `?jaspContainer`.

idem for `jaspState`, `jaspPlot`, `jaspTable`, and `jaspHtml`.


JorisGoosen:
Good idea!

PR #3025, opened 2019-01-31
unified the wrappers for jaspResults

vandenman:
Merges the two wrapper functions in common.R and zzzWrappers.R.

Key is the function `jaspResultsCalledFromJasp()`

If the jaspResultsModule object ever changes, this function needs to be adjusted accordingly. Currently the function checks if jaspResultsModule exists, is an S4 class, is of class "Module", and has only one slot called ".xData". It is possible to fool these checks, but I doubt this would happen on accident. If it ever comes up we could change the class of jaspResultsModule from "Module" to something more distinct.


JASP-R-Interface/JASP-R-Interface.pro
```diff
@@ -79,6 +79,7 @@ windows{
 
 ### making sure that writeImage.R is available to jaspEngine:
 SRC_WRITE_IMAGE = $${PWD}/jaspResults/R/writeImage.R
+SRC_WRITE_IMAGE = $${PWD}/jaspResults/R/zzzWrappers.R
```
JorisGoosen: This will overwrite SRC_WRITE_IMAGE (which should point to writeImage.R) with zzzWrappers.R. So that breaks the build, you just didn't see it because writeImage.R was still in your build-folder.

Mind you, I didn't test this because I'm writing it here in my browser it should work I think.

better would be to replace the relevant part by: 

```
### making sure that writeImage.R and zzzWrappers.R are available to jaspEngine:
SRC_WRITE_IMAGE = $${PWD}/jaspResults/R/writeImage.R
SRC_WRAPPERS = $${PWD}/jaspResults/R/zzzWrappers.R
DEST_DIR_AUX_R = $$OUT_PWD/$$DESTDIR/

auxillaryRFiles.path = $$INSTALLPATH
auxillaryRFiles.files = $${PWD}/jaspResults/R/writeImage.R
auxillaryRFiles.files = $${PWD}/jaspResults/R/zzzWrappers.R
INSTALLS += auxillaryRFiles

win32 {
    SRC_WRITE_IMAGE ~= s,/,\\,g
    SRC_WRAPPERS ~= s,/,\\,g
    DEST_DIR_AUX_R ~= s,/,\\,g

    copyRFiles.commands  += $$quote(cmd /c xcopy /S /I /Y $${SRC_WRITE_IMAGE} $${DEST_DIR_AUX_R});
    copyRFiles.commands  += $$quote(cmd /c xcopy /S /I /Y $${SRC_WRAPPERS} $${DEST_DIR_AUX_R});
}

unix {
    copyRFiles.commands += $(MKDIR) $$DEST_DIR_AUX_R ;
    copyRFiles.commands += cp $$SRC_WRITE_IMAGE $$DEST_DIR_AUX_R ;
    copyRFiles.commands += cp $$SRC_WRAPPERS $$DEST_DIR_AUX_R ;
}


! equals(PWD, $${OUT_PWD}) {
    QMAKE_EXTRA_TARGETS += copyRFiles
    POST_TARGETDEPS     += copyRFiles
}

```

JorisGoosen: So something like this?

JorisGoosen:
jaspResultsCalledFromJasp() seems like a good idea, I don't think the classdef will change at all because Rcpp::Modules (on which jaspResults is based) is no longer in active development anyway.

PR #3028, opened 2019-02-07
Bayesian Linear Regression now uses jaspResults

TimKDJ:
So we'll probably want to change the status handling and possibly the way the model is passed around.
Right now I do it "R-like", meaning I return and then pass it on explicitly. We could also use jaspResults[["stateModel"]] directly in the analyses. Thoughts?

### DIFF
--- a/JASP-Engine/JASP/R/regressionlinearbayesian.R
+++ b/JASP-Engine/JASP/R/regressionlinearbayesian.R
@@ -27,43 +27,43 @@ RegressionLinearBayesian <- function(jaspResults, dataset = NULL, options) {
   basregModel     <- .basregGetModel(basregContainer, dataset, options, ready)
 
   if (is.null(basregContainer[["modelComparisonTable"]]))
-    basregContainer[["modelComparisonTable"]]       <- .basregTableModelComparison(basregModel, options, position = 11)
+    .basregTableModelComparison(basregContainer, basregModel, options, position = 11)
   
   if (options$postSummaryTable || options$postSummaryPlot)
-    postSumContainer  <- .basregGetPosteriorSummaryContainer(basregContainer, position = 12)
+    postSumContainer <- .basregGetPosteriorSummaryContainer(basregContainer, position = 12)
     
   if (options$postSummaryTable || options$postSummaryPlot || options$plotCoefficientsPosterior)
-    postSumModel      <- .basregGetPosteriorSummary(basregContainer, basregModel, dataset, options, ready)
+    postSumModel <- .basregGetPosteriorSummary(basregContainer, basregModel, dataset, options, ready)
   
   if (options$postSummaryTable && is.null(basregContainer[["postSumContainer"]][["postSumTable"]]))
-    postSumContainer[["postSumTable"]]              <- .basregTablePosteriorSummary(postSumModel, basregModel, options, position = 121)
+    .basregTablePosteriorSummary(postSumContainer, postSumModel, basregModel, options, position = 121)
   
   if (options$postSummaryPlot && is.null(basregContainer[["postSumContainer"]][["postSumPlot"]]))
-    postSumContainer[["postSumPlot"]]               <- .basregPlotPosteriorSummary(postSumModel, options, position = 122)
+    .basregPlotPosteriorSummary(postSumContainer, postSumModel, options, position = 122)
   
   if (options$plotLogPosteriorOdds && is.null(basregContainer[["logPosteriorOddsPlot"]]))
-    basregContainer[["logPosteriorOddsPlot"]]       <- .basregPlotPosteriorLogOdds(basregModel, options, position = 13)
+    .basregPlotPosteriorLogOdds(basregContainer, basregModel, options, position = 13)
     
   if (options$plotResidualsVsFitted && is.null(basregContainer[["ResidualsVsFittedPlot"]]))
-    basregContainer[["ResidualsVsFittedPlot"]]      <- .basregPlotResidualsVsFitted(basregModel, position = 14)
+    .basregPlotResidualsVsFitted(basregContainer, basregModel, position = 14)
   
   if (options$plotModelProbabilities && is.null(basregContainer[["modelProbabilitiesPlot"]]))
-    basregContainer[["modelProbabilitiesPlot"]]     <- .basregPlotModelProbabilities(basregModel, position = 15)
+    .basregPlotModelProbabilities(basregContainer, basregModel, position = 15)
   
   if (options$plotModelComplexity && is.null(basregContainer[["modelComplexityPlot"]]))
-    basregContainer[["modelComplexityPlot"]]        <- .basregPlotModelComplexity(basregModel, position = 16)
+    .basregPlotModelComplexity(basregContainer, basregModel, position = 16)
     
   if (options$plotInclusionProbabilities && is.null(basregContainer[["inclusionProbabilitiesPlot"]]))
-    basregContainer[["inclusionProbabilitiesPlot"]] <- .basregPlotInclusionProbabilities(basregModel, position = 17)
+    .basregPlotInclusionProbabilities(basregContainer, basregModel, position = 17)
   
   if (options$plotQQplot && is.null(basregContainer[["qqPlot"]]))
-    basregContainer[["qqPlot"]]                     <- .basregPlotQQ(basregModel, position = 18)
+    .basregPlotQQ(basregContainer, basregModel, position = 18)
   
   if (options$plotCoefficientsPosterior && is.null(basregContainer[["postDistContainer"]]))
-    basregContainer[["postDistContainer"]]          <- .basregPlotsPosteriorDistribution(postSumModel, basregModel, options, position = 19)
+    .basregPlotsPosteriorDistribution(basregContainer, postSumModel, basregModel, options, position = 19)
   
   if (options$descriptives && is.null(jaspResults[["descriptivesTable"]]))
-    jaspResults[["descriptivesTable"]]              <- .basregTableDescriptives(dataset, options, ready, position = 2)
+    .basregTableDescriptives(jaspResults, dataset, options, ready, position = 2)
 }
 
 .basregReadData <- function(dataset, options) {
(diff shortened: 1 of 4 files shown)


JASP-Engine/JASP/R/regressionlinearbayesian.R
```diff
@@ -15,587 +15,360 @@
...
+    return()
+  
+  modelComparison <- createJaspTable(title = "Model Comparison")
+  jaspResults[["modelComparisonTable"]] <- modelComparison
+  
+  modelComparison$copyDependenciesFromJaspObject(jaspResults[["stateModel"]])
+  modelComparison$dependOnOptions(c("bayesFactorType", "bayesFactorOrder", "shownModels", "numShownModels"))
+  
+  
+  # modelComparison$addCitation <- list(
```
vandenman: It's a bit ugly but you can add citations individually:
```r
modelComparison$addCitation("Clyde, M. A. (2017). BAS: Bayesian Adaptive Sampling for Bayesian Model Averaging. (Version 1.4.7)[Computer software].")
modelComparison$addCitation("Clyde, M. A., Ghosh, J., & Littman, M. L. (2011). Bayesian adaptive sampling for variable selection and model averaging. Journal of Computational and Graphical Statistics, 20, 80-101.")
```

JASP-Engine/JASP/R/regressionlinearbayesian.R
```diff
@@ -15,587 +15,360 @@
...
+  
+  modelComparison$setExpectedRows(nRows)
+  
+  footnote <- basregModel[["interaction"]][["footnote"]]
+  if (!is.null(footnote)) {
+    modelComparison$addFootnote(message = footnote, symbol = "<em>Warning.</em>")
+  }
+  
+  if (sum(basregModel$nuisanceTerms) > 0) {
+    footnote <- paste("All models include ", paste(names(which(basregModel$nuisanceTerms)), collapse = ", "), ".", sep = "")
```
vandenman: `paste0`?

JASP-Engine/JASP/R/regressionlinearbayesian.R
```diff
@@ -15,587 +15,360 @@
...
-	null.model <- "Null model"
-	if (sum(nuisanceTerms) > 0) {
-		null.model <- paste("Null model (incl. ", paste(names(which(nuisanceTerms)), collapse = ", "), ")", sep = "")
-	}
-
-	# generate all model names
+  nuisanceTerms <- basregModel$nuisanceTerms
+  null.model <- "Null model"
+  if (sum(nuisanceTerms) > 0) {
+    null.model <- paste("Null model (incl. ", paste(names(which(nuisanceTerms)), collapse = ", "), ")", sep = "")
```
vandenman: `paste0`

JASP-Engine/JASP/R/regressionlinearbayesian.R
```diff
@@ -15,587 +15,360 @@
...
+  }
+  
+  # generate all model names
   allModelsVisited <- any(lengths(models) == 0) # analysis will chrash if TRUE
   model.names <- vector("character", length(models))
```
vandenman: `character(length(models))`?

JASP-Engine/JASP/R/regressionlinearbayesian.R
```diff
@@ -611,1138 +384,1065 @@ RegressionLinearBayesian <- function (
...
+  coefficients <- posteriorSummary[["coefficients"]]
+  probne0 <- posteriorSummary[["probne0"]]
+  loopIdx <- posteriorSummary[["loopIdx"]]
+  confInt <- posteriorSummary[["conf95"]]
+  
+  topm <- order(-basregModel$postprobs)[1:nModels]
+  mostComplex <- which.max(lengths(basregModel$which)[topm])
+  
+  for (i in loopIdx) {
+    coefficient <- .clean(coefficients[i])
```
vandenman: same thing about `.clean`

JASP-Engine/JASP/R/regressionlinearbayesian.R
```diff
@@ -611,1138 +384,1065 @@ RegressionLinearBayesian <- function (
...
+  
+  return()
+}
 
-    } else {
-        plot <- .makeEmptyPlot.basReg(title = title, status = status)
-    }
+.basregFillPlotInclusionProbabilities <- function(basregModel, jaspResults, options) {
+  
+  p <- try({
```
vandenman, line 686: In this and the other `try({...})` blocks the first 20 lines aren't really prone to erroring. Perhaps it's more clear to only put things in `try({...})` that can give an error?

JASP-Engine/JASP/R/regressionlinearbayesian.R
```diff
@@ -611,1138 +384,1065 @@ RegressionLinearBayesian <- function (
...
+    JASPgraphs::themeJasp(g, horizontal = TRUE, legend.position = "right") +
+      ggplot2::theme(
+        legend.title = ggplot2::element_text(size = .8*JASPgraphs::graphOptions("fontsize"))
+      )
     
+  })
+  
+  if (isTryError(p)) {
+    errorMessage <- paste("Plotting not possible:", .extractErrorMessage(p))
+    jaspResults[["inclusionProbabilitiesPlot"]]$error <- "badData" #FIXME: this should go
```
vandenman: In these (and other functions) why not pass the plotObject, rather than jaspResults? Should work exactly the same but is a bit more concise.

JASP-Engine/JASP/R/regressionlinearbayesian.R
```diff
@@ -15,587 +15,360 @@
...
+  .basregplotsPosteriorDistribution(posteriorSummary, basregModel, jaspResults, options, status)
+  
+  .basregPlotResidualsVsFitted(basregModel, jaspResults, options, status)
+  .basregPlotModelProbabilities(basregModel, jaspResults, options, status)
+  .basregPlotModelComplexity(basregModel, jaspResults, options, status)
+  .basregPlotInclusionProbabilities(basregModel, jaspResults, options, status)
+  .basregPlotQQ(basregModel, jaspResults, options, status)
+  
+  # this function still needs to be rewritten to ggplot (start in .plotImage.basReg):
+  .basregPlotPosteriorLogOdds(basregModel, jaspResults, options, status)
```
vandenman: This means there will be a hard crash (not an R crash) when a user tries to copy this plot as pdf or eps. So for the next release this should be fixed!

JASP-Engine/JASP/R/regressionlinearbayesian.R
TimKDJ: Yeah, this is indeed much nicer

JASP-Engine/JASP/R/regressionlinearbayesian.R
TimKDJ: Mkay. Maybe we can just omit the plot if no one gets it anyway...

JASP-Engine/JASP/R/regressionlinearbayesian.R
TimKDJ: Sorry?

JASP-Engine/JASP/R/regressionlinearbayesian.R
TimKDJ: Guess this method requires some work too ;)

JASP-Engine/JASP/R/regressionlinearbayesian.R
vandenman: It seems some comment disappeared. If I'm not mistaken, `.clean` is done automagically by jaspResults once you enter the data.

vandenman: also Joris just merged the error property PR so you could also try that out already

TimKDJ: I tested and .clean() is still necessary. For example `NULL` is passed on in json as `null` which javascript cannot deal with. I would also venture a guess that it also doesn't show a nice infinity symbol when `Inf` is passed.

I added the error handling. It seems to work nicely (nice job!), I have two  "issues" with it though:
1) `basReg$getError()` does not convey its meaning super nicely.  I guess this is kinda personal, something like`basReg$hasError()` or `basReg$error()` is clearer, but I understand why you named it this way -- to match the `setError()`.
2) we HAVE to use containers now to make it work, only the containers are always displayed as such in the output and this adds a visible grouping I don't think we really need. Also, and this is something you also encountered, you cannot use this container grouping to pass dependencies from the parent to the child, instead all children MUST have some sort of dependency set otherwise they get thrown out.

boutinb:
Is this OK now?

TimKDJ:
Nope.. First Don's PR needs to go through so I can incorporate that stuff.

TimKDJ:
@vandenman maybe you can help me out here. In the code there is a warning that states
```
paste("Prior model probabilities for models with interaction effects",
                                   "are obtained from a Bernoulli (p = 0.5) prior.",
                                   "We advice using a different model prior, or excluding interaction effects.")
```
This warning is included for `(options$modelPrior == "beta.binomial")` -- is this correct? Beta binomial is apparently a Bernoulli for n=1, but is that the case here? I'm going to guess it should also be included for the `(options$modelPrior == "Bernoulli")`, given that p=0.5?

vandenman:
BAS doesn't give this warning anymore, so maybe we can delete it? I'm not sure how the model probabilities should be computed for a beta-binomial with interactions, but this approach is not super weird. Perhaps @AlexanderLyNL has another idea?

AlexanderLyNL:
Let's discuss this in person, and get to root of why bas has this warning in the first case, then we can what changed in the meantime.

vandenman:
@AlexanderLyNL BAS does **not** have this warning. I think it did in the past and we just copied that. But sure let's discuss this when I'm back.

TimKDJ:
- I removed the warning, considering BAS doesn't have it anymore
- I changed the "limit number of models shown" to strictly enforce the specified number and not go num models + 1 (for null) as per Maarten's confusion
- I made the analysis use the nifty results-merging that Joris added

vandenman:
@TimKDJ great! Let me know if you want me to review it again.

TimKDJ:
Yeah, that'd be nice!

JASP-Engine/JASP/R/regressionlinearbayesian.R
```diff
@@ -15,1771 +15,1252 @@
...
-			title = "",
-			name = "",
-			PosteriorCoefficients = emptyPlot
-		))
-
-	}
-
-	return(plots)
+  if (options$samplingMethod == "MCMC") {
+    postLogOddsPlot$errorMessage <- "Cannot display Posterior Log Odds when sampling method is MCMC."
```
vandenman:
```suggestion
    postLogOddsPlot$setError("Cannot display Posterior Log Odds when sampling method is MCMC.")
```

PR #3029, opened 2019-02-07
JS shows empty axes when a plot has no data

TimKDJ:
So now the following will show an empty plot automatically:
```
createJaspPlot(title = "Some plot title", width = 300, height = 300)
```


TimKDJ:
@vandenman I guess we could also extend this to tables:
If an empty table is generated, automatically add 1 row of dots.
I find myself using $setExpectedRows() twice per table function, once if not ready or error (`$setExpectedRows(1)`) and once if actually filling the table (`$setExpectedRows(nRows)`).

vandenman:
Extending this to tables makes sense, but I'd rather see that changed on the cpp side. For example, when the results get send by jaspResults, if any table has 0 rows, add 1 row with ".". 

Oh and if the table gets filled quickly, do you really need the `$setExpectedRows(nRows)`?

TimKDJ:
Why do you want this in cpp?

vandenman:
Let me rephrase, I just don't want it in R. In particular `$setExpectedRows(1)` for empty tables can be done just as easy in jaspResults (e.g., before sending the json), but of course it can also be done in javascript if you prefer.

boutinb:
But this fix can be already pushed I suppose. The fix for table can be done after.

TimKDJ:
Yeah, it can be done separately.

PR #3032, opened 2019-02-08
added error property

vandenman:
A proof of concept for adding an error property to jaspResults. Some example behavior:

```r
rm(list = ls(all.names = TRUE))
library(jaspResults)

# example 1: errors are inherited at insertion
ct1 <- createJaspContainer()
ct1$getError() # FALSE
ct1$setError() 

tb1 <- createJaspTable()
ct1[["tb1"]] <- tb1
ct1[["tb1"]]$getError() # TRUE
tb1$getError()          # TRUE


# example 2: errors are propagated to children
jaspResults$setError()

tb2 <- createJaspTable()
ct1 <- createJaspContainer()

ct1[["tb2"]] <- tb2
ct1[["tb2"]]$getError() # FALSE

jaspResults[["ct1"]] <- ct1
jaspResults[["ct1"]][["tb2"]]$getError() # TRUE
tb2$getError()                           # TRUE

# example 3: if an error doesn't have an error it does not modify the error of a child
ct1 <- createJaspContainer("hoi1")
ct2 <- createJaspContainer("hoi2")
ct3 <- createJaspContainer("hoi3")

ct2$setError()

ct2[["hoi3"]] <- ct3
ct1[["hoi1"]] <- ct2

ct1$getError() # FALSE
ct2$getError() # TRUE
ct3$getError() # TRUE
```

The changes are not fully backwards compatible, because now you'd do `table$setError()` rather than `table$error <- string`. I suspect we won't have to make too many changes for that though.


JASP-R-Interface/jaspResults/src/jaspContainer.cpp
```diff
@@ -271,3 +272,10 @@ void jaspContainer::checkDependenciesChildren(Json::Value currentOptions)
 	for(auto & removeThis : removeThese)
 		_data.erase(removeThis);
 }
+
+void jaspContainer::setError(bool error) {
+	_error = error;
+	if (error)
```
JorisGoosen: I guess the assumption here is that there is no reason to ever set the error to false?

JASP-R-Interface/jaspResults/src/jaspContainer.h
```diff
@@ -43,7 +43,7 @@ class jaspContainer : public jaspObject
 	void		checkDependenciesChildren(Json::Value currentOptions) override;
 
 	void		completeChildren();
-
+	void		setError(bool error);
```
JorisGoosen: Because you override the virtual function on jaspObject you should add `override`, aka:
`void		setError(bool error) override;` otherwise you hide the original function and this might have unintended consequences.

JorisGoosen: I guess this should be the start of a request for changes.

JASP-R-Interface/jaspResults/src/jaspContainer.cpp
vandenman: Well, you can still set the error of a specific jaspObject to false, but it won't get passed to the children. This is sort of necessary, because if we remove the if statement above, and you'd put a child with error into a parent without error, the child's error gets overwritten and becomes false.

JASP-R-Interface/jaspResults/src/jaspContainer.h
vandenman: Good one! no idea how this stuff with `virtual` and `override` really works.

JASP-R-Interface/jaspResults/src/jaspContainer.cpp
JorisGoosen: Yeah, that is exactly why I wanted to set it only during the construction of the result-json (basically by changing virtual `Json::Value jaspObject::dataEntry()` to `Json::Value jaspObject::dataEntry(bool error)` and then setting the specific error on any object if either the parent passed error or if _error is true.

This way you don't need to worry about updating children/parents or whatever.

Because what happens now if you have a jaspContainer that has _error set and a table that is added later? First the table has it's error set to true by this insert function. You can however still set it false later right? I'm guessing that is not what we would want?

vandenman: if we want to forbid developers from turning `obj$error == TRUE` to `obj$error == FALSE`, we could change the `setError` function to do `if (!_error) _error = error`. 

I don't think it's good design to ever change the error from TRUE to FALSE, so if you really want to forbid that we can.

If we only update the children when making the json the problem is that you can't use the error information on the R-side in if statements (which is why this PR exists).

JorisGoosen: Or maybe we can just change the function `setError(bool error) -> setError()` and it always sets it to true?

JorisGoosen: And then the property error must also be read-only. (Which is easily done by removing the write option in jaspModuleRegistration)

vandenman: Hmm we can definitely ignore the bool, so `setError(bool error) _error = true`, but without the argument `bool error` the code doesn't want to compile. Probably it becomes ambiguous whether the call `jaspObj$error` is getting or setting the property. I guess we'd have to make it into a method and use `jaspObj$setError()` and `jaspObj$setError()` instead. But this is not intuitive for as R users. Thoughts?

JASP-R-Interface/jaspResults/src/jaspModuleRegistration.h
```diff
@@ -30,6 +30,7 @@ RCPP_MODULE(jaspResults)
 		.property("title",							&jaspObject_Interface::getTitle,	&jaspObject_Interface::setTitle,	"Set the title of this object")
 		.property("warning",						&jaspObject_Interface::getWarning,	&jaspObject_Interface::setWarning,	"Set a warning on this object to be shown in the output")
 		.property("position",						&jaspObject_Interface::getPosition,	&jaspObject_Interface::setPosition,	"Set the position of this object in it's container. By default this is at the end in the order of adding. You can specify any other value, they do not need to be next to each other or unique. The rule is: lower values (including negative) are higher in the container and when multiple objects in a container have the same position-value order is derived from adding-order.")
+  		.property("error",                          &jaspObject_Interface::getError,	&jaspObject_Interface::setError,	"Set an error on this object. Will make the output in JASP greyed out. Errors set on jaspContainers or jaspResults are propagated to children.")
```
JorisGoosen: In response to https://github.com/jasp-stats/jasp-desktop/pull/3032#discussion_r255084333

Just remove `,	&jaspObject_Interface::setError` here and then the property will be read-only.
Then setError can be added as a method.

While I agree that this is perhaps contrary to what an R-user would expect I think this is the best option. Because how confusing would it be if you set error to false and it remains true anyway?

JorisGoosen:
Yeah, having setError & getError is definitely more symmetrical, I am assuming you've tested this all and it works so I'll merge it.

PR #3035, opened 2019-02-11
added errorMessage to jaspObject

vandenman:
Implements a common errorMessage property on jaspObject. The goal of this is:

 - you can set an error message on jaspContainers.
 - this sets its error property and that of its children to true.
 - the error message get passed to the first inserted jaspObject.

For non-containers, the implementation is straigtforward. For containers, `_alreadyPassedErrorMessage` and `_passErrorMessageToNextChild` ensure that a container only passes an error message to the first child. If that child is another container, it will pass the error message to its first child.

If you update an error message of a container after setting it will pass the updated message to the first child. If it already passed an error message it does nothing. New error messages for tables/ plots will overwrite the current message. 

If we want to forbid attempts to updating error messages, we could add:
```c++
if (_errorMessage != "") 
  Rcpp::stop("Cannot overwrite existing error message")
```
I see no way to avoid this like we did with `setError()`. Some example code:

```r
library(jaspResults)
ct1 <- createJaspContainer("hoi1")
ct2 <- createJaspContainer("hoi2")
ct3 <- createJaspContainer("hoi3")
ct4 <- createJaspContainer("hoi4")

# example 1: create the object structure and then set the errormessage on highest container
tb1 <- createJaspTable("table1")
tb2 <- createJaspTable("table2")
ct2[["table1"]] <- tb1
ct2[["table2"]] <- tb2
ct1[["ct2"]] <- ct2
ct1$errorMessage <- "hoi!"

tb1$errorMessage # hoi!
tb2$errorMessage # ""

# error is set to TRUE
ct1$getError() # TRUE
ct2$getError() # TRUE
tb1$getError() # TRUE
tb2$getError() # TRUE

# example 2: first set the errormessage and then create the object structure
ct3$errorMessage <- "hoi@"
ct3[["ct4"]] <- ct4
ct4$errorMessage
tb3 <- createJaspTable("table3")
tb4 <- createJaspTable("table4")
ct4[["table3"]] <- tb3
ct4[["table4"]] <- tb4

tb3$errorMessage # hoi@
tb4$errorMessage # ""
tb3$getError() # TRUE
tb4$getError() # TRUE

```


TimKDJ:
> . why do you add a new method setErrorMessage, and do not extend the setError with a string argument like the setWarning method. It seems to me more logical for the user.

I kinda agree with this. It's nicer to go `jaspContainer$setError(errorMessage)` and then fetch the boolean status of error with either `jaspContainer$error()` or `jaspContainer$hasError()`. I don't think we need a separate `$setErrorMessage()` and `$setError()` -- if you set an error you should always specify a message.

vandenman:
TLDR: the last commit does the following:

- `setError(string)` is mandatory, `setError()` gives an error. `getError()` returns a boolean. The function `setError()` does exist on the C++ side but R developers have no access to it.

I can see some use of `setError()` without a message but let's deal with that when (if) we encounter a use case.

@boutinb 

> . why do you add a new method setErrorMessage, and do not extend the setError with a string argument like the setWarning method. It seems to me more logical for the user.
. When you overwrite setErrorMessage in jaspContainer, you should call the parent method first:
jaspObject::setErrorMessage(message);
in place of:
_error= true;
_errorMessage = message;
You have added 2 properties: _alreadyPassedErrorMessage & _passErrorMessageToNextChild
but I think they must be reset each time that _error is set: in setError or in convertFromJSON_SetFields

Good points, all changed! I removed the property `_alreadyPassedErrorMessage`, so that should not linger.

> I don't understand why the error message of a container has to be transmitted only to the first child of the container.

Imagine we have a container with 20 plots and a single error message. We could put the error message on all the plots, and then the user sees 20 error messages. That is very confusing, especially since all error messages are the same. So we only show the error on the first element, and set `_error` to true on all the other elements. Then, a user sees an error message on the first plot, and all other plots are greyed out.

> What I understand of the 2 properties _alreadyPassedErrorMessage & _passErrorMessageToNextChild is to transmit the error message to the first child of the container when the container has first no object, but later on gets an object. When does occur this situation?

If you first make an new container and already know it has an error, you can choose: (1) first put the error on there and then fill the container, or (2) fill the container and then  add the error. Both scenarios are possible and should work. The property `_alreadyPassedErrorMessage` made it such that you couldn't do this twice (e.g., `setError("hoi"); setError("doei")` implied the error message is "hoi"). Now you can do this repeatedly (but who will?).

PR #3039, opened 2019-02-19
implemented R6 classes

vandenman:
Creates R6 wrappers around the jaspResults objects.


vandenman:
Note that this doesn't work in JASP yet, because we need to copy more files / pollute the global namespace more.

vandenman:
Okay assuming Travis is happy the R6 classes are now working properly in both JASP and R, and this PR is ready for review. I would not include these changes in the alpha version though.

TimKDJ:
So do you reckon we should fix the existing code to deal with `createJaspPlot(error="bla")` or merge this for the new release?

Also, on an unrelated note, what do you think of merging all the dependency functions into one?
```
$setOptionMustBeDependency()
$setOptionMustContainDependency()
$copyDependenciesFromJaspObject()
$dependOnOptions()
```

Could be:

```
$dependsOn(
  options=c("opt1", "opt2"),
  optionsFromObject=jaspObj,
  optionContainsValue=list()
)
```

Besides possibly being a bit clearer, it would avoid multi-liners like:

```
table <- createJaspTable()
table$copyDependenciesFromJaspObject(jaspResults[["obj"]])
table$dependOnOptions("someOtherOpt")
```

As we'd get:

```
table <- createJaspTable()
table$dependsOn(options="someOtherOpt", optionsFromObject=jaspResults[["obj"]])
```

JorisGoosen:
This looks awesome!

JorisGoosen:
Can I merge this into development @boutinb ?

boutinb:
@vandenman Kun je de conflict fixen? Misschien door de commits te squashen?

PR #3044, opened 2019-02-21
Added QML documentation

TimKDJ:


TimKDJ:
@boutinb I didn't really talk about signals and the "Connecting Multiple Components" is a little thin, maybe you want to add some stuff after this is merged.

PR #3065, opened 2019-02-27
First full version of CFA in a dynamic module

vankesteren:
Still needs to be debugged / tested


vankesteren:
with help from @boutinb

Issue #309, opened 2015-02-16
Implement Bayesian One-Way ANOVA


mmarsman1:
Should we implement a Bayesian one-way ANOVA?

jonathon-love:
i don't think so. the main reason for providing a one way anova is so the user can specify bespoke contrasts; everything else they can do with the "normal" ANOVA.

PR #3090, opened 2019-03-07
Fixes Descriptives issues #304 and #309

TimKDJ:
The row.names returned from .readDataSetToEnd() were incorrect after a filter was applied and missing values were not omitted in the grouping variable in Descriptives.


vankesteren:
👍

PR #3109, opened 2019-03-15
Fix residual histogram scaling

koenderks:


JASP-Engine/JASP/R/regressionlinear.R
```diff
@@ -3938,11 +3938,11 @@ RegressionLinear <- function(dataset=NULL, options, perform="run", callback=func
 
     h <- hist(res, plot = FALSE)
     dens <- h$density
-    yhigh <- 1.2*max(h$density)
```
TimKDJ: `dens` is not a list, but a vector; it should be `max(c(h$density, dens))`

JASP-Engine/JASP/R/regressionlinear.R
```diff
@@ -3938,11 +3938,11 @@ RegressionLinear <- function(dataset=NULL, options, perform="run", callback=func
...
+		yhigh <- max(c(h$density, dens$y))
```
TimKDJ: `dens` is not a list, but a vector; it should be `max(c(h$density, dens))`

koenderks:
Fixed, but not as suggested in Tim's comment. The yhigh should be the maximum of the density line and the density of the histogram. Now this is implemented correctly.

AlexanderLyNL:
Can you fix the unit test?

koenderks:
> Can you fix the unit test?

The unit test is about the plot right? Is that not supposed to be?

AlexanderLyNL:
Update the svg?

koenderks:
> Update the svg?

Can you point me to where to do this?

AlexanderLyNL:
The test fails because the new svg looks different from the reference: 

/jasp-desktop/JASP-Tests/R/tests/figs/RegressionLinear/regressionlinear-residuals-histogram.svg

In my folder, this reference looks quite okay. What does the new one look like?

koenderks:
I'm not sure what data is used in the unit test to make that svg

AlexanderLyNL:
Let's talk about this tomorrow, perhaps @TimKDJ can help.

koenderks:
Tuesday will have to do, as I have to teach on Monday and Wednesday.

AlexanderLyNL:
@koenderks Could you also have "Standardized residuals" on by default?

TimKDJ:
Can you rebase? The test should be ok afterwards

koenderks:
Rebased

PR #3121, opened 2019-03-20
pass qgraph object rather than function

vandenman:
Should fix the recurring network bug on windows.

### DIFF
--- a/JASP-Engine/JASP/R/networkanalysis.R
+++ b/JASP-Engine/JASP/R/networkanalysis.R
@@ -1422,46 +1422,42 @@ NetworkAnalysis <- function (
 
 }
 
-.networkAnalysisOneNetworkPlot <- function() {
-
-	# eval(quote()) construction because this function is evaluated inside .writeImage()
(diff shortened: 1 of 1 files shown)


TimKDJ:
Fixes jasp-stats/jasp-issues#158, jasp-stats/jasp-issues#229, jasp-stats/jasp-issues#264 and possibly jasp-stats/jasp-issues#306

For future reference, the problem is as follows:
The network analysis plot is a `qgraph` object but we passed this to `writeimage()` as a function, which in turn transformed it to a `recordedPlot` and this `recordedPlot` was saved in the state. At the moment of retrieving this state the `recordedPlot`'s complexity was apparently too great for R's `load()` and so R crashed. The condition was not caught by R and therefore the `tryCatch()` did nothing, which is why it appeared as a hard crash in JASP. To fix this we pass `writeimage()` the `qgraph` object and consequently this is also what goes in the state.

Not really sure what causes the difference between loading the state in the JASP context vs interactive R; or mac vs windows, but it appears solved now.

PR #2990, opened 2019-01-03
[DO NOT MERGE] AF remaining features

Janto91:
This adds the remaining features from the Field list for Anova, Ancova, RM Anova, and logistic regression. However, there is some work on the code needed to get JASP running with these new features. The code also needs to be thoroughly tested and cross-checked with SPSS (see the Excel sheet I shared) before this can be merged.

PR #3142, opened 2019-04-01
andyField - RM ANOVA

Kucharssim:
This PR concludes https://github.com/jasp-stats/jasp-desktop/pull/2990

New features:
- chi square and df for mauchly's test of sphericity, and lower bound epsilon
- confidence intervals (bonferroni corrected) for post-hoc tests (pooled and not-pooled errors included)
- marginal means (including interactions)
- bootstrapping for marginal means

last feature missing for Field - contrasts for interactions. I guess that could be done while adding "custom" contrasts option @JohnnyDoorn?


JohnnyDoorn:
Looks great, just tried it! 
Is there any way to improve the state for the bootstrapping? Unchecking and checking the bootstrap option triggers new sampling - it might be nice to save the samples so the table just pops back up again. Or maybe a progress bar is nice for this, cause it can be quite slow? I don't think it's a necessity though, so if you advise against it I approve the PR.

Kucharssim:
Thanks! Strange, I think I had it right, maybe I habe broken the state with the last commit, I’ll look into it. The progress bar is a good idea, but I would prefer to postpone that for the release after this one and do it for all the bootstraping that we have scattered all over the place...

Kucharssim:
Seems that it was an easy fix, now it should be better!

JohnnyDoorn:
Awesome! And yeah I agree, it would be nice to have the whole thing rewritten in jaspresults anyway, then a progress bar is trivial (looking at myself to start rewriting this..)

PR #3150, opened 2019-04-05
Fixed the font size of saved images

TimKDJ:
It only works for ggplot objects; qgraph and base plots are still not ok. Hopefully those plots will be rewritten to ggplot soon, as I do not have a good solution for them.

Also I I did not include the tiff rescaling, as it only seemed to screw with the dimensions. @vankesteren: I think you wrote this code; why does tiff need rescaling, but not the other formats?

This PR still needs to be thoroughly tested on other platforms.

### DIFF
--- a/JASP-Engine/JASP/R/common.R
+++ b/JASP-Engine/JASP/R/common.R
@@ -2309,56 +2309,86 @@ saveImage <- function(plotName, format, height, width)
   error <- try({
     # Get file size in inches by creating a mock file and closing it
     pngMultip <- .fromRCPP(".ppi") / 96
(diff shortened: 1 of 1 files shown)


vankesteren:
I think when I wrote this code I was just programming by trial-and-error because this stuff is quite obscure.

vankesteren:
from memory: maybe because I was setting a specific resolution for tiff?

AlexanderLyNL:
Hi @akashrajkn @astefan1 @Kucharssim. Could you guys checkout this PR and see how this works on your OS?

Kucharssim:
On MacOS High Sierra 10.13.6 it works perfectly.

PR #3151, opened 2019-04-08
create R6 children if needed

vandenman:
Solves https://github.com/jasp-stats/INTERNAL-jasp/issues/178.

The R6 side kept a list of children (R6 objects) that was deleted after consecutive runs of an analysis. 

Now, when indexing in a container, if the c++ object exists but the R6 one does not the R6 version is recreated and added to the list of children. For this purpose the default constructors have a new argument, `jaspObject = NULL`. Developers should never touch this argument since they should use `createJaspTable(...)` rather than `jaspTableR$new(...)`.

Also, the method `getJaspObject` is now private instead of public. So `x$getJaspObject()` now becomes `private$getJaspObject(x)` and can only be called within R6 objects.


vandenman:
Exploratory Factor Analysis failed but that analysis doesn't even use jaspResults...

PR #3166, opened 2019-04-12
modified legend

vandenman:
Should fix https://github.com/jasp-stats/jasp-issues/issues/305

Descriptives now look like this:

afbeelding

The overlap of the x-axis labels is hard to fix. Also, the changes are partially in the theme and could be generalized to JASPgraphs, but I didn't do so yet as then we'd have to test many plots.


TimKDJ:
Are there other plots that use a legend? Because if that's the case it's probably better to implement it in JASPgraphs

vandenman:
Okay so after some looking around I made the change global. I also fixed the cut-off text for the axis labels.

before (g and y are cut off):
afbeelding


now:
afbeelding

vandenman:
@TimKDJ Travis is happy! Once I closed all running JASP instances I could reset my jasp-required-files branch (I couldn't before because JASP was using the packages). It was surprisingly easy to compare the visual differences with vdiffr and there were no major differences. The largest difference is that legends are now shown by default in the top-right of the plot rather than the middle right of the plot.

TimKDJ:
Awesome!

PR #3168, opened 2019-04-15
Bring back debug data set in data library on JASP Debug version.

FransMeerhoff:
- The debug data set was removed from data library, placed back again.
Solves https://github.com/jasp-stats/INTERNAL-jasp/issues/212


FransMeerhoff:
@Kucharssim 

Hi Simon, can you have a look at this PR
Two remarks in advance
- the indexdebug.json is not used anymore so I removed it.
- we use a debug flag in the json instead so I've added that to the update data library.R
I think the update data library needs some more changes to write the debug set to the index.json but I was not able to test that. Can you have a look? Thx

PR #3175, opened 2019-04-16
[DO NOT MERGE] Remove \u00a0 character from table headers

akashrajkn:
Fixes jasp-stats/INTERNAL-jasp#121


akashrajkn:
Need to test it on a windows system

akashrajkn:
This fix is not correct

PR #3202, opened 2019-04-25
PCA/EFA fixes

TimKDJ:
See jasp-stats/jasp-test-release#64

### DIFF
--- a/JASP-Engine/JASP/R/principalcomponentanalysis.R
+++ b/JASP-Engine/JASP/R/principalcomponentanalysis.R
@@ -135,10 +135,10 @@ mainFunctionPCAEFA <- function(type, dataset = NULL, options, perform = "run",
 	# Create fit measures tables:
 	results[["goodnessOfFit"]] <- .goodnessOfFit(analysisResults, options, perform)
 	# Create factor correlation table:
-	results[["factorLoadings"]] <- .getLoadings(analysisResults, dataset, options, perform, "pca")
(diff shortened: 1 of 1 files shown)


JASP-Engine/JASP/R/principalcomponentanalysis.R
```diff
@@ -279,12 +279,21 @@ mainFunctionPCAEFA <- function(type, dataset = NULL, options, perform = "run",
 
 	# Create JASP table:
 	Loadings <- list()
-	Loadings[["title"]] <- "Component Loadings"
+	Loadings[["title"]] <- ifelse(type == "pca", "Component Loadings", "Factor Loadings")
```
AlexanderLyNL: I think that this is a typical case that you'd like to use a standard if else construction, as the ifelse-function in R is vectorised and can lead to some unexpected behaviour:

ifelse(1:10 %% 2, 1:10, letters[1:3])

Other than that the code looks good, but I haven't done a checkout on it yet.

TimKDJ: Hadn't incorporated your suggestion yet. Fortunately in this case `type` is constant. We'll get rid of the ifelse's when we rewrite the analyses

(Although supplying a vector to `if` doesn't work well either, in the following example what will happen?
```
if (sample(c(T,F)))
  print("hi")
```
)

PR #3218, opened 2019-04-30
Fix image saving for windows

TimKDJ:
This should fix EPS/PDF saving for Windows and also fix the font problems in qgraph / base plots on OSX.

Basically OSX shouldn't use cairo devices and Windows shouldn't use the regular ones. There is a problem with the value of R_HOME as used by the default devices. I'm guessing it's something along the lines of the problem reported [here](https://bitbucket.org/rpy2/rpy2/issues/311/compatibility-problem-with-numpy-and-pdf).

R_HOME when asked for in R:
```
[1] "R home set to"

[1] "C:/JASP/build-debug-64/R"
```
But where it actually searches for the resources (after you specify the path to the `enc` file):
```
Warning in grDevices::pdf(file = filename, ..., version = version) :

afm file 'C:\Users\Tim\Documents/library/grDevices/afm/Helvetica.afm' could not be opened
```

This PR solves jasp-stats/INTERNAL-jasp#223

### DIFF
--- a/JASP-Engine/JASP/R/common.R
+++ b/JASP-Engine/JASP/R/common.R
@@ -2303,6 +2303,7 @@ saveImage <- function(plotName, format, height, width)
   relativePath <- paste0("temp.", format)
   
   error <- try({
+		
     # Get file size in inches by creating a mock file and closing it
     pngMultip <- .fromRCPP(".ppi") / 96
     png(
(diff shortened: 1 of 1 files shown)


TimKDJ:
@vandenman could you have a look on linux?

vandenman:
@TimKDJ Sure but I can't do that during work hours today.

TimKDJ:
No rush!

vandenman:
No issues on Linux, looks good to me!

TimKDJ:
Thanks for testing it

PR #3220, opened 2019-05-01
Make resized images retain their size on analysis rerun

TimKDJ:
Solves jasp-stats/jasp-test-release#93


TimKDJ:
Only works for jaspResults analyses btw. Another reason to quickly rewrite the remaining analyses :)

PR #3221, opened 2019-05-01
Avoid closing Analyses parameters form after clicking results

FransMeerhoff:
When running and anaylis and when - for example - you resize a plot, then the parameter window is closed. This is unwanted

### DIFF
--- a/JASP-Desktop/html/js/main.js
+++ b/JASP-Desktop/html/js/main.js
@@ -420,17 +420,7 @@ $(document).ready(function () {
 		var id = $(event.currentTarget).attr("id")
 		var idAsInt = parseInt(id.substring(3))
 
-		if (selectedAnalysisId == idAsInt && noteClicked === false) {
-				window.unselect()
-				jasp.analysisUnselected()
-		}
-		else if (selectedAnalysisId !== idAsInt && noteClicked === true) {
-			if (selectedAnalysisId !== -1) {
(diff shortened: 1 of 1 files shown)


JASP-Desktop/html/js/main.js
```diff
@@ -420,17 +420,7 @@ $(document).ready(function () {
...
-				jasp.analysisUnselected()
-		}
-		else if (selectedAnalysisId !== idAsInt && noteClicked === true) {
-			if (selectedAnalysisId !== -1) {
-				window.unselect()
-				jasp.analysisUnselected()
-			}
-		}
-		else {
+		if ((selectedAnalysisId == idAsInt && noteClicked === true) || (selectedAnalysisId !== idAsInt && noteClicked === false)) {
```
TimKDJ: The first part `(selectedAnalysisId == idAsInt && noteClicked === true)` is not necessary: if the analysis is selected, then select the analysis?

The second part `(selectedAnalysisId !== idAsInt && noteClicked === false)` is ok, but maybe we should think about the `noteClicked === false` bit, as it has consequences for the scrolling into view behaviour. This scrolling doesn't happen if an analysis is unselected (as would be the case here) but it does happen when an analysis is selected. Not very consistent in my opinion. I must add that in general I'm not a big fan of the scrolling when you click on a note. I don't see the purpose, really.

PR #3225, opened 2019-05-03
show footnote if intercept only model is selected

vandenman:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/100.

afbeelding


AlexanderLyNL:
The main conclusion of the ANOVA table is in the footnote. @EJWagenmakers Does this suffice? Or does it need to be conveyed more prominently? The table itself, the rows "Regression", "Residual" and the dashes might suggest that the numbers are still going to appear, which is not the case here. Instead of the table, we could also present just one row with the note in it.

PR #3229, opened 2019-05-06
 RM ANOVA - Fix sphericity test with ss type 1 

JohnnyDoorn:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/92


Kucharssim:
The fix seems correct!

Issue #137, opened 2015-01-07
The way we express interactions;     age:gender   or     age*gender


jonathon-love:
@raviselker

raviselker:
SPSS uses the "*" sign for interactions (see screen shot), so I think it's best to adopt this. Another reason to use the "*" sign for interactions is that an interaction is basically a multiplication of the variables in that interaction. The "*" sign is often used for multiplication while the ":" sign isn't (to my knowledge at least). Therefore, it makes more sense to use the "*" sign.

![image](https://cloud.githubusercontent.com/assets/4669197/5645404/3a4278d8-966f-11e4-9899-004239d7f497.png)

jonathon-love:
here are a few different options:

1. target + trialType + trialType:target
2. target + trialType + trialType*target
3. target + trialType + trialType * target
4. target + trialType + trialType∗target
5. target + trialType + trialType ∗ target
6. target + trialType + trialType ∗ target

i think i like 5.

what do you think @EJWagenmakers ?

EJWagenmakers:
I like the LaTeX \times sign, similar to an "x". This is the symbol that
":" and "*" try to emulate.
Cheers,
E.J.

********************************************
Eric-Jan Wagenmakers
Department of Psychological Methods, room 2.09
University of Amsterdam
Weesperplein 4
1018 XA Amsterdam
The Netherlands

Web: ejwagenmakers.com
Book: bayesmodels.com
Stats: jasp-stats.org
Email: EJ.Wagenmakers@gmail.com
Phone: (+31) 20 525 6420

“Man follows only phantoms.”
Pierre-Simon Laplace, last words
********************************************

jonathon-love:
1. target + trialType + trialType:target
2. target + trialType + trialType*target
3. target + trialType + trialType * target
4. target + trialType + trialType∗target
5. target + trialType + trialType ∗ target
6. target + trialType + trialType ∗ target
7. target + trialType + trialType×target
8. target + trialType + trialType × target
9. target + trialType + trialType × target

yeah,  i like 8.

raviselker:
In 8, the '×' sign could be mistaken for the letter x because you don't add any whitespace between the factors and the sign. So I would go for 6 or 9.

EJWagenmakers:
9 then

********************************************
Eric-Jan Wagenmakers
Department of Psychological Methods, room 2.09
University of Amsterdam
Weesperplein 4
1018 XA Amsterdam
The Netherlands

[...]

PR #3237, opened 2019-05-08
CFA - changes to unicode arrows & remove "first-order" from table titles

vankesteren:
Fixes [#137](https://github.com/jasp-stats/jasp-test-release/issues/137)

Also apparently OSX does not support full unicode. We should use `\u2192` (→) and `\u2194` (↔) because those do work, even though there are better-looking options


vankesteren:
I've added some improvements to CFA in general. Most notably, the bootstrap procedure is now interruptible.

PR #3241, opened 2019-05-08
Changed default behaviour of Bayes linear regression and added footno…

AlexanderLyNL:
…te for the beta-binomial

Addresses https://github.com/jasp-stats/INTERNAL-jasp/issues/213

@vandenman Did Merlise fixed the beta-binomial issue with nuisance parameters? If she did, please reject this PR. I changed the default on the model probabilities to the uniform prior.

### DIFF
--- a/JASP-Engine/JASP/R/regressionlinearbayesian.R
+++ b/JASP-Engine/JASP/R/regressionlinearbayesian.R
@@ -449,14 +449,21 @@ RegressionLinearBayesian <- function (
 	}
 
 	# select the type of model prior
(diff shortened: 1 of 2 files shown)


TimKDJ:
So.. what's the status on this? Has Merlise responded, or should we include a custom BAS package for the next release?

AlexanderLyNL:
I think that a custom build is most practical, as it also takes some time to get through to cran.

TimKDJ:
Alright so if we ship our own version of BAS then this PR can be closed, right?

PR #3247, opened 2019-05-10
Updated .checkMultinomialCounts() & changed Alt. hypotheses to Test Values

ASarafoglou:
- only observed counts must be integer values; expected counts can be either integers or proportions

- added a check that observed and expected counts vector cannot get too big. otherwise the computation takes up too much memory and the analysis crashes

- "Alt. hypothesis" is relabeled to "Test Values" which is consistent with the Bayesian Binomial Test. What the user is specifying in the multinomial test is actually the null hypothesis and not the alternative hypothesis


ASarafoglou:
I will add a helpfile for this analysis on Monday.

JASP-Engine/JASP/R/multinomialtest.R
```diff
@@ -643,7 +643,13 @@ MultinomialTest <- function (dataset = NULL, options, perform = "run",
     .quitAnalysis(paste0(variable, "variable contains negative values"))
   }
 
-  if (!all(counts == round(counts))) {
+  # only applies for observed counts, expected counts can be proportions
+  if (!expectedCounts & !all(counts == round(counts))) {
```
TimKDJ: I'm guessing you meant to use  `&&` here?

JASP-Engine/JASP/R/multinomialtest.R
```diff
@@ -643,7 +643,13 @@ MultinomialTest <- function (dataset = NULL, options, perform = "run",
...
     .quitAnalysis(paste0(variable, "variable must contain only integer values."))
   }
+
+  # a vector which is too big, ittakes up too much memory
+  if (sum(counts > 6e8)) {
```
TimKDJ: What do you base `6e8` on?

(Is it completely necessary to allocate a vector of size `sum(counts)`?)

ASarafoglou:
The function rep() breaks if the vector is equal or bigger than 7e8. The
way we conduct both the frequentist and the Bayesian multinomial test needs
the rep() function. However, we can (and should) improve this in the next
JASP release? I don't think it is necessary to have this argument in there,
but then we need to rewrite both analyses. I can try to fix it now, but I
suggest to put this in the To-Do for the next version?

JASP-Engine/JASP/R/multinomialtest.R
ASarafoglou: Corrected in latest commit.

PR #3248, opened 2019-05-10
Set changed analysis title directly in JS

TimKDJ:
Fixes jasp-stats/jasp-test-release#107

For the time being I changed the analysis name fallback (after providing `""` to the title) to be the last correct title, but the default title thing still needs to be implemented. I suppose we'd need a `defaultTitle` field in `analyses.json` stored in the JASP file?


JorisGoosen:
Well I'll take care of that in https://github.com/jasp-stats/jasp-test-release/issues/152, it should be fairly straightforward to implement.

PR #3254, opened 2019-05-13
Add error checks & implement them in regression analyses

TimKDJ:
Fixes the R side of things for jasp-stats/jasp-test-release/issues/131


JASP-Engine/JASP/R/commonerrorcheck.R
```diff
@@ -617,4 +620,107 @@
...
+      
+    components <- term$components
+    if (length(components) < 2)
+      next
+
+    numChildren <- 2^length(components) - 1
+    inclusion <- sapply(modelTerms, 
+      function(termToFind) {
+          componentsToFind <- termToFind$components
+          if (sum(componentsToFind %in% components) == length(componentsToFind))
```
vandenman: `return(sum(componentsToFind %in% components) == length(componentsToFind))`

JASP-Engine/JASP/R/commonerrorcheck.R
```diff
@@ -617,4 +620,107 @@
...
+  # Args:
+  #   dataset: JASP dataset.
+  #   target: String vector indicating the target variables.
+  #   grouping: String vector indicating the grouping variables.
+  #   groupingLevel: Vector indicating the level of each of the grouping variables.
+  
+  result <- list(error=FALSE, errorVars=NULL)
+  
+  findMissingValues <- function(x) {
+    return(any(is.na(x)))
```
vandenman: `anyNA` should do the trick

JASP-Engine/JASP/R/commonerrorcheck.R
```diff
@@ -617,4 +620,107 @@
...
+  result <- list(error=FALSE, errorVars=NULL)
+  
+  findNegativeValues <- function(x) {
+    return(any(x < 0))
+  }
+  
+  for (v in target) {
+    
+    if (is.factor(dataset[[.v(v)]])) { # Coerce factor to numeric.
+      dataset[[.v(v)]] <- as.numeric(as.character(dataset[[.v(v)]]))
```
vandenman: It's nice to deal with factors as well but perhaps we shouldn't because the `as.numeric(as.character(` may introduce NAs. And `findNegativeValues` does not seem to like those:
```r
any(c(NA, 1, 1) < 0)
[1] NA
any(c(NA, -1, 1) < 0) # I don't know why this does work..
[1] TRUE
```
If people want to use factors as numeric, I'd say they should just read them in as numeric.

An additional downside of the current approach is that this function will create a copy of the dataset (if there are any factors), rather than a copy of the data for the variable at hand. Probably that doesn't really matter though, since it only happens once.

TimKDJ: In general people shouldn't, but let's not limit anyone ;-) 
I now at least deal with the NA's

JASP-Engine/JASP/R/commonerrorcheck.R
TimKDJ: Love your knowledge of these niche R functions

PR #3268, opened 2019-05-16
Update to RM ANOVA effect sizes

JohnnyDoorn:
- Changed the qml of the Bayesian correlation to be more uniform to each other (and freq correlation), and the help files.
- Eta squared is now correct for RM ANOVA.
- Added generalized eta squared.


JohnnyDoorn:
SPSS does not report eta squared, but I checked with some other resources.

PR #3273, opened 2019-05-18
ReliabilityAnalysis - fix reverse scaled items

vandenman:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/52

@TimKDJ I added a unit test specifically for the data set where the results changed. I load this data set via 
```r
  datapath <- file.path(jasptools:::.pkgOptions$data.dir, "Data Library", "1. Descriptives", "Fear of Statistics.csv")
  data <- read.table(datapath, header = TRUE)
```
but it looks a bit hacky. Is there a better way? Copying the data set to `github/jasp-desktop/JASP-Tests/R/tests/datasets/` looks a bit weird because then we get copies of data sets.


TimKDJ:
With #3280 you can just type `jasptools::run("ReliabilityAnalysis", "Fear of Statistics.csv", options)`

TimKDJ:
If you rebase & change the dataset loading line then this PR is good to go

PR #3292, opened 2019-05-23
Small descriptives / binom fixes

TimKDJ:
Fixes jasp-stats/jasp-test-release#99 and jasp-stats/jasp-test-release#98

Today I learned that `max(numeric(0))` gives `-Inf`


AlexanderLyNL:
The supremum (max for bounded sets) of the empty set is - inf, and the infimum (min for bounded sets) is inf.

PR #3294, opened 2019-05-23
RM ANOVA - fix crash in posthoc when variance is very large

JohnnyDoorn:
Fix crash when variables have unusually large variance
Fix to indexing the fields object for z.ratio (now more dynamic because 7 does not always work)
@Kucharssim is this OK with you?


Kucharssim:
Sure!

PR #3298, opened 2019-05-24
Linear regression small fixes

Kucharssim:
@JohnnyDoorn in the end it was a bit more than one line, pls check out whether you agree with this :)

https://github.com/jasp-stats/jasp-test-release/issues/161


JohnnyDoorn:
Just checked this and it looks fine! So can be merged.

PR #3299, opened 2019-05-24
Prevent image adjustments if plot object is from an old file

TimKDJ:
If you have an old JASP file you should not be able to change ppi, background color, size or save it.
Usually this will break, so analyses must be recomputed first.

With this PR the user is asked if he wants to refresh the analysis/analyses, except when simply moving JASP between different dpi screens.
As the user isn't specifically requesting a ppi change I think that behaviour would be weird otherwise.

### DIFF
--- a/JASP-Desktop/mainwindow.cpp
+++ b/JASP-Desktop/mainwindow.cpp
@@ -188,7 +188,9 @@ void MainWindow::makeConnections()
 
 	connect(this,					&MainWindow::saveJaspFile,							this,					&MainWindow::saveJaspFileHandler,							Qt::QueuedConnection);
 	connect(this,					&MainWindow::imageBackgroundChanged,				_engineSync,			&EngineSync::imageBackgroundChanged							);
+	connect(this,					&MainWindow::ppiChanged,                        	_engineSync,			&EngineSync::ppiChanged                                     );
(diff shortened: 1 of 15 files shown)


JASP-Desktop/mainwindow.cpp
```diff
@@ -524,11 +525,47 @@ void MainWindow::dataSetChanged(DataSet * dataSet)
...
+
+
+void MainWindow::plotPPIChangedHandler(int ppi, bool wasUserAction)
+{
+    emit ppiChanged(ppi);
+
+	if (!hasOldAnalyses())
+	{
+		_engineSync->refreshAllPlots();
+		return;
```
JorisGoosen: It would be nicer to remove the `return` here and change `} if (wasUserAction)` to `else if(wasUserAction && MessageForwarder::showYesNo("Version incompatibility", "Your analyses were created in an older version of JASP, to change the PPI of the images they must be refreshed first.\n\nRefresh all analyses?"))`

TimKDJ:
- Fixed a bug where resized plots were not correctly retrieved from the state
- Added a signal to revert the image holder size if a user declines the analysis refresh
- Cleaned up the image editing stuff a little in analysis.js

PR #3308, opened 2019-05-28
Fix jasp-stats/jasp-test-release#177

TimKDJ:
Seems to work nicely now. @ASarafoglou: are the column names of the eprops table used anywhere? Because they are no longer base64 encoded once they leave `.generateExpectedProps`


ASarafoglou:
Yes they are shown again in the results table, but the user is not able to
adjust them (it’s just H_1, ... H_5).

Op di 28 mei 2019 om 06:10 schreef Tim de Jong <notifications@github.com>

TimKDJ:
I mean the Expected Counts variable, that one is user supplied

ASarafoglou:
Ah! No it is not used anymore.

Op di 28 mei 2019 om 08:12 schreef Tim de Jong <notifications@github.com>

PR #3127, opened 2019-03-24
andyField AN(C)OVA

Kucharssim:
Part of this PR: https://github.com/jasp-stats/jasp-desktop/pull/2990

Added features to AN(C)OVA:

- Šidák correction for post hoc tests
- Bootstrapping for post hoc tests
- Bootstrapping for marginal means (including interactions)
- (ANCOVA only) plotting covariate on the horizontal axis
- added unit tests for AN(C)OVA based on Field materials (checked with SPSS)

PR #3310, opened 2019-05-28
Types tweaking

AlexanderLyNL:
- ANCOVAs - Descriptives Plots: Allow covariates on the horizontal axis based on Simon's work: https://github.com/jasp-stats/jasp-desktop/pull/3127
- Restrict: Bayesian linear regression: Covariates to only take scale
- Restrict: CFA: Factors only take scale
- Reorder some symbols. For frequency analyses the ordinals are shown before scale in the counts

Continues:
https://github.com/jasp-stats/jasp-desktop/pull/3304

Discussion:
https://github.com/jasp-stats/INTERNAL-jasp/issues/278

### DIFF
--- a/Resources/ANOVA/qml/Ancova.qml
+++ b/Resources/ANOVA/qml/Ancova.qml
@@ -185,12 +185,10 @@ Form
 		
 		VariablesForm {
 			height: 200
-            AvailableVariablesList { name: "descriptivePlotsVariables"; title: qsTr("Factors");         source: ["fixedFactors", "covariates"]	}
-			AssignedVariablesList {	name: "plotHorizontalAxis";			title: qsTr("Horizontal Axis"); singleVariable: true	}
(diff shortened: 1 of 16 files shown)


AlexanderLyNL:
@JohnnyDoorn I now allow covariates to be used on the horizontal axis for the Descriptive Plots. Is this reasonable?

Kucharssim:
@AlexanderLyNL plotting covariate on the horizontal axis was implemented here: https://github.com/jasp-stats/jasp-desktop/pull/3127, so it is fine. But note that the horizontal axis can also display factors as well (the plot will change depending on whether it's a factor or a covariate).

AlexanderLyNL:
@Kucharssim Awesome! I was quite surprised that it worked and didn't know where this black magic came from. Great work. The horizontal axis now is able to take "everything" so I removed the symbol next to it again.

PR #3331, opened 2019-06-05
Show the error message when abtest fails

TimKDJ:
@quentingronau I also changed the minimum number of samples to 100 (like the Bayesian ANOVA's), do you agree with this?


quentingronau:
Yes, that is a good idea. With respect to the error messages: I think showing them is much better then what we had before, however, in the future, we could think about improving this further by matching the error messages (JASP-specific) with the names of the input options. For instance, in the package, "y1" is the number of successes in group 1, so the error message will refer to "y1" and not to how this is called in JASP.

TimKDJ:
Yeah that'd definitely be a good idea. I'll add some proper error handling once it becomes jaspResults.

PR #3349, opened 2019-06-11
Prevent analyses from adding duplicate citations

TimKDJ:


JorisGoosen:
The code looks fine, although it does mess up the order of citations. Is that a problem?

TimKDJ:
I figured that it'd make sense in this context, as references are always listed alphabetically

vankesteren:
I am a very happy boy with this update. 😄

PR #3353, opened 2019-06-12
Add the possibility to mix the terms of different sources for a VariablesList

boutinb:
Solves issue https://github.com/jasp-stats/jasp-issues/issues/384


boutinb:
@JohnnyDoorn Should this mixedModelTerms property must also be used for the Marginal Means? And also in the Bayesian version for Post Hoc Test and Descriptives Plots?

JohnnyDoorn:
Great! for now only for frequentist post hoc tests.

PR #3356, opened 2019-06-14
Fixes to bain module

TimKDJ:
I removed redundant `format`'s and `.clean`'s and tidied up some functions. Furthermore I fixed

- dependencies in a few plots (e.g., the bayes factor comparison plot of the independent samples ttest didn't update when you changed the grouping variable)
- error propagation anova's/regression: if the model cannot be computed then don't try to compute stuff dependent on the model (this fixes the unexpected errors)
- error propagation ttests: if results for one variable cannot be computed, still compute for others (this still needs a proper $addRows() that adds NaN's)
- test value in one sample ttest was not used and could not be negative
- whitespace in the model constraint text area no longer errors the analysis
- the styling to be more inline with the JASP R styleguide

The data reading and descriptives functions overlap and should probably be simplified further

Partially fixes jasp-stats/jasp-issues#380 (I'm not sure when variables are allowed to occur in multiple hypotheses; I think this is more of a bain package parser issue)

@koenderks maybe you can test the changes and we can see if you agree with them.

### DIFF
--- a/JASP-Engine/JASP/R/bainancovabayesian.R
+++ b/JASP-Engine/JASP/R/bainancovabayesian.R
@@ -25,47 +25,47 @@ BainAncovaBayesian	 <- function (jaspResults, dataset, options, ...) {
 	dataset <- readList[["dataset"]]
 	missingValuesIndicator <- readList[["missingValuesIndicator"]]
 	
+	bainContainer <- .bainGetContainer(jaspResults, deps=c("dependent", "fixedFactors", "covariates", "model"))
+	
 	### LEGEND ###
 	.bainLegendAncova(dataset, options, jaspResults)
 	
 	### RESULTS ###
-	.bainAncovaResultsTable(dataset, options, jaspResults, missingValuesIndicator, ready)
+	.bainAncovaResultsTable(dataset, options, bainContainer, missingValuesIndicator, ready)
 	
 	### BAYES FACTOR MATRIX ###
-	.bainBayesFactorMatrix(dataset, options, jaspResults, ready, type = "anova")
+	.bainBayesFactorMatrix(dataset, options, bainContainer, ready, type = "ancova")
 	
 	### COEFFICIENTS ###
-	.bainAncovaCoefficientsTable(dataset, options, jaspResults, ready)
+	.bainAncovaCoefficientsTable(dataset, options, bainContainer, ready)
 	
 	### BAYES FACTOR PLOT ###
-	.bainAnovaBayesFactorPlots(dataset, options, jaspResults, ready)
+	.bainAnovaBayesFactorPlots(dataset, options, bainContainer, ready)
 	
 	### DESCRIPTIVES PLOT ###
-	.bainAnovaDescriptivesPlot(dataset, options, jaspResults, ready, type = "ancova")
+	.bainAnovaDescriptivesPlot(dataset, options, bainContainer, ready, type = "ancova")
 }
 
-.bainAncovaResultsTable <- function(dataset, options, jaspResults, missingValuesIndicator, ready){
+.bainAncovaResultsTable <- function(dataset, options, bainContainer, missingValuesIndicator, ready){
 
-	if(!is.null(jaspResults[["bainTable"]])) return() #The options for this table didn't change so we don't need to rebuild it
+	if(!is.null(bainContainer[["bainTable"]])) return() #The options for this table didn't change so we don't need to rebuild it
 
-	variables 											<- c(options$dependent, options$fixedFactors, unlist(options$covariates))
-	bainTable                      	<- createJaspTable("Bain ANCOVA Result")
-	jaspResults[["bainTable"]]     	<- bainTable
-	bainTable$dependOn(options =c("dependent", "fixedFactors", "covariates", "model"))
+	variables <- c(options$dependent, options$fixedFactors, unlist(options$covariates))
+	bainTable <- createJaspTable("Bain ANCOVA Result")
 
-	bainTable$addColumnInfo(name="hypotheses", 				type="string", title="")
-	bainTable$addColumnInfo(name="BF", 						type="number", format="sf:4;dp:3", title= "BF.c")
-	bainTable$addColumnInfo(name="PMP1", 					type="number", format="sf:4;dp:3", title= "PMP a")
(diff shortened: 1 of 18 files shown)


JASP-Engine/JASP/R/common.R
```diff
@@ -2202,6 +2202,18 @@ as.list.footnotes <- function(footnotes) {
   return(NULL)
 }
 
+.suppressGrDevice <- function(plotFunc) {
```
TimKDJ: @vandenman I figured we could make this a convenience function. I think you also had to do the png trick a few times, right?

JASP-Engine/JASP/R/bainancovabayesian.R
```diff
@@ -122,200 +117,185 @@ BainAncovaBayesian	 <- function (jaspResults, dataset, options, ...) {
...
 			bainResult <- Bain::Bain_ancova_cm(X = inpt[[1]], dep_var = inpt[[2]], covariates = inpt[[3]], group = inpt[[4]], hyp = inpt[[5]])
-			jaspResults[["bainResult"]] <- createJaspState(bainResult)
-			jaspResults[["bainResult"]]$dependOn(options =c("dependent", "fixedFactors", "covariates", "model"))
+			bainContainer[["bainResult"]] <- createJaspState(bainResult)
 		})
 	}
 
-	if(class(p) == "try-error"){
-		bainTable$setError("An error occurred in the analysis. Please double check your variables.")
+	if (inherits(p, "try-error")) {
```
AlexanderLyNL: There's an isTryError function in common that does this if I remember correctly

JASP-Engine/JASP/R/bainanovabayesian.R
```diff
@@ -125,181 +119,193 @@ BainAnovaBayesian <- function (jaspResults, dataset, options, ...) {
...
-				jaspResults[["descriptivesPlot"]]$dependOn(options =c("descriptivesPlot", "fixedFactors", "dependent", "model"))
-				jaspResults[["descriptivesPlot"]]$position <- 4
-		}
-	} else if(options[["descriptivesPlot"]]){
-			errorPlot <- createJaspPlot(plot = NULL, title = "Descriptives Plot")
-			errorPlot$setError("Plotting not possible: No analysis has been run.")
-			jaspResults[["descriptivesPlot"]] <- errorPlot
-			jaspResults[["descriptivesPlot"]]$dependOn(options =c("descriptivesPlot", "fixedFactors", "dependent", "model"))
-			jaspResults[["descriptivesPlot"]]$position <- 4
+.bainAnovaDescriptivesPlot <- function(dataset, options, bainContainer, ready, type = "anova") {
```
AlexanderLyNL: Does this need to be very different from the descriptives plot we already have in the other analyses?

JASP-Engine/JASP/R/bainttestbayesianindependentsamples.R
```diff
@@ -230,27 +230,28 @@ BainTTestBayesianIndependentSamples <- function(jaspResults, dataset, options, .
...
 	  descriptivesTable$dependOn(options =c("variables", "descriptives", "descriptivesPlotsCredibleInterval", "groupingVariable"))
 		descriptivesTable$position <- 2
 
 	  descriptivesTable$addColumnInfo(name="v",                    title = "", type="string")
 		descriptivesTable$addColumnInfo(name="group",                title = "Group", type="string")
 	  descriptivesTable$addColumnInfo(name="N",                    title = "N", type="integer")
-	  descriptivesTable$addColumnInfo(name="mean",                 title = "Mean", type="number", format="sf:4;dp:3")
-	  descriptivesTable$addColumnInfo(name="sd",                   title = "sd", type="number", format="sf:4;dp:3")
-	  descriptivesTable$addColumnInfo(name="se",                   title = "se", type="number", format="sf:4;dp:3")
+	  descriptivesTable$addColumnInfo(name="mean",                 title = "Mean", type="number")
```
AlexanderLyNL: Am I correct that we do not need to formatting in general anymore, because formatting is controlled by jaspResults?

JASP-Engine/JASP/R/bainttestbayesianindependentsamples.R
```diff
@@ -122,103 +121,104 @@ BainTTestBayesianIndependentSamples <- function(jaspResults, dataset, options, .
...
+                               "hypothesis[type2]" = "H1: Smaller", "BF[type2]" = BF_01, "pmp[type2]" = PMP_1)
+        } else if (options$hypothesis == "_4type") {
+            row <-list(Variable=variable, "hypothesis[type1]" = "H1: Bigger", "BF[type1]"= "", "pmp[type1]" = PMP_1,
+                               "hypothesis[type2]" = "H2: Smaller", "BF[type2]" = BF_01, "pmp[type2]" = PMP_0)
+        } else if (options$hypothesis == "allTypes") {
             row <-list(Variable=variable,
                                "type[equal]" = "H0: Equal",
                                "BF[equal]"= "",
-                               "pmp[equal]" = .clean(PMP_0),
+                               "pmp[equal]" = PMP_0,
```
AlexanderLyNL: I guess that cleaning is not necessary anymore, because that's done by jaspResults. Is that correct?

JASP-Engine/JASP/R/bainanovabayesian.R
TimKDJ: No idea

JASP-Engine/JASP/R/bainttestbayesianindependentsamples.R
TimKDJ: `format="sf:4;dp:3"` is standard for number, so if you don't want anything fancy specifying `type="number"` is sufficient.

JASP-Engine/JASP/R/bainttestbayesianindependentsamples.R
TimKDJ: Yep jaspResults does this now in the aptly named macro `TO_INFINITY_AND_BEYOND`

JASP-Engine/JASP/R/common.R
vandenman: Well this looks like a fine way to solve it. Another idea is to always open a fake graphics device before an analysis starts and close it again after the analyses is finished (For the rare case that some analysis calls `par` or something like that).

PR #3358, opened 2019-06-18
 Adding Q-Q Plot to Descriptives

gibbona1:
Added a Q-Q Plot to the plots section of Descriptives. Can generate a plot for each variable (scale), and can also split and generate multiple plots per variable. Similar in style to ANOVA QQ plot.
2 changes to the R file.
QML change is a simple checkbox

### DIFF
--- a/JASP-Engine/JASP/R/descriptives.R
+++ b/JASP-Engine/JASP/R/descriptives.R
@@ -132,6 +132,8 @@ Descriptives <- function(jaspResults, dataset, options) {
       # remove missing values from the grouping variable
       dataset           <- dataset[!is.na(qqSplitFactor), ]
       for(var in variables){ 
+        if(!is.null(QQPlots[[var]]))
+          next
         deeperQQPlots <- createJaspContainer(paste0(var))
         deeperQQPlots$dependOn(optionContainsValue=list(variables=var))
         QQPlots[[var]] <- deeperQQPlots
@@ -1246,8 +1248,6 @@ Descriptives <- function(jaspResults, dataset, options) {
     p<-NULL
   }
   descriptivesQQPlot <- createJaspPlot(plot=p, width=400, aspectRatio=1, title=title)
-  if (!is.null(levelName))
-    descriptivesQQPlot$dependOn(optionContainsValue=list(variables=qqvar))
   return(descriptivesQQPlot)
 }
 
--- a/JASP-Engine/JASP/R/descriptives.R
+++ b/JASP-Engine/JASP/R/descriptives.R
@@ -111,6 +111,45 @@ Descriptives <- function(jaspResults, dataset, options) {
       }
     }
   }
+
+  # QQ plots
+  if (options$descriptivesQQPlot) {
+    if(is.null(jaspResults[["QQPlots"]])) {
+      if(length(variables)>1 || length(levels(dataset[[.v(splitName)]]))>1) #there will be more than one Q-Q Plot
+        jaspResults[["QQPlots"]] <- createJaspContainer("Q-Q Plots")
+      else #only one Q-Q Plot
+        jaspResults[["QQPlots"]] <- createJaspContainer("Q-Q Plot")
+      jaspResults[["QQPlots"]]$dependOn(c("descriptivesQQPlot", "variables", "splitby"))
+      jaspResults[["QQPlots"]]$position <- 8
+    }
+    QQPlots <- jaspResults[["QQPlots"]]
+    if(makeSplit) {
+      qqSplitFactor     <- dataset[[.v(splitName)]]
+      if(length(qqSplitFactor)==0)
+        return(createJaspPlot(error="Plotting is not possible: Variable only contains NA!", dependencies="splitby"))
+      #gives the different split values
+      qqSplitLevels     <- levels(qqSplitFactor)
+      # remove missing values from the grouping variable
+      dataset           <- dataset[!is.na(qqSplitFactor), ]
+      for(var in variables){ #allows multiple variables
+        #splits dataset according to split values
+        qqSplitData     <- split(dataset, qqSplitFactor)
+        for( lev in 1:length(qqSplitLevels)){
+          QQPlots[[paste0(var, lev)]] <- .descriptivesQQPlot(dataset=qqSplitData[[lev]], options=options, qqvar=var, levelName=qqSplitLevels[lev])
+        }
+      }
+    }
+    else { #no split
+      for(var in variables){
+        if(is.null(QQPlots[[var]])) {
+          QQPlots[[var]] <- .descriptivesQQPlot(dataset=dataset, options=options, qqvar=var)
+        }
+      }
+    }
+    QQPlots
+  }
+  QQPlots<-NULL
+
   return()
 }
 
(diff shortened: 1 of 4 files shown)


JASP-Engine/JASP/R/descriptives.R
```diff
@@ -1145,4 +1186,90 @@ Descriptives <- function(jaspResults, dataset, options) {
...
+
+  descriptivesQQPlot <- list()
+
+  if (!is.null(qqvar)) {
+
+    #to put a subtitle if there is a split
+    if(!is.null(levelName))
+      descriptivesQQPlot$title <- paste0(options$splitby, ": ", levelName)
+    else
+      descriptivesQQPlot$title <- " "
```
TimKDJ: We still want a title here: the variable name; otherwise we don't know which plot belongs to which variable.
But if you set the title in the beginning already just remove this here.

JASP-Engine/JASP/R/descriptives.R
```diff
@@ -1145,4 +1186,90 @@ Descriptives <- function(jaspResults, dataset, options) {
...
+  if (options$descriptivesQQPlot && is.null(options$variables))
+    return(list(result=NULL))
+
+  descriptivesQQPlot <- list()
+
+  if (!is.null(qqvar)) {
+
+    #to put a subtitle if there is a split
+    if(!is.null(levelName))
+      descriptivesQQPlot$title <- paste0(options$splitby, ": ", levelName)
```
TimKDJ: The other plots (e.g., distribution) simply use the level of the grouping variable as the title, to be consistent I think it's good to do that here as well.

JASP-Engine/JASP/R/descriptives.R
```diff
@@ -1145,4 +1186,90 @@ Descriptives <- function(jaspResults, dataset, options) {
...
-# </editor-fold> HELPER FUNCTIONS BLOCK
\ No newline at end of file
+.descriptivesQQPlot <- function(dataset, options,  qqvar, levelName=NULL) {
+
+  if (options$descriptivesQQPlot && is.null(options$variables))
+    return(list(result=NULL))
+
+  descriptivesQQPlot <- list()
+
+  if (!is.null(qqvar)) {
```
TimKDJ: I don't think `qqvar` can be `NULL`?

JASP-Engine/JASP/R/descriptives.R
```diff
@@ -1145,4 +1186,90 @@ Descriptives <- function(jaspResults, dataset, options) {
...
+      descriptivesQQPlot$title <- paste0(options$splitby, ": ", levelName)
+    else
+      descriptivesQQPlot$title <- " "
+
+    # Hardcode plot dimensions
+    options$descriptivesPlotWidthQQPlot <- 400
+    options$descriptivesPlotHeightQQPlot <- 400
+
+    descriptivesQQPlot$width  <- options$descriptivesPlotWidthQQPlot
+    descriptivesQQPlot$height <- options$descriptivesPlotHeightQQPlot
```
TimKDJ: You don't need these width and height specifications

JASP-Engine/JASP/R/descriptives.R
```diff
@@ -1145,4 +1186,90 @@ Descriptives <- function(jaspResults, dataset, options) {
...
+
+    # Hardcode plot dimensions
+    options$descriptivesPlotWidthQQPlot  <- 400
+    options$descriptivesPlotHeightQQPlot <- 400
+
+    descriptivesQQPlot$width  <- options$descriptivesPlotWidthQQPlot
+    descriptivesQQPlot$height <- options$descriptivesPlotHeightQQPlot
+    descriptivesQQPlot$custom <- list(width="descriptivesPlotWidthQQPlot", height="descriptivesPlotHeightQQPlot")
+    descriptivesQQPlot$data   <- NULL
+    p<-NULL
```
TimKDJ: Don't need to do this

JASP-Engine/JASP/R/descriptives.R
```diff
@@ -1145,4 +1186,90 @@ Descriptives <- function(jaspResults, dataset, options) {
...
+    descriptivesQQPlot$title <- "Q-Q Plot"
+
+    # Hardcode plot dimensions
+    options$descriptivesPlotWidthQQPlot  <- 400
+    options$descriptivesPlotHeightQQPlot <- 400
+
+    descriptivesQQPlot$width  <- options$descriptivesPlotWidthQQPlot
+    descriptivesQQPlot$height <- options$descriptivesPlotHeightQQPlot
+    descriptivesQQPlot$custom <- list(width="descriptivesPlotWidthQQPlot", height="descriptivesPlotHeightQQPlot")
+    descriptivesQQPlot$data   <- NULL
```
TimKDJ: Don't need the four lines above here; this is sort of a mix of old-style plotting and new-style plotting.
I totally understand that it is a little confusing that we have two distinct styles of plotting in JASP right now.

JASP-Engine/JASP/R/descriptives.R
```diff
@@ -111,6 +111,47 @@ Descriptives <- function(jaspResults, dataset, options) {
...
+    if(makeSplit) {
+      qqSplitFactor     <- dataset[[.v(splitName)]]
+      if(length(qqSplitFactor)==0)
+        return(createJaspPlot(error="Plotting is not possible: Variable only contains NA!", dependencies="splitby"))
+      #gives the different split values
+      qqSplitLevels     <- levels(qqSplitFactor)
+      # remove missing values from the grouping variable
+      dataset           <- dataset[!is.na(qqSplitFactor), ]
+      for(var in variables){ 
+        deeperQQPlots <- createJaspContainer(paste0(var))
```
TimKDJ: Here you can add `"variables"` as a dependency (and then you should also check if the specific container is `NULL` (meaning no plots were created for this variable)

JASP-Engine/JASP/R/descriptives.R
```diff
@@ -1145,4 +1186,90 @@ Descriptives <- function(jaspResults, dataset, options) {
...
 }
 
-# </editor-fold> HELPER FUNCTIONS BLOCK
\ No newline at end of file
+.descriptivesQQPlot <- function(dataset, options,  qqvar, levelName=NULL) {
+
+  if (options$descriptivesQQPlot && is.null(options$variables))
+    return(list(result=NULL))
+
+  descriptivesQQPlot <- list()
```
TimKDJ: `descriptivesQQPlot` can already be a jaspPlot here:
```
  title <- qqvar
  if (!is.null(levelName))
    title <- levelName
      
  descriptivesQQPlot <- createJaspPlot(plot=NULL, width=400, aspectRatio=1, title=title)
  if (!is.null(levelName))
    descriptivesQQPlot$dependOn(optionContainsValue=list(variables=var))
```

TimKDJ: So either the dependency for the plot is set on the container if it's grouped, or on the plot itself if it is not grouped

JASP-Engine/JASP/R/descriptives.R
```diff
@@ -1145,4 +1186,90 @@ Descriptives <- function(jaspResults, dataset, options) {
...
+    options$descriptivesPlotWidthQQPlot  <- 400
+    options$descriptivesPlotHeightQQPlot <- 400
+
+    descriptivesQQPlot$width  <- options$descriptivesPlotWidthQQPlot
+    descriptivesQQPlot$height <- options$descriptivesPlotHeightQQPlot
+    descriptivesQQPlot$custom <- list(width="descriptivesPlotWidthQQPlot", height="descriptivesPlotHeightQQPlot")
+    descriptivesQQPlot$data   <- NULL
+    p<-NULL
+  }
+  return(createJaspPlot(plot=p, width=400, aspectRatio=1, title=descriptivesQQPlot$title))
```
TimKDJ: If you create the jaspPlot in the start then simply return `descriptivesQQPlot`

JASP-Engine/JASP/R/descriptives.R
```diff
@@ -111,6 +111,47 @@ Descriptives <- function(jaspResults, dataset, options) {
...
+        #splits dataset according to split values
+        qqSplitData     <- split(dataset, qqSplitFactor)
+        for( lev in 1:length(qqSplitLevels)){
+          QQPlots[[var]][[paste0(var, lev)]] <- .descriptivesQQPlot(dataset=qqSplitData[[lev]], options=options, qqvar=var, levelName=qqSplitLevels[lev])
+        }
+      }
+    }
+    else { #no split
+      for(var in variables){
+        if(is.null(QQPlots[[var]])) {
```
TimKDJ: This will always be `NULL` right now, unless you change the dependencies as suggested, at which point this becomes a useful if-statement

TimKDJ:
Yes this is much better.

One last thing: can you make it "Q-Q **p**lot**s**" instead of "Q-Q Plot" to be inline with the other input option labels?

Screenshot 2019-06-26 at 10 20 24

TimKDJ:
Also remove the capital P please -- labels are always Capital followed by normal. While titles are Capital Followed By Other Capitals.

JASP-Engine/JASP/R/descriptives.R
```diff
@@ -111,6 +111,45 @@ Descriptives <- function(jaspResults, dataset, options) {
...
+      qqSplitFactor     <- dataset[[.v(splitName)]]
+      if(length(qqSplitFactor)==0)
+        return(createJaspPlot(error="Plotting is not possible: Variable only contains NA!", dependencies="splitby"))
+      #gives the different split values
+      qqSplitLevels     <- levels(qqSplitFactor)
+      # remove missing values from the grouping variable
+      dataset           <- dataset[!is.na(qqSplitFactor), ]
+      for(var in variables){ 
+        deeperQQPlots <- createJaspContainer(paste0(var))
+        deeperQQPlots$dependOn(optionContainsValue=list(variables=var))
```
TimKDJ: Maybe I'm missing it but are you checking if the plot was already created?
Because right now I think it still does the plotting again, only you don't see it because it doesn't remove the prior plot until the new one is created again.

JASP-Engine/JASP/R/descriptives.R
```diff
@@ -1146,4 +1185,70 @@ Descriptives <- function(jaspResults, dataset, options) {
...
+    # JASP theme
+    p <- JASPgraphs::themeJasp(p)
+
+  } else {
+
+    p<-NULL
+  }
+  descriptivesQQPlot <- createJaspPlot(plot=p, width=400, aspectRatio=1, title=title)
+  if (!is.null(levelName))
+    descriptivesQQPlot$dependOn(optionContainsValue=list(variables=qqvar))
```
TimKDJ: Same for this one; always check if something is `NULL` if you added dependencies to it; otherwise it will create it again despite the fact it already exists

JASP-Engine/JASP/R/descriptives.R
```diff
@@ -1246,8 +1248,6 @@ Descriptives <- function(jaspResults, dataset, options) {
     p<-NULL
   }
   descriptivesQQPlot <- createJaspPlot(plot=p, width=400, aspectRatio=1, title=title)
-  if (!is.null(levelName))
-    descriptivesQQPlot$dependOn(optionContainsValue=list(variables=qqvar))
```
TimKDJ: You can't remove this, because otherwise no dependencies are set when you do not have a grouping variable.

TimKDJ: Although it should totally be 
`if (is.null(levelName))` rather than  `if (!is.null(levelName))`

TimKDJ:
Good work!

PR #3366, opened 2019-06-20
added piechart to JASPgraphs

vandenman:
part 1 for jasp-stats/jasp-issues/issues/225.

the change to `JASP-Engine/JASPgraphs/man/plotQQnorm.Rd` is because apparently I didn't call `devtools::document` before pushing last time.

From `example("plotPieChart")`:

afbeelding

afbeelding

afbeelding

afbeelding

afbeelding

afbeelding

afbeelding


AlexanderLyNL:
Is there a possibility to have colour blind colours used by default? @koenderks can you have a quick look as you also made some of these plots.

koenderks:
Maybe use that color palette that we use in machine learning? That one was made for quantitative data analysis I believe. But color blind colors seems to me like it should be some kind of global option affecting all plots, doesnt it? I personally prefer the normal colors

vandenman:
I think a colorblind palette is a nice idea for plot editing, but it doesn't look very appealing:
```r
value <- c(25, 25, 50)
gg <- letters[1:3]
g <- plotPieChart(value, gg)
g + ggthemes::scale_fill_colorblind()
```
afbeelding

So I opted for `colorspace::qualitative_hcl()` instead, which is also used in the machine learning module. That looks like this:

afbeelding

EJWagenmakers:
Maybe we can have the user chose among palettes? (I actually liked the colorblind version)

Also, I don't get the numbers across the perimeter -- can we get rid of those and instead present the relevant category numbers (i.e., 25,25,50)? This may be problematic when the numbers become very small, so maybe make it optional to present the numbers in the legend, along the perimeter, or not at all.

Finally, can the legend be ordered from largest (on top) to smallest (on the bottom)? This is more intuitive.

Cheers
E.J.

TimKDJ:
Can you increment JASPgraphs' version number?

PR #3372, opened 2019-06-21
Make ribbonbar behave nicer with a custom menu

TimKDJ:
new <-> old
Screenshot 2019-06-21 at 18 27 17

- shadow no longer overlaps with the custom menu (part of jasp-stats/jasp-issues#381)
- the border of the pressed ribbonbar button no longer "misaligns" with the top/bottom of the ribbonbar


JorisGoosen:
But, to get rid of the shadow, isn't it easier to simply give the customMenu a much higher z-value?

TimKDJ:
Then the custom menu adds a shadow on top of the ribbonbar, which looks even worse

JorisGoosen:
Fair enough!

PR #3373, opened 2019-06-21
Range based for loop - deletion is dangerous

akashrajkn:
The existing code causes a crash if multiple variables are selected in `AssignedVariablesList`.

### DIFF
--- a/JASP-Desktop/analysis/options/optionstable.cpp
+++ b/JASP-Desktop/analysis/options/optionstable.cpp
@@ -45,8 +45,11 @@ Json::Value OptionsTable::asJSON() const
 
 void OptionsTable::deleteOldValues()
 {
-	for(Options *row : _value)
-		delete row;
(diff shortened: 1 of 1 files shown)


JASP-Desktop/analysis/options/optionstable.cpp
```diff
@@ -45,8 +45,11 @@ Json::Value OptionsTable::asJSON() const
 
 void OptionsTable::deleteOldValues()
 {
-	for(Options *row : _value)
-		delete row;
+	auto i = std::begin(_value);
+
+	while(i != std::end(_value)) {
+		_value.erase(i);
```
JorisGoosen: Why would you do this? 

`erase` does not delete a pointer... It simply removes the pointer from the list.
And we do clear afterwards? So this is pointless, the idea here was to avoid a memory leak caused by throwing away pointers to objects without deleting/destroying them..

PR #3378, opened 2019-06-25
Allow intercept in SEM model

TimKDJ:


AlexanderLyNL:
Works for me. 
- I've tried it with "x1 ~ 1" and "x1 ~ x2 + 1" and it works
- Tried this with "x1 ~~ 1" , which doesn't work, but neither does it in R
- Tried this with "x1 =~ 1" , which doesn't work, but neither does it in R

This should cover most of it, but perhaps @Kucharssim can correct me here.

PR #3393, opened 2019-06-28
Replace base64 encoded variables in SEM error message

TimKDJ:
Fix jasp-stats/jasp-test-release#233

### DIFF
--- a/JASP-Engine/JASP/R/semsimple.R
+++ b/JASP-Engine/JASP/R/semsimple.R
@@ -207,6 +207,15 @@
                      USE.NAMES = FALSE)])
 }
 
+.decodeVarsInMessage <- function(encodedVars, message) {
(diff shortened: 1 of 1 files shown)


JASP-Engine/JASP/R/semsimple.R
```diff
@@ -207,6 +207,15 @@
                      USE.NAMES = FALSE)])
 }
 
+.decodeVarsInMessage <- function(encodedVars, message) {
```
AlexanderLyNL: This is great for the problem at hand, but we would like to generalise this to other analyses don't we?

JorisGoosen: Yeah, we could just do this in the setErrorMessage of the jaspObjects/results?
I'm pretty sure we won't find anything else in there that will accidentally transforms into a known columnname after decodeing the base64...

PR #3416, opened 2019-07-09
Manova

TimKDJ:
I exchanged the string checking for a boolean (to be more inline with other analyses) and additionally I fixed the cases where:
- if the analysis was in an unready state and you checked either of the two boxes in Assumptions Checks you would get an unexpected error
- Shapiro computation would only catch one specific try-error instead of all possible try-errors


TimKDJ:
You ok with these changes @JohnnyDoorn?
I mainly set out to fix this:
Screenshot 2019-07-09 at 17 21 50

PR #3421, opened 2019-07-10
Matched rank biserial correlation now handles ties

JohnnyDoorn:


AlexanderLyNL:
For some reason descriptives fails. Could you also add a new unit test for the effect sizes?

TimKDJ:
Yeah the descriptives fail is unrelated; see jasp-stats/INTERNAL-jasp#356

PR #3427, opened 2019-07-12
Cyrillic Text in MatrixPlot

vandenman:
Fixes both issues mentioned in https://github.com/jasp-stats/jasp-issues/issues/407.

1. The cyrillic text in matrixplot. For some reason there were multiple `.v` and `.unv` calls in `.descriptivesMatrixplot`. I removed the `.unv` calls and replaced those by `options[["variables"]]` and did `.v` only once.

2. transparent plot backgrounds for figures. This was pretty easy, but I had to remake many .svgs. The only difference in the svg files is that `stroke: #FFFFFF; fill: #FFFFFF;` became `stroke: none;` (where `#FFFFFF` is the hex code for white).


TimKDJ:
Damn, we forgot to bump the JASPgraphs version number again. I'm making a PR in a bit and will do it

PR #3437, opened 2019-07-17
use JASPgraphs for making axis tick labels

vandenman:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/368 and a couple of other similar methods for creating axis tick labels.

I did a `Ctrl+Shift+F` for `xLabs <- vector("character", length(xticks))` and replaced the code everywhere I saw it with `xLabs <- JASPgraphs::axesLabeller(xticks)`, which should do the same. That guarantees a uniform way to create axes tick labels and we can adjust it for all plots at once if we want to. The change to `JASPgraphs::axesLabeller` should ensure that either all axis labels are displayed in scientific notation or none of them.

The only place where I did not change this is `commonAudit.R`, because there the labels never switch to scientific notation (I can also do it there if @koenderks thinks that's a good idea).

The `digits = 3` is skipped from `JASPgraphs::axesLabeller(xticks)`, because `digits = 3` is the default argument for that function.

### DIFF
--- a/JASP-Engine/JASP/R/ancova.R
+++ b/JASP-Engine/JASP/R/ancova.R
@@ -3176,30 +3176,14 @@ Ancova <- function(dataset=NULL, options, perform="run", callback=function(...) 
     xhigh <- max(pretty(xVar))
     xticks <- pretty(c(xlow, xhigh))
     
-    # format x labels
-    xLabs <- vector("character", length(xticks))
-    for (i in seq_along(xticks)) {
-      if (xticks[i] < 10^6) {
-        xLabs[i] <- format(xticks[i], digits= 3, scientific = FALSE)
-      } else {
-        xLabs[i] <- format(xticks[i], digits= 3, scientific = TRUE)
-      }
-    }
-    
-    # Format y ticks
+      # Format y ticks
     ylow <- min(pretty(yVar))
     yhigh <- max(pretty(yVar))        
     yticks <- pretty(c(ylow, yhigh))
     
-    # format y labels
-    yLabs <- vector("character", length(yticks))
-    for (i in seq_along(yticks)) {
-      if (yticks[i] < 10^6) {
-        yLabs[i] <- format(yticks[i], digits= 3, scientific = FALSE)
-      } else {
-        yLabs[i] <- format(yticks[i], digits= 3, scientific = TRUE)
-      }
-    }
+    # format axes labels
+    xLabs <- JASPgraphs::axesLabeller(xticks)
+    yLabs <- JASPgraphs::axesLabeller(yticks)
     
     p <- JASPgraphs::drawAxis(xName = "Theoretical Quantiles", yName = "Standardized Residuals", xBreaks = xticks, yBreaks = xticks, yLabels = xLabs, xLabels = xLabs, force = TRUE)
     p <- p + ggplot2::geom_line(data = data.frame(x = c(min(xticks), max(xticks)), y = c(min(xticks), max(xticks))), mapping = ggplot2::aes(x = x, y = y), col = "darkred", size = 1)
--- a/JASP-Engine/JASP/R/correlationbayesianpairs.R
+++ b/JASP-Engine/JASP/R/correlationbayesianpairs.R
@@ -1023,25 +1023,9 @@ CorrelationBayesianPairs <- function(dataset=NULL, options, perform="run", callb
         
     yticks <- pretty(c(ylow, yhigh))
     
-    # format x labels
-    xLabs <- vector("character", length(xticks))
-    for (i in seq_along(xticks)) {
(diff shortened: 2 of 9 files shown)


TimKDJ:
Can you install the latest jasptools version and see if it's workable when JASPgraphs gets updated?

vandenman:
What do you mean? Reinstall jasptools and `manageTestPlots()`? I did sort of hack this by not updating the version number of JASPgraphs until after I updated the figures. Oh and I get this a lot when I call `manageTestPlots()`:

```
Failed with error:  'Package 'Rcpp' version 1.0.1 cannot be unloaded:
 Error in unloadNamespace(package) : namespace 'Rcpp' is imported by 'httpuv', 'readxl', 'abtest', 'haven', 'plyr', 'tidyselect', 'gdtools', 'promises', 'BayesFactor', 'scales', 'openxlsx', 'dplyr', 'htmltools', 'later', 'vdiffr', 'fs', 'jaspResults' so cannot be unloaded
```

TimKDJ:
> What do you mean? Reinstall jasptools and manageTestPlots()?

Yeah, that's what I meant. I'm mainly just curious if it's workable or not.

> Oh and I get this a lot when I call manageTestPlots():

I've noticed this too, at times, I'm guessing it has something to do with the unholy trinity of (1) the fact that we have the `require(Rcpp)` at the start of every analysis, (2) that jasptools adjusts the order of the .libPaths() and (3) that you can have multiple different versions of Rcpp installed. Haven't looked into it too much yet though.

JASP-Engine/JASP/R/ancova.R
```diff
@@ -3176,30 +3176,14 @@ Ancova <- function(dataset=NULL, options, perform="run", callback=function(...)
...
-        yLabs[i] <- format(yticks[i], digits= 3, scientific = FALSE)
-      } else {
-        yLabs[i] <- format(yticks[i], digits= 3, scientific = TRUE)
-      }
-    }
+    # format axes labels
+    xLabs <- JASPgraphs::axesLabeller(xticks)
+    yLabs <- JASPgraphs::axesLabeller(yticks)
     
     p <- JASPgraphs::drawAxis(xName = "Theoretical Quantiles", yName = "Standardized Residuals", xBreaks = xticks, yBreaks = xticks, yLabels = xLabs, xLabels = xLabs, force = TRUE)
```
TimKDJ: Can't `JASPgraphs::drawAxis` take care of `JASPgraphs::axesLabeller(xticks)`? Are there situations we do not want this?

vandenman: I actually want to get rid of `drawAxis` and bring the JASPgraphs API closer to the usual ggplot2 code. I would prefer something like:
```r
ggplot2::ggplot(data = dat, mapping = ggplot2::aes(...)) + 
  # ggplot2::geom_point() but with our defaults
  JASPgraph::geom_point() +
  # mimics ggplot2::scale_x_continuous and has an argument called labels, which could default to axesLabeller
  JASPgraphs::scale_x_continuous(xName = ...) + 
  # creates the lines on the axes (mimic's base r, bty = "n")
  JASPgraphs::geom_rangeframe() + 
  # add the JASP theme
  JASPgraphs::themeJasp()
```
That's something for the future though, when I'm happy about how `JASPgraphs::geom_rangeframe()` works. I also want the API to be stable before I start to deprecate things and rewrite plots to a common style.

vandenman:
> Yeah, that's what I meant. I'm mainly just curious if it's workable or not.

Well forcing `manageTestPlots()` whenever JASpgraphs is updated is a bit much I think. In particular when you know only a few plots are affected that can be a bit tedious. But that's completely dependent on how easy/ difficulty it is to run all unit tests in general. My experience with it was fine though (except that I only updated the version number after updating the plots).

TimKDJ:
It'll only require you to validate changed plots though. If something didn't change, then you don't have to validate it either. I think in general it's good to always ensure that your changes didn't have any unintended side effects :)

JASP-Engine/JASP/R/ancova.R
TimKDJ: But.. then my question stands. Are there situations where we don't want the `axesLabeller` functionality? Because until you've done a rewrite it'll be something people will have to think of.

vandenman: well `axesLabeller` doesn't quite work for factors.
```r
JASPgraphs::axesLabeller(factor(letters[1:3]))
[1] "1" "2" "3"
attr(,"levels")
[1] "a" "b" "c"
```

Aside from that I think it can be used for everything, so there is no real reason not to make it the default (sorry I missed that the first time). I'll change `drawAxis()` so that it uses `axesLabeller` automatically if xBreaks or yBreaks are supplied.

PR #3440, opened 2019-07-19
Convert plot functions to recordedPlots in jaspPlot

TimKDJ:
It was not possible to save images that were created with plot functions in jaspResults.
(Plus it's better not to store the functions with their environments, as this can be quite sizable.)

I'm not entirely sure why you use `*.findName("")` on the lists after checking `*.containsElementNamed("")` @JorisGoosen, is this simply preference over `[""]` indexing? Or something else?


JorisGoosen:
Hey @TimKDJ I am not sure why I did that there. Maybe I missed that there was indexing by name? 
I suppose just indexing on it directly would be a bit cleaner actually.

You wanna change that in the PR?

TimKDJ:
Yeah changed it

PR #3442, opened 2019-07-22
Rewriting Contingency Tables to jaspResults

gibbona1:
**contingencytables.R**
Updated Contingency Tables to jaspResults  
Split up the computation of results and creation of tables. 
Each table now calls its own results function
Previously most errors were dealt with by leaving the table empty and adding a footnote. 
I replaced this with putting each result function inside a try() statement and using setError() if needed
Fixed a poor naming error in the main table's results function, which duplicated the "% within row" data 
when multiple options in Cells were selected. The previous function must have ignored this duplication, 
but this meant that the expected table in the unit test was wrong. See below.

**ContingencyTables.qml**
Now uses jaspResults

**test-contingencytables.R**
Main table needed to be adjusted because the previous code duplicated the "% within row" title and 
figures
Once I removed these the old table only disagreed with some character strings, like duplicated layer 
levels.
Chisquare table needed extra unicode expectations, but this might be because when I run the analysis 
the chi symbol is broken down into 3 messy unicode characters. If Travis doesn't do this I can just take 2 
unicode characters out of each part of the expected table and try again.

### DIFF
--- a/JASP-Engine/JASP/R/contingencytables.R
+++ b/JASP-Engine/JASP/R/contingencytables.R
@@ -311,6 +311,7 @@ ContingencyTables <- function(jaspResults, dataset, options, ...) {
     
     # Add columns to table
     .crossTabLayersColumns(crossTabGamma, analysis)
+    #crossTabGamma$addColumnInfo(name = "type[gammaCoef]",  title = "", type="string")
     crossTabGamma$addColumnInfo(name = "value[gammaCoef]", title = "Gamma",          
                                 type = "number")
     crossTabGamma$addColumnInfo(name = "Sigma[gammaCoef]", title = "Standard Error", 
@@ -347,6 +348,7 @@ ContingencyTables <- function(jaspResults, dataset, options, ...) {
     
     # Add columns to table
     .crossTabLayersColumns(crossTabKendallTau, analysis)
+    #crossTabKendallTau$addColumnInfo(name = "type[kTauB]",  title = "", type="string")
     crossTabKendallTau$addColumnInfo(name = "value[kTauB]", title = "Kendall's Tau-b ", 
                                      type = "number")
     crossTabKendallTau$addColumnInfo(name = "statistic[kTauB]", title = "Z", 
@@ -947,13 +949,6 @@ ContingencyTables <- function(jaspResults, dataset, options, ...) {
 }
 
 .crossTabTestsRows <- function(analysisContainer, var.name, groupList, options, ready, counts.fp) {
-  if(!options$chiSquared && 
-     !options$chiSquaredContinuityCorrection && 
-     !options$likelihoodRatio) 
-    return()
-  if(!is.null(analysisContainer[["resultsChisq"]]))
-    return()
-  
   tests.rows     <- list()
   group.matrices <- groupList$group.matrices
   groups         <- groupList$groups
@@ -1072,10 +1067,6 @@ ContingencyTables <- function(jaspResults, dataset, options, ...) {
     row <- .crossTabLayerNames(row, group)
     tests.rows[[length(tests.rows) + 1]] <- row
   }
-  chisq <- createJaspState(tests.rows)
-  table <- analysisContainer[["crossTabChisq"]]
-  chisq$dependOn(optionsFromObject = table)
-  analysisContainer[["resultsChisq"]] <- chisq
   return(tests.rows)
 }
 
@@ -1162,9 +1153,6 @@ ContingencyTables <- function(jaspResults, dataset, options, ...) {
 }
 
 .crossTabNominalRows <- function(analysisContainer, var.name, groupList, options, ready) {
-  if (!options$contingencyCoefficient && !options$phiAndCramersV && !options$lambda)
-    return()
-  
   nominal.rows   <- list()
   group.matrices <- groupList$group.matrices
   groups         <- groupList$groups
@@ -1212,9 +1200,6 @@ ContingencyTables <- function(jaspResults, dataset, options, ...) {
 }
 
 .crossTabGammaRows <- function(analysisContainer, var.name, groupList, options, ready) {
-  if (!options$gamma)
-    return()
-  
   ordinal.rows   <- list()
   group.matrices <- groupList$group.matrices
   groups         <- groupList$groups
@@ -1253,8 +1238,6 @@ ContingencyTables <- function(jaspResults, dataset, options, ...) {
 }
 
 .crossTabKendallsTauRows <- function(analysisContainer, var.name, groupList, options, ready) {
-  if (!options$kendallsTauB)
-    return()
   kendalls.rows  <- list()
   group.matrices <- groupList$group.matrices
   groups         <- groupList$groups
--- a/JASP-Engine/JASP/R/contingencytables.R
+++ b/JASP-Engine/JASP/R/contingencytables.R
@@ -463,8 +463,8 @@ ContingencyTables <- function(jaspResults, dataset, options, ...) {
   if(isTryError(res))
     table$setError(.extractErrorMessage(res))
   else
-    for (level in 1:length(res)) 
-      table$addRows(res[[level]])
+    for (level in 1:length(res$rows)) 
+      table$addRows(res$rows[[level]], rowNames = res$rownames)
 }
 
 .crossTabChisqAddColInfo <- function(fold, table, options, counts.fp = FALSE) {
@@ -599,7 +599,7 @@ ContingencyTables <- function(jaspResults, dataset, options, ...) {
   return(row)
 }
 
-.crossTabNominalMatrixToRow <- function(analysisContainer, row, counts.matrix, type, ready) {
+.crossTabNominalMatrixToRow <- function(analysisContainer, row, counts.matrix, type, ready, rowname) {
   if(type == "ContCoef") {
     row[["type[ContCoef]"]] <- "Contingency coefficient"
     val <- "contingency"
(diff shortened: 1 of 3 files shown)


TimKDJ:
@gibbona1 can you mark all my comments that you have solved as resolved?

JASP-Engine/JASP/R/contingencytables.R
```diff
@@ -15,1778 +15,1420 @@
...
+      crossTabChisq$addColumnInfo(name = "value[N]", title = "Value", 
+                                  type = "number", format = "sf:4;dp:2")
+    else 
+      crossTabChisq$addColumnInfo(name = "value[N]", title = "Value", 
+                                  type = "integer")
+    
+    crossTabChisq$addColumnInfo(name = "df[N]", title = "df")
+    crossTabChisq$addColumnInfo(name = "p[N]",  title = "p")
+    if (options$VovkSellkeMPR)
+      crossTabChisq$addColumnInfo(name = "MPR[N]", title = "VS-MPR",
```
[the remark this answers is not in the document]
gibbona1: This conversation was hidden so I didn't see it the first time I went through the comments. Fixed now

TimKDJ:
I did some testing and noticed the following issues (see version 0.10.2 to compare behaviour):
- [ ] nominal table: no longer has titles in the rows of the first column when `ready == FALSE`
- [ ] nominal table: it doesn’t add a footnote if calculation is impossible**
- [ ] chi-squared tests table: N is integer**
- [ ] chi-squared tests table: it doesn’t add footnote if calculation is impossible**
- [ ] contingency tables: there is a stray dot under “Total” when you specify a layer**
- [ ] contingency tables: the values should be integers if there is no counts variable and no options in the Cells tab are ticked
- [ ] contingency tables: it should not repeat the value of the left-most layer, but use `combine=TRUE` there too**
- [ ] log odds ratio table: adjusting confidence interval does not adjust the output
- [ ] Tables do not appear in the same location if they are created, removed and then created again  (e.g., first tick "X^2" -> tick "Log odds ratio (2x2 only)" -> untick "X^2" -> tick "X^2"; I expect the tables to be in the same order)

** as shown in the .jasp file: [contingencyRewrite.jasp.zip](https://github.com/jasp-stats/jasp-desktop/files/3435375/contingencyRewrite.jasp.zip)

You can use the above list to tick off things when you fix them.

gibbona1:
> I did some testing and noticed the following issues (see version 0.10.2 to compare behaviour):
> 
> * [X]  nominal table: no longer has titles in the rows of the first column when `ready == FALSE`
> * [X]  nominal table: it doesn’t add a footnote if calculation is impossible**
> * [X]  chi-squared tests table: N should be integer without a counts variable**
> * [X]  chi-squared tests table: it doesn’t add footnote if calculation is impossible**
> * [ ]  contingency tables: there is a stray dot under “Total” when you specify a layer**
> * [X]  contingency tables: the values should be integers if there is no counts variable and no options in the Cells tab are ticked
> * [ ]  contingency tables: it should not repeat the value of the left-most layer, but use `combine=TRUE` there too**
> * [X]  log odds ratio table: adjusting confidence interval does not adjust the output
> * [X]  Tables do not appear in the same location if they are created, removed and then created again  (e.g., first tick "X^2" -> tick "Log odds ratio (2x2 only)" -> untick "X^2" -> tick "X^2"; I expect the tables to be in the same order)
> 
> ** as shown in the .jasp file: [contingencyRewrite.jasp.zip](https://github.com/jasp-stats/jasp-desktop/files/3435375/contingencyRewrite.jasp.zip)
> 
> You can use the above list to tick off things when you fix them.

JASP-Engine/JASP/R/contingencytables.R
```diff
@@ -15,1778 +15,1412 @@
...
+                                    analyses$groups[[i]], analysisContainer, options, ready))
+    .crossTabSetErrorOrFill(res, crossTabNominal)
+  }
+}
 
-			} else {
+.crossTabGamma <- function(jaspResults, dataset, options, analyses, ready) {
+  if (!options$gamma)
+    return()
+  for (i in 1:length(analyses$a
```
vandenman:
```suggestion
      ss.matrix <- matrix(0, 
```
getting stuff from a namespace with `::` is good practice, but getting stuff out with `base::` is usually not necessary unless there is a naming clash (and then the question should be, why is there a naming clash?). Even CRAN doesn't require importing from `base` so I think those can just be omitted.

vandenman:
```suggestion
      ss.matrix  <- base::matrix(ss.table, nrow = nrow(ss.table), 
```
same stuff with nrow and ncol

vandenman:
```suggestion
          if (is.character(expected.matrix[1,1]))
```
?

vandenman:
```suggestion
      row <- colSums(counts.matrix)
```

vandenman: here us some code duplication.. maybe you could use a function to do:

```r
if (inherits(col.proportions.matrix[1,1], "character"))
  col.proportions <- col.proportions.matrix[1,]
else 
  colproportion <- apply(col.proportions.matrix, 2, base::sum)

col.proportions <- as.list(colproportion)
names(col.proportions) <- paste(names(col.proportions), "[col.proportions]", sep = "")

if (inherits(row.proportions.matrix[1,1], "character")) {
  col.proportions[["total[col.proportions]"]] <- ""
```
which is common to all 4 if statements.

vandenman: these two if statements can be joined

vandenman: these if statements can also be joined

JASP-Engine/JASP/R/contingencytables.R
```diff
@@ -15,1778 +15,1412 @@
...
-	}
-
-	list(row)
-
+# Preprocessing functions ----
+.crossTabReadData <- function(dataset, options) {
+  if (!is.null(dataset)) {
+    return(dataset)
+  } else {
+    layer.variables <- c()
```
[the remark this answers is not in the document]
gibbona1: layer
Unfortunately not as far as I can see

JASP-Engine/JASP/R/contingencytables.R
```diff
@@ -15,1778 +15,1412 @@
...
+  
+  for (layer in options$layers) {
+    layer.vars <- as.vector(layer$variables, "character")
+    analyses$analyses <- cbind(analyses$analyses, 
+                               rep(layer.vars, each = dim(analyses$analyses)[1]), 
+                      stringsAsFactors = FALSE)
+    names(analyses$analyses)[dim(analyses$analyses)[2]] <- layer$name
+  }
+  
+  analyses$analyses <- .dataFrameToRowList(analyses$analyses)
```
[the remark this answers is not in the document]
gibbona1: For the `analyses `and the `groups`, each subpart (`analysis` and `group`) needs to be a list. 
I thought it'd be better to keep the old way and create lists of lists rather than a making a list per row of the data.frame inside the `for` loops

JASP-Engine/JASP/R/contingencytables.R
gibbona1: search for apply(matrix, 2, sum)

TimKDJ:
@gibbona1 if you refresh`contingencyRewrite.jasp.zip` which I attached in my previous comment, you'll find a few errors. Can you fix these.

`object 'LogOR' not found`
`object 'analysisContainer' not found`

AlexanderLyNL:
It appears that you pushed a rebase on a branch that opened a PR. Could you please revert this?

JASP-Engine/JASP/R/contingencytables.R
```diff
@@ -15,1778 +15,1288 @@
...
+                            type = "number", format = "dp:1;pc")
+      }
+      if (options$percentagesColumn) {
+        private.name <- paste0(column.name,"[col.proportions]")
+        table$addColumnInfo(name = private.name, title = ".", 
+                            type = "number", format = "dp:1;pc")
+      }
+      if (options$percentagesTotal) {
+        private.name <- paste0(column.name,"[proportions]")
+        table$a
```
TimKDJ: For all footnotes: if they pertain to a specific row, can you ensure that a symbol is added to that row and the footnote itself refers to that row?
Right now they are added as general footnotes and users might not understand they refer to a row and not the entire table.

TimKDJ:
Rebase didn't go correctly here either

TimKDJ:
The footnotes still don't work correctly. The first thing I notice is that you pass in too many row names (as you only pass in 1 data row) and the second is that you specify the footnote symbol -- you do not have to do this, this is done automatically. See also https://github.com/jasp-stats/jasp-desktop/pull/3442#discussion_r321771985

gibbona1:
@TimKDJ the way I gather the data for each table, I need to add the layers to all the data before adding to the table then add the rows together to their table. Because of this, I can't add row by row inside the `for` loops, like the usual way. 
Is there a correct way to have the rownames as a list like I've done, but of the right length, or will this mess up because of the column folding?

TimKDJ:
Well you can add rows one by one and also add the rowNames one by one. Or you add the rows in bulk, in which case you also add the rowNames in bulk. I'm not sure I understand why rows and rowNames cannot be of equal length when they're added, I'm probably missing something though

TimKDJ:
But at this point it might be best to merge and I'll continue some work on it, so you can continue your work on the other analyses -- what do you say?

gibbona1:
Yeah it might be best to finish this so I can tweak the Bayesian contingency tables and make that PR.
Will I reset the commits so it's at the point where I had the footnote stuff done manually?

TimKDJ:
Alright, sounds good!
No, you don't have to do anything :)

PR #3443, opened 2019-07-23
Rewriting Multinomial Test to jaspResults

gibbona1:
**multinomialtest.R**
Updated Multinomial Test to jaspResults

**MultinomialTest.qml**
Now uses jaspResults

**test-multinomialtest.R**
changed unicode expectancy for main table, but this could just be my computer; if Travis fails I will change this back.
The descriptive plot test now fails with a slight margin difference, but I didn't want to update the plot and repeat what happened with Bayesian Binomial Test. Again if Travis fails here I will update the plot

### DIFF
--- a/JASP-Engine/JASP/R/multinomialtest.R
+++ b/JASP-Engine/JASP/R/multinomialtest.R
@@ -181,6 +181,11 @@ MultinomialTest <- function(jaspResults, dataset, options, ...) {
       dataframe <- cbind(dataframe, VovkSellkeMPR = .VovkSellkeMPR(dataframe$p))
     
     jaspResults[["chisqTable"]]$setData(dataframe)
+    
+    for (r in 1:length(chisqResults)) {
+      if (!is.null(chisqResults[[r]][["warn"]]))
+        jaspResults[["chisqTable"]]$addFootnote(chisqResults[[r]][["warn"]])
+    }
   }
 }
 
@@ -304,7 +309,7 @@ MultinomialTest <- function(jaspResults, dataset, options, ...) {
   chisqResults <- .chisquareTest(jaspResults, dataset, options)
   
   descriptivesTable <- createJaspTable(title = "Descriptives")
-  descriptivesTable$dependOn(c("factor", "counts", "exProbVar",  "hypothesis", "countProp",
+  descriptivesTable$dependOn(c("factor", "counts", "exProbVar",  "hypothesis", "countProp", "descriptives",
                                "confidenceInterval", "tableWidget", "confidenceIntervalInterval"))
 
   if(options$factor == ""){
--- a/JASP-Engine/JASP/R/multinomialtest.R
+++ b/JASP-Engine/JASP/R/multinomialtest.R
@@ -15,10 +15,30 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 
-MultinomialTest <- function (dataset = NULL, options, perform = "run",
-                             callback = function(...) 0,  ...) {
+MultinomialTest <- function(jaspResults, dataset, options, ...) {
+  
+  # Read dataset
+  dataset <- .multinomReadData(dataset, options)
+  
+  # Error checking
+  errors <- .multinomCheckErrors(dataset, options)
+  
+  # Compute the results
+  chisqResults      <- .chisquareTest(jaspResults, dataset, options, errors)
+  
+  multinomResults <- .multinomComputeResults(jaspResults, chisqResults, dataset, options, errors)
+  
+  # Output tables and plots
+  .chisqTable(                  jaspResults, multinomResults, options, errors)
+  .multinomialDescriptivesTable(jaspResults, chisqResults, multinomResults, options, errors)
+  .multinomialContainerPlots(   jaspResults, chisqResults, options, errors)
+  .multinomialDescriptivesPlot( jaspResults, chisqResults, options, errors)
+  
+  return()
+}
 
-  # First, we load the variables into the R environment
+# Preprocessing functions ----
+.multinomReadData <- function(dataset, options) {
   fact <- NULL
   asnum <- NULL
   if (options$factor != "") {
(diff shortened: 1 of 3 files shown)


Resources/Frequencies/qml/MultinomialTest.qml
```diff
@@ -25,7 +25,7 @@ import JASP.Theme 1.0
 
 Form
 {
-	usesJaspResults: false
+    usesJaspResults: true
```
vandenman:
```suggestion
```
`usesJaspResults: true` is the default, so you can just omit it.

JASP-Engine/JASP/R/multinomialtest.R
```diff
@@ -15,546 +15,447 @@
...
+.multinomCheckErrors <- function(dataset, options) {
+  # Test 1: Negatives and infinities
+  .hasErrors(dataset, 
+             type = c('negativeValues', 'infinity'), 
+             all.target = c(options$counts), 
+             exitAnalysisIfErrors = TRUE)
+  
+  # Test 2: Expected Counts
+  if (options$exProbVar != "" && options$counts == "") 
+    stop("Expected counts not supported without observed counts!")
```
vandenman: is there a particular reason to use `stop` over `.quitAnalysis`?

JASP-Engine/JASP/R/multinomialtest.R
```diff
@@ -15,546 +15,447 @@
...
+                chisq.test(x = dataTable, p = h, rescale.p = TRUE,
+                           simulate.p.value = options$simulatepval),
+                warning = function(w) warn <<- w$message
       )
       csr[["warn"]] <- warn
       return(csr)
     })
   }
+  # Save results to state
+  jaspResults[["stateChisqResults"]] <- createJaspState(chisqResults)
```
vandenman: I'm not a big fan of using the state rather than a return value (in particular in the subsequent function calls). I saw @TimKDJ did want this though. Could you explain the reason for this?

JASP-Engine/JASP/R/multinomialtest.R
```diff
@@ -15,546 +15,447 @@
...
-  #   perform: init or run
   #
   # Return:
-  #   Object (chisquare test)
-
+  #   No return, stores jaspResults Object (chisquare test)
+ 
+  # Take results from state if possible
+  if (!is.null(jaspResults[["stateChisqResults"]])) 
+    return(jaspResults[["stateChisqResults"]]$object)
```
vandenman: this `return` is redundant because the output is not assigned to anything.

JASP-Engine/JASP/R/multinomialtest.R
```diff
@@ -15,546 +15,447 @@
...
+  #   options: user options
+  #   chisqResults:
+  #   
+  # Return:
+  #   No return, stores jaspResults Object (main table results)
+  
+  results <- list()
+
+  # Compute Results
+  .chisquareTest(jaspResults, dataset, options)
```
vandenman:
```suggestion
  chisqResults <- .chisquareTest(jaspResults, dataset, options)
```
why not have `.chisquareTest` check the state and return the object if it exists, or otherwise compute and return the object? Now the next line is getting the object from state, although it could have been returned.

JASP-Engine/JASP/R/multinomialtest.R
```diff
@@ -15,546 +15,447 @@
...
+  # Get Results
+  chisqResults <- jaspResults[["stateChisqResults"]]$object
+  
+  if(!is.null(chisqResults)){
+    # fill in results one row at a time
+    for (r in 1:length(chisqResults)) {
```
vandenman: is there a reason for filling in the table by row? Here, everything is computed already right? So it won't be very slow to fill. I'd prefer a construction with `$setData(dataframe)` where `dataframe`, so for example:

```r
# extract relevant objects from chisqResults (this looks so convoluted is because its stored in an 
# inconvenient way, but that's probably a legacy code thingy).
dataframe <- do.call(rbind.data.frame, 
					 lapply(chisqResults, function(x) c(x[["parameter"]][["df"]], x[["p.value"]], x[["statistic"]][["X-squared"]])))
colnames(dataframe) <- c("chisquare", "df", "p")
dataframe <- cbind(case = names(chisqResults), dataframe)

if (options$VovkSellkeMPR)
	dataframe <- cbind(dataframe, VovkSellkeMPR = .VovkSellkeMPR(pVal))

jaspResults[["chisqTable"]]$setData(dataframe)
```

JASP-Engine/JASP/R/multinomialtest.R
```diff
@@ -15,546 +15,447 @@
...
+    message <- "Confidence intervals are based on independent binomial distributions."
+    results[["footnotes"]][["CImessage"]] <- message
+    if (any(is.nan(unlist(tableFrame[, c('lowerCI', 'upperCI')])))){
+      message <- "Could not compute confidence intervals."
+      results[["footnotes"]][["ciComputeError"]] <- message
+    }
+  }
+  
+  for (i in 1:nrow(tableFrame))
+    jaspResults[["descriptivesTable"]]$addRows(as.list(tableFrame[i,]))
```
vandenman:
```suggestion
    jaspResults[["descriptivesTable"]]$setData(tableFrame)
```
is a little cleaner.

JASP-Engine/JASP/R/multinomialtest.R
```diff
@@ -15,546 +15,447 @@
...
+  }
+  jaspResults[["descriptivesTable"]] <- descriptivesTable
+  if (!ready) 
+    return()
+  res <- try(.multinomDescriptivesTableFill(jaspResults, options, chisqResults))
+  .multinomialSetError(res, descriptivesTable)
+  if(options$confidenceInterval) {
+    message <- "Confidence intervals are based on independent binomial distributions."
+    descriptivesTable$addFootnote(message)
+    if (any(is.nan(unlist(descriptivesTable[["data"]][, c('lowerCI', 'upperCI')])))){
```
vandenman:
```suggestion
    if (anyNA(unlist(descriptivesTable[["data"]][, c('lowerCI', 'upperCI')]))){
```
should also work and is a little shorter :)

JASP-Engine/JASP/R/multinomialtest.R
```diff
@@ -15,546 +15,447 @@
...
+    list(ggplot2::geom_segment(data = d, 
+                               ggplot2::aes(x = x, y = y, xend = xend, yend = yend),
+                               size = 0.75, inherit.aes = FALSE),
+         ggplot2::scale_y_continuous(breaks = b))
+  }
+  
+  # Determine y-axis margin: If CIs could not be computed, use observed counts
+  plotFrame$yAxisMargin <- plotFrame$upperCI
+  for(i in 1:nrow(plotFrame))
+    if(plotFrame$upperCI[i] == 0)
```
vandenman: I can't judge if this will ever be exactly 0. If upperCI is a float, it could be better to test against a tolerance (e.g., `abs(plotFrame$upperCI) <= sqrt(.Machine$double.eps)`)

JASP-Engine/JASP/R/multinomialtest.R
TimKDJ: My main goal is to always show output to the user before you go into any (long) calculations. Consequently you must do markup first and then calculate and this calculation must be stored somewhere. The `state` is a natural place to keep the internal state (ha) of the analysis. 
However, I have no strong preference between either of these setups if that's what you mean:
```
tableFunction <- function(jaspResults) {
	<table stuff>
	
	model <- calcModel(jaspResults)
}

calcModel <- function(jaspResults) {
	if (!is.null(jaspResults$model))
		return(jaspResults$model$object)
		
	<calc model>
	
	jaspResults[["model"]] <- createJaspState(model)
	return(model)
}

tableFunction2 <- function(jaspResults) {
	<table stuff>
	
	calcModel2(jaspResults)
	model <- jaspResults[["model"]]
}

calcModel2 <- function(jaspResults) {
	if (!is.null(jaspResults$model))
		return()
		
	<calc model>
	
	jaspResults[["model"]] <- createJaspState(model)
	return()
}
```

JASP-Engine/JASP/R/multinomialtest.R
```diff
@@ -15,546 +15,440 @@
...
+  
+  if(!options$descriptives || !is.null(jaspResults[["descriptivesTable"]])) 
+    return()
+  
+  # Compute/get Results
+  chisqResults <- .chisquareTest(jaspResults, dataset, options)
+  
+  descriptivesTable <- createJaspTable(title = "Descriptives")
+  descriptivesTable$dependOn(c("factor", "counts", "exProbVar",  "hypothesis", "countProp",
+                               "confidenceInterval", "tableWidget", "confidenceIntervalInterval"))
```
TimKDJ: This should also depend on `"descriptives"`

JASP-Engine/JASP/R/multinomialtest.R
```diff
@@ -15,546 +15,440 @@
...
-         warn <<- w$message
-        }
+      csr  <- warn <- NULL
+      # need to improve this try statement
+      csr  <- withCallingHandlers(
+                chisq.test(x = dataTable, p = h, rescale.p = TRUE,
+                           simulate.p.value = options$simulatepval),
+                warning = function(w) warn <<- w$message
       )
       csr[["warn"]] <- warn
```
TimKDJ: You're not using these warning messages. Previously they were added as footnotes under the table.

JASP-Tests/R/tests/testthat/test-multinomialtest.R
```diff
@@ -42,3 +42,30 @@ test_that("Descriptives plot matches", {
...
   expect_equal_plots(testPlot, "descriptives-1", dir="MultinomialTest")
 })
+
+test_that("Analysis handles errors - Negative Values", {
+  options <- jasptools::analysisOptions("MultinomialTest")
+  options$factor <- "facExperim"
+  options$counts <- "contNormal"
+  results <- jasptools::run("MultinomialTest", "test.csv", options)
+  errorMsg <- results[["results"]][["errorMessage"]]
+  expect_is(errorMsg, "character")
```
TimKDJ: The msg at `results[["results"]][["errorMessage"]]` is always a character, whether the error was handled properly (through `.hasErrors()`) or not caught at all. It's better to test that `results[["status"]]` is equal to `"validationError"`

TimKDJ:
I think this one is finished once you do a rebase

PR #3447, opened 2019-07-24
fix typo in jasptool init

vandenman:
https://github.com/jasp-stats/jasp-desktop/commit/605390cf28891b6d0570496fd58679672efc32bc#diff-f9562ccfcced4a237ba5b1d4de419b43 changed `pathsToPackages` to `pathToPackages` but forgot one spot, which causes an error on Linux only.


TimKDJ:
Nice catch, thanks!

PR #3451, opened 2019-07-26
Improve the binomial tests

TimKDJ:
Considering these analyses are simple and can serve as examples of how to write a JASP analysis, I wanted them to be a bit more inline with the general style of other jaspResults analyses.

This also adds some additional error handling, fixes dependency issues in the Bayesian Binomial Test (e.g., "Prior" is not an interface option), adds .isNewGroup's and removes redudancy in some computation functions.

@AlexanderLyNL Descriptive plots were previously added to the Bayesian Binomial Test; in the interface we speak of a confidence interval, can we change this to credible interval just like we did for t-tests and ANOVA's? In other words, are confidence and credible intervals equivalent here (given a flat prior)?

### DIFF
--- a/JASP-Engine/JASP/R/binomialtest.R
+++ b/JASP-Engine/JASP/R/binomialtest.R
@@ -15,24 +15,17 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 
-BinomialTest <- function(jaspResults, dataset, options, ...) {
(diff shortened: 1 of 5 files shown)


AlexanderLyNL:
You have to change the prior beta(1, 1) to a beta(1/2, 1/2) then it's a credible interval that can be interpreted as a confidence interval.

TimKDJ:
Alright, so then we can just do the name change, I suppose

PR #3453, opened 2019-07-30
Bayesian t-test s with jaspResults 2

vandenman:
A second attempt of #3003, rewrite of Bayesian t-tests to jaspResults.
includes some improvements to JASPgraphs and quite a few new unit-tests.
It looks like quite a lot but most of the changes are in `JASP-Engine/JASP/R/commonbayesianttest.R`. 
Below a brief summary.

JASP (26 files):

3 .R analysis files
1 .R new file with common functions for Bayesian t-tests
3 .qml analysis files
3 .R unit-tests
13 .svg plots, I added unit-tests for the all plots WITH additional info and without additional info (previously we checked a mix).
1 .R  ttestplotfunctions.R for bain
1 .R  binomialtestbayesian.R , minor change due to API change in JASPgraphs
1 jasp-deps.txt version increment of JASPgraphs

JASPgraphs (8 files):

2 DESCRIPTION & NAMESPACE
2 .R PlotPriorAndPosterior & PlotRobustnessSequential
2 .R example files (for PlotPriorAndPosterior & PlotRobustnessSequential)
2 .Rd help files (for PlotPriorAndPosterior & PlotRobustnessSequential)


JohnnyDoorn:
Great!!

Functional comments:
- If you tick descriptives, maybe it's better to include the 95% CI in the table, as opposed to having to also tick the descriptives plotting box. Or maybe include that information in the BF table? This would make that table a bit more fleshed out (a pet peeve of mine) and you have all information in one place as opposed to scattered before and after plots.
- The plots are flickering when sampling for the Wilcoxon test - but maybe there is no way around that? Maybe set them to init state instead?
- The normal prior for the student ttest does not seem to work (one and two sample) - see screenshot
- The Savage Dickey dots do not line up in the plots for the Wilcoxon when using different priors, but I think that is because those should be grayed out, so that it takes only Cauchy priors (with the scale parameter free to vary).

ttestTwoSampleCrash

vandenman:
@JohnnyDoorn if you're happy then I think this is good be merged.

vandenman:
@boutinb @FransMeerhoff I think this can be merged!

PR #3454, opened 2019-08-05
Binomial Test Bayesian - Fixed error setting for main table and dependency issue

gibbona1:
Fixes issues with the main table where the error spotted did not call .quitAnalysis
Also fixed the dependency issue https://github.com/jasp-stats/INTERNAL-jasp/issues/396, now using "priorA" and "priorB" instead of "Prior"


vandenman:
@JorisGoosen I think this can be merged

TimKDJ:
.. why not use #3451? That pr implemented this

PR #3455, opened 2019-08-07
make CRI text adjustable

vandenman:
implements https://github.com/jasp-stats/INTERNAL-jasp/issues/397 and makes the text for credible interval and median adjustable for the prior and posterior plots.

the change to `JASP-Engine/JASPgraphs/man/PlotRobustnessSequential.Rd` is because I forgot to call `devtools::document()` last time.

@gibbona1 please let me know if this does all that you wanted!

### DIFF
--- a/JASP-Engine/JASPgraphs/DESCRIPTION
+++ b/JASP-Engine/JASPgraphs/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: JASPgraphs
 Type: Package
 Title: Custom graphs for JASP
-Version: 0.3.4
+Version: 0.3.5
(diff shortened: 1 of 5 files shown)


gibbona1:
The changes are good! Adjusting the Median text and CI works great.
For one of the analyses I'm rewriting I have "median Log OR = " as mediantxt, which is already pretty lengthy, and if I change it to say "long median Log OR = " there will be some text cutoff from the pizza plot. I don't think this is much to worry about though
priorposteriormediantext

vandenman:
@FransMeerhoff 	@boutinb I think this is good to be merged!

PR #3457, opened 2019-08-07
Reliability Analysis now uses jaspResults

gibbona1:
**reliabilityanalysis.R**
Now using jaspResults

**test-reliabilityanalysis.R**
Added error handling unit test

**ReliabilityAnalysis.qml**
removed `usesJaspResults: false` as default is `true`


JASP-Engine/JASP/R/reliabilityanalysis.R
```diff
@@ -15,479 +15,456 @@
...
+             perform = perform,
+             type    = "infinity",
+             infinity.target = options$variables,
+             exitAnalysisIfErrors = TRUE)
+} 
+
+# Results functions ----
+.reliabilityComputeResults <- function (jaspResults, dataset, options, ready) {
+  
+  if (!ready) return()
```
TimKDJ: Why not check the status of `ready` in the table function? Otherwise you have to do it in multiple places.

JASP-Engine/JASP/R/reliabilityanalysis.R
```diff
@@ -15,479 +15,456 @@
...
+  }
+  
+  if (options$missingValues == "excludeCasesListwise") 
+    exclwise = " listwise"
+  else 
+    exclwise = " pairwise"
+  
+  variables <- unlist(options$variables)
+  nObs      <- nrow(dataset)
+  nExcluded <- sum(!complete.cases(dataset))
```
TimKDJ: `dataset` can be `NULL`, in which case you get an error. You should probably have already returned at this point by checking `ready`

JASP-Engine/JASP/R/reliabilityanalysis.R
```diff
@@ -15,479 +15,456 @@
...
+.reliabilityScaleFill <- function(jaspResults, dataset, options, relyFit, ready) {
+  results   <- list()
+  nObs      <- nrow(dataset)
+  nVar      <- ncol(dataset)
+  variables <- unlist(options$variables)
+  options   <- .reliabilitySetAlphaNms(options)
+  
+  if(ready){
+    #Initialise all as null
+    alpha <- lambda <- omega <- glb <- rho <- mu <- sd <- lower <- upper <- NULL
```
TimKDJ: You don't have to pre-initialize these. In your jaspTable you can set `$showSpecifiedColumnsOnly <- TRUE` so only the columns which have their info set through `$addColumnInfo()` will be shown. All the other columns -- even those with data -- are discarded. This also removes the need for all the if statements below.

JASP-Engine/JASP/R/reliabilityanalysis.R
```diff
@@ -15,479 +15,456 @@
...
+        else # a useable value
+          glb  <- relyFit$glb
+      }
+      if (options[["confAlpha"]]) {
+        lower <- relyFit[["ciAlpha"]][1]
+        upper <- relyFit[["ciAlpha"]][2]
+      }
+    }
+  } else #Initialise all as "."
+    alpha <- lambda <- omega <- glb <- rho <- mu <- sd <- lower <- upper <- "."
```
TimKDJ: This is more of an old-style kinda thing. Returning a table without any rows generates one row of dots for you automatically.

JASP-Engine/JASP/R/reliabilityanalysis.R
```diff
@@ -15,479 +15,456 @@
...
+    only be calculated for three or more variables.")
+    scaleTable$addFootnote(message, symbol = "<em>Note.</em>")
+  }
+  
+  jaspResults[["scaleTable"]] <- scaleTable
+  
+  # Compute/get the results
+  relyFit <- .reliabilityComputeResults(jaspResults, dataset, options, ready)
+  
+  res <- try(.reliabilityScaleFill(jaspResults, dataset, options, relyFit, ready))
```
TimKDJ: I don't think `reliabilityScaleFill` can actually fail, so this `try` bit is not necessary

JASP-Engine/JASP/R/reliabilityanalysis.R
```diff
@@ -15,479 +15,456 @@
...
 
+.reliabilityItemTable <- function (jaspResults, dataset, options, ready) {
+  if (!(options$alphaItem || options$gutmannItem || options$itemRestCor ||
+        options$meanItem  ||options$mcDonaldItem || options$sdItem))
+    return()
+  
+  if (!is.null(jaspResults[["itemContainer"]][["table"]])) 
+    return()
+  
+  jaspResults[["itemContainer"]] <- createJaspContainer("Item Statistics")
```
TimKDJ: There can only be one item statistics table, right? Because then we don't need the container around it.

JASP-Tests/R/tests/testthat/test-reliabilityanalysis.R
```diff
@@ -60,10 +60,18 @@ test_that("Reverse scaled items match", {
...
          0.81017314606981, 0.831119838017159)
   )
+})
+
+test_that("Analysis handles errors - Infinity", {
+  options <- jasptools::analysisOptions("ReliabilityAnalysis")
+  options$variables <- c("debInf")
+  results <- jasptools::run("ReliabilityAnalysis", "test.csv", options)
+  errorMsg <- results[["results"]][["errorMessage"]]
+  expect_is(errorMsg, "character")
```
TimKDJ: Same comment as I made earlier; check for `"validationError"` in `results[["status"]]`

JASP-Engine/JASP/R/reliabilityanalysis.R
gibbona1: Will I remove the container then? I was trying to keep the layout the same as before

TimKDJ: Ah, yeah, I don't really understand why we have 
-> Item Statistics
--> Item Reliability Statistics

I think the outer container doesn't add any information and can just be removed to be inline with other analyses

TimKDJ:
I'm guessing you just haven't pushed the commit yet that has the resolved stuff?

JASP-Engine/JASP/R/reliabilityanalysis.R
```diff
@@ -15,479 +15,433 @@
...
+  }
+}
 
+.reliabilityItemTable <- function (jaspResults, dataset, options, ready) {
+  if (!(options$alphaItem || options$gutmannItem || options$itemRestCor ||
+        options$meanItem  ||options$mcDonaldItem || options$sdItem))
+    return()
+  
+  if (!is.null(jaspResults[["itemTable"]])) 
+    return()
```
TimKDJ: Combine with the statement above

JASP-Engine/JASP/R/reliabilityanalysis.R
```diff
@@ -15,479 +15,433 @@
...
+  jaspResults[["itemTable"]] <- itemTable 
+  
+  if(!ready)
+    return()
+  
+  # Compute/get the results
+  relyFit <- .reliabilityComputeResults(jaspResults, dataset, options)
+  
+  res <- try(.reliabilityItemFill(jaspResults, dataset, options, relyFit))
+  if(isTryError(res))
```
TimKDJ: Same comment about the `try()`

JASP-Engine/JASP/R/reliabilityanalysis.R
```diff
@@ -156,39 +153,34 @@ ReliabilityAnalysis <- function(jaspResults, dataset = NULL, options, ...) {
     sd     <- relyFit$total$sd
     rho    <- relyFit$total$average_r
     omega  <- relyFit$omega
-    if (relyFit$glb == "." || isTryError(relyFit$glb)) # unusable information
+    if (relyFit$glb == "." || isTryError(relyFit$glb)) { # unusable information
       glb  <- "."
-    else # a useable value
+      message <- "Greatest lower bound could not be calculated."
+      jaspResults[["scaleTable"]]$addFootnote(message, symbol = "Warning:")
```
TimKDJ: We don't want this if the warning on line 284 is already added though.

TimKDJ:
Cronbach's alpha does not change in the item statistics when you toggle between unstandardized and standardized. That's the last thing I can find

TimKDJ:
Could you undo the branch merge you did? Right now your PR has 142 commits..

PR #3462, opened 2019-08-09
Bugfixes in Machine Learning Module

vandenman:
Fixes everything in https://github.com/jasp-stats/INTERNAL-jasp/issues/398, except for the code duplication and the unit-tests.

- [x] parallelization should be avoided or we need to give users some warning
- [x] debMiss99 breaks all 5 clustering analyses
- [x] Bug in LDA

the parallelization issue is avoided by overwriting the function in the `gbm` packages with functions that don't create a parallel backend. This is very hacky, but the only thing I could come up with.

### DIFF
--- a/JASP-Engine/JASP/R/mlRegressionBoosting.R
+++ b/JASP-Engine/JASP/R/mlRegressionBoosting.R
@@ -79,9 +79,6 @@ mlRegressionBoosting <- function(jaspResults, dataset, options, ...) {
 
   trees <- base::switch(options[["modelOpt"]], "optimizationManual" = options[["noOfTrees"]], "optimizationOOB" = options[["maxTrees"]])
   
-  assignFunctionInPackage(fakeGbmCrossValModelBuild, "gbmCrossValModelBuild", "gbm")
-  assignFunctionInPackage(fakeGbmCrossValErr, "gbmCrossValErr", "gbm")
-
   bfit <- gbm::gbm(formula = formula, data = train, n.trees = trees,
                                shrinkage = options[["shrinkage"]], interaction.depth = options[["intDepth"]],
                                cv.folds = noOfFolds, bag.fraction = options[["bagFrac"]],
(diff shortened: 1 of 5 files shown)


JASP-Engine/JASP/R/mlRegressionBoosting.R
```diff
@@ -78,6 +78,65 @@ mlRegressionBoosting <- function(jaspResults, dataset, options, ...) {
...
+      model <- cv.models[[index]]
+      model$valid.error * in.group[[index]]
+    }, double(n.trees))
+    if (is.matrix(cv.error))
+      return(rowSums(cv.error)/nTrain)
+    else
+      return(cv.error / nTrain)
+  }
+  
+  assignFunctionInPackage <- function(fun, name, package) {
```
AlexanderLyNL: Are we going to use this hack more often? I think it might be good to have these assignments some where more prominent, so we remember that we did this. Once it's revised, we can then remove these more easily. In particular,

assignFunctionInPackage(fakeGbmCrossValModelBuild, "gbmCrossValModelBuild", "gbm")
assignFunctionInPackage(fakeGbmCrossValErr, "gbmCrossValErr", "gbm")

JASP-Engine/JASP/R/mlRegressionBoosting.R
```diff
@@ -79,62 +79,6 @@ mlRegressionBoosting <- function(jaspResults, dataset, options, ...) {
...
-  
-  assignFunctionInPackage <- function(fun, name, package) {
-    # a horrible hack to avoid a parallel backend from being registered
-    ns <- getNamespace(package)
-    unlockBinding(name, ns)
-    assign(name, fun, ns)
-    lockBinding(name, ns)
-  }
-
   assignFunctionInPackage(fakeGbmCrossValModelBuild, "gbmCrossValModelBuild", "gbm")
```
AlexanderLyNL: Move these two assignments to that separate file, then we have them all together.

vandenman: done!

PR #3464, opened 2019-08-13
Machine learning fixes 2

vandenman:
Fixes the following:

- [x] scaling data (e.g., substracting the mean) happens for factors!
- [x] rescaling of qml does not happen for data split preferences, zoom JASP and look at the consequences for the labels.

In addition, I moved the data split prefernces to a single qml component rather than copy pasting it 8 times.

### DIFF
--- a/Resources/Machine Learning/qml/mlClassificationBoosting.qml
+++ b/Resources/Machine Learning/qml/mlClassificationBoosting.qml
@@ -148,9 +148,9 @@ Form {
                         max: 95 
                     }
 
-                    Text {
+                    Label {
                         text: qsTr("of all data")
-                        enabled: true
+                        alwaysEnabled: true
                     }
                 }
             }
@@ -204,9 +204,9 @@ Form {
                         max: 95
                     }
 
-                    Text {
+                    Label {
                         text: qsTr("for validation data")
-                        enabled: true
+                        alwaysEnabled: true
                     }
                 }
             }
@@ -226,9 +226,9 @@ Form {
                         fieldWidth: 30
                     } 
 
-                    Text {
+                    Label {
                         text: qsTr("folds")
-                        enabled: true
+                        alwaysEnabled: true
                     }
                 }
             }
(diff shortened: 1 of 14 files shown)


Resources/Machine Learning/qml/mlClassificationBoosting.qml
```diff
@@ -126,114 +127,8 @@ Form {
...
-
-                    Text {
-                        text: qsTr("folds")
-                        enabled: true
-                    }
-                }
-            }
-        }
-    }
+    ML.DataSplit {leaveOneOutVisible: false}
```
koenderks: Maybe add some spacing to make it more readable and consistent with all the other qml files and code

Resources/Machine Learning/qml/mlClassificationKnn.qml
```diff
@@ -113,119 +114,7 @@ Form {
...
-                }
-            }
-
-            RadioButton { 
-                text: qsTr("Leave-one-out")                 
-                name: "validationLeaveOneOut"
-            }
-        }
-    }
+    ML.DataSplit{}
```
koenderks: same

Resources/Machine Learning/qml/mlClassificationLda.qml
```diff
@@ -165,92 +166,7 @@ Form {
...
-
-                    Text {
-                        text: qsTr("for validation data")
-                        enabled: true
-                    }
-                }
-            }
-        }
-    }
+    ML.DataSplit{leaveOneOutVisible: false; kFoldsVisible: false}
```
koenderks: etc.

vandenman:
@koenderks 	I changed the spacing, although it was not much different from e.g., `AssignedVariablesList { name: "variables";	title: qsTr("Variables") }` inside `descriptives.qml`. Also, I wasn't quite sure why you changed the indenting from the other jasp files, or if there was an scheme behind it. What you did is a lot closer to QT's [QML guidelines](https://doc.qt.io/qt-5/qml-codingconventions.html), but then again QT also recommends one liners such as `Rectangle { color: "blue"; width: parent.width / 3 }`. Maybe I'm just overthinking guidelins though :)

PR #3471, opened 2019-08-14
fixes to machine learning

vandenman:
Fixes:

https://github.com/jasp-stats/INTERNAL-jasp/issues/424
https://github.com/jasp-stats/INTERNAL-jasp/issues/423
https://github.com/jasp-stats/INTERNAL-jasp/issues/421
https://github.com/jasp-stats/INTERNAL-jasp/issues/420

@AlexanderLyNL I moved the `assignFunctionInPackage` back to the functions because otherwise they are ignored (at least on Linux).

@koenderks the change to regularized regression is because glmnet just can't handle characters. I have no idea how they handle factors, as they convert everything to numeric(ish).

### DIFF
--- a/JASP-Engine/JASP/R/commonMachineLearningRegression.R
+++ b/JASP-Engine/JASP/R/commonMachineLearningRegression.R
@@ -302,7 +302,7 @@
 
   regressionResult <- jaspResults[["regressionResult"]]$object
   
-  predPerformance <- data.frame(true = regressionResult[["testReal"]], predicted = regressionResult[["testPred"]])
+  predPerformance <- data.frame(true = c(regressionResult[["testReal"]]), predicted = regressionResult[["testPred"]])
 
   allBreaks <- JASPgraphs::getPrettyAxisBreaks(predPerformance[, 1], min.n = 4)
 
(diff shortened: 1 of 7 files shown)


JASP-Engine/JASP/R/commonMachineLearningRegression.R
```diff
@@ -302,7 +302,7 @@
 
   regressionResult <- jaspResults[["regressionResult"]]$object
   
-  predPerformance <- data.frame(true = regressionResult[["testReal"]], predicted = regressionResult[["testPred"]])
+  predPerformance <- data.frame(true = c(regressionResult[["testReal"]]), predicted = regressionResult[["testPred"]])
```
koenderks: Is this on purpose?

vandenman: Yes, it is required AND on purpose!
```r
a <- matrix(1)
colnames(a) <- "hoi"
> data.frame(true = a)
  hoi
1   1
> data.frame(true = c(a))
  true
1    1
```
the next bit in the code is a ggplot with `aes(... = true)` so the version without `c(...)` would have the wrong names.

PR #3472, opened 2019-08-15
Update requiredPackages.R

FransMeerhoff:
For the time being the required package list we need for JASP required files should also contain the packages used by the dynamic modules.


TimKDJ:
Not needed at this time; dynamic modules are not shipped

PR #3481, opened 2019-08-22
Remove Rcpp errors and unnecessary messages in RStudio

TimKDJ:
@JorisGoosen I've wanted to remove some startup messages anyway. So maybe this fixes the install of jaspResults on linux, as `packageVersion()` is no longer called. (I can't test on linux at the moment)


JorisGoosen:
Hmm, it fails for me on loading with:
```
installing to /usr/lib/R/library/00LOCK-jaspResults/00new/jaspResults/libs
** R
** byte-compile and prepare package for lazy loading
Attempting to load the environment ‘package:jaspResults’
Error in getClass(class(x)) : 
  “Rcpp_jaspResultsClass” is not a defined class
Calls: <Anonymous> -> $ -> envRefInferField -> getClass
** help
*** installing help indices
** building package indices
** testing if installed package can be loaded from temporary location
Attempting to load the environment ‘package:jaspResults’
Error: package or namespace load failed for ‘jaspResults’:
 .onAttach failed in attachNamespace() for 'jaspResults', details:
  call: .External(structure(list(name = "InternalFunction_invoke", address = <pointer: (nil)>, 
  error: NULL value passed as symbol address
Specified environment not found: using '.GlobalEnv' instead
Error: package or namespace load failed for ‘jaspResults’:
 .onAttach failed in attachNamespace() for 'jaspResults', details:
  call: .External(structure(list(name = "InternalFunction_invoke", address = <pointer: (nil)>, 
  error: NULL value passed as symbol address
Error: loading failed
Execution halted
ERROR: loading failed
* removing ‘/usr/lib/R/library/jaspResults’
```

By the way, you should up the JASP-R-Interface version in JASP.pro to make sure windows users also know they need to rebuild it.

JorisGoosen:
On the other hand, this might be cause again by R 3.6.0 so don't worry about it too much.

TimKDJ:
> By the way, you should up the JASP-R-Interface version in JASP.pro to make sure windows users also know they need to rebuild it.

Done!

Will figure out the other problem at some later point

PR #3486, opened 2019-08-27
fixed .scaleNumericData and added fallback

vandenman:
fixes https://github.com/jasp-stats/jasp-test-release/issues/288

adding `, drop = FALSE` avoids the subsetting from converting the data.frame to a factor, so now the correct method gets called. 

I also added `.scaleNumericData.default` as a fallback function, just as a precaution.

@koenderks I have some issues with JASP atm, I checked the bug disappeared with jasptools, can you verify the bug is gone in JASP?

### DIFF
--- a/JASP-Engine/JASP/R/commonMachineLearningRegression.R
+++ b/JASP-Engine/JASP/R/commonMachineLearningRegression.R
@@ -427,6 +427,12 @@
   return(x)
 }
 
+.scaleNumericData.default <- function(x, center = TRUE, scale = TRUE) {
+  warning(".scaleNumericData.default was called! you probably didn't want that!")
+  # just do nothing
+  return(x)
+}
+
 .regressionAddValuesToData <- function(options, jaspResults, ready){
   if(!ready || !options[["addValues"]] || options[["valueColumn"]] == "")  return()
 
--- a/JASP-Engine/JASP/R/commonMachineLearningRegression.R
+++ b/JASP-Engine/JASP/R/commonMachineLearningRegression.R
@@ -413,6 +413,10 @@
 }
 
 .scaleNumericData.matrix <- function(x, center = TRUE, scale = TRUE) {
+  if (!is.numeric(x)) {
+    warning(sprintf("Object passed to .scaleNumericData.matrix was not numeric!"))
+    return(x)
+  }
   x <- scale(x, center, scale)
   attr(x, which = "scaled:center") <- NULL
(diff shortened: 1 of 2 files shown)


JASP-Engine/JASP/R/commonMachineLearningRegression.R
```diff
@@ -427,6 +427,12 @@
   return(x)
 }
 
+.scaleNumericData.default <- function(x, center = TRUE, scale = TRUE) {
+  warning(".scaleNumericData.default was called! you probably didn't want that!")
```
TimKDJ: Shouldn't this just fail hard? I could imagine you might get some unexpected values/bugs otherwise

vandenman: Well.. it failed hard before. `.scaleNumericData` has three methods that are the only ones that should ever be called:

- `.scaleNumericData.data.frame`, automatically handles the variable types in the data.frame.
- `.scaleNumericData.matrix`, assumes you put in numeric data and otherwise crashes.
- `.scaleNumericData.numeric`, obviously numeric.

It can happen that someone writes: `.scaleNumericData(myDataframe[, predictors])`. Now, if `predictors` has length 1 then R (annoyingly) simplifies the result and instead of a data.frame you pass a vector with some type to `.scaleNumericData`. If this vector contains numeric data, all is well, otherwise you get an error. 

As a fallback I now added `.scaleNumericData.default`, since all the cases where `.scaleNumericData` should actually do something are accounted for (I hope). That way, character/ factor data won't cause a crash. 

Another way to avoid this error is to enforce `drop = FALSE`. For example, `.scaleNumericData(myDataframe[, predictors, drop = FALSE])`. However, whenever we forget about this a bug appears. To be realistic, at some point we will forget about this.

So that gives two options:
1. be strict about `drop = FALSE` (good idea anyway I think).
2. have a fallback to avoid crashes when people didn't do `drop = FALSE`, possibly with a warning.

I don't really see how `.scaleNumericData.default` could introduce unexpected bugs, but maybe you have an idea? If you don't like the fallback idea we can also just remove it.

TimKDJ: Ah, I see. In that case I think you could actually remove the warning, but then you should also not have `.matrix` crash if it's character. Because then the behaviour is consistent: either you supply invalid data and it does nothing, or your data is (partially) valid and it's scaled

PR #3496, opened 2019-08-30
added Rcode as textType

vandenman:
QML component `TextArea` now accepts `textType: "Rcode"`. This uses the whitelist to check if the R-code is safe.

Changes:
 - r_functionwhitelist.cpp and .h to JASP-Common.
 - r_functionwhitelist.h now includes the definition of `filterException`.
 - `boundqmltextarea.cpp` has some new `else if` statements for Rcode.

This type is needed for the JAGS module, but it could also be useful for any other analysis that wants to run a user's R code.


JASP-Desktop/widgets/boundqmltextarea.cpp
```diff
@@ -19,6 +19,7 @@
 #include "boundqmltextarea.h"
 #include "../analysis/analysisform.h"
 #include "qmllistviewtermsavailable.h"
+#include "../../JASP-Common/r_functionwhitelist.h"
```
JorisGoosen: I think this can just be `#include "r_functionwhitelist.h"` because JASP-Common is already an `INCLUDEPATH`.

JASP-Engine/rbridge.cpp
```diff
@@ -22,6 +22,7 @@
 #include "appinfo.h"
 #include "tempfiles.h"
 #include "log.h"
+#include "../JASP-Common/r_functionwhitelist.h"
```
JorisGoosen: `#include "r_functionwhitelist.h"` again.

JASP-Engine/rbridge.h
```diff
@@ -33,7 +33,7 @@
 #include <boost/function.hpp>
 #include "../JASP-Common/dataset.h"
 #include "../JASP-R-Interface/jasprcpp_interface.h"
-#include "r_functionwhitelist.h"
+#include "../JASP-Common/r_functionwhitelist.h"
```
JorisGoosen: ;)

vandenman:
I checked this:

afbeelding
afbeelding
afbeelding

JorisGoosen:
Top!

PR #3498, opened 2019-08-31
Add classical estimation analysis to audit module

koenderks:


Resources/Audit/qml/classicalEstimation.qml
```diff
@@ -0,0 +1,224 @@
...
+						id: 		explanatoryText
+						text:	 	qsTr("Enable")
+						name: 		"explanatoryText"
+						checked: 	true
+					}
+
+					HelpButton 
+					{ 
+						helpPage:			"Audit/explanatoryText"
+						toolTip: 			"Show explanatory text and formulas"
```
JorisGoosen: No qsTr()?

JorisGoosen:
I did not actually check all the R code though

koenderks:
That is fine, I actually noticed I missed a dependency somewhere so I will double check all of this again when I add it.

PR #3501, opened 2019-09-01
Fix https://github.com/jasp-stats/jasp-test-release/issues/293

koenderks:
adding column to data when there are missing values

image

### DIFF
--- a/JASP-Engine/JASP/R/commonMachineLearningClassification.R
+++ b/JASP-Engine/JASP/R/commonMachineLearningClassification.R
@@ -839,11 +839,9 @@
 
   if(is.null(jaspResults[["classColumn"]])){
     predictions <- as.character(classificationResult[["classes"]])
-    print(predictions)
     classColumn <- rep(NA, jaspResults[["lengthOriginalDataset"]]$object)
     classColumn[jaspResults[["indexOfCompleteCases"]]$object] <- predictions
     classColumn <- factor(classColumn)
-    print(classColumn)
     jaspResults[["classColumn"]] <- createJaspColumn(columnName=options[["classColumn"]])
     jaspResults[["classColumn"]]$dependOn(options = c("classColumn", "noOfNearestNeighbours", "trainingDataManual", "distanceParameterManual", "weights", "scaleEqualSD", "modelOpt",
                                                             "target", "predictors", "seed", "seedBox", "modelValid", "maxK", "noOfFolds", "modelValid", "holdoutData", "testDataManual",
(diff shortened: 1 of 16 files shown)


JASP-Engine/JASP/R/commonMachineLearningClassification.R
```diff
@@ -835,7 +838,12 @@
   classificationResult <- jaspResults[["classificationResult"]]$object
 
   if(is.null(jaspResults[["classColumn"]])){
-    classColumn <- classificationResult[["classes"]]
+    predictions <- as.character(classificationResult[["classes"]])
+    print(predictions)
```
vandenman:
```suggestion
```

JASP-Engine/JASP/R/commonMachineLearningClassification.R
```diff
@@ -835,7 +838,12 @@
...
+    classColumn <- rep(NA, jaspResults[["lengthOriginalDataset"]]$object)
+    classColumn[jaspResults[["indexOfCompleteCases"]]$object] <- predictions
+    classColumn <- factor(classColumn)
+    print(classColumn)
```
vandenman:
```suggestion
```

JASP-Engine/JASP/R/commonMachineLearningRegression.R
```diff
@@ -26,13 +26,21 @@
...
-    dataset <- .readDataSetToEnd(columns = variables.to.read, exclude.na.listwise = variables.to.read)
+    dataset <- .readAndAddCompleteRowIndices(dataset, variables.to.read)
   }
   if(length(unlist(options[["predictors"]])) > 0 && options[["target"]] != "" && options[["scaleEqualSD"]])
-    dataset[,.v(c(options[["predictors"]], options[["target"]]))] <- .scaleNumericData(dataset[,.v(c(options[["predictors"]], options[["target"]]))])
+    dataset[,.v(c(options[["predictors"]], options[["target"]]))] <- .scaleNumericData(dataset[,.v(c(options[["predictors"]], options[["target"]])), drop = FALSE])
   return(dataset)
 }
 
+.readAndAddCompleteRowIndices <- function(dataset, variables.to.read){
```
JorisGoosen: This function looks deliciously compact. Maybe we could use this in Descriptives as well? 

Or maybe move this to common to give everybody easy access to this functionality?
@TimKDJ what do you think?

TimKDJ: Yeah we can definitely add it. In that case it would be better to take the `.readDataSetToEnd()` out of the function though, as this differs per analysis.

@koenderks why does `.readAndAddCompleteRowIndices()` need the `dataset` argument?

PR #3503, opened 2019-09-03
Fix issues for machine learning

koenderks:
+ Remove factors from decision boundary plot (fixes error in https://github.com/jasp-stats/jasp-test-release/issues/289)
+ Remove classes that are not included in test set from evaluation metrics table and roc curve plot (this caused the error in https://github.com/jasp-stats/jasp-test-release/issues/278)

### DIFF
--- a/JASP-Engine/JASP/R/mlClassificationBoosting.R
+++ b/JASP-Engine/JASP/R/mlClassificationBoosting.R
@@ -141,40 +141,7 @@ mlClassificationBoosting <- function(jaspResults, dataset, options, ...) {
   }
 
   # Calculate AUC
-  lvls <- levels(factor(test[, .v(options[["target"]])]))
-  auc <- numeric(length(lvls)) 
-
-  predictorNames <- .v(options[["predictors"]])
-  AUCformula <- formula(paste("levelVar", "~", paste(predictorNames, collapse=" + ")))
-
-  for(i in 1:length(lvls)){
-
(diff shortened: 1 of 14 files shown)


JASP-Engine/JASP/R/mlClassificationBoosting.R
```diff
@@ -168,8 +168,12 @@ mlClassificationBoosting <- function(jaspResults, dataset, options, ...) {
     score <- predict(bfitAUC, newdata = test, n.trees = noOfTrees, type = "response")
     actual.class <- test[,.v(options[["target"]])] == lvls[i]
 
-    pred <- ROCR::prediction(score, actual.class)
-    auc[i] <- ROCR::performance(pred, "auc")@y.values[[1]]
+    if(length(levels(factor(actual.class))) == 2){
```
TimKDJ: Can we do something about these `for(i in 1:length(lvls)){` loops?
They do the same basic thing and share a lot of code. Which is exemplified by the fact you needed to copy-paste this fix 4 times.

(Also, above this it does ` typeData <- typeData[, -column]` twice)

koenderks: like?

TimKDJ: I made a pull request to your repo: https://github.com/koenderks/jasp-desktop/pull/11
This removes all the duplication. Should be checked though!

PR #3505, opened 2019-09-04
Ignore filter column names that are in quotes

TimKDJ:
Partially fixes jasp-stats/jasp-issues#469

@JorisGoosen I can do it in a different way, but this seemed easiest.
Not sure if we should merge before release


JASP-Engine/rbridge.cpp
```diff
@@ -745,13 +744,43 @@ std::string	rbridge_encodeColumnNamesToBase64(const std::string & filterCode)
...
+    std::vector<int> positions;
+    bool inString = false;
+    char delim;
+    for (std::string::size_type pos = 0; pos < filterBase64.length(); ++pos)
+    {
+		if (!inString && filterBase64.substr(pos, columnName.length()) == columnName)
+		{
+        	positions.push_back(int(pos));
+			continue;
+		}
```
JorisGoosen: Can we not get rid of the continue and just do an `} else if (filterBase64[pos] etc`?

Like:
```
for (std::string::size_type pos = 0; pos < filterBase64.length(); ++pos)
	if (!inString && filterBase64.substr(pos, columnName.length()) == columnName)
		positions.push_back(int(pos));
	else if (filterBase64[pos] == '"' || filterBase64[pos] == '\'')
	{
		if (inString && filterBase64[pos] == delim)
			inString = false;
		else if (!inString)
		{
			delim = filterBase64[pos];
			inString = true;
		}
	}

return positions;
}
```

JorisGoosen: I think this is more readable an clearly shows that you cannot get to the second if-block if you add a position.

JorisGoosen:
The code looks fine, besides my comment at the `continue` and the fact that you forgot to up the JASP R Interface minor version :p

I do think we should merge it into development because it is minor but useful!

PR #3516, opened 2019-09-06
Fix https://github.com/jasp-stats/jasp-test-release/issues/289

koenderks:
Add an error in the decision matrix when there is only 1 numeric variable remaining after removing all factors from the variables.


JASP-Engine/JASP/R/commonMachineLearningClassification.R
```diff
@@ -334,6 +334,11 @@
   variables <- variables[ !sapply(dataset[, .v(variables)], is.factor) ] # remove factors from boundary plot
   l <- length(variables)
 
+  if(l == 1){
```
TimKDJ: can `l` be `0` at this point, too?

koenderks: Yes, will fix by replacing with <= 1

koenderks:
Now also changes some input variable types for analyses and fixes error in boosting deviance plot

koenderks:
And adds data files for tims blogpost and jasp file for mine

PR #3518, opened 2019-09-09
reintroduce function needed by Bain

vandenman:
fixes https://github.com/jasp-stats/jasp-test-release/issues/297

function `.base_breaks_y3` is taken from the history of JASP:
https://github.com/jasp-stats/jasp-desktop/blob/2576fdd3337a460abb7a47fce52bcee62cc37033/JASP-Engine/JASP/R/ttestbayesianindependentsamples.R

### DIFF
--- a/JASP-Engine/JASP/R/ttestplotfunctions.R
+++ b/JASP-Engine/JASP/R/ttestplotfunctions.R
@@ -2335,3 +2335,13 @@
   list(ggplot2::geom_segment(data=d, ggplot2::aes(x=x, y=y, xend=xend, yend=yend), inherit.aes=FALSE, size = 1),
        ggplot2::scale_y_continuous(breaks=c(min(b), testValue, max(b))))
 }
+
+# only used by bain!
+.base_breaks_y3 <- function(x) {
+
+	ci.pos <- c(x$ciLower, x$ciUpper)
+	b <- pretty(ci.pos)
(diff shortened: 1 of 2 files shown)


JASP-Engine/JASP/R/ttestplotfunctions.R
```diff
@@ -2335,3 +2335,13 @@
   list(ggplot2::geom_segment(data=d, ggplot2::aes(x=x, y=y, xend=xend, yend=yend), inherit.aes=FALSE, size = 1),
        ggplot2::scale_y_continuous(breaks=c(min(b), testValue, max(b))))
 }
+
+# only used by bain!
+.base_breaks_y3 <- function(x) {
```
TimKDJ: Does BAIN have a reason to be so "special"? Or can it use the same stuff as normal t-tests?

vandenman: Probably not. The other functions, `.base_breaks_y2` and `.base_breaks_y4` also use a testValue argument, which could be `NULL` I guess.

TimKDJ: Ok then I think we can also do `.base_breaks_y4 <- function(x, testValue=NULL) {` and have it call `_y4` instead of `_y3`?
`c(5, NULL, 3)` returns `c(5, 3)` so that should work

PR #3519, opened 2019-09-09
Standardize number of decimals in priors across Bayesian analyses

TimKDJ:
See jasp-stats/INTERNAL-jasp#460


AlexanderLyNL:
I think this looks good, but I was able to set r to be 0 in 
- Bayes ANCOVA
Screen Shot 2019-09-09 at 22 01 59

- Bayes ANOVA
Screen Shot 2019-09-09 at 22 01 53

- Bayes Lin regression: Also lower bound alpha please
Screen Shot 2019-09-09 at 22 01 39

- SumStat Bayes lin regression
Screen Shot 2019-09-09 at 22 02 45

I think we should lower bound it by, say, 0.003

AlexanderLyNL:
I think this looks good, but I was able to set r to be 0 in 
- Bayes ANCOVA
Screen Shot 2019-09-09 at 22 01 59

- Bayes ANOVA
Screen Shot 2019-09-09 at 22 01 53

- Bayes Lin regression: 
Screen Shot 2019-09-09 at 22 01 39
  - Also lower bound alpha please
  - And also lower bound the parameters of the beta binomial 

- SumStat Bayes lin regression
Screen Shot 2019-09-09 at 22 02 45

I think we should lower bound it by, say, 0.003.

TimKDJ:
Apparently the `alpha` parameter of Bayesian Linear Regression has never worked..
I've corrected the bug that caused this, but noticed that for `alpha <= 2` we cannot compute anything meaningful. In the documentation of BAS we find:
> For Zellner's g-prior, alpha = g, for the Liang et al hyper-g or hyper-g-n method, recommended choice is alpha are between (2 < alpha < 4), with alpha = 3 the default.

Shall I protect our users and set the min to 2, max 4, exclusive?

TimKDJ:
Btw is there a reason for `0.003`? If it's simply to exclude the bound, we have a parameter for that

AlexanderLyNL:
- Yes for the alpha's in Liang et al. 
- For correlation I found 0.003 because it was just good enough for overflows. When the width it zero, you have large number divided by a large number which makes the 2F1 shit itself. There's no real reason why it isn't 0, other than numerical stability for the correlation. Perhaps you can just set it > 0 for the other analyses.

AlexanderLyNL:
Sorry that I only noticed it just now, but it's possible in 
- Bayesian linear regression 
- Additional Options
- Model Prior - Bernoulli 
  - p = 0
  - p = 1

Could you restrict it so p in (0,1)? 

Screen Shot 2019-09-10 at 13 22 05

PR #3522, opened 2019-09-10
Some first bug fixes

koenderks:
+ Add random forest help file (https://github.com/jasp-stats/jasp-test-release/issues/271)
+ Add value to density based k-distance plot
+ Add error for clustering with more clusters than data points (https://github.com/jasp-stats/jasp-test-release/issues/334)
+ Fix issue with random forest training data used per tree (https://github.com/jasp-stats/jasp-test-release/issues/332)
+ Show warning when R-squared cannot be computed due to zero variance in the predictions (https://github.com/jasp-stats/jasp-test-release/issues/337)

### DIFF
--- a/JASP-Engine/JASP/R/commonMachineLearningClassification.R
+++ b/JASP-Engine/JASP/R/commonMachineLearningClassification.R
@@ -35,43 +35,51 @@
   return(dataset)
 }
 
-.errorHandlingClassificationAnalyses <- function(dataset, options){
+.errorHandlingClassificationAnalyses <- function(dataset, options, type){
   predictors                <- unlist(options['predictors'])
   predictors                <- predictors[predictors != ""]
   target                    <- NULL
   if(options[["target"]] != "")
     target                  <- options[["target"]]
   variables.to.read         <- c(predictors, target)
+
+  checkNearestNeighbors <- function( ){
+    if(type == "knn"){
+      # Adjust for too much nearest neighbors (nn > nTrain) before the analysis starts
+      nn <- base::switch(options[["modelOpt"]], "optimizationManual" = options[["noOfNearestNeighbours"]], "optimizationError" = options[["maxK"]])
+      if(options[["testSetIndicatorVariable"]] != "" && options[["holdoutData"]] == "testSetIndicator"){
+        nTrain <- length(which(dataset[, .v(options[["testSetIndicatorVariable"]])] == 0))
+      } else {
+        nTrain <- ceiling(nrow(dataset) - nrow(dataset)*options[['testDataManual']])
+      }
+      if(options[["modelOpt"]] == "optimizationError"){
+        if(options[["modelValid"]] == "validationManual")
+          nTrain <- ceiling(nTrain - nTrain*options[['validationDataManual']])
+        if(options[["modelValid"]] == "validationKFold")
+          nTrain <- ceiling(nTrain - nTrain / options[["noOfFolds"]])
+        if(options[["modelValid"]] == "validationLeaveOneOut")
+          nTrain <- nTrain - 1
+      }
+      if(nn >= nTrain)
+        return(paste0("You have specified more nearest neighbors than observations in the training set. Please choose a number lower than ", nTrain, "."))
+    }
+  }
+
   errors <- .hasErrors(dataset, perform, type = c('infinity', 'observations'),
-                       all.target = variables.to.read,
+                       all.target = variables.to.read, custom = checkNearestNeighbors,
                        observations.amount = "< 2",
                        exitAnalysisIfErrors = TRUE)
 
-  dataset <- na.omit(dataset)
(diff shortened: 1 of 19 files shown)


JASP-Engine/JASP/R/commonMachineLearningClassification.R
```diff
@@ -873,7 +873,7 @@
     typeData <- cbind(train, levelVar = factor(levelVar))
     typeData <- typeData[, -which(colnames(typeData) == .v(options[["target"]]))]
     
-    score <- .calcAUCScore(AUCformula, test, typeData, levelVar, options, ...)
+    score <- .calcAUCScore(AUCformula, train, test, typeData, levelVar, options, ...)
```
TimKDJ: If you change this to
`score <- .calcAUCScore(AUCformula, train=train, test=test, typeData=typeData, levelVar=levelVar, options=options, ...)`
Then you don't have to specify `train` in the S3Methods that don't actually use that parameter (=`ldaClassification`, `boostingClassification`, `knnClassification`)

Will make it a little easier to make changes in the future.

JASP-Engine/JASP/R/commonMachineLearningClassification.R
```diff
@@ -909,16 +909,16 @@
...
 
-.calcAUCScore.knnClassification <- function(AUCformula, test, typeData, nn, distance, weights, ...) {
+.calcAUCScore.knnClassification <- function(AUCformula, train, test, typeData, nn, distance, weights, ...) {
   kfit_auc <- kknn::kknn(formula = AUCformula, train = typeData, test = test, k = nn, distance = distance, kernel = weights, scale = FALSE)
   score <- predict(kfit_auc, test, type = 'prob')[, 'TRUE']
   return(score)
 }
 
-.calcAUCScore.randomForestClassification <- function(AUCformula, test, typeData, levelVar, options, dataset, noOfTrees, noOfPredictors, ...) {
+.calcAUCScore.randomForestClassification <- function(AUCformula, train, test, typeData, levelVar, options, dataset, noOfTrees, noOfPredictors, ...) {
```
TimKDJ: `dataset` still needed?

JASP-Engine/JASP/R/commonMachineLearningClustering.R
```diff
@@ -115,7 +120,16 @@
 
   clusteringTable$addCitation("Hartigan, J. A., & Wong, M. A. (1979). Algorithm AS 136: A k-means clustering algorithm. Journal of the Royal Statistical Society. Series C (Applied Statistics), 28(1), 100-108.")
 
-  if(!ready) clusteringTable$addFootnote(message="Please provide at least 2 variables.", symbol="<i>Note.</i>")
+  if(type != "densitybased"){
+    clusters  <- base::switch(options[["modelOpt"]], "validationManual" = options[["noOfClusters"]], "validationOptimized" = options[["maxClusters"]])
+    tooMuch   <- clusters > (nrow(dataset) - 1)
+
+    if(!ready && tooMuch)
+      clusteringTable$addFootnote(message = "You have specified more clusters than distinct data points", symbol="<b>Warning.</b>")
```
TimKDJ: This is a user input error -- like when entering infinity or zero variance, etc. Shouldn't this be handled by `.hasErrors()` as is the custom?

TimKDJ: So something like
```
.errorHandlingClusteringAnalyses <- function(dataset, options, type) {
  predictors <- unlist(options$predictors)
  if (length(predictors[predictors != ""]) == 0L)
    return()

  checkClusters <- function() {
    if (type != "densitybased") {
      clusters  <- base::switch(options[["modelOpt"]], "validationManual" = options[["noOfClusters"]], "validationOptimized" = options[["maxClusters"]])
      if (clusters > (nrow(dataset) - 1))
        return("You have specified more clusters than distinct data points")
    }
  }
  
  .hasErrors(dataset, type = c('infinity', 'observations'), custom=checkClusters, all.target = predictors,
            observations.amount = "< 2", exitAnalysisIfErrors = TRUE)

  return()
}
```

where the function now requires `type`

TimKDJ: And if you set `ready` before you called `.errorHandlingClusteringAnalyses` you would not need to check the number of predictors

koenderks: Hmm, point is that JASP crashes quite hard when you do this right? Isn't it better that we just show empty tables?

JASP-Engine/JASP/R/commonMachineLearningClassification.R
koenderks: No, got it

JASP-Engine/JASP/R/commonMachineLearningClassification.R
koenderks: yes, done

PR #3529, opened 2019-09-10
Fix errors in Bayesian t-tests

vandenman:
fixes https://github.com/jasp-stats/jasp-test-release/issues/343
fixes https://github.com/jasp-stats/jasp-test-release/issues/344 by using JASPgraphs::themeJaspRaw()
afbeelding

fixes https://github.com/jasp-stats/jasp-test-release/issues/345 by adding a NaN

> 1. We're by default always placing a 3 decimal value in the footnote (so if testValue = 0, the footnote reads 0.000). This should only be a 3 decimal value if that is what the user entered.

Agreed, but it's pretty hard to tell the difference between 1, 1.0, 1.00, and 1.000. My solution right now is to drop any trailing 0s. So the previous all become 1.

> 2. It looks quite strange to have a footnote indicator without a cell value. I think we should put NaN there if a value cannot be computed, like it was beforehand

Agreed, added the NaN:
afbeelding

fixes https://github.com/jasp-stats/jasp-test-release/issues/346

> Is it by design that the user prior is now the dotted line as opposed to the solid line?

Nope, good catch! Should be fixed now.
afbeelding

### DIFF
--- a/JASP-Engine/JASP/R/commonbayesianttest.R
+++ b/JASP-Engine/JASP/R/commonbayesianttest.R
@@ -1094,29 +1094,14 @@
 
 	pd <- ggplot2::position_dodge(.2)
 
-	p <-	ggplot2::ggplot(summaryStat, mapping = mapping) +
+	p <-	JASPgraphs::themeJasp(ggplot2::ggplot(summaryStat, mapping = mapping) +
 			ggplot2::geom_errorbar(ggplot2::aes(ymin=ciLower, ymax=ciUpper), colour="black", width=.2, position=pd) +
 			ggplot2::geom_line(position=pd, size = .7) +
 			ggplot2::geom_point(position=pd, size=4) +
-			xlab + ylab +
-			ggplot2::theme_bw() +
-			ggplot2::theme(panel.grid.minor=ggplot2::element_blank(), plot.title = ggplot2::element_text(size=18),
-				panel.grid.major=ggplot2::element_blank(),
-				axis.title.x = ggplot2::element_text(size=18,vjust=-.2), axis.title.y = ggplot2::element_text(size=18,vjust=-1),
-				axis.text.x = ggplot2::element_text(size=15), axis.text.y = ggplot2::element_text(size=15),
-				panel.background = ggplot2::element_rect(fill = 'transparent', colour = NA),
-				plot.background = ggplot2::element_rect(fill = 'transparent', colour = NA),
-				legend.background = ggplot2::element_rect(fill = 'transparent', colour = NA),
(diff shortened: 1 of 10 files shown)


TimKDJ:
Looks good besides the footnote that goes missing.
On a somewhat unrelated note, the number of decimals you give is a little random:
Bayes factor (new <-> old)
Screenshot 2019-09-11 at 10 30 11
Bayes factor (new; old does 3 decimals everywhere)
Screenshot 2019-09-11 at 10 32 10

And is this difference between the old and new plots due to sampling (note the maxBF and placement of the dots)?
new <-> old
Screenshot 2019-09-11 at 10 35 58

vandenman:
I'll have a look at the footnote!

> 1. On a somewhat unrelated note, the number of decimals you give in the plots is a little random: Bayes factor

Actually, they are now all reported with the same number of [Significant figures](https://en.wikipedia.org/wiki/Significant_figures). I think this is more natural, although it may be less visually appealing.

> 2. And is this difference between the old and new plots due to sampling (note the maxBF and placement of the dots)?

Yes. For inequality hypotheses, we call `.oneSidedTtestBFRichard` which calls `BayesFactor::ttestBF(x=x, y=y, paired=paired, posterior=TRUE, ...`. There shouldn't be any need to sample from the posterior to compute this, but that's how it's been for a while now (and I didn't want to touch any of that). You actually see in both screenshots that the user prior isn't even on the line... (but that's okay I guess?)

TimKDJ:
> Actually, they are now all reported with the same number of Significant figures. I think this is more natural, although it may be less visually appealing.

Ah. Yes. That makes sense

TimKDJ:
Descriptives plots are no longer removed if the corresponding variable is

vandenman:
Hmm, this happened because jaspContainers without dependencies are never thrown away. The fix is to give them a random dependency. I thought this bug was fixed in the past, but apparently not.

TimKDJ:
It is by design that something without a dependency is not thrown away. This was actually a deliberate change from always throwing things away when it has no dependency. Otherwise we couldn't do the "inherit from parent" dependency structure

vandenman:
That's all fine, but when a container doesn't have dependencies and it's childeren do, the dependencies of the childeren never get checked. So even though the plots had `plot$dependOn(optionContainsValue = list(variables = var))`, this dependency was never actually checked, because their parent didn't have a dependency.

TimKDJ:
Okay, so at the moment the state for descriptives plots is not quite there yet.

1. Add contNormal
2. Tick "Descriptives" (the plot one)
3. Remove contNormal
4. Add contNormal

PR #3541, opened 2019-09-13
[NOT FOR 0.11] Prevent images from flickering when analyses update

TimKDJ:
This should make the updates in the results window much more smooth when there are images. No more flickering!

This still needs to be tested more thoroughly, especially for the old-style analyses. Hard requirement of this fix: image names are never recycled.


TimKDJ:
I have to test whether I need to add an analysis id

PR #3543, opened 2019-09-16
Add jasp files for machine learning

koenderks:
A bit uninformed though...but it's something. May have to put more information in these later.


koenderks:
Added in-depth explanations to the JASP files, I think they look good!

PR #3558, opened 2019-09-25
adding piechart

FBartos:
Fixes https://github.com/jasp-stats/jasp-issues/issues/225

### DIFF
--- a/JASP-Engine/JASP/R/descriptives.R
+++ b/JASP-Engine/JASP/R/descriptives.R
@@ -1303,7 +1303,6 @@ Descriptives <- function(jaspResults, dataset, options) {
 
   if (any(is.infinite(column))) {
     plotObj$setError("Plotting is not possible: Variable contains infinity")
-    plotObj$plotObject    <- JASPgraphs::themeJasp(JASPgraphs::drawAxis(xName = variable, xBreaks = 1:5, yBreaks = 1:5))
   }
   else if (length(column) < 3) {
     plotObj$setError("Plotting is not possible: Too few rows (left)")
--- a/JASP-Engine/JASP/R/descriptives.R
+++ b/JASP-Engine/JASP/R/descriptives.R
@@ -171,6 +171,9 @@ Descriptives <- function(jaspResults, dataset, options) {
     piePlots <- jaspResults[["pieCharts"]]
 
     for (var in variables) {
+      # skip non-categorical variables
+      if(is.double(dataset.factors[[.v(var)]]))next
+      
       if(is.null(piePlots[[var]])) {
         if (makeSplit) {
           piePlots[[var]] <- .descriptivesPieChart(dataset = splitDat.factors, options = options, variable = var)
@@ -1284,14 +1287,14 @@ Descriptives <- function(jaspResults, dataset, options) {
     plotResult$dependOn(optionContainsValue=list(variables=variable))
 
     for (l in split) {
-      plotResult[[l]] <- .descriptivesPieChart_SubFunc(column=dataset[[l]][[.v(variable)]], variable=variable, width=options$plotWidth, height=options$plotHeight, displayDensity = options$distPlotDensity, title = l)
+      plotResult[[l]] <- .descriptivesPieChart_SubFunc(column=dataset[[l]][[.v(variable)]], variable=variable, width=options$plotWidth, height=options$plotHeight, title = l)
       plotResult[[l]]$dependOn(optionsFromObject=plotResult)
     }
 
     return(plotResult)
   } else {
     column <- dataset[[.v(variable)]]
-    aPlot <- .descriptivesPieChart_SubFunc(column=column[!is.na(column)], variable=variable, width=options$plotWidth, height=options$plotHeight, displayDensity = options$distPlotDensity, title = variable)
+    aPlot <- .descriptivesPieChart_SubFunc(column=column[!is.na(column)], variable=variable, width=options$plotWidth, height=options$plotHeight, title = variable)
     aPlot$dependOn(options="splitby", optionContainsValue=list(variables=variable))
 
     return(aPlot)
@@ -1306,7 +1309,6 @@ Descriptives <- function(jaspResults, dataset, options) {
   }
   else if (length(column) < 3) {
     plotObj$setError("Plotting is not possible: Too few rows (left)")
-    plotObj$plotObject    <- JASPgraphs::themeJasp(JASPgraphs::drawAxis(xName = variable, xBreaks = 1:5, yBreaks = 1:5))
   }
   else if (length(column) > 0) {
     tb  <- as.data.frame(table(column))
--- a/JASP-Engine/JASP/R/descriptives.R
+++ b/JASP-Engine/JASP/R/descriptives.R
@@ -158,6 +158,28 @@ Descriptives <- function(jaspResults, dataset, options) {
       }
     }
   }
+
+  # Pie charts
+  if (options$descriptivesPiechart) {
+
+    if(is.null(jaspResults[["pieCharts"]])) {
+      jaspResults[["pieCharts"]] <- createJaspContainer("Pie charts")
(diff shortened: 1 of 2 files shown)


JASP-Engine/JASP/R/descriptives.R
```diff
@@ -1253,4 +1275,45 @@ Descriptives <- function(jaspResults, dataset, options) {
   return(descriptivesQQPlot)
 }
 
+.descriptivesPieChart <- function(dataset, options, variable) {
+  if (options$splitby != "" ) {
+    # return a collection
+    split <- names(dataset)
+
+    plotResult <- createJaspContainer(title=variable)
+    plotResult$dependOn(options="splitby", optionContainsValue=list(variables=variable))
```
vandenman:
```suggestion
    plotResult$dependOn(optionContainsValue=list(variables=variable))
```
`options="splitby"` is already a dependency of the container.

JASP-Engine/JASP/R/descriptives.R
```diff
@@ -1253,4 +1275,45 @@ Descriptives <- function(jaspResults, dataset, options) {
...
+    return(aPlot)
+  }
+}
+
+.descriptivesPieChart_SubFunc <- function(column, variable, width, height, displayDensity, title) {
+  plotObj <- createJaspPlot(title=title, width=width, height=height)
+
+  if (any(is.infinite(column))) {
+    plotObj$setError("Plotting is not possible: Variable contains infinity")
+    plotObj$plotObject    <- JASPgraphs::themeJasp(JASPgraphs::drawAxis(xName = variable, xBreaks = 1:5, yBreaks = 1:5))
```
vandenman:
```suggestion
```
If you set an error, there should be no need to fill the plotObject.

Resources/Descriptives/qml/Descriptives.qml
```diff
@@ -62,6 +62,7 @@ Form
 				}
 			}
 			CheckBox {				name: "descriptivesQQPlot";		label: qsTr("Q-Q plots")						}
+            CheckBox {				name: "descriptivesPiechart";	label: qsTr("Pie charts")						}
```
vandenman:
```suggestion
			CheckBox {				name: "descriptivesPiechart";	label: qsTr("Pie charts")						}
```
rather nitpicky, QML uses tabs rather than spaces:
afbeelding

JASP-Engine/JASP/R/descriptives.R
```diff
@@ -158,6 +158,28 @@ Descriptives <- function(jaspResults, dataset, options) {
       }
     }
   }
+
+  # Pie charts
+  if (options$descriptivesPiechart) {
+
+    if(is.null(jaspResults[["pieCharts"]])) {
+      jaspResults[["pieCharts"]] <- createJaspContainer("Pie charts")
```
vandenman: Perhaps a little bit more concise:
```r
if(is.null(jaspResults[["pieCharts"]])) {
      jaspResults[["pieCharts"]] <- createJaspContainer("Pie charts", dependencies = c("splitby", "descriptivesPiechart"), position = 9)
```

JASP-Engine/JASP/R/descriptives.R
```diff
@@ -1253,4 +1275,45 @@ Descriptives <- function(jaspResults, dataset, options) {
...
+  } else {
+    column <- dataset[[.v(variable)]]
+    aPlot <- .descriptivesPieChart_SubFunc(column=column[!is.na(column)], variable=variable, width=options$plotWidth, height=options$plotHeight, displayDensity = options$distPlotDensity, title = variable)
+    aPlot$dependOn(options="splitby", optionContainsValue=list(variables=variable))
+
+    return(aPlot)
+  }
+}
+
+.descriptivesPieChart_SubFunc <- function(column, variable, width, height, displayDensity, title) {
```
vandenman:
```suggestion
.descriptivesPieChart_SubFunc <- function(column, variable, width, height, title) {
```
argument `displayDensity` is unused.

JASP-Engine/JASP/R/descriptives.R
FBartos: ok, but the original version is used trough the whole script. Should I rewrite the rest of descriptives.R to be like this suggestion?

JASP-Engine/JASP/R/descriptives.R
FBartos: true, thanks for pointing it out

JASP-Engine/JASP/R/descriptives.R
FBartos: thanks, forgot to delete it

JASP-Engine/JASP/R/descriptives.R
FBartos: noted

Resources/Descriptives/qml/Descriptives.qml
FBartos: I see, I will use tabs next time

FBartos:
1. Maybe we could add a dropdown box with predefined palets and pass them to the JASPgraphs::JASPcolors 
2. Yes, you are right, should I disable the plots for continuous variables or limit them to a maximum number of categories?

vandenman:
> 1. Maybe we could add a dropdown box with predefined palets and pass them to the JASPgraphs::JASPcolors

Sounds good.

> 2. Yes, you are right, should I disable the plots for continuous variables or limit them to a maximum number of categories?

Let's show the plot as long as the variable type is ordinal or nominal.

JASP-Engine/JASP/R/descriptives.R
vandenman: Nah that's overkill. There isn't really a preferred way of doing this (yet).

JASP-Engine/JASP/R/descriptives.R
```diff
@@ -171,6 +171,9 @@ Descriptives <- function(jaspResults, dataset, options) {
     piePlots <- jaspResults[["pieCharts"]]
 
     for (var in variables) {
+      # skip non-categorical variables
```
FBartos: skipping all continuous variables - is there another way to find the variable specification from JASP? (I couldn't find it in "options")

vandenman: I don't think there is, this solution looks good to me.

FBartos:
> > 1. Maybe we could add a dropdown box with predefined palets and pass them to the JASPgraphs::JASPcolors
> 
> Sounds good.

ok, but there must be a change in "JASPgraphs::JASPcolors" because the only accessible palet now is colorblind (I don't know how to access or edit "JASPgraphs_data" object which seems to hold the pallet options?)
`
    if (!palette %in% names(JASPgraphs_data)) {
        stop(sprintf("palette was %s but must be one of %s", 
            as.character(palette), names(JASPgraphs_data)))
`

> > 1. Yes, you are right, should I disable the plots for continuous variables or limit them to a maximum number of categories?
> 
> Let's show the plot as long as the variable type is ordinal or nominal.

But this still doesn't solve a problem of an extremely big legend in case of adding a categorical variable with many levels - id

Should I also add a cap on the maximum length of the variable names to prevent this? Showing only the first 10 characters for example?
image

vandenman:
I've made some changes to `JASPgraphs::JASPcolors` on another branch exactly for that reason, but that is not quite ready yet (which is why I think we should add it in a future PR).

An extremely long factor name that messes up a plot is a known problem. There is no real solution for it (yet). Although we can figure out the size of the legend beforehand, we don't really know where to go from there (adjusting the size of the plot could work but is a bit icky). If people have a ridiculously long factor name the easy solution is that they just change the name.

FBartos:
ok, so, it seems to be all for now?

vandenman:
@FBartos Yes, looks good to me!

TimKDJ:
Nice, congrats on your first official contribution @FBartos :)

PR #3560, opened 2019-10-01
Rewriting T tests with jaspResults

gibbona1:
**commonTTest.R**
I have moved everything here essentially, as there is a lot of overlap.

**ttestonesample.R**, **ttestonesample.R**, **ttestonesample.R**
Now just has a `ready `and `ttestRunAnalysis()` function

**test-ttestindependentsamples.R**, **test-ttestonesample.R**, **test-ttestpairedsamples.R**
Minor changes due to containers, removing redundant elements in test tables, and changed error checking from footnotes to errorMessage


JASP-Engine/JASP/R/commonTTest.R
```diff
@@ -18,60 +18,1615 @@
...
+        return()
+      datasetErrorCheck <- data.frame(dataset[[.v(p1)]] - dataset[[.v(p2)]])
+      colnames(datasetErrorCheck) <- .v(paste0("Difference between ", p1, " and ", p2))
+      .hasErrors(datasetErrorCheck, 
+                 type = "variance", 
+                 exitAnalysisIfErrors = TRUE)
+      .hasErrors(dataset,
+                 type = c('observations', 'variance', 'infinity'),
+                 all.target = c(p1, p2), 
+                 observations.amount  = c('< 3', '> 5000'),
```
TimKDJ: You should not check for `> 5000`; the only place where this should be checked is in the normality test.
The reason why we need to check it for the normality test is the limit defined here:
```
> stats::shapiro.test
function (x) 
{
    DNAME <- deparse(substitute(x))
    stopifnot(is.numeric(x))
    x <- sort(x[complete.cases(x)])
    n <- length(x)
    if (is.na(n) || n < 3L || n > 5000L) 
        stop("sample size must be between 3 and 5000")
    rng <- x[n] - x[1L]
    if (rng == 0) 
        stop("all 'x' values are identical")
    if (rng < 1e-10) 
        x <- x/rng
    res <- .Call(C_SWilk, x)
    RVAL <- list(statistic = c(W = res[1]), p.value = res[2], 
        method = "Shapiro-Wilk normality test", data.name = DNAME)
    class(RVAL) <- "htest"
    return(RVAL)
}
```

JASP-Engine/JASP/R/commonTTest.R
```diff
@@ -18,60 +18,1615 @@
...
+                 observations.amount  = c('< 3', '> 5000'), 
+                 exitAnalysisIfErrors = TRUE)
+    }
+  }
+  else if(type == "one-sample") {
+    for (variable in options$variables)
+      .hasErrors(dataset, 
+                 type = c('observations', 'variance', 'infinity'),
+                 all.target = variable,
+                 observations.amount  = c('< 3', '> 5000'),
```
TimKDJ: Same adjustment

JASP-Engine/JASP/R/commonTTest.R
TimKDJ: Also check for < 2 instead of < 3, like it used to be before

JASP-Engine/JASP/R/commonTTest.R
```diff
@@ -18,60 +18,1615 @@
...
+    wantsWelchs <- options$welchs
+    wantsWilcox <- options$mannWhitneyU
+    whichTests  <- list("1" = wantsStudents, "2" = wantsWelchs, "3" = wantsWilcox)
+    percentConfidenceEffSize  <- options$descriptivesEffectSizeConfidenceIntervalPercent
+    percentConfidenceMeanDiff <- options$descriptivesMeanDiffConfidenceIntervalPercent
+  }
+  allTests <- c(wantsStudents, wantsWilcox)
+  if(type == "one-sample")
+    allTests <- c(allTests, wantsZtest)
+  onlyTest <- sum(allTests) == 1
```
TimKDJ: This is a copy of the lines at the start of `.ttestMainFill()`; can you put this in a function to prevent the duplication?

JASP-Engine/JASP/R/commonTTest.R
```diff
@@ -18,60 +18,1615 @@
...
+                  "meanDifference", "meanDiffConfidenceIntervalCheckbox",
+                  "meanDiffConfidenceIntervalPercent", "hypothesis", "VovkSellkeMPR")
+  pairList        <- c("pairs", "wilcoxonSignedRank")
+  oneSampleList   <- c("zTest", "testValue")
+  independentList <- c("groupingVariable", "effectSizesType", "welchs",
+                             "descriptivesMeanDiffConfidenceIntervalPercent")
+  ttest$dependOn(c(dependList, pairList, oneSampleList, independentList))
+  ttest$showSpecifiedColumnsOnly <- TRUE
+  
+  if(type == "paired") {
```
TimKDJ, line 76: I would recommend splitting this up in one sample table markup, paired samples table markup, etc. These functions could then be placed in their respective files (as it's not "common" functionality), and be called from this function. I think otherwise it might be tricky to maintain this function in the long run.

JASP-Engine/JASP/R/commonTTest.R
```diff
@@ -18,60 +18,1615 @@
...
+    whichTests  <- list("1" = wantsStudents, "2" = wantsWelchs, "3" = wantsWilcox)
+    percentConfidenceEffSize  <- options$descriptivesEffectSizeConfidenceIntervalPercent
+    percentConfidenceMeanDiff <- options$descriptivesMeanDiffConfidenceIntervalPercent
+  }
+  allTests <- c(wantsStudents, wantsWilcox)
+  onlyTest <- sum(allTests) == 1
+  
+  direction <- .ttestMainGetDirection(options$hypothesis)
+  
+  if(type == "paired") {
```
TimKDJ: Same here. Better to split it up and place the `if paired; else if independent else if one sample`code belonging to each analysis in their dedicated analysis file

I think when you can't distill any common code (because it's of course better to make code common to all cases, but this is not always possible), then the code should not be in the common file

JASP-Engine/JASP/R/commonTTest.R
```diff
@@ -18,60 +18,1615 @@
...
+  ttestNormalTable$addFootnote(message)
+  
+  container[["ttestNormalTable"]] <- ttestNormalTable
+ 
+  res <- try(.ttestNormalFill(container, dataset, options, ready, type))
+  .ttestSetError(res, ttestNormalTable)
+}
+
+.ttestNormalFill <- function(container, dataset, options, ready, type) {
+  if(type == "paired") {
```
TimKDJ: Some more splitting up here

JASP-Engine/JASP/R/commonTTest.R
```diff
@@ -18,60 +18,1615 @@
...
+        
+      } else 
+        row <- list(v = variable, W = ".", p = ".")
+      row[[".isNewGroup"]] <- rowNo == 1
+      container[["ttestNormalTable"]]$addRows(row)
+    }
+  }
+}
+
+.ttestIndependentEqVarTable <- function(jaspResults, dataset, options, ready, type){
```
TimKDJ: This can be placed in the independent samples ttest file

JASP-Engine/JASP/R/commonTTest.R
```diff
@@ -18,60 +18,1615 @@
...
+  equalityVariance$addColumnInfo(name = "F",  type = "number")
+  equalityVariance$addColumnInfo(name = "df", type = "integer")
+  equalityVariance$addColumnInfo(name = "p",  type = "pvalue")
+  
+  container[["equalityVariance"]] <- equalityVariance
+  res <- try(.ttestIndependentEqVarFill(container, dataset, options, ready))
+  .ttestSetError(res, equalityVariance)
+}
+
+.ttestIndependentEqVarFill <- function(container, dataset, options, ready){
```
TimKDJ: And this can also go the independent samples t-test file

JASP-Engine/JASP/R/commonTTest.R
```diff
@@ -18,60 +18,1615 @@
...
+    container[["table"]]$addRows(data)
+  }
+}
+
+.ttestDescriptivesPlot <- function(jaspResults, dataset, options, ready, type) {
+  if(!options$descriptivesPlots)
+    return()
+  .ttestDescriptivesContainer(jaspResults, options)
+  container <- jaspResults[["ttestDescriptives"]]
+  container[["title"]] <- createJaspContainer("Descriptives Plots")
```
TimKDJ: This looks a little strange to me. `title` is a container?

JASP-Engine/JASP/R/commonTTest.R
```diff
@@ -18,60 +18,1615 @@
...
+    for(variable in options$variables) {
+      title <- variable
+      .ttestDescriptivesPlotSetup(subcontainer, dataset, options, ready, type, title, variable)
+    }
+  }
+  return()
+}
+
+.ttestDescriptivesPlotFill <- function(dataset, options, type, var) {
+  if(type == "paired") {
```
TimKDJ: Separated and in their own file

JASP-Engine/JASP/R/commonTTest.R
```diff
@@ -18,60 +18,1615 @@
...
+    .ttestCheckErrors(        dataset, options, type)
+  }
+  # Output tables (each calls its own results function)
+  .ttestMainTable(  jaspResults, dataset, options, ready, type)
+  .ttestNormalTable(jaspResults, dataset, options, ready, type)
+  if(type == "independent")
+    .ttestIndependentEqVarTable(jaspResults, dataset, options, ready, type)
+  # Descriptives
+  .ttestDescriptivesTable(jaspResults, dataset, options, ready, type)
+  .ttestDescriptivesPlot( jaspResults, dataset, options, ready, type)
```
TimKDJ: I don't suppose we can use the bayesian functions for this?

JASP-Engine/JASP/R/commonTTest.R
```diff
@@ -18,60 +18,1615 @@
...
+    if (!is.null(groups) && groups == "") 
+      groups <- NULL
+    if(type %in% c("one-sample", "independent"))
+      depvars <- unlist(options$variables)
+    else if (type == 'paired') {
+      depvars <- unlist(options$pairs)
+      depvars <- depvars[depvars != ""]
+    }
+    exclude <- NULL
+    if (options$missingValues == "excludeListwise")
```
TimKDJ, line 35: It's important that all elements `$dependOn()` on the missing value handling, because going from "Exclude cases analysis by analysis" to "Exclude cases listwise" should change literally all output

JASP-Engine/JASP/R/commonTTest.R
```diff
@@ -18,60 +18,1615 @@
...
+      nameOfLocationParameter <- "Mean Difference"
+      nameOfEffectSize <- "Cohen's d"
+    } else {
+      testStat <- "Statistic"
+      nameOfLocationParameter <-  "Location Parameter"
+      nameOfEffectSize <-  "Effect Size"
+    }
+    
+    ## if the user wants all tests, add a column called "Test"
+    if (sum(allTests) == 2)
```
TimKDJ: This should be 3 for one sample and independent samples

JASP-Engine/JASP/R/commonTTest.R
TimKDJ: For all output elements please set their `position` so it's deterministic in the output

JASP-Tests/R/tests/testthat/test-ttestindependentsamples.R
```diff
@@ -82,22 +80,22 @@ test_that("Analysis handles errors", {
   options$variables <- "debInf"
   options$groupingVariable <- "contBinom"
   results <- jasptools::run("TTestIndependentSamples", "test.csv", options)
-  notes <- unlist(results[["results"]][["ttest"]][["footnotes"]])
-  expect_true(any(grepl("infinity", notes, ignore.case=TRUE)), label = "Inf check")
+  msg <- results[["results"]][["errorMessage"]]
```
TimKDJ: I don't think this is a good change. Each t-test row is independent, so one bad row should not affect the others. It's definitely harder to implement like that, but it's the philosophy we try to follow in most analyses.

JASP-Engine/JASP/R/commonTTest.R
```diff
@@ -18,60 +18,1615 @@
...
+  if(!options$descriptivesPlots)
+    return()
+  .ttestDescriptivesContainer(jaspResults, options)
+  container <- jaspResults[["ttestDescriptives"]]
+  container[["title"]] <- createJaspContainer("Descriptives Plots")
+  subcontainer <- container[["title"]]
+  if(type == "paired") {
+    for(pair in options$pairs) {
+      title <- paste(pair, collapse=" - ")
+      .ttestDescriptivesPlotSetup(subcontainer, dataset, options, ready, type, title, pair)
```
TimKDJ: There is a check missing to see if a plot was already created. As it stands the plots are always recreated on every analysis run. 
There's also a missing`descriptivesPlotsConfidenceInterval` dependency

JASP-Engine/JASP/R/commonTTest.R
```diff
@@ -18,60 +18,1615 @@
...
+    
+    ttest$addColumnInfo(name = "v",      type = "string", title = "")
+    ttest$addColumnInfo(name = testStat, type = "number")
+    if(wantsStudents)
+      ttest$addColumnInfo(name = "df",     type = "integer")
+    ttest$addColumnInfo(name = "p",      type = "pvalue")
+    
+    ## if the user wants more than one test, add a column called "Test"
+    if (sum(allTests) > 1) 
+      ttest$addColumnInfo(name = "test", type = "string", title = "Test")
```
TimKDJ: This one should go earlier, the test column was next to `"v"`

JASP-Engine/JASP/R/commonTTest.R
```diff
@@ -18,60 +18,1615 @@
...
+                confIntEffSize[2] <- Inf
+              else if (direction == "less")
+                confIntEffSize[1] <- -Inf
+              
+              confIntEffSize <- sort(confIntEffSize)
+            }
+          }
+          
+          ## same for all tests
+          p     <- as.numeric(r$p.value)
```
TimKDJ: `p` is not being updated correctly in the one sample t-test when you toggle the `Alt. Hypothesis` (ditto with vovksellke, but that's because it uses `p`)

JASP-Engine/JASP/R/commonTTest.R
```diff
@@ -18,60 +18,1615 @@
...
+  onlyTest <- sum(allTests) == 1  
+  
+  title <- .ttestMainTitle(type)
+  
+  # Create table
+  ttest <- createJaspTable(title = title)
+  dependList <- c("effectSize", "effSizeConfidenceIntervalCheckbox", "variables",
+                  "effSizeConfidenceIntervalPercent", "students", "mannWhitneyU",
+                  "meanDifference", "meanDiffConfidenceIntervalCheckbox",
+                  "meanDiffConfidenceIntervalPercent", "hypothesis", "VovkSellkeMPR")
```
TimKDJ: `stddev` is missing as a dependency for the one-sample ttest

JASP-Engine/JASP/R/ttestindependentsamples.R
```diff
@@ -15,978 +15,589 @@
...
-
-	ttest[["data"]] <- ttest.rows
-	ttest
+.ttestIndependentMainTable <- function(jaspResults, dataset, options, ready, type) {
+  if (!is.null(jaspResults[["ttest"]])) 
+    return()
+  optionsList <- .ttestOptionsList(options, type)
+  # Create table
+  ttest <- createJaspTable(title = "Independent Samples T-Test")
+  dependList <- c("effectSize", "effSizeConfidenceIntervalCheckbox", "variables",
```
TimKDJ, line 43: This misses a dependency on the actual value of the effect size confidence interval (e.g., `95`)

JASP-Engine/JASP/R/ttestindependentsamples.R
```diff
@@ -15,978 +15,589 @@
...
+  #at least one variable and one grouping variable
+  ready <- length(options$variables) > 0 && options$groupingVariable != ""
+  type  <- "independent"
+  if(ready) {
+    dataset <- .ttestReadData(dataset, options, type)
+    .ttestCheckErrors(        dataset, options, type)
+  }
+  # Output tables (each calls its own results function)
+  .ttestIndependentMainTable(  jaspResults, dataset, options, ready, type)
+  .ttestIndependentNormalTable(jaspResults, dataset, options, ready, type)
```
TimKDJ, line 28: The assumption check tables should appear above the descriptives. At the moment they are above descriptives in paired samples, but below descriptives in one sample and independent samples

JASP-Engine/JASP/R/commonTTest.R
```diff
@@ -18,60 +18,327 @@
...
+      p2 <- pair[[2]]
+      if(is.null(p1) || is.null(p2))
+        return()
+      datasetErrorCheck <- data.frame(dataset[[.v(p1)]] - dataset[[.v(p2)]])
+      colnames(datasetErrorCheck) <- .v(paste0("Difference between ", p1, " and ", p2))
+      .hasErrors(datasetErrorCheck, 
+                 type = "variance", 
+                 exitAnalysisIfErrors = TRUE)
+      .hasErrors(dataset,
+                 type = c('observations', 'variance', 'infinity'),
```
TimKDJ: Maybe my comment in https://github.com/jasp-stats/jasp-desktop/pull/3560#discussion_r331474754 was not very clear. If you perform these error checks here, then the analysis won't show everything that can still be computed.

And you get this difference, between how it used to be and how you changed it:
New <-> Old
Screenshot 2019-11-05 at 12 13 25

Do you think you can make the new sitation as robust as the old one? Otherwise I can also do it at a later otherwise

JASP-Tests/R/tests/testthat/test-ttestindependentsamples.R
```diff
@@ -76,25 +76,6 @@ test_that("Descriptives plot matches", {
 
 test_that("Analysis handles errors", {
   options <- jasptools::analysisOptions("TTestIndependentSamples")
-
-  options$variables <- "debInf"
```
TimKDJ, line 82: Now the original checks can be placed back, right?

JASP-Engine/JASP/R/ttestonesample.R
```diff
@@ -343,9 +356,21 @@ TTestOneSample <- function(jaspResults, dataset = NULL, options, ...) {
...
     variables = "."
+  
   for (variable in variables) {
     
     if (ready) {
+      errors <- .hasErrors(dataset, 
+                           message = 'short', 
+                           type = c('observations', 'variance', 'infinity'),
+                           all.target = variable,
+                           observations.amount = c('> 5000'))
```
TimKDJ: Where is this > 5000 coming from? I don't think we had this before

JASP-Engine/JASP/R/ttestpairedsamples.R
```diff
@@ -176,16 +178,30 @@ TTestPairedSamples <- function(jaspResults, dataset = NULL, options, ...) {
     p1 <- pair[[1]]
     p2 <- pair[[2]]
     
+    errors <- .hasErrors(dataset,
+                         type = c('observations', 'variance', 'infinity'),
```
TimKDJ, line 184: `message = "short"`

JASP-Engine/JASP/R/ttestpairedsamples.R
```diff
@@ -176,16 +178,30 @@ TTestPairedSamples <- function(jaspResults, dataset = NULL, options, ...) {
...
+      if (!identical(errors, FALSE)){
+        jaspResults[["ttest"]]$addFootnote(errors$message)
+        isFirst <- (rowNo %% 2 == 1 && sum(optionsList$allTests) == 2) || (sum(optionsList$allTests) == 1)
+        row <- list(v1  = ifelse(isFirst, p1, ""),
+                    sep = ifelse(isFirst, "-", ""),
+                    v2  = ifelse(isFirst, p2, ""))
+        row[[testStat]] <- NaN
+        jaspResults[["ttest"]]$addRows(row)
+        next
+      }
```
TimKDJ: This adds too many rows for each error; always # statistics + 1

JASP-Engine/JASP/R/ttestonesample.R
gibbona1: I reverted the Normality error check to `observations.amount = c('< 3', '> 5000')` as it was previously

TimKDJ: But this is the descriptives plot?

JASP-Engine/JASP/R/ttestonesample.R
```diff
@@ -388,6 +413,15 @@ TTestOneSample <- function(jaspResults, dataset = NULL, options, ...) {
 }
 
 .ttestOneSampleDescriptivesPlotFill <- function(dataset, options, variable) {
+  errors <- .hasErrors(dataset, 
+                       message = 'short', 
+                       type = c('observations', 'variance', 'infinity'),
+                       all.target = variable,
+                       observations.amount = c('> 5000'))
```
TimKDJ: Ah put it on the wrong one -- this > 5000

TimKDJ:
You have a bunch of files in here right now that you don't want (csv's), I'm guessing the rebase did not go exactly as planned

JASP-Engine/JASP/R/ttestonesample.R
gibbona1: Yeah sorry, changed this back to `observations.amount = c('< 2')`

JASP-Engine/JASP/R/ttestpairedsamples.R
TimKDJ: This is not fixed, note the third row:
Screenshot 2019-11-20 at 11 22 11

JASP-Engine/JASP/R/ttestpairedsamples.R
```diff
@@ -15,619 +15,472 @@
...
+    if(length(pair) < 2) next
+    p1 <- pair[[1]]
+    p2 <- pair[[2]]
+    
+    errors <- .hasErrors(dataset,
+                         type = c('observations', 'variance', 'infinity'),
+                         all.target = c(p1, p2), 
+                         observations.amount  = c('< 2', '>5000'))
+    if(!identical(errors, FALSE)) {
+      container[["ttestNormalTable"]]$addFootnote(errors$message, colnames = c("W", "p"), rowNames = paste(p1, p2, sep = "-"))
```
TimKDJ: colNames* (note the capital N)

JASP-Engine/JASP/R/ttestonesample.R
```diff
@@ -15,663 +15,443 @@
...
+                         observations.amount = '< 2')
+    for (test in seq_len(length(optionsList$whichTests))) {
+      currentTest <- optionsList$whichTests[[test]]
+      
+      ## don't run a test the user doesn't want
+      if (!currentTest) 
+        next
+      if (!identical(errors, FALSE)){
+        jaspResults[["ttest"]]$addFootnote(errors$message, colNames = c("p", "m", testStat), rowNames = variables)
+        row <- list(v = variable, p = "NaN", m = "NaN", lowerCI = "", upperCI = "")
```
TimKDJ: I'd just use the `testStat` column, to be consistent with the independent and paired samples ttests
(or do the other two like this)

PR #3567, opened 2019-10-03
Rewrite PCA for JASP results V2 (V1: #3491)

vankesteren:
Continuation of the PCA rewrite.


vankesteren:
Fixed, the entire analysis should only run with more than 1 variable.

PR #3568, opened 2019-10-03
Remove timer from Bayesian audit workflow

koenderks:
Did not test properly yet, as JASP keeps crashing on my computer. Just removed the timer.


koenderks:
Seems with the hotfix the other errors disappeared. The frequentist one works now, but the Bayesian workflow filter doesn't update well. It seems this is the only difference between the two, so this pull should fix that.

PR #3575, opened 2019-10-04
Add support for TSV files

larsgw:
See https://github.com/jasp-stats/jasp-issues/issues/511

### DIFF
--- a/JASP-Common/utils.cpp
+++ b/JASP-Common/utils.cpp
@@ -42,6 +42,7 @@ const char* Utils::getFileTypeString(const Utils::FileType &fileType)
 	{
 	case Utils::csv:		return "csv";
 	case Utils::txt:		return "txt";
+	case Utils::tsv:		return "tsv";
 	case Utils::sav:		return "sav";
 	case Utils::ods:		return "ods";
 	case Utils::jasp:		return "jasp";
--- a/JASP-Common/utils.h
+++ b/JASP-Common/utils.h
@@ -28,7 +28,7 @@ class Utils
 public:
 	friend class PreferencesModel;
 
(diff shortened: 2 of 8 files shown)


larsgw:
The current version of JASP requires Qt 5.13 if I am not mistaken, at least since https://github.com/jasp-stats/jasp-desktop/commit/c19738cfc07a62fe78d86fdc21bd1646869946ef introduced [`QTransposeProxyModel`](https://doc-snapshots.qt.io/qt5-dev/qtransposeproxymodel.html). However, that version isn't available yet on Debian. Can someone else build and test this? Also, am I correct that there is no test suite for importing/exporting?

vandenman:
Thank you for your contribution!

just an FYI, it should be possible to use Qt 5.13 on Debian. If you go to 
https://www.qt.io/download-qt-installer?hsCtaTracking=99d9dd4f-5681-48d2-b096-470725510d34%7C074ddad0-fdef-4e53-8aa8-5e8a876d6ab4

I get the following:

image

After installing you should get the Qt Maintenance tool which should allow you to install Qt 5.13 (at least, this works for me on Ubuntu 18.04, even though the default `apt install qt-...` gives me 5.9).

image

larsgw:
I'm still getting

```
/home/larsgw/Projects/jasp/jasp-desktop/JASP-Desktop/data/columnsmodel.h:4:10: fatal error: QTransposeProxyModel: File or directory does not exist
 #include <QTransposeProxyModel>
          ^~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

when running

    QTDIR=/path/to/Qt/5.13.1 make -j8

boutinb:
Thank you very much for this! You are really welcome to help us!

PR #3581, opened 2019-10-08
Fisher fix

FBartos:
Fixes https://github.com/jasp-stats/jasp-issues/issues/500

- adds a menu for the alternative hypothesis selection in the same style as in Bayesian Contingency Tables (BCT)
image
- adds p-values for Fisher's exact test
- adds a note to the resulting table with the direction of the alternative hypothesis (as in BCT)
 
I would consider renaming the checkbox from "Logs odds ratio (2x2 only)" to something like "Logs odds ratio & Fisher's exact test (2x2 only)" because it might be hard to find how to do the Fisher's exact test.

### DIFF
--- a/JASP-Engine/JASP/R/contingencytables.R
+++ b/JASP-Engine/JASP/R/contingencytables.R
@@ -578,20 +578,6 @@ ContingencyTables <- function(dataset=NULL, options, perform="run", callback=fun
 	ordinal.footnotes <- .newFootnotes()
 	kendalls.footnotes <- .newFootnotes()
 
-	# Fisher test table footnote
-	if(length(group.matrices) >= 1 & options$oddsRatio & perform == "run" & options$oddsRatioHypothesis != "two.sided"){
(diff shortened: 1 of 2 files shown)


TimKDJ:
Contingency tables is currently undergoing a rewrite (see https://github.com/jasp-stats/jasp-desktop/pull/3442). I'm afraid the work you did here can't be merged, as the rewrite will change the entire structure of the analysis.

Considering the sweeping changes currently being made, I would recommend checking on [this list](https://github.com/jasp-stats/INTERNAL-jasp/projects/10) if the analysis is "safe" to alter (i.e., not being rewritten) or by asking me

TimKDJ:
Btw, once 3442 is merged, you can attempt this again, but it might actually be better to start from scratch rather than trying to rebase this branch

FBartos:
Oh sorry, this possibility didn't occur to me. I will do it once again the 3442 is done.

PR #3584, opened 2019-10-09
CFA bugfix: inadvertent equality constraints in large models

vankesteren:
With large cfa models, equality constraints were inadvertently added to the model. This is now fixed.
(λ_11_1 without underscores is the same as λ_1_11)


TimKDJ:
Don't need to do it in this PR but can you disable the whitespace trimming?

PR #3588, opened 2019-10-10
updates to JASPgraphs

vandenman:
Major updates to JASPgraphs:

- `cowplot` is no longer a dependency since we use `gridExtra` for combining ggplots.
- `R6` is a new dependency although it was already imported by `scales`.
- unit tests now work for combined plots, such as the prior-posterior plots and the matrix plots.
- the axes lines are created when the plot is drawn (necessary for plot editing).
- the y-axis scaling for sequential and robustness plots is allocating more space for the arrows that indicate evidence for H0/ H1.
- some updates to documentation and removal of deprecated/ commented out code.

All test files for the plots were remade (hence the many changes files).

EDIT:

- Also modifies the theme slightly so that certain components are transparent. Previously, the prior posterior plot had some white hollow rectangles if you selected a new analysis but a plot from a previous analysis was still visible.

Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/456


JASP-Tests/R/tests/testthat/helper-plots.R
```diff
@@ -6,8 +6,25 @@ expect_equal_plots <- function(test, name, dir) {
...
+  skip_if_grob(test)
+
+  if (inherits(test, "JASPgraphsPlot")) {
+    subplots <- test$subplots
+    nms <- names(subplots)
+    if (is.null(nms))
+      nms <- seq_along(subplots)
+
+    for (i in seq_along(subplots)) {
+      skip_if_grob(subplots[[i]]) # This means if one out of 20 plots is a grob, we skip the test!
```
TimKDJ: So this is technically not necessary anymore, right?

JASP-Engine/JASPgraphs/R/JASPgraphsPlot.R
```diff
@@ -0,0 +1,42 @@
+#' @importFrom ggplot2 is.ggplot
+
+JASPgraphsPlot <- R6::R6Class(
+  classname = "JASPgraphsPlot",
```
TimKDJ, line 4: Nice, R6 lends itself well for this.

JASP-Engine/JASPgraphs/R/JASPgraphsPlot.R
```diff
@@ -0,0 +1,42 @@
...
+  public = list(
+    initialize = function(subplots, plotFunction = reDrawJASPgraphsPlot, ...) {
+
+      if (!all(vapply(subplots, is.ggplot, TRUE)))
+        stop("all subplots should be of class ggplot!")
+      if (!is.function(plotFunction))
+        stop("plotFunction should be a function!")
+      plotArgs <- list(...)
+      if (!length(names(plotArgs)) == length(plotArgs))
+        stop("... should be a named list")
```
TimKDJ: `stop("all your arguments must be named")`? It's not a list they're providing, right?

JASP-Engine/JASPgraphs/R/JASPgraphsPlot.R
```diff
@@ -0,0 +1,42 @@
...
+        plotArgs[["names"]] <- paste0("plot", seq_along(subplots))
+
+      self$subplots     <- subplots
+      self$plotFunction <- plotFunction
+      self$plotArgs     <- plotArgs
+    },
+    plot = function(...) self$plotFunction(self$subplots, self$plotArgs, ...),
+    plotFunction = NULL,
+    plotArgs     = NULL,
+    subplots     = NULL
```
TimKDJ, line 25: Are we certain `subplots` will never change its structure? Because if it might, you could add a getter to be able to keep the API consistent in the face of change.

JASP-Engine/JASPgraphs/R/ggMatrixPlot.R
```diff
@@ -453,21 +486,15 @@ ggMatrixPlot.default <- function(plotList = NULL, nr = NULL, nc = NULL,
...
-  # gridExtra::grid.arrange(totalGraph)
-  # totalGraph <- gridExtra::arrangeGrob(grobs = t(plotList),
-  # 																		 nrow = d[1L], ncol = d[2L],
-  # 																		 widths = width, heights = height)
-  dev.off()
-  if (file.exists(f))
-    try(file.remove(f))
-  class(totalGraph) <- c("JASPgraphs", class(totalGraph), "ggMatrixplot")
+  
+  totalGraph <- JASPgraphsPlot$new(
```
TimKDJ, line 490: Just for my understanding, if people want to use this on a single ggplot rather than multiple, will it fail?
If so, something like `JASPgraphsCombinedPlot` might be a more apt name

vandenman: Single ggplots will also work! People just have to provide a custom function to the argument `plotFunction`. For example, you could pass a single ggplot and then do `plotFunction = print`.

JASP-Engine/JASPgraphs/R/JASPgraphsPlot.R
vandenman: Well, I thought about it and there's three reasons why I did it like this (for now):

1. I could not come up with a single situation where you'd want to have `subplots` behave differently. 
2. We can still modify the behavior and keep the same API by making `subplots` and active function.
3. JASPgraphsPlot is not exported at the moment. So only changes within JASPgraphs and one change `helper-plots.R` would be needed.

JASP-Engine/JASPgraphs/R/JASPgraphsPlot.R
vandenman: good point, I'll change this.

JASP-Tests/R/tests/testthat/helper-plots.R
vandenman: Technically not, I'll remove it.

PR #3589, opened 2019-10-11
Error message for identical variables + Scatterplots + Bugfix plotEditing

vandenman:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/290 and adds a unit test for this case.

Error message:

image


JASP-Engine/JASP/R/descriptives.R
```diff
@@ -1317,4 +1327,57 @@ Descriptives <- function(jaspResults, dataset, options) {
...
+  variablesB64 <- .v(variables)
+  # remove non-numeric variables
+  numerics <- sapply(dataset[variablesB64], is.double)
+  variables    <- variables[numerics]
+  variablesB64 <- variablesB64[numerics]
+
+  nvar <- length(variables)
+  if (nvar < 2L) {
+    msg <- if (length(numerics) > 1L) { # basically all user variables have the wrong type...
+      "These plots can only be shown for scale variables."
```
TimKDJ: Kinda split on this. On the one hand, it provides useful information to the user. On the other hand, we do not do this anywhere else in JASP. Do you reckon it will be too confusing to show an empty plot until the plot is ready to be computed?

JASP-Engine/JASP/R/descriptives.R
```diff
@@ -1317,4 +1327,57 @@ Descriptives <- function(jaspResults, dataset, options) {
...
+    return()
+  }
+
+  for (i in 1:(nvar - 1L)) for (j in (i + 1L):nvar) {
+    v1 <- variables[i]
+    v2 <- variables[j]
+    name <- paste(v1, "-", v2)
+    if (is.null(jaspContainer[[name]])) {
+      scatterPlot <- createJaspPlot(title = name)
+      scatterPlot$dependOn(optionContainsValue = list(variables = c(v1, v2)))
```
TimKDJ, line 1369: I don't think this works right now, but we should definitely make it work...

JASP-Engine/JASP/R/descriptives.R
```diff
@@ -1317,4 +1327,57 @@ Descriptives <- function(jaspResults, dataset, options) {
 
   return(plotObj)
 }
+
+.descriptivesScatterPlots <- function(jaspContainer, dataset, variables, split, options) {
+
+  if (!is.null(split) && split != "")
+    group <- dataset[, .v(split)]
```
TimKDJ, line 1337: We should add a legend if there is a split, so it's clear what each color corresponds to

JASP-Engine/JASPgraphs/R/JASPScatterPlot.R
```diff
@@ -0,0 +1,129 @@
...
+  if (is.null(group)) {
+    df <- data.frame(x = x, y = y)
+    mapping <- aes(x, y)
+  } else {
+    group <- factor(group)
+    df <- data.frame(x = x, y = y, g = group)
+    mapping <- aes(x = x, y = y, group = g, color = g, fill = g)
+  }
+
+  geomSmooth <- if (addSmooth) geom_smooth(formula = y ~ x, method = "loess", se = addSmoothCI) else NULL
```
TimKDJ: You reckon we should add a straight line option, too?

TimKDJ: I can also totally hear people demanding a different CI level...

TimKDJ (changes requested):
Ha, sneaky scatterplot in there as well. It looks very nice!
The legend is a must; the others are comments

vandenman:
~~oh oops, the scatterplot was not meant to be in this PR! Until JAGS has been merged, the color scheme can't be adjusted so I wanted to wait with that for a bit. Seem like I accidentally pushed to the wrong branch..~~

~~I'll move those changes to a different branch and update the PR accordingly.~~

Eh, I'll just leave the scatterplots in here. I fixed almost everything, only the legend needs some tweaking but perhaps we should discuss that in person.

TimKDJ:
Ok.. so that solves it. Let me know when you're finished with edits (and don't forget to remove the print statements)

vandenman:
@TimKDJ so this PR got waaayyy out of hand...

The things it should do are:

- Check for identical variables in the ANCOVA
- add scatterplots to descriptives
- fix a bug in the plotediting functions and clean up some incorrectly changed files.

Let me know what you think!

JASP-Engine/JASP/R/descriptives.R
```diff
@@ -1317,4 +1329,69 @@ Descriptives <- function(jaspResults, dataset, options) {
...
+
+.descriptivesScatterPlots <- function(jaspContainer, dataset, variables, split, options) {
+
+  if (!is.null(split) && split != "") {
+    group <- dataset[, .v(split)]
+    legendTitle <- split
+    if (options[["showLegend"]]) {
+      oldPalette <- JASPgraphs::getGraphOption("palette")
+      on.exit(JASPgraphs::setGraphOption("palette", oldPalette))
+      JASPgraphs::setGraphOption("palette", options[["colorPalette"]])
```
TimKDJ: This seems kinda error prone. People will forget to do this.
Can't we pass this in as a parameter to graphing functions if we don't want to make persistent changes to JASPgraphs' state?

TimKDJ:
Ok so like we discussed:

1. remove the html text that tells you why it can't plot for the time being
2. if you add a nominal variable then it errors
```

.descriptivesPieChart(dataset = dataset.factors, options = options, variable = var)

.descriptivesPieChart_SubFunc(column = column[!is.na(column)], variable = variable, width = options$plotWidth, height = options$plotHeight, title = variable)

JASPgraphs::plotPieChart(tb[, 2], tb[, 1], legendName = variable)

JASPcolors(length(unique(group)))

stop('palette must be character!')
```

JASP-Engine/JASP/R/descriptives.R
TimKDJ: And here we can move it to common

TimKDJ (approved):

PR #3591, opened 2019-10-14
New model prior for Bayesian linear regression

vandenman:
Changes to Bayesian linear regression per @EJWagenmakers executive order:

- Uniform model prior is the new default.
- Added two new model priors:
  - "Wilson" based on Consonni et al., (2018) and Wilson et al., (2010)
  - "Castillo" based on Castillo et al., (2015).
- Both priors are special cases of the beta-binomial prior and their use is explained through a footnote on the main table. 
  - Wilson: `The Wilson model prior corresponds to a beta binomial prior with α = 1 and β = p * λ and corresponds to an approximate penalization equal to log(λ + 1) in log-odds scale for each additional covariate added to the model (Consonni et al., 2018; Wilson et al., 2010).` 
  - Castillo: `The Castillo model prior corresponds to a beta binomial prior with α = 1 and β = p^u and is suitable  for sparse regression when there are more covariates than observations (Castillo et al., 2015).`.
- Added unit tests, references, and documentationt about these model priors to the analysis.


AlexanderLyNL:
I don't like the uniform model prior as default. The beta-binomial is okay, except when p becomes relatively large compared to n. In the middle we would find the Wilson prior then, why not use that?

vandenman:
@AlexanderLyNL please talk to @EJWagenmakers about that.

AlexanderLyNL:
In general, we do this via full enumeration don't we? Can we add this as an option. At the moment, we have to do this, by setting BAS = 0. Furthermore, wouldn't it be nice to add the co-authors of the papers to options? WICSS and CSvdV.

vandenman:
> In general, we do this via full enumeration don't we? Can we add this as an option. At the moment, we have to do this, by setting BAS = 0. 

Full enumeration only works for p <= 20. What would you propose after that? Quit the analysis? The BAS method does full enumeration by default, if p > 20 the sampling without replacement procedure (BAS) is used. I also don't really see how this comment is related to the current PR, so perhaps it should be disucssed in a separate issue.

> Furthermore, wouldn't it be nice to add the co-authors of the papers to options? WICSS and CSvdV.

I'm not sure if you mean the internal name for the option, or the name shown to the users. Either way:

1. We credit the co-authors since they appear in the citations.
2. Nobody (no programmer, no user) aside from you will known what that abbreviation means. You could write their full names but I think that will just be too long. I'm also not really sure what the gain is, since the co-authors are credited already. Currently, the footnote underneath the main table should make clear that the name of the prior is a direct reference to a paper.

AlexanderLyNL (approved):
I did not look into the code directly, but talked to EJ about the naming conventions and he's happy about them.

TimKDJ:
What about the prior though? Are we rolling with a uniform? I thought Scott & Berger had a little to say about that in https://arxiv.org/abs/1011.2333

Resources/Regression/qml/RegressionLinearBayesian.qml
```diff
@@ -220,9 +220,26 @@ Form {
...
+					childrenArea.columnSpacing: 1
+					DoubleField { name: "wilsonParamLambda"; label: qsTr("λ"); defaultValue: 1; inclusive: "maxOnly"}
+				}
+				RadioButton
+				{
+					value: "Castillo"
+					label: qsTr("Castillo")
+					childrenOnSameRow: true
+					childrenArea.columnSpacing: 1
+					DoubleField { name: "castilloParamU"; label: qsTr("u"); defaultValue: 1; inclusive: "maxOnly"; min: 1}
```
TimKDJ: You should either specify `inclusive: "yes"` or set the minimum to be higher than 1

PR #3596, opened 2019-10-15
JaspResults rewrite ANOVA/ANCOVA/RM ANOVA

JohnnyDoorn:
Rewrites to jasp results.
Added df and option to pool/unpool standard error for contrasts
Some more tweaks and fixes.


TimKDJ:
Can you rebase? The plot editing merge broke the unit tests, but they're fixed now

PR #3606, opened 2019-10-18
Feature request: Fisher's Exact Test

### LINKED ISSUE #500: Feature request: Fisher's Exact Test
I would be very interested if JASP could provide Fisher's Exact Test as part of the Contingency Table statistics. Particularly for small samples, having an option to select Fisher's Exact Test for any number of rows and columns, along with selecting the option for a one-sided or a two-sided test, could be very useful.

I believe Fisher's Exact Test is available in R using fisher.test. I saw a previous post about this feature request going back to 2017 and just wondered if it might still be considered for implementation in a future release. Many thanks in advance!

John

FBartos:
Fixes https://github.com/jasp-stats/jasp-issues/issues/500
the second try of adding p-values and one-sided p-values for Fisher exact test

- adds a menu for the alternative hypothesis selection in the same style as in Bayesian Contingency Tables (BCT)
image
- adds p-values for Fisher's exact test
- adds a note to the resulting table with the direction of the alternative hypothesis (as in BCT) - I'm extracting the dimnames from the generated data matrix, couldn't come up with a more elegant way
 
I would consider renaming the checkbox from "Logs odds ratio (2x2 only)" to something like "Logs odds ratio & Fisher's exact test (2x2 only)" because it might be hard to find how to do the Fisher's exact test now.

questions:
1) is it possible to add a column to only one row? I tried to add p-values only for the row with the Fisher exact test, however, it's started filling it in the log odds ratio row (now I'm adding the column to both rows and filling it with "" in the logs odds column.

### DIFF
--- a/JASP-Engine/JASP/R/contingencytables.R
+++ b/JASP-Engine/JASP/R/contingencytables.R
@@ -235,7 +235,7 @@ ContingencyTables <- function(jaspResults, dataset, options, ...) {
 
     # Create table
     crossTabLogOdds <- createJaspTable(title = "Log Odds Ratio")
-    crossTabLogOdds$dependOn(c("oddsRatio", "oddsRatioConfidenceIntervalInterval"))
+    crossTabLogOdds$dependOn(c("oddsRatio", "oddsRatioConfidenceIntervalInterval", "oddsRatioHypothesis"))
     crossTabLogOdds$showSpecifiedColumnsOnly <- TRUE
     crossTabLogOdds$position <- 3
     
@@ -251,6 +251,9 @@ ContingencyTables <- function(jaspResults, dataset, options, ...) {
     analysis <- as.list(analysis)
     # Compute/get Group List
     groupList <- .crossTabComputeGroups(dataset, options, analysisContainer, analysis, ready)
+    
+    # Add note for one-sided tests
+    .crossTabLogOddsNote(crossTabLogOdds, groupList, options, ready)
     
     res <- try(.crossTabOddsRatioRows(analysisContainer, analysis$rows, groupList, options, ready))
     .crossTabSetErrorOrFill(res, crossTabLogOdds)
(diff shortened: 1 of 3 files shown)


TimKDJ:
Can you have a look at [this document](https://github.com/jasp-stats/jasp-desktop/blob/development/JASP-Tests/README.md) to make our automatic unit tests happy?

With respect to your question:
Yes, that's definitely possible. But it might be a little more tricky in this case, because there is column folding. If you come by this week we can have a look

FBartos:
Thanks for pointing that out. I updated the test for odds ratio table to reflect the added column.

I will stop by the office on Tuesday/Wednesday to find out how to do the column thing.
(And probably will change the code and the test to reflect the changes once more).

JASP-Engine/JASP/R/contingencytables.R
```diff
@@ -487,6 +490,9 @@ ContingencyTables <- function(jaspResults, dataset, options, ...) {
                                 overtitle = ci.label, type = "number", format = "dp:3")
   table$addColumnInfo(name = paste0("up[", fold, "]"),    title = "Upper", 
                                 overtitle = ci.label, type = "number", format = "dp:3")
+  table$addColumnInfo(name = paste0("p[", fold, "]"),     title = "p",
```
vandenman, line 493: can't this be `type = "pvalue"`?

vandenman (changes requested):
> "Logs odds ratio (2x2 only)" to something like "Logs odds ratio & Fisher's exact test (2x2 only)"

I agree with the name change. An alternative might be separate checkboxes for each test.

Aside from that, the only comment is that we use tabs in QML and not spaces. See [this link](https://doc.qt.io/qtcreator/creator-indenting-code.html) to change your default settings for Qt creator.

The rest looks good to me!

JASP-Engine/JASP/R/contingencytables.R
FBartos: yes, it was supposed to be type = "pvalue", thanks for catching it

FBartos:
> > "Logs odds ratio (2x2 only)" to something like "Logs odds ratio & Fisher's exact test (2x2 only)"
> 
> I agree with the name change. An alternative might be separate checkboxes for each test.

I tried a few more alternatives and I liked this one the most:
image


> Aside from that, the only comment is that we use tabs in QML and not spaces. See [this link](https://doc.qt.io/qtcreator/creator-indenting-code.html) to change your default settings for Qt creator.

Thanks, that explains how my tabs got converted into spaces.

vandenman (approved):

PR #3611, opened 2019-10-21
Bayesian contingency tables rewrite to jaspResults

gibbona1:
**contingencytablesbayesian.R**
Now uses jaspResults
I've removed the code for the main table and am just calling from `contingencytables.R`

**test-contingencytablesbayesian.R**
Changed the main table unit test to the one in `test-contingencytables.R`


JASP-Engine/JASP/R/contingencytablesbayesian.R
```diff
@@ -15,2522 +15,660 @@
...
-	if (perform == "run") {
-
-		list(results=results, status="complete", state=new.state, keep=keep)
-
-	} else {
-
-		list(results=results, status="inited", state=new.state, keep=keep)
-	}
+ContingencyTablesBayesian <- function(jaspResults, dataset = NULL, options, ...) {
+  ready <- length(options$rows > 0) && length(options$columns > 0)
```
TimKDJ:
```suggestion
  ready <- length(unlist(options$rows)) > 0 && length(unlist(options$columns)) > 0
```

JASP-Engine/JASP/R/contingencytablesbayesian.R
```diff
@@ -15,2522 +15,660 @@
...
-	}
-
-	list(plot=odds.ratio.plot, state=new.plot.state, keep=plot[["png"]], complete=complete)
+# Log Odds Plot
+.contTabBasLogOddsPlot <- function(jaspResults, options, analyses, ready) {
+  if(!options$plotPosteriorOddsRatio)
+    return()
+  oddsRatioPlotContainer <- createJaspContainer("Log Odds Ratio Plots")
+  dependList <- c("plotPosteriorOddsRatio", "hypothesis", "samplingModel",
+                  "plotPosteriorOddsRatioAdditionalInfo", "priorConcentration")
```
TimKDJ: Shouldn't there be a dependency on the counts and layers?

JASP-Engine/JASP/R/contingencytablesbayesian.R
```diff
@@ -15,2522 +15,660 @@
...
+        oddsRatioSubContainer <- createJaspContainer(plot.title)
+      }
+      analysis <- analyses[i,]
+      title    <- ""
+      
+      if(length(options$rows) > 0 || length(options$columns) > 0)
+        title <- paste0(analysis$rows, " - ", analysis$columns)
+      logOddsPlot <- createJaspPlot(title = title, width = 530, height = 400)
+      oddsRatioSubContainer[["plot"]] <- logOddsPlot
+      oddsRatioPlotContainer[[paste0("plots", i, "sub", g)]] <- oddsRatioSubContainer
```
TimKDJ, line 206: And shouldn't this then have a dependency on the rows/columns?
Otherwise you don't need the statement at line 168.

Something like
```
$dependOn(optionContainsValue(list(rows=analysis$rows, columns=analysis$columns))
```

TimKDJ (changes requested):
Great that you reused a lot of code from the frequentist contingency tables!

A couple of short code comments and I noticed the following while testing:
- [ ] Switching "Row order" from "Ascending" to "Descending" doesn’t change the row order
- [ ] If you add several rows, then the additional contingency tables should be under the existing ones (order should also be deterministic, I think I saw the plots jump around)
- [ ] I don't think the log odds ratio plots needs a legend as it only has one line
- [ ] Setting “Sample” to “hypergeometrics (2x2 only)” and “Alt. Hypothesis” to “Group one < Group two” crashes the analysis: `Error in .contTabBasComputeBF(options, grp.mat[[g]], ready): object 'prop.consistent' not found`
- [ ] If you add only a column variable and tick “Column order” -> “Descending” you get a crash: `Error in rev(lvls, decreasing = TRUE): unused argument (decreasing = TRUE)` (same with row variable and row order)

gibbona1:
> Great that you reused a lot of code from the frequentist contingency tables!
> 
> A couple of short code comments and I noticed the following while testing:
> 
> * [x]  Switching "Row order" from "Ascending" to "Descending" doesn’t change the row order
> * [x]  If you add several rows, then the additional contingency tables should be under the existing ones (order should also be deterministic, I think I saw the plots jump around)
> * [x]  I don't think the log odds ratio plots needs a legend as it only has one line
> * [x]  Setting “Sample” to “hypergeometrics (2x2 only)” and “Alt. Hypothesis” to “Group one < Group two” crashes the analysis: `Error in .contTabBasComputeBF(options, grp.mat[[g]], ready): object 'prop.consistent' not found`
> * [x]  If you add only a column variable and tick “Column order” -> “Descending” you get a crash: `Error in rev(lvls, decreasing = TRUE): unused argument (decreasing = TRUE)` (same with row variable and row order)

@vandenman does `JASPgraphs::PlotPriorAndPosterior` allow for the legend not to be added?

TimKDJ:
@vandenman is @gibbona1's request possible?

vandenman:
Sorry for the late reply. That is possible:
```r
library(JASPgraphs)
# ?PlotPriorAndPosterior # for more info

n <- 100
x <- seq(-3, 3, length.out = n)
dfLines <- data.frame(
  x = x,
  y = c(dnorm(x, 0, 1), dnorm(x, 1, .5)),
  g = rep(c("Prior", "Posterior"), each = n)
)

BF10 <- dnorm(0, 1, .5) / dnorm(0, 0, 1)
p <- PlotPriorAndPosterior(dfLines, BF = BF10, bfType = "BF10")
p$subplots[[3]] <- p$subplots[[3]] + ggplot2::theme(legend.position = "none")
p
```
image

gibbona1:
Perfect, thanks @vandenman, works great!

TimKDJ:
> - [ ]  If you add several rows, then the additional contingency tables should be under the existing ones (order should also be deterministic, I think I saw the plots jump around)

This one is not implemented yet; right now you get:
```
[tables first row/column pair]
[plots all row/column pairs]
[tables second row/column pair]
```
It should be like it was before
```
[tables all row/column pairs]
[plots all row/column pairs]
```

We should probably make it a little more clear altogether... but let's just return to the old situation for now

TimKDJ:
>     * [ ]  Setting “Sample” to “hypergeometrics (2x2 only)” and “Alt. Hypothesis” to “Group one < Group two” crashes the analysis: `Error in .contTabBasComputeBF(options, grp.mat[[g]], ready): object 'prop.consistent' not found`

This option still doesn't play nicely:
Screenshot 2019-11-20 at 11 50 34

TimKDJ:
I will make the last change

PR #3612, opened 2019-10-22
moved color palette and changed default

vandenman:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/535

The new default for color palette is `colorblind`.

The position of colorpalette dropdown is now in the top left above `Distribution plots`.


AlexanderLyNL (approved):
Wonderful, this looks very nice. I like how you added all the new features. I also like how you'd retained the meaty colour on the violin plot, but shouldn't that also change if we change the colour palette?

TimKDJ:
@vandenman see Lexi's comment. Should the violin change?

vandenman:
I guess it should.

TimKDJ:
Alright, then I'll wait with merging

vandenman:
color palette now also works for the box plot. I've made some small changes to the box plot as well (it now uses `themeJaspRaw`) but if the unit tests pass then I think this should be good to merge.

image

TimKDJ (changes requested):
The color schemes work nicely for all but `ggplot2` -- if you choose that one the piecharts don't play ball:

[colorsPiecharts.jasp.zip](https://github.com/jasp-stats/jasp-desktop/files/3808169/colorsPiecharts.jasp.zip)

TimKDJ (approved):

PR #3619, opened 2019-10-24
Jasp results summary stats

ASarafoglou:
Rewrote Analyses in Summary Stats Module to JASP results (except Bayesian Correlation Pairs).


Resources/Summary Statistics/qml/SummaryStatsBinomialTestBayesian.qml
```diff
@@ -24,7 +24,7 @@ import JASP.Widgets 1.0
 
 Form 
 {
-	usesJaspResults: false
+	usesJaspResults: true
```
TimKDJ: You can just omit `usesJaspResults: true` -- this is the default

JASP-Engine/JASP/R/summarystatsbinomialtestbayesian.R
```diff
@@ -15,348 +15,269 @@
...
-							keep = keep)
-				)
+SummaryStatsBinomialTestBayesian <- function(jaspResults, dataset = NULL, options, ...) {
+  
+  # Reading in a datafile is not necessary
+  # Check for possible errors
+  .checkErrors.summarystats.binomial(options)
+  
+  # Compute the results and create main results table
+  summaryStatsBinomialResults <- .summaryStatsBinomialMainFunction(jaspResults, options)
```
TimKDJ, line 25: The inconsistency in function names (`.checkErrors.summarystats.ttest` <-> `.summaryStatsTTestMainFunction`) makes me a little sad

JASP-Engine/JASP/R/summarystatsbinomialtestbayesian.R
```diff
@@ -15,348 +15,269 @@
...
-	return(returnPlot)
+# Execute Bayesian binomial test ----
+.summaryStatsBinomialMainFunction <- function(jaspResults, options) {
+  
+  # This function is the main workhorse, and also makes the table
+  if (is.null(jaspResults[["binomialContainer"]])) {
+    jaspResults[["binomialContainer"]] <- createJaspContainer()
+    jaspResults[["binomialContainer"]]$dependOn(c("successes", "failures", "betaPriorParamA", "betaPriorParamB", "testValue", "hypothesis"))
+  }
+
```
TimKDJ, line 43: Suggestion:

If you add
`container <- jaspResults[["binomialContainer"]]`
here, then you can use `container` instead of `jaspResults[["binomialContainer"]]` and your function becomes quite a bit more legible

JASP-Engine/JASP/R/summarystatsbinomialtestbayesian.R
```diff
@@ -15,348 +15,269 @@
...
+  bayesianBinomialTable$addCitation(.summaryStatsCitations[c("Jeffreys1961", "OHagan2004", "Haldane1932")])
+  
+  message <- hypothesisList$message
+  if (!is.null(message)) bayesianBinomialTable$addFootnote(message)
+  
+  bayesianBinomialTable$addColumnInfo(name = "successes", title = "Successes" , type = "integer")
+  bayesianBinomialTable$addColumnInfo(name = "failures" , title = "Failures"  , type = "integer")
+  bayesianBinomialTable$addColumnInfo(name = "theta0"   , title = "Test value", type = "number")
+  bayesianBinomialTable$addColumnInfo(name = "BF"       , title = bfTitle     , type = "number")
+  bayesianBinomialTable$addColumnInfo(name = "pValue"   , title = "p"         , type = "number")
```
TimKDJ:
```suggestion
  bayesianBinomialTable$addColumnInfo(name = "pValue"   , title = "p"         , type = "pvalue")
```

JASP-Engine/JASP/R/summarystatsregressionlinearbayesian.R
```diff
@@ -15,1294 +15,327 @@
...
+  
+  # check if number of covariates is correct in H1
+  if(options$numberOfCovariatesAlternative!=0 && options$sampleSize!=0 && ((options$sampleSize - options$numberOfCovariatesAlternative) < 2)) {
+    
+    .quitAnalysis("Number of Covariates must be less than N-1 (sample size minus 1)")
+    
+  }
+  
+  # check if number of covariates is correct in H0
+  if(options$numberOfCovariatesNull==0 && options$unadjustedRSquaredNull==0) {
```
TimKDJ: This if statement is in direct contrast with the one it encapsulates. I think it can be deleted altogether?

JASP-Engine/JASP/R/commonsummarystatsttestbayesian.R
```diff
@@ -157,1709 +77,559 @@
...
+  if(analysis == "independentSamples")
+    ttestTableData$n2 <- n2
+  ttestTableData$BF     <- BFlist[[options$bayesFactorType]]
+  ttestTableData$error  <- ttestResults$properror
+  ttestTableData$pValue <- ttestResults$pValue[[hypothesis]]
+  
+  # check whether % error could be computed
+  if(is.na(ttestTableData$error) || is.null(ttestTableData$error)){
+    ttestTableData$error <- NaN
+    ttestTableMessage    <- "Proportional error estimate could not be computed."
```
TimKDJ, line 332: This should be a column (or cell, whichever, it only has 1 cell anyway) specific footnote, rather than a general table footnote.

TimKDJ (changes requested):
I think you did a good job :)
Do have a few comments though:

**Binomial:**
- [ ] The additional info of the Prior posterior plot no longer takes into account what the "Bayes Factor"/"Alt. Hypothesis" is. It should change if you go from `BF10` to `BF01`, and also if you go from `> Test value` to `< Test value`
 
---

**Regression:**
- [ ] BF results are way different, but error term is the same:
Screenshot 2019-10-31 at 11 06 13

- [ ] User prior (gray dot) is not on the line:
Screenshot 2019-10-31 at 11 07 38

---

**ttest independent samples:**
- [ ] The Bayes factor robustness check plot does not change when you change the type of the "Bayes Factor"

- [ ] Results seem incorrect for informed priors (they appear to be the correct for the other ttests)
Screenshot 2019-10-31 at 11 59 15

TimKDJ:
Btw, please fix the code comments in all files (if I made a comment about something in file A I did not make the same comment again in file B).

TimKDJ:
@ASarafoglou how is the updated code coming along?

ASarafoglou:
I post my progress by updating this comment:

 **Binomial:**
 
 * [x]  The additional info of the Prior posterior plot no longer takes into account what the "Bayes Factor"/"Alt. Hypothesis" is. It should change if you go from `BF10` to `BF01`, and also if you go from `> Test value` to `< Test value`
 
**Regression:** 
* [x]  BF results are way different, but error term is the same:
* [x]  User prior (gray dot) is not on the line.
        Corrected. Also the robustness plot additional info changes now with the BF type.

 **ttest independent samples:**
 
* [x]  The Bayes factor robustness check plot does not change when you change the type of the "Bayes Factor"
* [x]  Results seem incorrect for informed priors (they appear to be the correct for the other ttests)

Resources/Summary Statistics/qml/SummaryStatsBinomialTestBayesian.qml
ASarafoglou: I deleted this line.

JASP-Engine/JASP/R/summarystatsbinomialtestbayesian.R
ASarafoglou: Nobody wants to see you being sad :(
I fixed this for .checkErrors.summarystats.ttest() and .hypothesisType.summarystats.binomial()

(will do this for the other summary stats analyses as well)

JASP-Engine/JASP/R/summarystatsbinomialtestbayesian.R
ASarafoglou: Corrected.

JASP-Engine/JASP/R/summarystatsregressionlinearbayesian.R
ASarafoglou: You are right. I corrected it.

JASP-Engine/JASP/R/commonsummarystatsttestbayesian.R
ASarafoglou: Fixed.

ASarafoglou:
@TimKDJ I think I am done with the rewrite..??

giphy

TimKDJ:
SumStats correlation expects the '.getBayesfactorTitle.summarystats.ttest' function (see [travis](https://travis-ci.org/jasp-stats/jasp-desktop/builds/611939073?utm_source=github_status&utm_medium=notification)), could you add this to that R file? At least until Lexi's rewrite is finished

JASP-Tests/R/tests/testthat/test-summarystatscorrelationbayesianpairs.R
```diff
@@ -1,29 +1,16 @@
-context("SummaryStatsCorrelationBayesianPairs")
+context("Summary Statistics Bayesian Correlation Pairs")
```
TimKDJ: Can you undo the changes you made to this file?
This is what fails the unit tests right now and considering you didn't rewrite correlation these changes seem unnecessary.

TimKDJ (changes requested):
The changes you made look good (mostly)! Did notice the following though:
Binomial:
- [ ] Ticking “Prior and posterior” crashes the analysis: Error in .summaryStatsBinomialMainFunction(jaspResults, options): object 'container' not found

Regression:
- [ ] If you create a bayes factor robustness check plot, then it says “Cauchy prior width” on the x-axis rather than “r scale”
- [ ] The user prior in the additional info portion of the plot appears to be incorrect (related to the gray dot not being on the line, which is also the case here).

TTest independent samples:
- [ ] Same as binomial: Ticking “Prior and posterior” crashes the analysis: Error in .summaryStatsBinomialMainFunction(jaspResults, options): object 'container' not found

All of these points can be found in [rewriteSumStats.jasp.zip](https://github.com/jasp-stats/jasp-desktop/files/3869016/rewriteSumStats.jasp.zip)

JASP-Tests/R/tests/testthat/test-summarystatscorrelationbayesianpairs.R
ASarafoglou: corrected.

ASarafoglou:
Binomial:

* [x]  Ticking “Prior and posterior” crashes the analysis: Error in .summaryStatsBinomialMainFunction(jaspResults, options): object 'container' not found
 
Regression:

* [x]  If you create a bayes factor robustness check plot, then it says “Cauchy prior width” on the x-axis rather than “r scale”
* [x]  The max BF in the additional info portion of the plot appears to be incorrect (related to the gray dot not being on the line, which is also the case here).

@TimKDJ I don't see that the user prior is incorrect. Can you give an example where this occurs? I just saw that the max BF01 was actually not the correct number, so I fixed that instead (for both the t-tests and the linear regression)

TTest independent samples:
 
* [x]  Same as binomial: Ticking “Prior and posterior” crashes the analysis: Error in .summaryStatsBinomialMainFunction(jaspResults, options): object 'container' not found

TimKDJ (changes requested):
Sorry for my horribly late response!
Just small things now ;)

Independent Samples TTest
- [ ] Robustness plot should have no dots on the line if there is no legend (like it used to be previously)

- [ ] The piechart on top of the prior posterior plot does not look ok (seems to be caused by the next comment):
Screenshot 2019-12-10 at 09 48 48

- [ ] The bayes factor in the prior posterior plot is whatever type of Bayes Factor is selected (e.g., BF10 or Log(BF10)), but this is 1) not reflected in the label — it should say Log(BF) if that’s what we’re showing; and 2) the value does not change when you change the Bayes Factor type after you create the plot. I think this should never be the log value though?

---
Paired Samples TTest
- [ ] The label of the robustness plot says log(BF0+) but I don’t think that’s what it’s actually showing:
old <-> new
Screenshot 2019-12-10 at 10 00 18

---
Binomial
- [ ] Space between “Population proportion” and “theta”
Screenshot 2019-12-10 at 10 34 06

ASarafoglou:
Independent Samples TTest:

- [x] Robustness plot should have no dots on the line if there is no legend (like it used to be previously)

- [x] The piechart on top of the prior posterior plot does not look ok (seems to be caused by the next comment)

- [x] The Bayes factor in the prior posterior plot is whatever type of Bayes Factor is selected (e.g., BF10 or Log(BF10)), but this is 1) not reflected in the label — it should say Log(BF) if that’s what we’re showing; and 2) the value does not change when you change the Bayes Factor type after you create the plot. I think this should never be the log value though?

The Label is not adapting and I don't allow log Bayes factors in the prior posterior plot.

Paired Samples TTest:
- [x] The label of the robustness plot says log(BF0+) but I don’t think that’s what it’s actually showing:
old <-> new

Binomial:

- [x] Space between “Population proportion” and “theta”

TimKDJ (approved):

TimKDJ:
Here comes the.. um.. merge shark!
giphy

ASarafoglou:
Hahaha amazing! xD

PR #3620, opened 2019-10-25
Rewrite linear regression with jaspResults

TimKDJ:
Additionally,

- Made the interface more consistent with logistic regression (renamed and moved "include constant in equation", moved "vovk sellke" and renamed the section "Options" to "Method Specification")
- Changed the forward regression method to start from a null model and then add predictors
- Changed the enter method to name models "H0" and "H1" (like logistic regression) and to always show the intercept model (this was inconsistent previously)
- Changed the numbering of models for stepping procedures (going from 1 to n, rather than sometimes 0 to n or 1 to n)
- Removed pretty much all `try()` statements as they weren't caught anyway :|
- Added more footnotes that state for which table the null model isn't included
- Fixed a bug where the null model is duplicated in backward regression if it contained a nuisance term
- Related to ^ fixed a bug where not all predictors were shown in the Collinearity Diagnostics table
- Fixed a bug where all plots would fail if you added an interaction term -> tick "Partial plots" -> tick other plots

There are 3 `TODO: figure out` statements in the code (just ctrl+f) for scenarios that I was not sure are correctly handled in the old code; it would be great if you could give me input on those @vandenman, @Kucharssim

### DIFF
--- a/JASP-Desktop/components/JASP/Controls/CheckBox.qml
+++ b/JASP-Desktop/components/JASP/Controls/CheckBox.qml
@@ -47,6 +47,9 @@ JASPControl
 			property alias	alignChildrenTopLeft:	childControlsArea.alignChildrenTopLeft
 
 	signal clicked();
+	function click() { control.toggle(); }
+	function toggle() { control.toggle(); }
+
 
 	CheckBox
 	{
--- a/JASP-Desktop/components/JASP/Controls/RadioButton.qml
+++ b/JASP-Desktop/components/JASP/Controls/RadioButton.qml
@@ -58,6 +58,8 @@ JASPControl
 		}
 	}
 
+	function toggle() { control.toggle() }
+
 	RadioButton
 	{
 		id:					control
--- a/JASP-Engine/JASP/R/binomialtest.R
+++ b/JASP-Engine/JASP/R/binomialtest.R
@@ -56,24 +56,6 @@ BinomialTest <- function(jaspResults, dataset = NULL, options, ...) {
     exitAnalysisIfErrors = TRUE
   )
 
-  # Error check 2: 0 observations for a level of a variable
-  for (variable in options$variables) {
-
-    column <- dataset[[.v(variable)]]
-    data   <- column[!is.na(column)]
-    levels <- levels(data)
-
-    for (level in levels) {
-      .hasErrors(
-        dataset              = data[data == level],
-        perform              = "run",
-        type                 = "observations",
-        observations.amount  = "< 1",
-        exitAnalysisIfErrors = TRUE
-      )
-    }
-  }
-
 }
 
 # Results functions ----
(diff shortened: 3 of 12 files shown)


vandenman:
I couldn't find any bugs! Looks good to me. Just to reply to the TODOs:

> TODO: figure out what the rule is here, if the same is added and then removed, what should the algorithm do? Stop altogether, or ignore it in the next iteration ignoring in the next iteration seems wrong because you might enter an infinite loop (Add A, Remove A | (Ignore A), Add B, Remove B | (Ignore B), Add A, remove A) stop if the same predictor was added and then removed

I think the algorithm should just stop.

> TODO: figure out if we can show partial plots for non-interaction-term-predictors if any of the other predictors have an interaction term

I have no idea, that's a substantial question.

> TODO: maybe add footnote to casewise diagnostics if there are no cases to show?

Good idea!

> TODO: figure out if this is correct (seems wrong).. the last model fit does not necessarily contain all predictors

Well, you could sort the model by number of predictors so that this assumption always holds. However, the current comment/ code inside that function provides me with too little information to really judge what's going on.

vandenman (approved):

TimKDJ:
> Well, you could sort the model by number of predictors so that this assumption always holds. However, the current comment/ code inside that function provides me with too little information to really judge what's going on.

Suppose we use backwards stepping to obtain our models. It starts with all predictors, but as they're all a poor fit, we drop them all, and the final model only contains the intercept. What used to happen is that it still showed "Residuals vs. Covariates" plots for all predictors, but it used the residuals from the intercept only model.

This seems wrong; it should use the residuals from the full model. That's easy enough to fix of course. But I guess my main question is, should it even make these plots, if the model of inference (the final model) does not have any predictors? Because all other plots are specifically related to whatever the "final" model is.

Hope that clarifies it a bit

vandenman:
That definitely clarifies! In that case, I'd argue that there shouldn't be a plot. It's pretty misleading if only that plot doesn't correspond to the final model.

Kucharssim (changes requested):
Ok, the code is very good! I found only one issue in the bootstrap, where the confidence interval for the first term gets copied for all terms in the model. Otherwise the analysis seems fine.

Screen Shot 2019-11-01 at 12 17 48

(Also from the User friendliness perspective, maybe it could be worthwhile to have a state for the bootstrap results so that one does not have to resample when enabling/disabling confidence intervals, etc... This is not necessary but would be nice I guess)

Kucharssim:
> > TODO: figure out if we can show partial plots for non-interaction-term-predictors if any of the other predictors have an interaction term
> 
> I have no idea, that's a substantial question.

Yeah I lean towards not to do it unless we know what are we doing :) Here are some pointers https://stats.stackexchange.com/q/265448. It seems that it is possible but making sure that it is easily interpretable could be tricky.

TimKDJ:
I fixed the CI of the bootstrap table and added a state to it. Also followed the suggestions you guys gave.

PR #3622, opened 2019-10-25
jaspResults rewrite - Correlation

Kucharssim:
On top of that, adds some functionality:

- reporting sample size: https://github.com/jasp-stats/jasp-issues/issues/528 and https://github.com/jasp-stats/jasp-issues/issues/245
- choice between list-wise/pair-wise exclusion: https://github.com/jasp-stats/jasp-issues/issues/244
- pairwise plots (although seeing Don's improvement of this in descriptives this was probably overkill :D): https://github.com/jasp-stats/jasp-issues/issues/298
- assumption checks: https://github.com/jasp-stats/jasp-issues/issues/488
- add correlation heatmap plot (cannot find the issue but I am pretty sure @EJWagenmakers asked me to do this like a year ago :D)

- prepared some structure for partial correlations but this is still only in debug mode since we need to do some research on non-parametric confidence intervals.


Resources/Regression/qml/Correlation.qml
```diff
@@ -23,13 +23,14 @@ import JASP.Theme 1.0
 
 Form
 {
-	usesJaspResults: false
+    usesJaspResults: true
```
TimKDJ: This line can go

Resources/Regression/qml/Correlation.qml
```diff
@@ -23,13 +23,14 @@ import JASP.Theme 1.0
...
 
 	VariablesForm
 	{
 		height: Theme.smallDefaultVariablesFormHeight
 		AvailableVariablesList { name: "allVariablesList" }
 		AssignedVariablesList { name: "variables"; title: qsTr("Variables"); suggestedColumns: ["ordinal", "scale"] }
+        AssignedVariablesList { name: "conditioningVariables"; title: qsTr("Condition on"); suggestedColumns: ["ordinal", "scale"]; debug: true }
```
TimKDJ: Qml uses tabs :)

JASP-Engine/JASP/R/correlation.R
```diff
@@ -14,644 +14,961 @@
...
-        for (variable.name in variables) {
-          fields[[length(fields)+1]] <- list(name=paste0(variable.name, "[", test, "-upperCI]"), title=variable.name, type="number", format="dp:3")
-          fields[[length(fields)+1]] <- list(name=paste0(variable.name, "[", test, "-lowerCI]"), title=variable.name, type="number", format="dp:3")
+.corrAssumptions <- function(jaspResults, dataset, options, ready, corrResults){
+  # uses .multivariateShapiroComputation from manova.R
+  if(isFALSE(options$multivariateShapiro) && isFALSE(options$pairwiseShapiro)) return()
+  if(!is.null(jaspResults[['assumptionsContainer']])) return()
+  
+  assumptionsContainer <- createJaspContainer(title = "Assumption checks")
+  assumptionsContainer$dependOn(c("multivariateShapiro", "pairwiseShapiro",
```
TimKDJ: If you set `"multivariateShapiro", "pairwiseShapiro"` on their respective tables instead of the container, then you don't recompute multivariate when you toggle pairwise and vice versa. You'd have to check for the existence of the tables tho

JASP-Engine/JASP/R/correlation.R
```diff
@@ -14,644 +14,961 @@
...
+        }
+        
+        for(i in seq_along(shapiroResult)){
+          if(!is.null(shapiroResult[[i]]$errors)){
+            shapiroTable$setError(shapiroResult[[i]]$errors)
+            break
+          }
+          shapiroTable$addRows(list(vars = names(shapiroResult)[i],
+                                    W    = shapiroResult[[i]]$result$statistic,
+                                    p    = shapiroResult[[i]]$result$p.value))
```
TimKDJ: This would go nicely into a function of its own (same for pairwiseShapiro)

JASP-Engine/JASP/R/correlation.R
```diff
@@ -14,644 +14,961 @@
...
+    plot$dependOn(c("variables", "conditioningVariables", "missingValues", "pearson", "spearman", "kendallsTauB", 
+                    "flagSignificant"))
+    plot$position <- 4
+    jaspResults[['heatmaps']] <- plot
+    
+    if(ready) plot$plotObject <- .corrPlotHeatmap(tests, options, corrResults)
+  } else{
+    heatmaps <- createJaspContainer(title = "Heatmaps")
+    heatmaps$dependOn(c("variables", "conditioningVariables", "missingValues", "pearson", "spearman", "kendallsTauB",
+                        "flagSignificant"))
```
TimKDJ, line 944: Why does this have a dependency on "flagSignificant"?

TimKDJ (changes requested):
Well done Simon! Cool new features too

Some minor code comments and I noticed the following during testing:
- [ ] Vovksellke has the same symbol as the significance indicator which is rather confusing
- [ ] We no longer have the footnote “all tests one-tailed, for positive correlation. Also the "* p < .05, etc" no longer has the “one-tailed” addition at the end. Not sure if we care about either, just want to make sure they're not forgotten
- [ ] You get additional stars under the table (in the screenshot 3, but depends on the hypothesis):
Screenshot 2019-10-31 at 14 23 51

JASP-Engine/JASP/R/correlation.R
Kucharssim: because if flagSignificant == TRUE, the heatmap highlights significant correlations...

JASP-Engine/JASP/R/correlation.R
Kucharssim: Yeah, thanks for the suggestion!

JASP-Engine/JASP/R/correlation.R
Kucharssim: I moved these parts into their own functions

Kucharssim:
Ooops, I should have squashed the commits I guess... Does it hurt this way?

Ok, I addressed all points raised, except this one:
> * [ ]  You get additional stars under the table (in the screenshot 3, but depends on the hypothesis):

Can I actually add a footnote without printing anything under the table? That is, I use addFootnote to add symbols *, **, and *** inside the table to flag the significant correlations, but then the symbol is also added below. I did not figure out how to do that; if it's not possible, should I make an issue for @JorisGoosen?

JorisGoosen:
@Kucharssim we can squash it here ;)

As far as the footnotes without footnotes go: I don't think this is supported right now. But if you make an issue for it we can add something like: "a footnote without a message is not listed below the table".

The point about the stars underneath the table is then solved right? 
And then we can squash+merge it.

JASP-Engine/JASP/R/correlation.R
TimKDJ: Ah okay, missed that

JASP-Engine/JASP/R/correlation.R
```diff
@@ -107,7 +107,17 @@ Correlation <- function(jaspResults, dataset, options){
     .corrInitCorrelationTable(mainTable, options, variables)
   }
   
-  if(options[['flagSignificant']]) mainTable$addFootnote(message = "p < .05, ** p < .01, *** p < .001", symbol = "*")
+  if(options[['hypothesis']] == "correlatedPositively"){
+    mainTable$addFootnote(message = "All tests one-tailed, for positive correlation")
+    additionToFlagSignificant <- ", one-talied"
+  } else if(options[['hypothesis']] == "correlatedNegatively"){
+    mainTable$addFootnote(message = "All tests one-tailed, for negative correlation")
+    additionToFlagSignificant <- ", one-talied"
```
TimKDJ: 2x one-tailed instead of one-talied

TimKDJ:
Just a typo fix and we're all good

JASP-Engine/JASP/R/correlation.R
Kucharssim: oh christ :D

TimKDJ (approved):

PR #3623, opened 2019-10-28
Meta analysis rewrite to jaspResults

gibbona1:
**classicalmetaanalysis.R**
Rewrite was fairly extensive, especially since all the plots had to be done in ggplot

**test-classicalmetaanalysis.R**
Added tests for all tables and plots, but this had to be created from the rewritten file


JASP-Engine/JASP/R/classicalmetaanalysis.R
```diff
@@ -1,774 +1,773 @@
...
-  funnelPlot <- state$funnelPlot
-  diagnosticsPlot <- state$diagnosticsPlot
-  profilePlot <- state$profilePlot
-  trimfillPlot <- state$trimfillPlot
-
-  ### This file interfaces the metafor::rma function and associated diagnostic functions
-  ### (but it restricts its multipurposeness to one central purpose: fitting a
-  ### meta-regression to effect sizes and their standard errors and providing diagnostics.)
-
-  ### options contains:
```
raoelg, line 63: Why remove this options description section?

JASP-Engine/JASP/R/classicalmetaanalysis.R
```diff
@@ -1,774 +1,773 @@
...
-    trimfillPlot = c(fitOpts, "trimfillPlot")
-  )
+ClassicalMetaAnalysis <- function(jaspResults, dataset = NULL, options, ...) {
+  
+  ready <- options$dependent != "" && options$wlsWeights != ""
+  if(ready) {
+    dataset <- .metaAnalysisReadData(dataset, options)
+    .metaAnalysisCheckErrors(dataset, options)
+  }
+  # Output tables
```
raoelg, line 24: So everything is now move to side effect functions?

raoelg (approved):

TimKDJ:
@vandenman could you take a gander at the ggplot implemetation of the base plots?

Resources/Meta Analysis/qml/ClassicalMetaAnalysis.qml
```diff
@@ -88,7 +86,7 @@ Form
 		Group
 		{
 			title: qsTr("Plots")
-			CheckBox { name: "trimfillPlot";			text: qsTr("Trim-fill analysis")	}
+            CheckBox { name: "trimFillPlot";			text: qsTr("Trim-fill analysis")	}
```
vandenman, line 95: we use tabs in QML and not spaces. See [this link](https://doc.qt.io/qtcreator/creator-indenting-code.html) to change your default settings for Qt creator.

JASP-Engine/JASP/R/classicalmetaanalysis.R
```diff
@@ -1,774 +1,1308 @@
...
-    egger = c(fitOpts, "funnelPlotAsymmetryTest"),
-    influ = c(fitOpts, "residualsCasewiseDiagnostics"),
-    fsn.fit = c(fitOpts, "plotResidualsCovariates"),
-    forestPlot = c(fitOpts, "forestPlot"),
-    funnelPlot = c(fitOpts, "funnelPlot"),
-    diagnosticsPlot = c(fitOpts, "plotResidualsDependent", "plotResidualsQQ"),
-    profilePlot = c(fitOpts, "plotResidualsPredicted"),
-    trimfillPlot = c(fitOpts, "trimfillPlot")
-  )
+ClassicalMetaAnalysis <- function(jaspResults, dataset = NULL, options, ...) {
```
vandenman, line 18: Why is the copyright gone?

JASP-Engine/JASP/R/classicalmetaanalysis.R
```diff
@@ -1,774 +1,1308 @@
...
+                          labs    = slabs,
+                          ES      = rma.fit$yi,
+                          ci.int  = ci.int,
+                          ci.lb   = ci.lb,
+                          ci.ub   = ci.ub,
+                          shape   = rep(15, k),
+                          size    = psize)
+  
+  if(!rma.fit$int.only){
+    #TODO: these need to be polygons/segments as in forest.rma
```
vandenman: Is this a leftover TODO or does it still need to be done?

vandenman (changes requested):
Overall I think the code looks fine but there are some bugs (see below). These are probably not all bugs in the current PR, I just stopped after a while. Let me know once they are fixed so I can take another look!


List of bugs:
- [ ] Changing `Method` doesn't update the output. Is it added to the depends?
- [ ] Add two covariates. Under Model Terms, add their interaction. In the table "Coefficients", the last row contains one of the names of the interaction but is empty otherwise.
image

- [ ] Perhaps the error message of `rma.fit <- tryCatch(...)` should be prefixed with "The metafor package crashed with the following error". Mainly since it's error messages can be a little bit.. odd for people using JASP.
image
image

- [ ] Funnel plot can crash weirdly:
image

- [ ] Forest plot can crash weirdly:
image

- [ ] Casewise diagnostics cause missing value error:
image

- [ ] Diagnostics plot can crash weirdly. Probably related to the forst plot crash since the message is the same.
image

- [ ] Covariance matrix bug (perhaps caused by collinear variables):
image

- [ ] I can toggle `Confidence intervals` but it only does something when `Estimates` is on. Maybe it should be grayed out whenever `Estimates` is unchecked? (I'm aware this behavior was also in the previous release, but I'm not very fond of it).

- [ ] Regression test for funnel plot asymmetry is still base64 encoded:
image

TimKDJ:
How's this one coming along @gibbona1?

gibbona1:
@TimKDJ still having a problem with switching `b64()` and `d64()` to `.v()` and `.unv()` in 
`.metaAnalysisComputeModel()`.
Would you be able to have a look? I pushed the latest version

TimKDJ:
> @TimKDJ still having a problem with switching `b64()` and `d64()` to `.v()` and `.unv()` in
> `.metaAnalysisComputeModel()`.
> Would you be able to have a look? I pushed the latest version

Replace
```
  if (ready) {
    if(length(options$modelTerms) > 0) {
      .vmodelTerms <- NULL
      for (i in seq_along(options$modelTerms)) {
        components   <- options$modelTerms[[i]]$components
        .vmodelTerms <- c(.vmodelTerms, .v(components)) 
        #browser()
        #if (length(components) == 1)
        #  .vmodelTerms <- c(.vmodelTerms, .v(components[[1]]))
        #else {
        #  components.unlisted <- unlist(components)
        #  term.base64         <- paste0(.v(components.unlisted), collapse = ":")
        #  .vmodelTerms        <- c(.vmodelTerms, term.base64)
        #}
      }
      # infer model formula
      #.vmodelTerms <- .v(options$modelTerms)
      .vmodelTerms <- .vmodelTerms[.vmodelTerms != ""]
      #browser()
      formula.rhs <- as.formula(.vmodelTerms)
    } else
      formula.rhs <- NULL
```
with 
```
  if (ready) {
    if (length(options$modelTerms) > 0)
      formula.rhs <- formula(as.modelTerms(.v(options$modelTerms)))
    else
      formula.rhs <- NULL
```

and you should be good to go

gibbona1:
> * [X]  Changing `Method` doesn't update the output. Is it added to the depends?
> * [X]  Add two covariates. Under Model Terms, add their interaction. In the table "Coefficients", the last row contains one of the names of the interaction but is empty otherwise.
> * [X]  Perhaps the error message of `rma.fit <- tryCatch(...)` should be prefixed with "The metafor package crashed with the following error". Mainly since it's error messages can be a little bit.. odd for people using JASP.
> * [X]  Funnel plot can crash weirdly:
> * [X]  Forest plot can crash weirdly:
> * [X]  Casewise diagnostics cause missing value error:
> * [X]  Diagnostics plot can crash weirdly. Probably related to the forst plot crash since the message is the same.
> * [X]  Covariance matrix bug (perhaps caused by collinear variables):
> * [X]  I can toggle `Confidence intervals` but it only does something when `Estimates` is on. Maybe it should be grayed out whenever `Estimates` is unchecked? (I'm aware this behavior was also in the previous release, but I'm not very fond of it).
> * [X]  Regression test for funnel plot asymmetry is still base64 encoded:

Everything seems to work in R, but I'm getting some `object 'model term' not found` errors when I build JASP and test there (looks like it's from `formula(as.modelTerms(.v(options$modelTerms)))`).
meta-modelTerms

TimKDJ:
> Everything seems to work in R, but I'm getting some `object 'model term' not found` errors when I build JASP and test there (looks like it's from `formula(as.modelTerms(.v(options$modelTerms)))`).

I'm about to make some fixes to the encoding decoding functions which will solve this. However for these fixes to work you must undo your change to classicalmetaanalysis.qml:
```
ModelTermsList { width: parent.width * 5 / 9 }
```
is wrong, there are no null model terms. Replace it with
```
AssignedVariablesList  { name: "modelTerms"; title: qsTr("Model Terms"); listViewType: "Interaction"}
```

TimKDJ:
@vandenman could you have another look at this?

vandenman (changes requested):
Overall there were a couple bugs but I think it looks fine and works smoothly.

Bugs:

- [ ] invalid 'times' argument
image

Enter contNormal and contGamma as Effect Size and Effect Size Standard Error. Tick Funnel plot.

- [ ] Diagnostic plots
image

Same variables as above, tick Diagnostic plots under Diagnostics.

- [ ] Changing intercept does not update parameter estimates

image

add contcor1 as covariate, notice that ticking and unticking Include intercept under Model does not change the table of parameter estimates.s

There are some other cosmetic annoyances here and there (e.g., interactions effects are shown with `:` in tables) but I think @TimKDJ and I can sort those out later (we have to do some stuff for the translations anyway).

@gibbona1 can you fix the three bugs listed above?
@TimKDJ after that I think this PR is ready to be merged.

gibbona1:
> Bugs:
> 
> * [X]  invalid 'times' argument
> 
> * [X]  Diagnostic plots
> 
> * [X]  Changing intercept does not update parameter estimates

TimKDJ:
@vandenman: could you check it again?

@gibbona1: you'll need to remake the plots with the latest JASPgraphs versions

gibbona1:
> @vandenman: could you check it again?
> 
> @gibbona1: you'll need to remake the plots with the latest JASPgraphs versions

@TimKDJ I have JASPgraphs 0.4.5, jasptools 0.8.5, jaspResults 1.12
Are these all good?

vandenman:
[metaanalysisBugs.jasp.zip](https://github.com/jasp-stats/jasp-desktop/files/4098195/metaanalysisBugs.jasp.zip)

image

Funnel plot and diagnostics plot are still broken?

TimKDJ:
Yeh the dependencies are correct. I noticed in a [previous PR](https://github.com/jasp-stats/jasp-desktop/pull/3710/files) that some axis stuff needs to be made specific, because otherwise they appear different across platforms, namely:
- set the `limits` argument of the `scale_x_continuous()` and `scale_y_continuous()`
- use `JASPgraphs::getPrettyAxisBreaks()` to get the xbreaks and ybreaks

gibbona1:
@vandenman Funnel plot and diagnostics plot seem to work for me
Screenshot 2020-01-25 18 58 32

vandenman:
Okay, those plots appear to work, there was a package version issue on my end. At least we'll know that updating the dependencies will crash the module.

vandenman (approved):

PR #3628, opened 2019-10-31
decode plots

vandenman:
- Decodes all plots.
- Plots are now saved and shown in JASP as `.svg` files.
- Changed `root.get("name", ...` into `root.get("data",...` because the path where JASP looked to copy the file was completely wrong for me ("name" was the name jaspResults uses).
- There is a small bug in the network analysis plot function where some weird interaction with the decoding causes the circles around nodes to disappear. Upon resizing the plot, however, the bug disappears. Since I'm also rewriting the network analysis module, I'll fix that then.
- Reworked the decoding functions, they now work for `character`, `list`, `data.frame`, and `factor`.
- Added manual changes to CFA/ SEM. This is just replacing `.unv` with `decodeColumnName` where necessary. Perhaps we should disucss with @vankesteren  what the best approach is to take there.


JASP-Desktop/mainwindow.cpp
```diff
@@ -634,24 +634,25 @@ void MainWindow::_analysisSaveImageHandler(Analysis* analysis, QString options)
...
-		if(root["type"].asString() != "png")
+		if(root["type"].asString() != "svg")
 		{
 			root["finalPath"] = finalPath.toStdString();
 			analysis->saveImage(root);
 		}
 		else
 		{
-			QString imagePath = QString::fromStdString(TempFiles::sessionDirName()) + "/" + root.get("name", Json::nullValue).asCString();
+			QString imagePath = QString::fromStdString(TempFiles::sessionDirName()) + "/" + root.get("data", Json::nullValue).asCString();
```
JorisGoosen, line 655: Ah yeah I missed that one!

JASP-Engine/JASP/R/common.R
```diff
@@ -2224,106 +2224,106 @@ as.list.footnotes <- function(footnotes) {
...
-    type <- "quartz"
-
-
-  if (ggplot2::is.ggplot(plot) || inherits(plot, c("gtable", "ggMatrixplot", "JASPgraphs"))) {
-
-    pngMultip <- ppi / 96
+  
+  # convert width & height from pixels to inches. ppi = pixels per inch. 4 is a magic number.
+  width  <- width  / ppi * 4
+  height <- height / ppi * 4
```
JorisGoosen: Didn't we just do this at 2239 -> 2242 ?

JASP-R-Interface/jaspResults/R/writeImage.R
```diff
@@ -146,3 +138,48 @@ redrawPlotJaspResults <- function(rec_plot)
...
+}
+
+.decodeplot.JASPgraphsPlot <- function(x) {
+  for (i in seq_along(x$subplots))
+    x$subplots[[i]] <- .decodeplot(x$subplots[[i]], returnGrob = FALSE)
+  return(x)
+}
+
+.decodeplot.gg <- function(x, returnGrob = TRUE) {
+  # TODO: do not return a grid object!
```
JorisGoosen: What is the consequence of returning a grid object here?

JASP-Engine/JASP/R/common.R
```diff
@@ -2224,106 +2224,106 @@ as.list.footnotes <- function(footnotes) {
...
     backgroundColor <- .fromRCPP(".imageBackground")
     ppi             <- .fromRCPP(".ppi")
   }
-
+  
+  # convert width & height from pixels to inches. ppi = pixels per inch. 4 is a magic number.
+  if (units == "pixels") {
+    width  <- width  / ppi * 4
+    height <- height / ppi * 4
+  }
```
JorisGoosen: Ref for the comment at 2260/1

JorisGoosen:
Mostly looks fine, except for those two questions.

JASP-R-Interface/jaspResults/R/writeImage.R
vandenman: Right now, there isn't any difference in terms of output because .decodeplot is always called before drawing and we save the encoded plot in the state. 

It's probably better to save the decoded plot (and only decode once), but we can't do that because then we get a grid object and we can't get options for plot editing from a grid object. 

I'd rather return a ggplot object, but I'm returning a grid object right now because I don't know exactly how to decode certain ggplot functions that draw text (e.g., what if people put the name of a column as an axis tick mark? Thats's possible and I just don't know a catch all decoding method).

JASP-Engine/JASP/R/common.R
vandenman: good catch!

vandenman:
hmm seems like the unit tests cannot find .decodeplot if it's in writeImage... I'll move it to columnEncoding.

vandenman:
another thing is that we need [svglite](https://cran.r-project.org/web/packages/svglite/index.html) in the required files for the unit tests to work.

PR #3630, opened 2019-11-01
V0.11.1 fixes

lklende:
An issue
The scale reliability Statistics on the Reliability Analysis page of Descriptive statistic calculate wrong values for mean and SD. The values are calculated without taking into considertion reversed values. See Field's Fear of Statistics example.


TimKDJ:
https://github.com/jasp-stats/jasp-issues/issues/540

PR #3633, opened 2019-11-06
Bayesian t-tests: properly convert factors to numeric

### LINKED ISSUE #529: Incorrect means in descriptives from t-test
* JASP version: 0.11.1
* OS name and version: Mac OSX
* Analysis: Bayesian t-test
* Bug description: Incorrect means in descriptives from t-test that do not match the correct means from the descriptives toolbox
* Add JASP files as a zip: 
* [Screenshot:](url) 
image

* Expected behaviour:
* Steps to reproduce:

vandenman:
Fixes https://github.com/jasp-stats/jasp-issues/issues/529.


JohnnyDoorn (approved):
Haha classic R scenario with the as.numeric(factors), nice catch!

PR #3615, opened 2019-10-22
Implement Bain CRAN package and make unit tests

koenderks:

PR #3634, opened 2019-11-06
Fix some bugs in bain and implement suggested descriptives updates

koenderks:
As requested in https://github.com/jasp-stats/jasp-desktop/pull/3615


koenderks:
I think the unit tests will fail actually, I think I also have to rewrite these right @TimKDJ?

koenderks:
@TimKDJ Could you take another look at this?

koenderks:
@TimKDJ Could you take another look at this?

TimKDJ:
I'll have a look today somewhere

JASP-Engine/JASP/R/bainttestbayesianindependentsamples.R
```diff
@@ -238,9 +238,8 @@ BainTTestBayesianIndependentSamples <- function(jaspResults, dataset, options, .
 
 	  descriptivesTable$addColumnInfo(name="v",                    title = "", type="string")
 		descriptivesTable$addColumnInfo(name="group",                title = "Group", type="string")
-	  descriptivesTable$addColumnInfo(name="N",                    title = "N", type="integer")
+	  descriptivesTable$addColumnInfo(name="N",                    title = "n", type="integer")
```
TimKDJ: We call this `N` in other places (as per https://blog.apastyle.org/apastyle/2011/08/the-grammar-of-mathematics-writing-about-variables.html)

JASP-Engine/JASP/R/bainttestbayesianindependentsamples.R
```diff
@@ -238,9 +238,8 @@ BainTTestBayesianIndependentSamples <- function(jaspResults, dataset, options, .
...
 	  descriptivesTable$addColumnInfo(name="mean",                 title = "Mean", type="number")
-	  descriptivesTable$addColumnInfo(name="sd",                   title = "Std. Deviation", type="number")
```
TimKDJ, line 243: Any reason in particular why we're dropping the standard deviation?

JASP-Engine/JASP/R/bainregressionlinearbayesian.R
```diff
@@ -141,16 +144,23 @@ BainRegressionLinearBayesian <- function(jaspResults, dataset, options, ...) {
...
 	groups <- as.character(sum_model[["Parameter"]])
-	estim <- sum_model[["Estimate"]]
+	N <- sum_model[["n"]]
+	mu <- sum_model[["Estimate"]]
 	CiLower <- sum_model[["lb"]]
 	CiUpper <- sum_model[["ub"]]
-	SE <- (CiUpper - CiLower)/(2 * qnorm(0.975))
 
-	d <- data.frame(v = groups, mean = estim, SE = SE, CiLower = CiLower, CiUpper = CiUpper)
+	# Deduct standard error from 95 percent confidence interval
```
TimKDJ: Did you mean `Deduce` here? That comment threw me off a little :p

JASP-Engine/JASP/R/bainttestbayesianindependentsamples.R
```diff
@@ -238,9 +238,8 @@ BainTTestBayesianIndependentSamples <- function(jaspResults, dataset, options, .
...
 	  descriptivesTable$addColumnInfo(name="se",                   title = "Std. Error", type="number")
```
TimKDJ: Come to think of it I don't think "Std. Error" is an APA abbreviation of standard error -- this should be "SE" or "Standard Error"

TimKDJ (changes requested):
Very nice, way more consistent like this. Some minor nitpicky comments but that's it

JASP-Engine/JASP/R/bainregressionlinearbayesian.R
koenderks: Hehe you are right... I will change that

JASP-Engine/JASP/R/bainttestbayesianindependentsamples.R
koenderks: Should we keep it consistent? Or also change here?
image

JASP-Engine/JASP/R/bainttestbayesianindependentsamples.R
koenderks: Yeah, it is probably not what people are interested in when calling bain. All we do is compare means, so then mean + se.mean + conf.int is important. Also, for coefficients it would not be standard to report standard deviations, standard practice would be the standard error. I would say, for standard deviations, visit descriptive statistics. Bain would then just be for inference. What do you think?

JASP-Engine/JASP/R/bainttestbayesianindependentsamples.R
TimKDJ: Nice so we have SE, Std. Error and Std. Error of Mean.. :|
Maybe for now just be consistent with all the other t-tests we have and use SE?

JASP-Engine/JASP/R/bainttestbayesianindependentsamples.R
TimKDJ: I don't think it hurts anyone by being there, but I have no strong opinion either way, so it's up to you

koenderks: Alright, leaving it out then. If Herbert really wants it back, I'll put it back :p

JASP-Engine/JASP/R/bainttestbayesianindependentsamples.R
koenderks: Alright, will do!

TimKDJ (approved):

PR #3637, opened 2019-11-07
Stop testthat from terminating early

vandenman:
Line 41 of https://github.com/r-lib/testthat/blob/master/R/reporter-progress.R shows that it's actually possible to change the maximum number of failures for `testthat`. I changed it to 1000, so that practically the unit tests will never fail early (unless you mess up fantastically).

In case you consider testing this, if the limit is reached a new file won't be tested, but the current file will still be finished.


TimKDJ:
Heh that's neat. I never looked into it, but it did bother me

PR #3648, opened 2019-11-13
Dont initialize the webengine before creating the application

TimKDJ:
Can't build atm...

### DIFF
--- a/JASP-Desktop/main.cpp
+++ b/JASP-Desktop/main.cpp
@@ -272,10 +272,10 @@ int main(int argc, char *argv[])
 
 			QLocale::setDefault(QLocale(QLocale::English)); // make decimal points == .
 
-			QtWebEngine::initialize();
 
 			JASPTIMER_START("JASP");
 			Application a(argc, argv, filePathQ, unitTest, timeOut, save, logToFile);
+			QtWebEngine::initialize();
 			try
 			{
 				int exitCode = a.exec();


JorisGoosen:
Ok, i moved that because Qt gave me a message that initializing the webengine after qcoreapplication was deprecated.

And it does work on windows and linux. But if it doesn't for you then...

TimKDJ:
Heh.. I got `QtWebEngine::initialize() must be called after the construction of the application object.`

JorisGoosen:
[The docs say](https://doc.qt.io/qt-5/qtwebengine.html): 
> void QtWebEngine::initialize()
>
>Sets up an OpenGL Context that can be shared between threads. This has to be done before QGuiApplication is created and before window's QPlatformOpenGLContext is created.


So, I don't know what to do anymore :p

Right now, with your change it isn't called at all right?

TimKDJ:
Does it matter that we use `QApplication` rather than `QGuiApplication`?

JorisGoosen:
Well [QApplication is derived from QGuiApplication](https://doc.qt.io/qt-5/qapplication.html) so that shouldn't matter.

I am getting confused about this though..
[Here](https://doc.qt.io/qt-5/qtwebengine-overview.html#embedding-web-content-into-qt-quick-applications) a code snippet that is recommended for qtquick applications:
```C++
int main(int argc, char *argv[])
{
    QGuiApplication app(argc, argv);

    QtWebEngine::initialize();

    QQmlApplicationEngine engine;
    engine.load(QUrl("qrc:/main.qml"));

    return app.exec();
}
``` 

But that conflicts directly with the above doc of QtWebEngine::initialize()...

I guess I'll put it back where it was and I'll just ignore the warning :p

JorisGoosen:
No, I'll leave it as you did. It will be better if in the future (for translations perhaps) we call loadQML() multiple times.

JorisGoosen:
Aah yeah, I see it again, on Windows with qt 5.13.2:
`QtWebEngine::initialize() called with QCoreApplication object already created and should be call before. This is depreciated and may fail in the future.`

TimKDJ:
That's almost amusing. Might be something we should ask the Qt developers

PR #3651, opened 2019-11-13
changed QML_IMPORT_PATH

vandenman:
Previously there was always this (even though JASP would build just fine):

image

and after this PR it's gone:

image

this also means that it technically should be possible to use the QML designer that is shipped with qt creator, although it failed for all qml files that I tried.


JorisGoosen:
Sounds amazing!

And our qml files won't work with designer... I think you are not allowed to have any javascript (well, complicated javascript anyway) in there, because then it doesn't know what to do anymore.

PR #3652, opened 2019-11-13
Frequentist (RM) AN(C)OVA rewrites 3

JohnnyDoorn:
Already incorporated Don's review on the ANCOVA, still waiting for RM ANOVA to be reviewed.


TimKDJ:
For future reference: it's better to keep everything in one pull request so we can track its progression (now I need to look up the two closed pr's to see what has been dealt with already)

JohnnyDoorn:
> For future reference: it's better to keep everything in one pull request so we can track its progression (now I need to look up the two closed pr's to see what has been dealt with already)

I know, but I messed something up with git apparently, and I needed to rebase because my testplots were not working and I figured something might have been changed in development (else these unit tests would keep on failing). 
This one still contains all relevant changes. Don looked at the ANOVA and only had some minor changes, which are in here now.

Resources/ANOVA/qml/Ancova.qml
```diff
@@ -130,55 +142,66 @@ Form
...
+
 		Group
 		{
 			title: qsTr("Correction")
 			CheckBox { name: "postHocTestsTukey";		label: qsTr("Tukey"); checked: true	}
 			CheckBox { name: "postHocTestsScheffe";		label: qsTr("Scheffe")				}
 			CheckBox { name: "postHocTestsBonferroni";	label: qsTr("Bonferroni")			}
 			CheckBox { name: "postHocTestsHolm";		label: qsTr("Holm")					}
-            CheckBox { name: "postHocTestsSidak";       label: qsTr("Šidák")                }
+            		CheckBox { name: "postHocTestsSidak";       label: qsTr("Šidák")                }
```
vandenman: We use tabs and not spaces in QML files, see [this link](https://doc.qt.io/qtcreator/creator-indenting-code.html) to change your default settings for Qt creator. Please edit this in all QML files (even if they used spaces before, they shouldn't have).

vandenman (changes requested):
Nice that integers are now always displayed as integers in (almost) all tables!

I like that you moved the effect size buttons to a different spot (it took me a minute to find them in the current release).

- [ ] Variables forms have different heights
  - Anova.qml: no height
  - Ancova.qml: height:400
  - Anovarepeatedmeasures.qml: height: 520.

- [ ] The Levenes table does not display df as integers:
image


- [ ] Clicking Display error bars in Ancova crashes:
image

Note that the plot in the current release also uses colors (don't ask me why though):
image


Perhaps we can think of a more appropriate solution that reuses the scatter plots from descriptives.

- [ ] The effect size for the between subjects effects is different from before:

image
(left old, right new)

JASP-Tests/R/tests/testthat/test-ancova.R
```diff
@@ -344,13 +344,13 @@ test_that("Field - Chapter 6 results match", {
   
   
   # descriptive plots
-  plot1 <-  results$state$figures[[1]]$obj
+  plot1 <-  results$state$figures[[1]]$obj$subplots$mainPlot
```
vandenman, line 347: you should be able to pass `results$state$figures[[1]]$obj` directly to `expect_equal_plots`. Internally, that will recognize it as a JASPgraphsplot and test each subplot independently.

JASP-Engine/JASP/R/anovarepeatedmeasures.R
```diff
@@ -504,6 +504,8 @@ AnovaRepeatedMeasures <- function(jaspResults, dataset = NULL, options) {
     return()
   
   betweenTable <- createJaspTable(title = "Between Subjects Effects", position = 2)
+  betweenTable$dependOn <- c("effectSizeEstimates", "effectSizeEtaSquared", "effectSizePartialEtaSquared", 
+                             "effectSizeGenEtaSquared", "effectSizeOmegaSquared", "VovkSellkeMPR")
```
vandenman: This is why your unit tests fail:
```r
  betweenTable$dependOn <- c("effectSizeEstimates", "effectSizeEtaSquared", "effectSizePartialEtaSquared", 
                             "effectSizeGenEtaSquared", "effectSizeOmegaSquared", "VovkSellkeMPR")
```
should be
```r
betweenTable$dependOn(c("effectSizeEstimates", "effectSizeEtaSquared", "effectSizePartialEtaSquared", 
                        "effectSizeGenEtaSquared", "effectSizeOmegaSquared", "VovkSellkeMPR"))
```

vandenman (changes requested):

JASP-Tests/R/tests/testthat/test-ancova.R
JohnnyDoorn: I tried this but it does not work so Im keeping it like this

JohnnyDoorn:
Descriptive plots are now updated (uses the scatterplot from Descriptives when the x-variable is covariate).
The changed eta squared is correct, this was wrong for mixed ANOVA (https://forum.cogsci.nl/index.php?p=/discussion/3210/0/#Form_Comment)
The df are now integers for Levene's
The variable form heights are not changed, because of the different layouts for the analyses I think it makes sense like they are now.

JASP-Tests/R/tests/testthat/test-ancova.R
vandenman: Did you update both JASPgraphs and jasptools? Because at lines 98-105 of [test-ttestbayesianindependentsamples.R](https://github.com/jasp-stats/jasp-desktop/blob/development/JASP-Tests/R/tests/testthat/test-ttestbayesianindependentsamples.R) it works fine for the t-tests.

TimKDJ (changes requested):
Haven't gotten around to looking at the code, but I did some testing already:

AN(C)OVA:
- [x] The footnote “Bootstrapping based on 1000successful replicates” requires a space between the number and successful
- [x] Descriptives table has N as a double, this could be an integer
- [x] There are output elements that lack a dependency on the WLS weights (at least the main table, not sure what other output elements actually depend on it)
- [x] The analysis doesn’t like it when you include a covariate in post hoc tests and do not perform bootstrapping: `Error in qr.default(object@linfct[., , drop = FALSE]): NA/NaN/Inf in foreign function call (arg 1)`**
- [x] The standard posthoc table is missing a dependency on the “Standard” checkbox (and the other post hoc tables do have a dependency on it, which I don’t think they need?)
- [x] If you open the “Beer Goggles” .jasp file from the data library you’re met with an error: `Error in `[[<-.data.frame`(`*tmp*`, 'F value', value = numeric(0)): replacement has 0 rows, data has 2`
- [x] The flag significant correlations adds multiple footnotes in and below the table (see https://github.com/jasp-stats/INTERNAL-jasp/issues/562):
Screenshot 2019-11-20 at 16 49 26

- [x] Shouldn't the non-standard posthoc tests have a dependency on “Flag Significant Comparisons”?
- [x] Post hoc crashes when I set the confidence interval to 30 and untick/tick Dunnett**
- [x] Should the “Confidence intervals” checkbox be indented under “Standard”? It appears to have an effect on the other post hoc test tables as well. Although if you make changes the other post hoc tables do not update correctly (e.g., changing conf int from 95 to 50 does not update the other tables, however if you set it to 50 before you make any tables, then all tables will show 50)

** = [rewriteANOVA.jasp.zip](https://github.com/jasp-stats/jasp-desktop/files/3870354/rewriteANOVA.jasp.zip)

---

RM ANOVA:
- [x] The container title “Repeated Measures ANOVA” seems a bit redundant, because that’s the same as the title of the analysis which is right above it
Screenshot 2019-11-20 at 17 55 18

- [x] The pvalue of the contrasts table should have `type = "pvalue"` (also noticed it in simple main effects and conover’s post hoc, so it might be missing at a few spots)
- [ ] The conover’s post hoc table is missing a dependency on the conover’s post hoc tests checkbox



--- 
Some suggestions:
- [x] The analyses have multiple progressbars — it might be nice to give them a label so the user knows what is happening when a given progressbar is shown (just specify the label in `startProgressbar(expectedTicks, label)`, e.g., `label = "Estimating marginal means"`)
- [ ] It would be nice if the “Comparison” could be the same for each table (now it’s 0 1, 0 - 1, 1 - 0); also the standard table lacks the “Comparison” column name:
Screenshot 2019-11-20 at 15 25 31

AlexanderLyNL:
Is it necessary to call the bootstrap replicates successful? Are we rejecting some bootstrap samples?

vandenman:
@AlexanderLyNL Yes, the bootstraps can fail. This is just to distinguish between failed and successful replicates (previously it just indicated the number, e.g., "based on 1000 bootstrap replicates").

AlexanderLyNL:
Why not just call them the effective bootstrap samples? Successful sounds as if things were going wrong.

JohnnyDoorn:
That's what I suggested too, but the argument in favor of "successful" is that it is a little bit easier to understand for the layman, albeit less accurate.

vandenman:
@AlexanderLyNL This was the previous situation:
image

A user wants 1000 bootstraps and sees in the table that only 999 (or some other number) were done. There was no indication as to why this.

AlexanderLyNL:
Okay, makes sense, go for it.

JASP-Engine/JASP/R/ancova.R
```diff
@@ -650,566 +318,386 @@ Ancova <- function(dataset=NULL, options, perform="run", callback=function(...)
...
   
-  list(model = model, singular = singular)
+  # Save model to state
+  anovaContainer[["model"]] <- createJaspState(object = model$model)
 }
 
-.anovaTable <- function(options, model, status, singular) {
+.anovaResult <- function(anovaContainer, options) {
+
+  model <- anovaContainer[["model"]]$object
```
TimKDJ, line 352: It seems to me there should be a check at this point to see if there is an error on the container (possibly set in `.anovaModelContainer()` on line 336) before using `model`

JASP-Engine/JASP/R/ancova.R
```diff
@@ -650,566 +318,386 @@ Ancova <- function(dataset=NULL, options, perform="run", callback=function(...)
...
-      anova[["title"]] <- paste("ANOVA - ", options$dependent, sep = "")
-      
-    } else {
-      
-      anova[["title"]] <- "ANOVA"
-      
+    result <- base::tryCatch(stats::anova(model),error=function(e) e, warning=function(w) w)
+    
+    if (!is.null(result$message) && result$message == "ANOVA F-tests on an essentially perfect fit are unreliable")
+      stop(result$message)
```
TimKDJ: This is not caught inside the analysis. Did you mean to use `$setError()`?

JASP-Engine/JASP/R/ancova.R
```diff
@@ -650,566 +318,386 @@ Ancova <- function(dataset=NULL, options, perform="run", callback=function(...)
...
+  anovaTable$setExpectedSize(rows = length(options$modelTerms) * length(corrections))
   
-  if (options$sumOfSquares == "type1") {
-    
-    .addFootnote(footnotes, text = "Type I Sum of Squares", symbol = "<em>Note.</em>")
+  # here we ask for the model to be computed
+  .anovaResult(anovaContainer, options)
+
+  if ((options$homogeneityBrown || options$homogeneityWelch) && length(options$modelTerms) > 1) {
+    anovaTable$setError("The Brown-Forsythe and Welch corrections are only available for one-way ANOVA")
```
TimKDJ, line 553: [Question]: performing a multi-way anova with these corrections used to render the entire analysis useless and prevented all other output elements from showing. At the moment this would only stop this very table. Is that intended?

JASP-Engine/JASP/R/ancova.R
```diff
@@ -1903,1361 +1379,334 @@ Ancova <- function(dataset=NULL, options, perform="run", callback=function(...)
...
+  qqPlot <- createJaspPlot(title = "Q-Q Plot", width = options$plotWidthQQPlot, height = options$plotHeightQQPlot)
   
-  gamesVariables <- unique(unlist(options$postHocTestsVariables))
-  dependentVar <- dataset[[ .v(options$dependent) ]]
-  postHocInterval  <- options$confidenceIntervalIntervalPostHoc
+  # now we assign the plot to jaspResults
+  anovaContainer[["assumptionsContainer"]][["qqPlot"]] <- qqPlot
   
-  gamesTables <- list()
+  if (!ready)
```
TimKDJ: And a check to see if the container has an error. 

This goes for all output elements that use parts of `anovaContainer[["model"]]$object`  -- the `(!ready)` check should be expanded to check for container errors too

JASP-Engine/JASP/R/ancova.R
```diff
@@ -1224,491 +712,528 @@ Ancova <- function(dataset=NULL, options, perform="run", callback=function(...)
...
-      pSidak <- "."
-      effectSize <- "."
-      md <- "."
-      se  <- "."
-      t <- "."
-      p  <- 1
-      uprBound <- "."
-      lwrBound <- "."
+      ## Computation
+      bootstrapPostHoc <- try(boot::boot(data = dataset, statistic = .bootstrapPostHoc,
```
TimKDJ, line 846: The `try()` around `boot::boot()` should be handled (or removed). As it stands, if a `try-error` occurs in `boot::boot()` then the analysis continues but it will result in a more cryptic error being thrown in `summary(bootstrapPostHoc)`

JASP-Engine/JASP/R/ancova.R
```diff
@@ -1224,491 +712,528 @@ Ancova <- function(dataset=NULL, options, perform="run", callback=function(...)
...
+      bootstrapPostHocConf <- t(sapply(1:nrow(bootstrapSummary), function(comparison){
+        res <- try(boot::boot.ci(boot.out = bootstrapPostHoc, conf = options$confidenceIntervalIntervalPostHoc, type = "bca",
+                                 index = comparison)[['bca']][1,4:5])
+        if(!inherits(res, "try-error")){
+          return(res)
+        } else if(identical(attr(res, "condition")$message, "estimated adjustment 'a' is NA")){
+          ci.fails <<- TRUE
+          return(c(NA, NA))
+        } else{
+          return(res)
```
TimKDJ: Sort of the same situation here. If a `try-error` occurs then it returns this unhandled (unless it has a specific message) and this will cause problems further down

JASP-Engine/JASP/R/ancova.R
```diff
@@ -1903,1361 +1379,334 @@ Ancova <- function(dataset=NULL, options, perform="run", callback=function(...)
...
-          row[["t"]] <- .clean(r$t.ratio[index])
-          row[["p"]] <- .clean(r$p.value[index])
-        }
-        
-        if(cases[k,nCol] == lvls[[ nCol ]][1]) {
-          row[[".isNewGroup"]] <- TRUE
-        } else {
-          row[[".isNewGroup"]] <- FALSE
+      ci.fails <- FALSE
+      bootstrapMarginalMeansCI <- t(sapply(1:nrow(bootstrapSummary), function(index){
```
TimKDJ, line 1449: Same two comments about the try-catching here (and also for the rm anova)

JASP-Engine/JASP/R/anovarepeatedmeasures.R
```diff
@@ -560,33 +227,46 @@ AnovaRepeatedMeasures <- function(dataset=NULL, options, perform="run", callback
   
   main <- paste("(",paste(unlist(terms.base64), collapse=" + "),")", sep="")
   termsBS <- paste("(",paste(termsBS.base64, collapse=" + "),")", sep="")
-  errorRM <- paste("Error(",paste("subject/(", termsRM.base64, ")",sep="", collapse=" + "),")",sep="")
+  errorRM <- paste("Error(",paste("Xc3ViamVjdA/(", termsRM.base64, ")",sep="", collapse=" + "),")",sep="")
```
TimKDJ: We're actually moving away from base64 encoding, so this will break once you rebase. If you use `subject` and `dependent` directly, instead of their base64 equivalents, then you should be fine. 

The new encoding only encodes/decodes the actual column names. So calling `.v("dependent")` will return `"dependent"` which is perfect here

JASP-Engine/JASP/R/anovarepeatedmeasures.R
```diff
@@ -970,2554 +1151,693 @@ AnovaRepeatedMeasures <- function(dataset=NULL, options, perform="run", callback
...
-      list(name="GG", type="number", format="sf:4;dp:3", title="Greenhouse-Geisser \u03B5"),
-      list(name="HF", type="number", format="sf:4;dp:3", title="Huynh-Feldt \u03B5"),
-      list(name="LB", type="number", format="sf:4;dp:3", title="Lower Bound \u03B5"))
-    
+    return(rep(NA, termLength))
   }
+}
+
+.rmAnovaFriedmanTable <- function(rmAnovaContainer, longData, options, ready) {
+  if (!is.null(rmAnovaContainer[["nonparametricContainer"]]) || length(options$friedmanWithinFactor) == 0 || !ready)
```
TimKDJ: This `!ready` can be removed as it's also checked on line 1459

TimKDJ (changes requested):
Code structure is nice in both the rm & regular anova!
Did find a few things that could be improved

TimKDJ:
I only tested the things I found last time to see if they are fixed. All are except for this one:

> RM ANOVA:

>     * [ ]  The conover’s post hoc table is missing a dependency on the conover’s post hoc tests checkbox

I also noticed the following 3 things:

- [ ] The symbol for the bca interval and significance is the same, the bca one should be something else:
Screenshot 2019-12-12 at 13 27 42

- [ ] Do we need both of these footnotes?
Screenshot 2019-12-12 at 13 30 05

- [ ] In the ANOVA qml form -> Assumption Checks; you got rid of the "Sphericity corrections" checkbox and just used the three correction checkboxes directly; the RM ANOVA still has this top level checkbox. Would be good to make this consistent

JASP-Engine/JASP/R/ancova.R
JohnnyDoorn: Correct - the rest of the analysis uses the non-corrected model anyway, the corrections are just for the main ANOVA and become unavailable as there are more than 1 factors specified.

JASP-Engine/JASP/R/ancova.R
JohnnyDoorn: I think this should be left as is - if it fails then it will print a footnote, which I think will suffice in this case. If something really wonky goes on in bootstrapping it will be caught by the earlier try-catch/setError construction.

TimKDJ: The footnote is only added when the error message contains `"estimated adjustment 'a' is NA"`, but the error message might be something completely different. The `else` captures that situation and that situation will sort of crash and burn ;)

JASP-Engine/JASP/R/ancova.R
```diff
@@ -1224,491 +720,525 @@ Ancova <- function(dataset=NULL, options, perform="run", callback=function(...)
...
+                                         postHocVarIndex = postHocVarIndex),
+                              silent = TRUE)
       
-      mainPostResults <- matrix(nrow=1)
-    }
-    
-    for (i in 1:nrow(mainPostResults)) {
+      if (class(bootstrapPostHoc) == "try-error") {
+        postHocStandardContainer[[thisVarName]]$setError(bootstrapPostHoc)
+        return()
```
TimKDJ: `next()` instead of `return()` (there are other tables that might work fine)

JASP-Engine/JASP/R/ancova.R
```diff
@@ -1903,1361 +1385,339 @@ Ancova <- function(dataset=NULL, options, perform="run", callback=function(...)
...
       
-      if(options$marginalMeansCIAdjustment == "bonferroni") {
-        adjMethod <- "bonferroni"
-      } else if(options$marginalMeansCIAdjustment == "sidak") {
-        adjMethod <- "sidak"
-      } else {
-        adjMethod <- "none"
+      if (class(bootstrapMarginalMeans) == "try-error") {
+        marginalMeansContainer[[thisVarName]]$setError(bootstrapMarginalMeans)
+        return()
```
TimKDJ: `next()`

JASP-Engine/JASP/R/ancova.R
```diff
@@ -1903,1361 +1385,339 @@ Ancova <- function(dataset=NULL, options, perform="run", callback=function(...)
...
+      bootstrapMarginalMeansCI <- t(sapply(1:nrow(bootstrapSummary), function(index){
+        res <- try(boot::boot.ci(boot.out = bootstrapMarginalMeans, conf = 0.95, type = "bca",
+                                 index = index)[['bca']][1,4:5])
+        if(!inherits(res, "try-error")){
+          return(res)
+        } else if(identical(attr(res, "condition")$message, "estimated adjustment 'a' is NA")){
+          ci.fails <<- TRUE
+          return(c(NA, NA))
+        } else{
+          return(res)
```
TimKDJ: This `boot.ci` should get the same treatment as the others did (no return of the try error and no message checking)

JASP-Engine/JASP/R/anovarepeatedmeasures.R
```diff
@@ -970,2554 +1162,698 @@ AnovaRepeatedMeasures <- function(dataset=NULL, options, perform="run", callback
...
+      bootstrapMarginalMeans <- try(boot::boot(data = dataset, statistic = .bootstrapMarginalMeansRmAnova, 
+                                               R = options[["marginalMeansBootstrappingReplicates"]],
+                                               options = options,
+                                               nRows = nRows,
+                                               termLength = length(term),
+                                               formula = formula), silent = TRUE)
+
+      if (class(bootstrapMarginalMeans) == "try-error") {
+        marginalMeansContainer[[termBase64]]$setError(bootstrapMarginalMeans)
+        return()
```
TimKDJ: `next()`

JASP-Engine/JASP/R/anovarepeatedmeasures.R
```diff
@@ -766,118 +845,231 @@ AnovaRepeatedMeasures <- function(dataset=NULL, options, perform="run", callback
...
+  
+  postHocTable$addColumnInfo(name="contrast_A", title=" ", type="string", combine = TRUE)
+  postHocTable$addColumnInfo(name="contrast_B", title=" ", type="string")
+  
+  postHocTable$addColumnInfo(name="estimate", title="Mean Difference", type="number")
+  
+  if (options$confidenceIntervalsPostHoc || makeBootstrapTable) {
+    
+    if (makeBootstrapTable) {
+      thisOverTitle <- paste0(options$confidenceIntervalIntervalPostHoc * 100, "% bca\u002A CI")
```
TimKDJ, line 1020: This was supposed to become the cross, right?

TimKDJ (changes requested):
Laatste loodjes ;)

JASP-Engine/JASP/R/anovarepeatedmeasures.R
TimKDJ: And this goes for all the bca stuff everywhere

TimKDJ:
Can you solve the merge conflicts? Then we'll be good to go

PR #3654, opened 2019-11-13
Use a custom condition to break out of an R analysis

TimKDJ:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/462

jaspResults will be able to break out of try statements, e.g.,
```
startProgressbar(1000)
try({
  for (i in 1:1000) {
    Sys.sleep(0.1)
    progressbarTick()
  }
})
<this point will not be reached>
```


JASP-R-Interface/jaspResults/src/jaspResults.cpp
```diff
@@ -216,8 +214,8 @@ void jaspResults::checkForAnalysisChanged()
 	{
 		jaspPrint("Polling for analysis changes found a change, analysis should restart!");
 		setStatus("changed");
-		static Rcpp::Function stop("stop");
-		stop(_analysisChangedErrorMessage);
+		static Rcpp::Function signalAnalysisAbort = jaspResults::isInsideJASP() ? Rcpp::Function("signalAnalysisAbort") : Rcpp::Environment::namespace_env("jaspResults")["signalAnalysisAbort"];
```
JorisGoosen, line 217: Funky!

JorisGoosen:
Looks good!

You will need to increment the JASP R Interface version number though...

TimKDJ:
> Looks good!
> 
> You will need to increment the JASP R Interface version number though...

I guess I'm going to need a post-it in the middle of my screen reminding me of this...

JorisGoosen:
Yeah that is annoying, I'll check and see if we can get rid of the whole JASP-R-Interface being split off from the rest. Then you can just use your screen for it's display properties :p

PR #3657, opened 2019-11-14
Rewrite NetworkAnalysis

vandenman:
Rewrite of network analysis to jaspResults.

The widget `InputListView` that assigns variables to groups doesn't work properly at the moment and is therefore only available in debug.


JASP-Engine/JASP/R/networkanalysis.R
```diff
@@ -15,1760 +15,1356 @@
...
+  
+  options[["layoutInvalid"]] <- FALSE
+  # some sanity checks on the layout
+  if (!is.null(options[["layoutXData"]]) && (length(options[["layoutXData"]]) < length(variables) || length(options[["layoutYData"]]) < length(variables)))
+    options[["layoutInvalid"]] <- TRUE
+  
+  options[["nGraphs"]] <- length(dataset)
+  options[["ready"]] <- length(options[["variables"]]) > 2L
+  options[["readyBootstrap"]] <- options[["ready"]] && options[["bootstrapOnOff"]] && options[["numberOfBootstraps"]] > 0L
+  return(options)
```
TimKDJ: Have to admit I'm not the biggest fan of mucking about with the options list. Especially considering you don't use most things you add to it more than 2 times and some only once. 

If I saw `options[["nGraphs"]]` in an analysis I would immediately think I will be able to find this in the qml file, which is not the case. Whereas if I saw something like `.networkAnalysisNumGraphs(dataset) > 1` I would understand what's up.

Up to you though. It's not wrong, just unexpected.

JASP-Engine/JASP/R/networkanalysis.R
```diff
@@ -15,1760 +15,1356 @@
...
+  
+  bootstrapType <- options[["BootstrapType"]]
+  substr(bootstrapType, 1, 1) <- toupper(substr(bootstrapType, 1, 1)) # capitalize first letter
+  
+  table <- createJaspTable(title = "Bootstrap summary of Network", position = position)
+  table$addColumnInfo(name = "type",               title = "Type", type = "string")
+  table$addColumnInfo(name = "numberOfBootstraps", title = "Number of bootstraps", type = "integer")
+  table[["type"]]               <- bootstrapType
+  table[["numberOfBootstraps"]] <- options[["numberOfBootstraps"]]
+  return(table)
```
TimKDJ: I think you mean to assign this to the bootstrapContainer rather than returning it

TimKDJ (changes requested):
Nicely structured code; I have little to comment there.

From testing:
- [ ] In the qml file “nFolds” should have a default >= 3 to avoid the red border
- [ ] The “info” column of the “Summary of Network” table seems.. useless
- [ ] The error check for < 10 in each group after grouping seems gone as it now crashes hard**
- [ ] If you perform bootstrapping and then toggle the bootstrap type, the edges and centrality plots disappear never to return (this does not happen when you only have the bootstrap plots visible, you also need all other plots and tables visible— I suspect one of them screws it up, but didn’t test which exactly)**
- [ ] The footnote “Bootstrapping is only done when 'Bootstrap network' is ticked and 'Number of Bootstraps' is larger than 0." does not go away when you satisfy both criteria (same for the min edge strength > max edge strength)
- [ ] `Please drag all variables to a particular type under "Estimator options"` should that be `"Analysis Options"`?
- [ ] Should the weights matrix really be that low in the output? Considering how prominent it was before

** = [networkRewrite.jasp.zip](https://github.com/jasp-stats/jasp-desktop/files/3847491/networkRewrite.jasp.zip)

vandenman:
:heavy_check_mark: In the qml file “nFolds” should have a default >= 3 to avoid the red border
:heavy_check_mark: The “info” column of the “Summary of Network” table seems.. useless
:heavy_check_mark: The error check for < 10 in each group after grouping seems gone as it now crashes hard**
:heavy_check_mark: The footnote “Bootstrapping is only done when 'Bootstrap network' is ticked and 'Number of Bootstraps' is larger than 0." does not go away when you satisfy both criteria (same for the min edge strength > max edge strength)
:heavy_check_mark:`Please drag all variables to a particular type under "Estimator options"` should that be `"Analysis Options"`?

Good points, I implemented all the above.

:heavy_check_mark: Should the weights matrix really be that low in the output? Considering how prominent it was before

Eehh, honestly, I don't think anybody will look at it. Still, I've moved it to be above the plots and below the last table (so that all tables are grouped).


:grey_question: If you perform bootstrapping and then toggle the bootstrap type, the edges and centrality plots disappear never to return (this does not happen when you only have the bootstrap plots visible, you also need all other plots and tables visible— I suspect one of them screws it up, but didn’t test which exactly)**

I could not replicate this.. I had all tables and plots visible but the bootstrap plots always stayed visible. Perhaps you can show me in person?

JASP-Engine/JASP/R/networkanalysis.R
vandenman: Good catch!

JASP-Engine/JASP/R/networkanalysis.R
vandenman: I'm still not too happy about modifying the options either.. `nGraphs` is just the number of graphs (I hope that is clear). `.networkAnalysisNumGraphs(dataset)` would just be `length(dataset)` which I would find a horrible function to write/ use. Somehow, I just want to store an additional set of options in an easily accessible object. In the ttests, I used a list called `derivedOptions`, but here there would only be three options. I could hide these options in the model object of course, but I don't think they really belong there.

TimKDJ: Well to be honest I would just use `length(dataset) > 1`, but I figured you wanted to "namespace" it somehow considering your current solution.

Strongly advocate either using a dedicated list for this, just doing the if statements on the fly, or creating small functions

vandenman: Okay, I changed it to recompute everything on the fly.

TimKDJ:
I'll see if I can find a more reliable reproduction path. (If I can't, we'll have to have a look at the state that is in the .jasp file I attached earlier, because that contains the problem)

JASP-Engine/JASP/R/networkanalysis.R
```diff
@@ -15,1760 +15,1360 @@
...
-	} else {
+}
 
-		dataset <- .vdf(dataset, columns.as.numeric = variables, columns.as.factor = varsAsFactor)
+# plots ----
+.networkAnalysisPlotContainer <- function(mainContainer, network, options) {
+  
+  plotContainer <- mainContainer[["plotContainer"]]
+  if (is.null(plotContainer)) {
+    plotContainer <- createJaspContainer(position = 4, dependencies = c("abbreviateLabels", "abbreviateNoChars"))
```
TimKDJ: They're both at 4 now, for me it shows the plotcontainer first

TimKDJ (approved):
Can't reproduce the bug anymore myself either :-)

PR #3659, opened 2019-11-14
SEM rewrite

JohnnyDoorn:
Removed the options for multiple models (discussed with Tim/Erik-Jan/Don). I will add this functionality when we have a decent (tabbed) widget for this (we will make a separate issue for this).

### DIFF
--- a/JASP-Engine/JASP/R/semsimple.R
+++ b/JASP-Engine/JASP/R/semsimple.R
@@ -60,15 +60,9 @@ SEMSimple <- function(jaspResults, dataset = NULL, options) {
   
   .semMardiasCoefficientTable(semContainer, options, dataset, ready)
 
-
-  
   .lavCreatePathDiagram(semContainer, options, ready)
-  return()  
-  if (options$groupingVariable != "")
-    meta[[13]] <- list(name="pathDiagram", type="collection", meta="image")
-  else
-    meta[[13]] <- list(name="pathDiagram", type="image")
   
+  return()  
 }
 
 .getSemContainer <- function(jaspResults) {
@@ -345,10 +339,14 @@ SEMSimple <- function(jaspResults, dataset = NULL, options) {
   
   semRMSEATable$addColumnInfo(name="model", title="", type="string")
   semRMSEATable$addColumnInfo(name="rmsea", title="RMSEA", type="number")
-  semRMSEATable$addColumnInfo(name="rmsea.ci", title="90 Percent Confidence Interval", type="string")
+  thisOverTitle <- "90% CI"
+  semRMSEATable$addColumnInfo(name="rmsea.ci.lower", type = "number", title = "Lower",
+                              overtitle = thisOverTitle)
+  semRMSEATable$addColumnInfo(name="rmsea.ci.upper", type = "number", title = "Upper",
+                              overtitle = thisOverTitle)
   semRMSEATable$addColumnInfo(name="rmsea.pvalue", title="p-value RMSEA <= 0.05 ", type="number")  
   
-  semRMSEATable$setData(c(list(model="Model"), semFitMeasures[c('logl', 'unrestricted.logl', 'npar', 'aic', 'bic', 'bic2')]))
+  semRMSEATable$setData(c(list(model="Model"), semFitMeasures[c('rmsea', 'rmsea.ci.lower', 'rmsea.ci.upper', 'rmsea.pvalue')]))
   
   semContainer[["fitMeasuresRMSEA"]] <- semRMSEATable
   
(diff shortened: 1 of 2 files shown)


JASP-Engine/JASP/R/semsimple.R
```diff
@@ -15,497 +15,201 @@
...
-    if (identical(errorCheck, FALSE)) {
-      if (perform == "run" && ncol(dataset) > 0 && !nrow(dataset) > ncol(dataset)) {
-        errorCheck <- list(error=TRUE, message="Not more cases than number of variables. Is your data a variance-covariance matrix?")
-      }
-    }
-  }
-
-  inputCorrect <- identical(errorCheck, FALSE)
-  ######################
+    semContainer <- createJaspContainer(title = "Structural Equation Modeling<br/><span style='color:#888888;font-family:monospace;font-size:12px;font-weight:normal;'>Powered by lavaan.org</span>")
```
TimKDJ: I thought we removed this on purpose to not give lavaan a special status? (i.e., we have no "power by BayesFactor" or the like)

JASP-Engine/JASP/R/semsimple.R
```diff
@@ -15,497 +15,201 @@
...
+                            "outputMardiasCoefficients", "outputModificationIndices", 
+                            "outputModificationIndicesHideLowIndices", "outputModificationIndicesHideLowIndicesThreshold",
+                            "groupingVariable", "eq_loadings", "eq_intercepts", "eq_residuals", "eq_residualcovariances",
+                            "eq_means", "eq_thresholds", "eq_regressions", "eq_variances", "eq_lvcovariances", 
+                            "includeMeanStructure", "assumeFactorsUncorrelated", "fixExogenousCovariates", 
+                            "factorStandardisation", "fixManifestInterceptsToZero", "fixLatentInterceptsToZero",
+                            "omitResidualSingleIndicator", "residualVariances", "correlateExogenousLatents", 
+                            "addThresholds", "addScalingParameters", "correlateDependentVariables", "addPathDiagram",
+                            "outputpathdiagramstandardizedparameter", "Data", "errorCalculation", "estimator", 
+                            "emulation"))
```
TimKDJ: This approach seems a little heavy handed. Quite a few options in here are not shared dependencies.
From what I could tell the checkboxes "Additional fit measures" until "Modification Indices/threshold" (under "Statistics") are distinct output elements that are not dependent on one another. You could put these dependencies on the respective output element itself

JASP-Engine/JASP/R/semsimple.R
```diff
@@ -525,487 +229,646 @@ SEMSimple <- function(dataset=NULL, options, perform="run", callback=function(..
...
-  if (is.null(semResults)) {
-    sem_anova <- structure(list(Df=c(0, NA), AIC=c(NA, NA), BIC=c(NA, NA),
-                                Chisq = c(0, NA), `Chisq diff`=c(NA, NA),
-                                `Df diff` = c(NA, NA), `Pr(>Chisq)` = c(NA, NA)),
-                           .Names = c("Df", "AIC", "BIC", "Chisq", "Chisq diff", "Df diff", "Pr(>Chisq)"),
-                           row.names = c("Saturated", options$modelName),
-                           class = c("anova", "data.frame"),
-                           heading = "Chi Square Test Statistic (unscaled)\n")
+.semMainTable <- function(semContainer, options, dataset, ready) {
+  if (!ready || !is.null(semContainer[["semFitTable"]]))
```
TimKDJ: Remove the `!ready` here and then on line 483 put (while also removing the table assignment at the end of the function) this:
```
semContainer[["semFitTable"]] <- semFitTable

if (!ready)
  return()
```
This ensures that the table has a nice empty placeholder for when the analysis is not ready (otherwise it immediately returns and all we get is whitespace in the output).

Please do this for all the tables/plot

JASP-Engine/JASP/R/semsimple.R
```diff
@@ -525,487 +229,646 @@ SEMSimple <- function(dataset=NULL, options, perform="run", callback=function(..
...
+  semFitTable <- createJaspTable(title = "Chi Square Test Statistic (unscaled)")
+  
+  semFitTable$addColumnInfo(name="Model", title = "", type="string")
+  semFitTable$addColumnInfo(name="Df", title = "df", type="number")
+  # semFitTable$addColumnInfo(name="DfDiff", title = "&#916;df", type="number")
+  semFitTable$addColumnInfo(name="AIC", type="number")
+  semFitTable$addColumnInfo(name="BIC", type="number")
+  semFitTable$addColumnInfo(name="Chisq", title = "&#967;&sup2;", type="number")
+  # semFitTable$addColumnInfo(name="ChisqDiff", title = "&#916;&#967;&sup2;", type="number")
+  semFitTable$addColumnInfo(name="PrChisq", title = "p", type="number")
```
TimKDJ: For all p values put `type = "pvalue"`, this ensures they behave properly (showing < 0.001, etc.)

TimKDJ (changes requested):
Thanks a lot for doing this Johnny. Really helped me out :)

A couple of code comments and I noticed the following while testing:
- [ ] Analysis crashes if you tick "R-squared"**
- [ ] Analysis crashes if you untick all checkboxes**
- [ ] Analysis crashes if you select a grouping variable**
- [ ] In the "Parameter Estimates" and "Modification Indices" tables the abbreviations are still base64 encoded
- [ ] The "Other Fit Indices" table no longer appears to have the "Expected Cross-Validation Index" column
- [ ] The "Covariances (lower triangle) / correlations (upper triangle)" table is (1) duplicated and (2) the table itself has no title. This only seems to be the case when you tick "Means" under "Equality Constraints"**
- [ ] If you untick "Path diagram" then the path diagram remains

** = [semRewrite.zip](https://github.com/jasp-stats/jasp-desktop/files/3852525/semRewrite.zip)

JASP-Engine/JASP/R/semsimple.R
TimKDJ: This ensures we get the right hand situation after the rewrite
placeholders

JohnnyDoorn:
The error handling was not working correctly, so I think everything should work now:
-  In the "Parameter Estimates" and "Modification Indices" tables the abbreviations are still base64 encoded 
Don told me this is going to be fixed in general so Im going to leave this for now
- The "Other Fit Indices" table no longer appears to have the "Expected Cross-Validation Index" column
Lavaan does not provide this value (anymore?)

I also like the powered by lavaan. I think it should be the case that other analyses copy this style and make the R-packages more prominent, rather than the other way aorund

TimKDJ:
This is merged in https://github.com/jasp-stats/jasp-desktop/pull/3692 (I rebased your branch and made some small (encoding) fixes)

PR #3662, opened 2019-11-16
Corrieresults

AlexanderLyNL:
- Remove correlationBayesianPairs.qml
- Still need to fix up the Robustness plot dots colours
- I still need to add bcorPackage as a package on my own repository instead of as a separate file
- The latter will help us separate the name space of the old functions and the new packag functions

Wishlist:
- General matrix plot for descriptives, correlation, correlationbayesian
- When the matrix plot is big, the waity circle is hard to see @tim
- Match up functions between correlation and correlationbayesian
- Match up functions between correlationbayesian and ttests
- I need to make a wrapper for t-test and do some computations there to simplify stuff
- Perhaps use data.table for ggplot2 stuff. I suspect some speed issues there
- A separate plotRobustness functionality in JASPgraphs. My poor attempt is plotRobustnessCor, which is covfefe sad
- Make qml mapper or item dictionary to map the arguments of correlationbayesianOld and correlationbayesianpairs to correlationbayesian


Resources/Summary Statistics/qml/SummaryStatsCorrelationBayesianPairs.qml
```diff
@@ -24,58 +24,71 @@ import JASP.Widgets 1.0
 
 Form 
 {
-	usesJaspResults: false
+    usesJaspResults: true
 
-	IntegerField { name: "sampleSize"; label: qsTr("Sample size") }
```
TimKDJ, line 29: Can you keep the names as is in the summary stats module? That way we don't need backwards compatibility stuff everywhere

Resources/Regression/qml/CorrelationBayesianOld.qml
```diff
@@ -0,0 +1,95 @@
+//
```
TimKDJ: You can remove this entire file. Github is very good at preserving changes ;)

Resources/Regression/qml/CorrelationBayesian.qml
```diff
@@ -23,73 +23,118 @@ import JASP.Widgets 1.0
...
-	}
 
-	Group
-	{
-		title: qsTr("Correlation Coefficient")
-		CheckBox { name: "pearson";			label: qsTr("Pearson's rho"); checked: true	}
-		CheckBox { name: "spearman";		label: qsTr("Spearman"); debug: true		}
-		CheckBox { name: "kendallsTauB";	label: qsTr("Kendall's tau-b")				}
-	}
+    VariablesForm
```
TimKDJ, line 27: Please use tabs for .qml files

JASP-Engine/JASP/R/bCorPackage.R
```diff
@@ -0,0 +1,2236 @@
+# 1. Main test function --------
+bcor.test <- function(x, y, alternative=c("two.sided", "less", "greater"),
```
TimKDJ, line 2: I think you showed that this already a functional R package? If you host this on your github as a package then we can just do `install_github()` and put it in required packages. That way you keep complete control over your own package (i.e., you don't need to make pull requests to work on the package)

(I won't look at the package files further, your package, your code style ;))

JASP-Engine/JASP/R/bfCommon.R
```diff
@@ -0,0 +1,131 @@
...
+  if (!options[["reportBayesFactors"]]) 
+    itemNames <- setdiff(itemNames, "bf")
+  
+  if (!options[["ci"]])
+    itemNames <- setdiff(itemNames, c("lowerCi", "upperCi"))
+  
+  return(itemNames)
+}
+
+.bCorRowNames <- function(options, itemNames, method=c("pearson", "kendall", "spearman")) {
```
TimKDJ: Quite a few of these are very specific to correlation. Maybe move to your commoncorrelation file?

JASP-Engine/JASP/R/bfCommon.R
```diff
@@ -0,0 +1,131 @@
...
+      bfTitle <- "BF\u2080\u208A"
+    } else {
+      bfTitle <- "BF\u2080\u208B"
+    }
+  }
+  return(bfTitle)
+}
+
+
+.bSelectItems <- function(options) {
```
TimKDJ, line 46: I would have no idea what this does. A little more descriptive name/some documentation would help

JASP-Engine/JASP/R/commoncor.R
```diff
@@ -0,0 +1,320 @@
+.getPairsLength <- function(options) {
+  pairs <- options[["pairs"]]
+  nPairs <- length(pairs)
+  
+  if (nPairs!=0) {
+    lastPair <- pairs[[nPairs]]
```
TimKDJ, line 6: Any pair can be incomplete

JASP-Engine/JASP/R/commoncor.R
```diff
@@ -0,0 +1,320 @@
...
+    x <- rank(x)
+    y <- rank(y)
+  }
+  
+  # TODO(Alexander): Ask Simon: Why does it return an error if I add drawAxes = drawAxes? 
+  # 
+  .plotScatter(xVar = x, yVar = y, xBreaks = xBreaks, yBreaks = yBreaks, xName = xName, yName = yName) 
+}
+
+# # TODO(Alexander): Remove this, since it's stolen from Simon
```
TimKDJ: If this is not going to be used, remove (github will keep its copy of old code)

JASP-Engine/JASP/R/correlationbayesianOld.R
```diff
@@ -0,0 +1,2864 @@
+#
```
TimKDJ: Remove the entire file

JASP-Engine/JASP/R/summarystatscorrelationbayesianpairs.R
```diff
@@ -8,1287 +8,244 @@
 #
 # This program is distributed in the hope that it will be useful,
 # but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# MERCHANTABILITY or FITNESS FOR A P          ARTICULAR PURPOSE.  See the
```
TimKDJ: Don't think this was intended

AlexanderLyNL:
See the points in the last commit, and in particular:

Screen Shot 2019-11-18 at 10 48 15

JASP-Engine/JASP/R/correlationbayesian.R
```diff
@@ -15,2850 +15,1312 @@
...
-  # respectively, Christian Robert (2015). As the proposal is defined on Fisher transform atanh(rho), 
-  # we have a Jocobian. The Jacobian in the upper part is (1-rhoCurrent^2)^(-1) and in the lower part
-  # is (1-rhoProp^2)^(-1).
-  # Note that since prior(rhoProp) \propto (1-rhoProp^2)^(alpha-1), we can drop the one due 
-  # to the Jacobian term of the denominator ending up in the numerator.
-  # 
-  # For the likelihood we use Jeffreys's approximation 
-  #   rho \propto (1-rho^2)^((n-1)/2)*(1-rho*r)^((3-2*n)/2)
-  # 
-  # The log prior is (alpha-1)*log(1-rho^2) # where
```
TimKDJ, line 1035: Noticed quite a few of these (`&`) -- considering you don't want to do elementwise comparisons, use `&&`.
Same for `|` in your if-statements.

JASP-Engine/JASP/R/correlationbayesian.R
```diff
@@ -15,2850 +15,1312 @@
...
+    if (is.null(dataset)) 
+      dataset <- .corBayesReadData(dataset, options)
+  }
 
-.stretchedBeta <- function(rho, alpha, beta) {
-	result <- 1/2*dbeta((rho+1)/2, alpha, beta)
-	return(result)
+  # 3. Table: Get, compute and fill -----------
+  #
+  allBfObjects <- .makeTableCorBayes("jaspResults"=jaspResults, "allBfObjects"=allBfObjects,
```
TimKDJ: It feels unnatural that `makeTableCorBayes` doesn't return.. a table?

Is there a reason you're creating one massive state object? Because you don't need to do that anymore (you can create as many state objects as you'd like)... It's great because then you can compartmentalize your code and don't couple everything so tightly.

JASP-Engine/JASP/R/correlationbayesian.R
```diff
@@ -15,2850 +15,1312 @@
...
+
+  # d. Compute ----
+  #
+  allBfObjects <- .computeCorBayes("allBfObjects"=allBfObjects, "dataset"=dataset,
+                                   "options"=options, "computePlots"=FALSE,
+                                   "purpose"="matrix", "ready"=ready)
+
+  # e. Fill ----
+  #
+  allBfObjects <- .fillCorBayesTable("table"=corBayesTable, "options"=options,
```
TimKDJ: Same here, what does the filling of a table have to do with this object?
Here I'd expect a table to be returned or nothing (because of pass by reference of the table)

TimKDJ (changes requested):
Hey there it is! :)

Didn't encounter a single crash while testing -- nice. Did note the following though:
- [x] To be consistent can you use long dashes in the matrix table just like it used to be + how frequentist correlation is now (`"\u2014"`)
- [x] For the flag supported correlation you should do a set footnote without a message to prevent additional footnotes from showing up under the table (see https://github.com/jasp-stats/INTERNAL-jasp/issues/562):
corRewriteFootnote
- [x] "Density" label should be smaller in bayesian correlation matrix. It's disproportionally large in comparison to the legend and almost overlaps:
corRewriteAxisLabel
- [x] I think it might bemore sensible to set the source of the individual pairs to be the variables entered in the main analysis (otherwise people have to search multiple times): `AvailableVariablesList { name: "allVariablesList2" source: "variables" }`
- [x] Messages in prior posterior plot look a lot different than robustness and sequential analysis:
corRewriteErrMsg
- [x] In the scatter plot it would be nice if we could get the variables as labels on the axes again (like it used to be), so the plots can be used without needing the var1-var2 context of the JASP container
- [x] Color legend of robustness plot is not correct (white/black in legend and red/gray in plot):
corRewriteRobustness
- [ ] The analysis is slower, but not that big of a deal (e.g., switching from missing values type used to be recalculated much quicker). Maybe worthwhile to profile it at some point.

PR #3664, opened 2019-11-19
Major bain changes

koenderks:
Okay, so what I have done here also touches upon https://github.com/jasp-stats/INTERNAL-jasp/issues/575 on figuring out the correctness of bain credible intervals.

- I have rewritten all the t-tests to make them more efficient and fitting in the current bain code format.
- All (adjusted) means, N, standard errors and credible intervals are directly read from bain using their credible interval method. This results in differing standard errors in ANOVA (which we don't show btw) than when calculated for individual groups, due to the pooled standard deviation. Bain also returns slightly different credible intervals for t-tests than our Bayesian t-tests, but returns the same standard deviation.
- Only where appropriate (descriptives tables) standard deviations are shown. These are always calculated from the data. 
- I have changed all non-sensical option names in the qml files to fitting names

I might update this pull request depending on what herbert says on the credible intervals, but I think this is probably a step in the right way and easy to adjust like this.


koenderks:
@TimKDJ Herbert will visit next week somewhere, so it is best to leave this PR open until then. Most likely this is the way that it will be implemented, but I will check with Herbert just to be sure.

Might still adjust the adjusted means and the credible intervals (since they differ slightly from JASP)

koenderks:
I have rewritten the state system in the t-test so that it works most optimal now. I will meet with Herbert this Wednesday and will come back on the credible intervals after that.

koenderks:
@TimKDJ Can you take a look at this (maybe tomorrow)? Herbert is urgently asking for a new Bain nightly. I have implemented the latest `bain` package from CRAN in the last commit, so the module should be completely finished now.

@FransMeerhoff This pull request does require the latest version of `bain` from CRAN. Can you add this (the latest is version 0.2.3) to the R framework?

AlexanderLyNL:
@JorisGoosen also needs to know this, so he can update the flatpak build. 

"This pull request does require the latest version of bain from CRAN. Can you add this (the latest is version 0.2.3) to the R framework?"

koenderks:
Note: unit tests fail because latest `bain` package is missing

Resources/BAIN/qml/BainAncovaBayesian.qml
```diff
@@ -28,21 +28,20 @@ Form
...
-			suggestedColumns: ["scale"]
+			allowedColumns: ["scale"]
 		}
 
 		AssignedVariablesList {
 			name: "fixedFactors"
 			title: qsTr("Fixed Factors")
 			singleVariable: true
-			suggestedColumns: ["ordinal", "nominal"]
+			allowedColumns: ["ordinal", "nominal"]
```
TimKDJ: This will now disallow `nominalText` from being entered, which is a little weird for an ANOVA.

The suggested columns allow more than they imply, so you'll have to be careful with this conversion.
From the docs:
```
If `allowedColumns` is empty and `suggestedColumns` is specified, then the following rules apply:
    * `scale` allows Nominal Integer and Ordinal columns
    * `nominal` allows all Nominal columns (Integer or String), and Ordinal column
```

Resources/BAIN/qml/BainTTestBayesianIndependentSamples.qml
```diff
@@ -47,11 +48,11 @@ Form
 			title: qsTr("Hypothesis Test")
 			name: "hypothesis"
 
-			RadioButton { text: qsTr("Equal vs. not equal")                     ; name: "groupsNotEqual" ; checked: true}
-			RadioButton { text: qsTr("Equal vs. bigger")                        ; name: "groupOneGreater" }
-			RadioButton { text: qsTr("Equal vs. smaller")                       ; name: "groupTwoGreater" }
-			RadioButton { text: qsTr("Bigger vs. smaller")                      ; name: "_4type" }
-			RadioButton { text: qsTr("Equal vs. bigger vs. smaller")            ; name: "allTypes" }
+			RadioButton { text: qsTr("Equal vs. not equal")                     ; name: "equalNotEqual" ; checked: true}
```
TimKDJ, line 51: Changing names creates a bit of a problem for people that try to open older .jasp files. They will be greeted by a big red error box, because `groupsNotEqual` does not exist anymore. And this goes for each and every one of the components you rename.

So in general this is not recommended. However, @JorisGoosen is currently working on a workaround for this exact situation. Basically you'll be able to include an upgrade.json (or something) where you specify that `groupsNotEqual` has now become `equalNotEqual`. If you wish to keep these name changes, you'll need to make sure you create such a file before we release 0.12

JASP-Engine/JASP/R/bainttestbayesianindependentsamples.R
```diff
@@ -24,335 +24,357 @@ BainTTestBayesianIndependentSamples <- function(jaspResults, dataset, options, .
...
+	all.variables 									<- unlist(options$variables)
+	grouping   										<- options$groupingVariable
+	read.variables 									<- c(all.variables, grouping)
+	if (options[["groupingVariable"]] == "")
+		grouping <- NULL
+
+	if (is.null(dataset)) {
+						trydata                	<- .readDataSetToEnd(columns.as.numeric=all.variables)
+						missingValuesIndicator 	<- .unv(names(which(apply(trydata, 2, function(x) { any(is.na(x))} ))))
+            dataset 								<- .readDataSetToEnd(columns.as.numeric=all.variables, columns.as.factor=grouping, exclude.na.listwise=read.variables)
```
TimKDJ: Outlining gone wrong? ;)

JASP-Engine/JASP/R/bainttestbayesianindependentsamples.R
```diff
@@ -24,335 +24,357 @@ BainTTestBayesianIndependentSamples <- function(jaspResults, dataset, options, .
...
-
-				} else {
-						n <- length(groupDataOm)
-						row <- data.frame(v = variable, group = "", N = n, mean = "", sd = ".", se = "", lowerCI = "", upperCI = "")
-				}
-				descriptivesTable$addRows(row)
-			}	
-  	}
+		
+		bainAnalysis <- .bainTwoSampleState(variable, options, dataset, bainContainer)
```
TimKDJ, line 320: This might be a `try-error` right? It's handled as such in `bainIndependentSamplesResultsTable()` anyway

JASP-Engine/JASP/R/bainttestbayesianindependentsamples.R
```diff
@@ -24,335 +24,357 @@ BainTTestBayesianIndependentSamples <- function(jaspResults, dataset, options, .
...
-	readList <- list()
-  readList[["dataset"]] <- dataset
-  readList[["missingValuesIndicator"]] <- missingValuesIndicator
-	return(readList)
+	if (!ready || bainContainer$getError())
+		return()
+
+	for (variable in options[["variables"]]) {
+
+		bainAnalysis <- .bainTwoSampleState(variable, options, dataset, bainContainer)
```
TimKDJ: Also here; this might be an error object

JASP-Engine/JASP/R/bainttestbayesianindependentsamples.R
```diff
@@ -24,335 +24,357 @@ BainTTestBayesianIndependentSamples <- function(jaspResults, dataset, options, .
...
-			g1 <- levels[1]
-			g2 <- levels[2]
-		}
+	overTitle <- paste0(100 * options[["credibleInterval"]], "% Credible Interval")
+	descriptivesTable$addColumnInfo(name="lowerCI",              title = "Lower", type="number", overtitle = overTitle)
+	descriptivesTable$addColumnInfo(name="upperCI",              title = "Upper", type="number", overtitle = overTitle)
+	
+	bainContainer[["descriptivesTable"]] <- descriptivesTable
+
+	if (!ready || bainContainer$getError())
```
TimKDJ: I can't actually find a `setError()` on the container. Is this done in some common file?
If you don't do this anywhere then `bainContainer$getError()` will always be `FALSE`

JASP-Engine/JASP/R/bainttestbayesianindependentsamples.R
```diff
@@ -24,335 +24,357 @@ BainTTestBayesianIndependentSamples <- function(jaspResults, dataset, options, .
...
+	if (options[["groupingVariable"]] == "")
+		grouping <- NULL
+
+	if (is.null(dataset)) {
+						trydata                	<- .readDataSetToEnd(columns.as.numeric=all.variables)
+						missingValuesIndicator 	<- .unv(names(which(apply(trydata, 2, function(x) { any(is.na(x))} ))))
+            dataset 								<- .readDataSetToEnd(columns.as.numeric=all.variables, columns.as.factor=grouping, exclude.na.listwise=read.variables)
+    }
+	
+	.hasErrors(dataset, type="factorLevels",
```
TimKDJ: I think it's better to be in line with the way other analyses do this: error checking and reading data are separate steps. Which imo makes sense because `readDataBainTwoSample` does not convey `checkErrors`

JASP-Engine/JASP/R/bainttestbayesianindependentsamples.R
```diff
@@ -24,335 +24,357 @@ BainTTestBayesianIndependentSamples <- function(jaspResults, dataset, options, .
...
 
-	if (isTryError(p)) {
-		bainTable$addRows(list(Variable=variable), rowNames=variable)
-		bainTable$addFootnote(message=paste0("Results not computed: ", .extractErrorMessage(p)), colNames="Variable", rowNames=variable)
-		progressbarTick()
-		next
-	}
+		if (isTryError(bainAnalysis)) {
+			bainTable$addRows(list(Variable=variable), rowNames=variable)
+			bainTable$addFootnote(message=paste0("Results not computed: ", .extractErrorMessage(p)), colNames="Variable", rowNames=variable)
```
TimKDJ: `.extractErrorMessage(bainAnalysis)`

JASP-Tests/R/tests/testthat/test-bainanovabayesian.R
```diff
@@ -43,13 +43,13 @@ test_that("Descriptives Plot matches", {
 test_that("Descriptive Statistics table results match", {
   table <- results[["results"]][["bainContainer"]][["collection"]][["bainContainer_descriptivesTable"]][["data"]]
   expect_equal_tables(table,
-                      list(60, 48.2744318767126, 49.8666666666667, 7.46169123980563, 0.963300196876966,
+                      list(60, 48.2744318767126, 49.8666666666667, 7.46169123980563, 0.812379616418173,
```
TimKDJ, line 46: So this is the result of the change to credible interval calculations?

TimKDJ (changes requested):
Sorry for taking a bit there.

Some minor comments; looks good in general but the error handling might require a bit of attention.
I only made the comments in one analysis, but they are applicable to all.

Also, when testing I noticed the scaling of the axis of the descriptives plot does not seem to work super well:
(cut off)
Screenshot 2019-12-17 at 13 41 58
(doesn't stretch far enough)
Screenshot 2019-12-17 at 13 44 54

JASP-Tests/R/tests/testthat/test-bainanovabayesian.R
koenderks: No, this is the result from removing columns for individual variables and showing only the difference between the pairs.

Resources/BAIN/qml/BainAncovaBayesian.qml
koenderks: Added "nominalText" to the allowedcolumns for anova, ancova and independent samples t-test

Resources/BAIN/qml/BainTTestBayesianIndependentSamples.qml
koenderks: That is too bad, because some bain analyses won't work anyway anymore since bain changed its model specification (anova, ancova, regression) to not include a `.`  anymore. This will break all of these .jasp file anyway...

JASP-Engine/JASP/R/bainttestbayesianindependentsamples.R
koenderks: Not really? I don't see this. Changed it anyway.

Resources/BAIN/qml/BainTTestBayesianIndependentSamples.qml
TimKDJ: Well I mean you could do a simple conversion in the code between option names.
But like I said there will be a workaround available, you'll just need to make a mental note to implement that ;)

JASP-Engine/JASP/R/bainttestbayesianindependentsamples.R
koenderks: I'm not sure what you mean? This is a try-error right?

koenderks: I see what you mean now!

TimKDJ: In `.bainTwoSampleState()` you put a `try()` around the main calculation. So that calculation can obviously fail (and you handle this possible failure well in `bainIndependentSamplesResultsTable`), however in this function you don't handle this possible failure. If you call `bainSummary <- summary(bainAnalysis, ci = options[["credibleInterval"]])` on an object of class `try-error` you'll get bad things.

Resources/BAIN/qml/BainTTestBayesianIndependentSamples.qml
koenderks: Yes, might be nice to have some sort of template for that when the time comes

JASP-Engine/JASP/R/bainttestbayesianindependentsamples.R
koenderks: Implemented this now

JASP-Engine/JASP/R/bainttestbayesianindependentsamples.R
koenderks: fixed

JASP-Engine/JASP/R/bainttestbayesianindependentsamples.R
koenderks: Implemented

JASP-Engine/JASP/R/bainttestbayesianindependentsamples.R
koenderks: Implemented

JASP-Engine/JASP/R/bainttestbayesianindependentsamples.R
koenderks: Removed them in the t-test files

koenderks:
Implemented all this feedback in the latest commit

koenderks:
Ugh okay, I see that I need to change the unit test plots for the descriptive plots (with the better axis). I'll do it today.

koenderks:
Cannot create the unit tests because jasptools does not work, see https://github.com/jasp-stats/INTERNAL-jasp/issues/629.

koenderks:
Okay, I hope this fixes the unit tests

PR #3670, opened 2019-11-25
FormulaField for math expressions

vandenman:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/578

This PR introduces the widget `FormulaField` in QML where people can type math expressions in addition to single numbers.

The current implementation is somewhat suboptimal because the code is parsed multiple times, both in C++ and in R, but it's the best we can at the moment.

On the R side, I introduced the functions `.parseRCodeInOptions` which parses the code if it is valid and `.RCodeInOptionsIsOk` to test if the R code in the options is valid. These two functions may seem a little bit redundant, but this way we can easily change the C++/ QML API and only have to change two functions in R.

As a minimal example, FormulaField is now used in the Bayesian binomial test.

### DIFF
--- a/JASP-Desktop/components/JASP/Controls/FormulaField.qml
+++ b/JASP-Desktop/components/JASP/Controls/FormulaField.qml
@@ -23,11 +23,11 @@ import JASP 1.0
 
 TextField
 {
-	id: formulaField
-	property double defaultValue:		0
+					id:					formulaField
+					defaultValue:		"0"
 	property double	min:				-Infinity
 	property double	max:				Infinity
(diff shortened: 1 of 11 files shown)


vandenman:
Ok.. for some unknown reason the build stopped. Can somebody trigger a restart? Also, @JorisGoosen did you change something to the rights of my account? Because I'm no longer allowed to restart travis.

JorisGoosen:
It seems the build started again already?

Also, I did not change anything to your rights.

JorisGoosen (approved):
Looks good!

PR #3673, opened 2019-11-28
Changed size calculation of svg plots

vandenman:
So we used (see https://github.com/jasp-stats/jasp-desktop/blob/ad4d3615c2db2279d2d8d86b8d1dd8898b84d244/JASP-R-Interface/jaspResults/R/writeImage.R):

```r
pngMultip <- ppi / 96
ggplot2::ggsave(
		...
    	width     = width  * pngMultip,
    	height    = height * pngMultip,
```

Usually, `inch * ppi = pixels` or `inch = pixels/ ppi`. Width used to be in pixels: `width(px) * ppi / 96`. So dividing by ppi gives: `width (inches) = width(px) / 96`, which is what this PR implements.

My flatpak jasp doesn't work right now so I cannot compare this to a previous jasp version.. :(

Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/598

### DIFF
--- a/JASP-Engine/JASP/R/common.R
+++ b/JASP-Engine/JASP/R/common.R
@@ -2263,9 +2263,11 @@ openGrDevice <- function(...) {
     ppi             <- .fromRCPP(".ppi")
   }
   
-  # convert width & height from pixels to inches. ppi = pixels per inch. 96 is a magic number inherited from the past.
-  width  <- width  / 96
-  height <- height / 96
+  # convert width & height from pixels to inches. ppi = pixels per inch. 72 is a magic number inherited from the past.
+  # originally, this number was 96 but svglite scales this by (72/96 = 0.75). 0.75 * 96 = 72.
+  # for reference see https://cran.r-project.org/web/packages/svglite/vignettes/scaling.html
+  width  <- width  / 72
+  height <- height / 72
   image <- list()
   
   # TRUE if called from analysis, FALSE if called from editImage
--- a/JASP-Engine/JASP/R/common.R
+++ b/JASP-Engine/JASP/R/common.R
@@ -2263,9 +2263,9 @@ openGrDevice <- function(...) {
     ppi             <- .fromRCPP(".ppi")
   }
   
-  # convert width & height from pixels to inches. ppi = pixels per inch. 4 is a magic number.
-  width  <- width  / ppi * 4
-  height <- height / ppi * 4
+  # convert width & height from pixels to inches. ppi = pixels per inch. 96 is a magic number inherited from the past.
+  width  <- width  / 96
+  height <- height / 96
   image <- list()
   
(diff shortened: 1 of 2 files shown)


TimKDJ:
Now the labels became quite big:
your change <-> 0.11
svgsize

Are we sure we don't need the ppi, considering we're converting from pixels to inches and then back to pixels (in JavaScript)?

vandenman:
That looks like the descriptives plot from Anova, which is not using jaspResults yet. The fix is only implemented for plots in jaspResults, although I don't see the need to do that since we're getting rid of all that code once everything is translated to jaspResults.

TimKDJ:
It's the Bayesian ANOVA though

vandenman:
ah that changes things.

vandenman:
ok so I tweaked the scaling a bit based on https://cran.r-project.org/web/packages/svglite/vignettes/scaling.html

The scaling is now 1.5 times smaller than before.

> Are we sure we don't need the ppi, considering we're converting from pixels to inches and then back to pixels (in JavaScript)?

Well, I'm definitely not sure about that. But then the question is rather, why did we in the past use the ppi to determine the size of an image that should be 300 x 300 pixels? I guess that we secretly did some screen-specific conversions on the image size. However, using the standard conversion from inches to pixels simply means that ppi drops out. Also note that inches is format for image size that is (or should be) ppi independent (i.e., for all screens, we should render images using the same width and height in inches because the web engine should handle the rest). I still cannot test this in the latest release btw, but I'll ask Joris for some help with that.

JASP-Engine/JASP/R/common.R
```diff
@@ -2263,9 +2263,9 @@ openGrDevice <- function(...) {
     ppi             <- .fromRCPP(".ppi")
   }
   
-  # convert width & height from pixels to inches. ppi = pixels per inch. 4 is a magic number.
-  width  <- width  / ppi * 4
-  height <- height / ppi * 4
+  # convert width & height from pixels to inches. ppi = pixels per inch. 96 is a magic number inherited from the past.
+  width  <- width  / 96
+  height <- height / 96
```
TimKDJ: Can you change this to 72 as well?

TimKDJ:
So it looks sorta okay again right now. The letters seem a bit more bold and the dots are quite a bit more prominent

new <-> old
Screenshot 2019-12-06 at 09 59 32

I guess at this point in time it becomes personal preference? What do you think

vandenman:
Another avenue to explore is adapting the pointsize argument of `svglite`. That should only affect the fonts, but I don't know why this argument would work differently for `svglite` compared to all other graphics devices (the default is always 12). Anyway I hope this solution is ok for now.

TimKDJ:
Well it at least makes the plots okay again ;)
Maybe once you can build you can mess around with it a little

PR #3678, opened 2019-12-02
remove plot environments from ggplot2 objects

vandenman:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/601.

If a plot fails, the plot environment is kept (and saved) as a fallback. An error is shown on a failing plot, but only in developer mode.

### DIFF
--- a/JASP-R-Interface/jaspResults/R/writeImage.R
+++ b/JASP-R-Interface/jaspResults/R/writeImage.R
@@ -98,12 +98,14 @@ writeImageJaspResults <- function(width=320, height=320, plot, obj=TRUE, relativ
           limitsize = FALSE # only necessary if users make the plot ginormous.
         )
 
-        image[["error"]] <- paste(
-          "ggplot could not be rendered without the enclosing environment.",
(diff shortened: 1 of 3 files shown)


TimKDJ:
So what the is the gain in terms of the size of our data library?

vandenman:
I wasn't able to test this completely because some data library files crashed due to encoding problems (I could try again now though). The difference will vary enormously from analysis to analysis though.

I'll run it again and post the results here (or I'll open more issues if things fail).

TimKDJ:
Yeah I think it needs to be clear that this is useful to implement, because we'll be inconveniencing people that want to develop for us

JorisGoosen:
So regardless if we add this to JASP later on, I would suggest we do not add it before release.

vandenman:
> So regardless if we add this to JASP later on, I would suggest we do not add it before release.

Definitely not for this release! And we should only do this if there is a substantial gain. I'll close this for now though to make the PR list a bit cleaner

PR #3686, opened 2019-12-06
Bayes multinomial

ASarafoglou:
The analysis crashed for data sets in the short format (i.e., one factor level per row). I fixed this my deleting some redundant functions in the bayesian multinomial (.checkCountsMultinomial()). Instead, I merged the error checks for the frequentist and bayesian multinomial test which also checks the counts in the multinomial dataset.


TimKDJ:
I noticed this quite a while ago, but forgot to fix it. Thanks for this!

PR #3697, opened 2019-12-18
QuillJS editor

akashrajkn:
Replaces jasp note-editor with [quill js](https://quilljs.com/) editor. I tested it only on my system - so there could be bugs.

It is compatible with the existing jasp files. But sometimes the _markdown-to-delta conversion_ (delta format is similar to json that quilljs uses) has trouble figuring out breaks/newlines in the existing notes.

Fixes jasp-stats/INTERNAL-jasp#340, jasp-stats/INTERNAL-jasp#360, jasp-stats/jasp-issues#293, jasp-stats/jasp-issues#249, jasp-stats/jasp-issues#36

Example below:

quill


boutinb:
Hi @akashrajkn Could you solve the last problems this week?

akashrajkn:
Yes. I'll be in the office this Wednesday.

PR #3710, opened 2020-01-04
Add constants to LDA coefficients table

koenderks:
As requested in https://github.com/jasp-stats/jasp-issues/issues/575


koenderks:
Weirdly enough, the unit tests do not fail on my machine...

boutinb:
@koenderks Do we use the right package?

koenderks:
I believe so. The unit tests only fails for 4 plots (in which the axis is different)? Could it be a different version of JASPgraphs?

boutinb:
@koenderks Do you know why Travis still fails

koenderks:
No

TimKDJ:
If you don't change the .svg files then there is no problem for me.

I use these versions, do they differ for you?
> packageVersion('JASPgraphs')
[1] ‘0.4.3’
> packageVersion('jasptools')
[1] ‘0.8.4’
> packageVersion('jaspResults')
[1] ‘1.12’

PR #3714, opened 2020-01-06
[WIP] use grDevices::svg to deal with base plots that open multiple pages

vandenman:
Switch svg backend to `grDevices`. Note that this does not affect ggplot2 plots because those are saved via `ggsave` which internally uses `svglite`. 

Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/618 (in my opinion) and
https://github.com/jasp-stats/INTERNAL-jasp/issues/620 (no error)

### DIFF
--- a/JASP-Engine/JASP/R/common.R
+++ b/JASP-Engine/JASP/R/common.R
@@ -2246,10 +2246,16 @@ as.list.footnotes <- function(footnotes) {
 }
 
 openGrDevice <- function(...) {
-  if (jaspResultsCalledFromJasp())
-    svglite::svglite(...)
-  else
+  if (jaspResultsCalledFromJasp()) {
+    dots <- list(...)
+    if ("file" %in% names(dots)) {
+      dots[["filename"]] <- dots[["file"]]
+      dots[["file"]]     <- NULL
+    }
+    do.call(grDevices::svg, dots)
+  } else {
     grDevices::png(..., units="in", res = 72, type = ifelse(Sys.info()["sysname"] == "Darwin", "quartz", "cairo"))
+  }
 }
 
 .writeImage <- function(width=320, height=320, plot, obj = TRUE, relativePathsvg = NULL,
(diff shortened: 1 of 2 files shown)


TimKDJ:
From what I recall the ggplot's had the same issue, but it's just not as "in your face" because at the moment there are no huge round elements in any of them. I can see this becoming an issue though (and to be fair, I think it is already an issue, because our plots are not as pretty as they were).

Isn't it easier to just switch back to png?

vandenman:
Well for many reasons, svgs are just nicer to work with (e.g., easier to edit outside of JASP, don't need to resize all plots if a user moves jasp from one window to another). We can always switch back to png as a last resort, but let's just see if we can make this work. I just got flatpak working again and had a small discussion with @JorisGoosen and @vankesteren on this so I'll do some more testing.

TimKDJ:
> easier to edit outside of JASP

I'm not saying we shouldn't offer the possibility to export svg's though (we did and we still do). Just talking about the format that we show within JASP

> don't need to resize all plots if a user moves jasp from one window to another

Yeah it's nice if we don't have to do this, but we did already have the functionality so...

> We can always switch back to png as a last resort, but let's just see if we can make this work.

I mean if you can make sure the plots are just as kick ass as they've always been, then it's a nice improvement. But we shouldn't settle for anything less, because I don't see an earth shattering argument in favour of making the switch.

vankesteren:
I agree with Tim

vandenman:
> I'm not saying we shouldn't offer the possibility to export svg's though (we did and we still do).

JASP 0.11.1 cannot export to svg, I guess the same goes for older versions. But I agree that if we want to give the option to export to svg, we should do it properly. Then again if we do it properly anyway, we might as well make it the default because of the benefits mentioned above. 
Altogether I see three reasonable options:

1. use png as default and do not offer svg (easiest way out).
2. use png as default and offer poorly tested svg.
3. use svg as default and properly test it/ fix it.

I'd prefer 3, then 1, then 2.

EJWagenmakers:
I'm in favor of any thing you guys agree on, just as long as the quality of the plots is not compromised, because that is key.

PR #3715, opened 2020-01-06
The sync file is sometimes lost

boutinb:


JASP-Desktop/widgets/filemenu/filemenu.cpp
```diff
@@ -408,13 +408,33 @@ void FileMenu::setSyncRequest(const QString& path)
...
-	clearSyncData();
+	bool exist = false;
 
-	return false;
+	if (QFileInfo::exists(path))
+	{
+		int trials = 0;
+		while (Utils::getFileSize(path.toStdString()) == 0 && trials < 5)
+		{
+			// sometimes it is temporarly empty...
```
JorisGoosen, line 421: If you are going to add this weird timer and retry thing, can you make sure we are not wasting time while loading jasp-files?

boutinb: Yes. I split the 2 if statements in 2: first it checks whether the file uberhaupt exists, then it checks whether the file has data, and it loops only if the first condition is met.
I will add comments.

PR #3717, opened 2020-01-07
Make TableView smoothier

boutinb:
smoothie

### DIFF
--- a/JASP-Desktop/components/JASP/Controls/TableView.qml
+++ b/JASP-Desktop/components/JASP/Controls/TableView.qml
@@ -126,6 +126,7 @@ JASPControl
 				model:					null
 				itemHorizontalPadding:	0
 				itemVerticalPadding:	8 * preferencesModel.uiScale
+				cacheItems:				false
 
 				viewportX: myFlickable.visibleArea.xPosition * width
 				viewportY: myFlickable.visibleArea.yPosition * height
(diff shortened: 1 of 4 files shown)


JASP-Desktop/components/JASP/Controls/TableView.qml
```diff
@@ -212,42 +183,16 @@ JASPControl
 						anchors.topMargin:		- 8 * preferencesModel.uiScale
 						fieldHeight:			parent.height + 16 * preferencesModel.uiScale
 						fieldWidth:				parent.width
-						visible:				false
+						visible:				true
 						useExternalBorder:		false
 						value:					itemText
 						useLastValidValue:		false
+						focus:					true
```
JorisGoosen: why this?
it's a delegate and now they all get the focus set...

JASP-Desktop/qquick/datasetview.cpp
```diff
@@ -78,10 +78,15 @@ void DataSetView::setRolenames()
...
 }
 
-void DataSetView::modelDataChanged(const QModelIndex &, const QModelIndex &, const QVector<int> &)
+void DataSetView::modelDataChanged(const QModelIndex &index, const QModelIndex &, const QVector<int> &)
 {
-	modelAboutToBeReset();
-	calculateCellSizes();
+	QSizeF calcedSize = getTextSize(_model->data(index).toString());
+
+	if (_cellSizes[size_t(index.column())].width() < calcedSize.width())
```
JorisGoosen: Now you are only comparing the current size of the cell with the new size of the cell. But the actual comparison should be made on the basis of the width of the whole column. 

Like: `_model->headerData(col, Qt::Orientation::Horizontal, _roleNameToRole["maxColString"]).toString();`
If you put that in a function you could reuse that easily, and then also use it at line 147

JASP-Desktop/qquick/datasetview.cpp
```diff
@@ -137,14 +142,11 @@ void DataSetView::calculateCellSizes()
 	_colXPositions.resize(_model->columnCount());
 	_cellTextItems.clear();
 
-	_metricsFont = QFontMetricsF(_font);
-
 	for(int col=0; col<_model->columnCount(); col++)
 	{
 		QString text = _model->headerData(col, Qt::Orientation::Horizontal, _roleNameToRole["maxColString"]).toString();
```
JorisGoosen: Here you can use that function ^^

JASP-Desktop/qquick/datasetview.cpp
```diff
@@ -539,7 +541,10 @@ void DataSetView::storeTextItem(int row, int col, bool cleanUp)
 
 	textItem->item->setVisible(false);
 
-	_textItemStorage.push(textItem);
+	if (_cacheItems)
```
JorisGoosen:
```
if (_cacheItems)	_textItemStorage.push(textItem);
else				delete textItem;
```
? ;)

JASP-Desktop/qquick/datasetview.cpp
boutinb: OK?

PR #3720, opened 2020-01-08
Fix two bugs with engine crashes

JorisGoosen:
- First was that exit in "Exit" or "Report" didnt actually exit...
- Second was that an analysiscrash no longer recovered properly


JASP-Engine/JASP/R/descriptives.R
```diff
@@ -16,6 +16,7 @@
 #
 
 Descriptives <- function(jaspResults, dataset, options) {
+.crashPlease()
```
vandenman: Travis fails because you left this in... :P

JorisGoosen: Pretty good reason :p

JorisGoosen:
@boutinb could you have a look at this please?

PR #3722, opened 2020-01-08
FormulaField does not work in a ListView

boutinb:
The error message should stay inside the ListView


JASP-Desktop/analysis/analysisform.cpp
```diff
@@ -574,19 +587,26 @@ void AnalysisForm::addControlError(QQuickItem* control, QString message, bool te
 			_controlErrorMessageCache.append(controlErrorMessageItem);
 		}
 
+		QQuickItem* container = this;
```
JorisGoosen, line 600: `QQuickItem * container = this;` is prettier right? No problem though :p

JASP-Desktop/analysis/analysisform.cpp
```diff
@@ -574,19 +587,26 @@ void AnalysisForm::addControlError(QQuickItem* control, QString message, bool te
...
 
-	JASPControlBase* jaspControl = qobject_cast<JASPControlBase*>(control);
-	if (jaspControl)
-	{
-		if (warning)
-			jaspControl->setHasWarning(true);
-		else
-			jaspControl->setHasError(true);
-	}
+	if (warning)
```
JorisGoosen, line 615:
```
if (warning)	control->setHasWarning(true);
else			control->setHasError(true);
```
Is also prettier I think

JASP-Desktop/analysis/boundqmlitem.cpp
```diff
@@ -21,7 +21,9 @@
 
 void BoundQMLItem::runRScript(const QString &script, bool whiteListedVersion)
 {
-	form()->runRScript(script, name(), whiteListedVersion);
+	QString id = parentListControl() ? (parentListControl()->name() + "." + parentListControlKey() + "." + name()) : name();
```
JorisGoosen, line 24: I assume parentListControlKey() is always returning something sensible whenever parentListControl() != nulltpr?

JorisGoosen: Ooh yeah I see already

JASP-Desktop/components/JASP/Controls/IntegerField.qml
```diff
@@ -28,7 +28,7 @@ TextField
 	property bool	negativeValues:	false
 	property int	min:			negativeValues ? -2147483647 : 0 // 2^32 - 1
 	property int	max:			2147483647
-	property string inclusive:		"Yes"
+	property string inclusive:		"yes"
```
JorisGoosen: why is this a string anyway?

JASP-Desktop/widgets/boundqmltextinput.cpp
```diff
@@ -299,19 +299,21 @@ void BoundQMLTextInput::_setFormulaOptions(std::string formula, bool valid)
...
-	double min		= getItemProperty("min").toDouble();
-	double max		= getItemProperty("max").toDouble();
-	bool inclusive	= getItemProperty("inclusive").toBool();
-
-	bool tooSmall = inclusive ? result < min : result <= min;
-	bool tooLarge = inclusive ? result > max : result >= max;
+	double min			= getItemProperty("min").toDouble();
+	double max			= getItemProperty("max").toDouble();
+	QString inclusive	= getItemProperty("inclusive").toString();
+	bool includeMin = (inclusive == "yes" || inclusive == "minOnly");
```
JorisGoosen: ow right.
I think it makes more sense to have an `includeMin`, `includeMax`.
The idea of people having to enter random strings seems quite errorprone to me. And if we are going to have peopl enter strings we should at least force it to always be lowercase or something.

JASP-Desktop/widgets/boundqmltextinput.cpp
```diff
@@ -299,19 +299,21 @@ void BoundQMLTextInput::_setFormulaOptions(std::string formula, bool valid)
...
+	bool tooSmall = includeMin ? result < min : result <= min;
+	bool tooLarge = includeMax ? result > max : result >= max;
 	bool inBounds = !(tooSmall || tooLarge);
 
 	if (!inBounds)
 	{
 		QString end;
-		if (tooSmall)	end = (inclusive ? "&ge; " : "&gt; ") + getItemProperty("min").toString();
-		else			end = (inclusive ? "&le; " : "&lt; ") + getItemProperty("max").toString();
+		if (tooSmall)	end = (includeMin ? "&ge; " : "&gt; ") + getItemProperty("min").toString();
```
JorisGoosen, line 315: See, even you think it is a good idea ;)

JorisGoosen (changes requested):
Mostly looks fine, the style-things are just there because I noticed them.
The `inclusive` thing as a string I do not like at all. It could be replaced with either two booleans that are intially bound to inclusive, or we could add an enum with `inclusiveType = {minMax, minOnly, maxOnly, none}` or something like that.

JASP-Desktop/analysis/analysisform.cpp
```diff
@@ -574,42 +597,44 @@ void AnalysisForm::addControlError(QQuickItem* control, QString message, bool te
...
 
 void AnalysisForm::addControlErrorSet(JASPControlBase *control, bool add)
 {
 	if (!control) return;
 
-	if (add)
-		_jaspControlsWithErrorSet.insert(control);
-	else
-		_jaspControlsWithErrorSet.remove(control);
+	if (add)	_jaspControlsWithErrorSet.insert(control);
```
JorisGoosen, line 625: 👍

JASP-Desktop/analysis/jaspcontrolbase.h
```diff
@@ -44,7 +45,17 @@ class JASPControlBase : public QQuickItem
 		, ComponentsList
 	};
 
+	enum class Inclusive	{ MinMax, MinOnly, MaxOnly, None };
```
JorisGoosen: Maybe it would be good (like we just discussed) to have each enum start where the last one left off?
Aka:
```
enum class Inclusive	{ None = 0, MinMax, MinOnly, MaxOnly };
enum class DropMode		{ DropNone  = static_cast<int>(Inclusive::MaxOnly) + 1, DropInsert, DropReplace };
enum class ListViewType { AssignedVariables  = static_cast<int>(DropMode::DropReplace) + 1, Interaction, AvailableVariables, RepeatedMeasures, Layers, AvailableInteraction };
enum class AssignType	{ AssignDefault  = static_cast<int>(ListViewType::AssignedVariables) + 1, AssignCross, AssignMainEffects, AssignInteraction, AssignAll2Way, AssignAll3Way, AssignAll4Way, AssignAll5Way };
```

JorisGoosen:
Besides the question on the starting value of each enum I would say it is good to go though!
And that starting number isn't essential, I just think it would make debugging easier and also possible error-checking later on

PR #3723, opened 2020-01-08
Fix partial plots

### LINKED ISSUE #581: Partial Plots don't display regression line
* JASP version:  Version 0.11.1

* OS name and version:  Win 10
* Analysis: 
* Bug description:  Partial plots display a horizontal regression line in multiple linear regression.  Instead, they should display the same slope as the multiple regression coefficient.  Please correct this, or tell me how I can change a setting to correct it.  Thank you.
* Expected behaviour:  See above.

Kucharssim:
- Fixes https://github.com/jasp-stats/jasp-issues/issues/581

- Adds unit tests for plots in linear regression


vandenman:
Looks good to me, great that you added unit tests! After some digging, it seems that this was broken in https://github.com/jasp-stats/jasp-desktop/commit/b443c40ff857c46429b57648f176581773573c81#diff-1f5589c0ecb10b824bb5fee1f8163dec (Nov 17, 2018).

vandenman (approved):

blogfrey:
Thank you very much!  I am a new JASP user as well as a new GitHub user.  Is there / will there be an updated version of the software that I can download to implement the fix you did?  Thanks.  -Dan

boutinb:
We should have a new official version (0.12) begin februari. For the time being you can still use the nightly version (http://static.jasp-stats.org/Nightlies/) tomorrow (if the build did not fail). BUT this build is not at all an official version, it is just a test version that might have many other bugs...

blogfrey:
Hello Bruno,

That is great, thank you!

Dan

PR #3726, opened 2020-01-09
Progressbar in correlation matrix

JorisGoosen:
Correlation is incredibly slow in calculating concordance variance and recalculates it every time...

It now stores this variance in the state and I also tried to optimize it slightly but that barely makes a difference.
To do so would require a full rewrite.
I also added a progressBar to be able to cancel it.

### DIFF
--- a/JASP-Engine/JASP/R/correlation.R
+++ b/JASP-Engine/JASP/R/correlation.R
@@ -267,14 +267,14 @@ Correlation <- function(jaspResults, dataset, options){
            correlatedNegatively = "less",
            correlatedPositively = "greater")[options$hypothesis]
   
-  if(length(options$conditioningVariables) == 0){
-    pcor <- FALSE
-  } else{
-    pcor <- TRUE
-  }
+  pcor <- !length(options$conditioningVariables) == 0
   
   results <- list()
+  startProgressbar(length(vpair) * nrow(dataset))
+
   for(i in seq_along(vpair)){
+
+
     # some variable pairs might be reusable, so we don't need to compute them again
     if(!is.null(jaspResults[[vpair[i]]])) {
       results[[vpair[i]]] <- jaspResults[[vpair[i]]]$object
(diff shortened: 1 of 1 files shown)


Kucharssim:
Just a remark from top of my head: It does recompute it every time when confidence intervals need to be changed (i.e., either confidence level is changed, or hypothesis is changed), otherwise there should be already a state (e.g., when a new variable is added to the variable list, the old variable pairs are not computed again).

Alternatives/additions to this fix might be:
1) take out the computations of the CIs out of the helper .corr.test function and return just the results of the concordances computations. The CIs then can be computed later (I might need to do this anyways if we decide to have bootstrapping for CIs). I don’t think this would mean a full rewrite though...
2) for large n fall back on normal approximation for the Kendall's tau. Not sure whether we want to go this way, maybe @AlexanderLyNL and @JohnnyDoorn have opinions about that?

JorisGoosen:
Hmm, it really does seem like it recomputes the kendall tau concordances every time, but maybe I was mistaken.

A thing I was wondering anyway is, if the kendall takes so long, wouldn't it be better to only compute this when it is requested?

Kucharssim:
> Hmm, it really does seem like it recomputes the kendall tau concordances every time, but maybe I was mistaken.

Maybe the mistake is on my part in the implementation, but I remember that the intent was there.

> A thing I was wondering anyway is, if the kendall takes so long, wouldn't it be better to only compute this when it is requested?

Oh yeah, that was the third point that I wanted to mention, but forgot about it while writing the comment :)

JorisGoosen:
> > Hmm, it really does seem like it recomputes the kendall tau concordances every time, but maybe I was mistaken.
> 
> Maybe the mistake is on my part in the implementation, but I remember that the intent was there.
I didn't check it very well and it was hard to focus on it because it takes so long and I still didnt set jaspTools up properly (and thus was running it through JASP).


> > A thing I was wondering anyway is, if the kendall takes so long, wouldn't it be better to only compute this when it is requested?
> 
> Oh yeah, that was the third point that I wanted to mention, but forgot about it while writing the comment :)
Haha ok :p

JorisGoosen:
Btw. I took out the state thing and my "optimizations" because apparently they gave a different results than the unit test expected and I did not want to spend more time on it.

PR #3727, opened 2020-01-09
catch shapiro error when all values are identical

### LINKED ISSUE #553: Bug Shapiro-Wilk Test
* JASP version: 0.11.1.0
 Windows
* Analysis:  Shapiro-Wilk Test
* Bug description:
This analysis terminated unexpectedly.
Error in shapiro.test(param): all 'x' values are identical
Stack trace

.descriptivesDescriptivesTable(dataset, options, jaspResults, numberMissingSplitBy = numberMissingSplitBy)

.descriptivesDescriptivesTable_subFunction(column, list(Variable = variable, Level = splitLevels[l]), options, shouldAddNominalTextFootnote, shouldAddModeMoreThanOnceFootnote)

.descriptivesDescriptivesTable_subFunction_OptionChecker(options$shapiro, na.omitted, function(param) {
    shapiro.test(param)$statistic
})

.clean(function_to_use(na.omitted))

function_to_use(na.omitted)

shapiro.test(param)

stop('all 'x' values are identical')
* Expected behaviour:

Kucharssim:
Fixes https://github.com/jasp-stats/jasp-issues/issues/553

### DIFF
--- a/JASP-Engine/JASP/R/descriptives.R
+++ b/JASP-Engine/JASP/R/descriptives.R
@@ -294,7 +294,7 @@ Descriptives <- function(jaspResults, dataset, options) {
         stats$addRows(subReturn$resultsCol, rowNames = paste0(variable, l))
         
         if (subReturn$shouldAddIdenticalFootnote)
-          stats$addFootnote(message = "All values are identical",
+          stats$addFootnote(message = gettext("All values are identical"),
                             colNames = c("Skewness", "Kurtosis", "Shapiro-Wilk", "P-value of Shapiro-Wilk"),
                             rowNames = paste0(variable, l))
       }
@@ -310,7 +310,7 @@ Descriptives <- function(jaspResults, dataset, options) {
       stats$addRows(subReturn$resultsCol, rowNames = variable)
       
       if (subReturn$shouldAddIdenticalFootnote)
-        stats$addFootnote(message = "All values are identical",
+        stats$addFootnote(message = gettext("All values are identical"),
                           colNames = c("Skewness", "Kurtosis", "Shapiro-Wilk", "P-value of Shapiro-Wilk"),
                           rowNames = variable)
     }
@@ -354,8 +354,8 @@ Descriptives <- function(jaspResults, dataset, options) {
   resultsCol[["Std. Error of Kurtosis"]]  <- .descriptivesDescriptivesTable_subFunction_OptionChecker(options$kurtosis,          na.omitted, .descriptivesSEK)
   resultsCol[["Skewness"]]                <- .descriptivesDescriptivesTable_subFunction_OptionChecker(options$skewness,          na.omitted, .descriptivesSkewness)
   resultsCol[["Std. Error of Skewness"]]  <- .descriptivesDescriptivesTable_subFunction_OptionChecker(options$skewness,          na.omitted, .descriptivesSES)
-  resultsCol[["Shapiro-Wilk"]]            <- .descriptivesDescriptivesTable_subFunction_OptionChecker(options$shapiro,           na.omitted, function(param) { res <- try(shapiro.test(param)$statistic); ifelse(inherits(res, "try-error"), NA, res) })
-  resultsCol[["P-value of Shapiro-Wilk"]] <- .descriptivesDescriptivesTable_subFunction_OptionChecker(options$shapiro,           na.omitted, function(param) { res <- try(shapiro.test(param)$p.value);   ifelse(inherits(res, "try-error"), NA, res) })
+  resultsCol[["Shapiro-Wilk"]]            <- .descriptivesDescriptivesTable_subFunction_OptionChecker(options$shapiro,           na.omitted, function(param) { res <- try(shapiro.test(param)$statistic); if(isTryError(res)) NA else res })
+  resultsCol[["P-value of Shapiro-Wilk"]] <- .descriptivesDescriptivesTable_subFunction_OptionChecker(options$shapiro,           na.omitted, function(param) { res <- try(shapiro.test(param)$p.value);   if(isTryError(res)) NA else res })
   resultsCol[["Range"]]                   <- .descriptivesDescriptivesTable_subFunction_OptionChecker(options$range,             na.omitted, function(param) { range(param)[2] - range(param)[1]})
   resultsCol[["Minimum"]]                 <- .descriptivesDescriptivesTable_subFunction_OptionChecker(options$minimum,           na.omitted, min)
   resultsCol[["Maximum"]]                 <- .descriptivesDescriptivesTable_subFunction_OptionChecker(options$maximum,           na.omitted, max)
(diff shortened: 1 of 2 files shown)


JASP-Engine/JASP/R/descriptives.R
```diff
@@ -291,7 +291,12 @@ Descriptives <- function(jaspResults, dataset, options) {
         shouldAddNominalTextFootnote      <- subReturn$shouldAddNominalTextFootnote
         shouldAddModeMoreThanOnceFootnote <- subReturn$shouldAddModeMoreThanOnceFootnote
         
-        stats$addRows(subReturn$resultsCol)
+        stats$addRows(subReturn$resultsCol, rowNames = paste0(variable, l))
+        
+        if (subReturn$shouldAddIdenticalFootnote)
+          stats$addFootnote(message = "All values are identical",
```
vandenman:
```suggestion
          stats$addFootnote(message = gettext("All values are identical"),
```
to immediately add the translation :)

JASP-Engine/JASP/R/descriptives.R
```diff
@@ -302,7 +307,12 @@ Descriptives <- function(jaspResults, dataset, options) {
       shouldAddNominalTextFootnote      <- subReturn$shouldAddNominalTextFootnote
       shouldAddModeMoreThanOnceFootnote <- subReturn$shouldAddModeMoreThanOnceFootnote
 
-      stats$addRows(subReturn$resultsCol)
+      stats$addRows(subReturn$resultsCol, rowNames = variable)
+      
+      if (subReturn$shouldAddIdenticalFootnote)
+        stats$addFootnote(message = "All values are identical",
```
vandenman:
```suggestion
        stats$addFootnote(message = gettext("All values are identical"),
```
again translation.

JASP-Engine/JASP/R/descriptives.R
```diff
@@ -329,7 +339,9 @@ Descriptives <- function(jaspResults, dataset, options) {
   if (base::is.factor(na.omitted) && (options$mean || options$mode || options$median || options$minimum || options$standardErrorMean || options$iqr || options$mad || options$madrobust || options$kurtosis || options$shapiro || options$skewness || options$percentileValuesQuartiles || options$variance || options$standardDeviation || options$percentileValuesPercentiles || options$sum || options$maximum)) {
     shouldAddNominalTextFootnote <- TRUE
   }
-    
+  
+  shouldAddIdenticalFootnote <- all(na.omitted[1] == na.omitted) && (options$skewness || options$kurtosis || options$shapiro)
```
vandenman, line 343: so `==` is very strict... so strict that you'll notice it when `1 + 1e-16 == 1` is true but `1 + 2e-16 == 1` returns false. This basically means that this will never be true unless you actually duplicate the data. I'm not sure if this is the right implementation for this. I would consider at least the following two alternatives:

1. `isTRUE(all.equal(na.omitted[1], na.omitted))`. This uses `tolerance = sqrt(.Machine$double.eps)` which is also what most optimization algorithms use. Still I think this will rarely pass.
2. `var(na.omitted) < some_magic_number`. Most likely the failure is related to the absence of variance, so we could also just test that.

Both of these allow us to tweak the tolerance.

JASP-Engine/JASP/R/descriptives.R
```diff
@@ -342,8 +354,8 @@ Descriptives <- function(jaspResults, dataset, options) {
   resultsCol[["Std. Error of Kurtosis"]]  <- .descriptivesDescriptivesTable_subFunction_OptionChecker(options$kurtosis,          na.omitted, .descriptivesSEK)
   resultsCol[["Skewness"]]                <- .descriptivesDescriptivesTable_subFunction_OptionChecker(options$skewness,          na.omitted, .descriptivesSkewness)
   resultsCol[["Std. Error of Skewness"]]  <- .descriptivesDescriptivesTable_subFunction_OptionChecker(options$skewness,          na.omitted, .descriptivesSES)
-  resultsCol[["Shapiro-Wilk"]]            <- .descriptivesDescriptivesTable_subFunction_OptionChecker(options$shapiro,           na.omitted, function(param) { shapiro.test(param)$statistic })
-  resultsCol[["P-value of Shapiro-Wilk"]] <- .descriptivesDescriptivesTable_subFunction_OptionChecker(options$shapiro,           na.omitted, function(param) { shapiro.test(param)$p.value })
+  resultsCol[["Shapiro-Wilk"]]            <- .descriptivesDescriptivesTable_subFunction_OptionChecker(options$shapiro,           na.omitted, function(param) { res <- try(shapiro.test(param)$statistic); ifelse(inherits(res, "try-error"), NA, res) })
```
vandenman:
```suggestion
  resultsCol[["Shapiro-Wilk"]]            <- .descriptivesDescriptivesTable_subFunction_OptionChecker(options$shapiro,           na.omitted, function(param) { res <- try(shapiro.test(param)$statistic); if(isTryError(res)) NA else res })
```

JASP-Engine/JASP/R/descriptives.R
```diff
@@ -342,8 +354,8 @@ Descriptives <- function(jaspResults, dataset, options) {
...
+  resultsCol[["P-value of Shapiro-Wilk"]] <- .descriptivesDescriptivesTable_subFunction_OptionChecker(options$shapiro,           na.omitted, function(param) { res <- try(shapiro.test(param)$p.value);   ifelse(inherits(res, "try-error"), NA, res) })
```
vandenman: same about `isTryError` and `if` instead of `ifelse`.

vandenman (changes requested):
So.. there is also a function inside `.hasErrors` to just crash the analysis if identical variables are detected. Perhaps that's too extreme, but then again, identical variables tend to mess up most statistical functions. Did you consider just quitting early? Ultimately it doesn't matter too much so the choice is up to you. Crashing early does mean that code deep down does not need to deal with these kinds of corner cases.

I left some comments, mostly about style and one thing about how to check for equality.

vandenman:
Also, since this is a corner case that is very easy to break in future updates, a small unit-test would be nice.

Kucharssim:
> So.. there is also a function inside `.hasErrors` to just crash the analysis if identical variables are detected. Perhaps that's too extreme, but then again, identical variables tend to mess up most statistical functions. Did you consider just quitting early? Ultimately it doesn't matter too much so the choice is up to you. Crashing early does mean that code deep down does not need to deal with these kinds of corner cases.

Yeah I thought about that, but since we are in descriptives, I would say that it would be nice to report the descriptive stats as they are (at least those that do not crash when variance is 0), even if the variable has 0 variance...

> I left some comments, mostly about style and one thing about how to check for equality.

Thanks! I’ll resolve them tomorrow.

JASP-Engine/JASP/R/descriptives.R
Kucharssim: > This basically means that this will never be true unless you actually duplicate the data.

Which is what we want, I think.

Basically I wanted to flag the condition for which `shapiro.test()` actually returns an error, and that is when `diff(range(x)) == 0`, which is effectively what `all(na.omitted[1] == na.omitted)` does except that the former would fail if x is a factor. If the range is very very small, the `shapiro.test()` just rescales the variable (which does not affect shapiro statistic). The alternatives which you proposed are nice but they would also flag cases for which we actually can compute the shapiro wilk, so I tend to think that my code is what we want for this case. I am happy to be convinced otherwise, but right now I don’t see a reason to have a more lenient check...

vandenman (approved):

PR #3733, opened 2020-01-13
small bugfix in sequential plot

vandenman:
Small fix for @sophieberkhout

Empty plots didn't get removed from robustness and sequential plots and that caused a crash. No plots in JASP were affected since those either use all subplots or none.


TimKDJ:
Can you make the commit message a little more clear?

PR #3740, opened 2020-01-14
Jags updates

vandenman:
updates for JAGS

- model is now loaded correctly when opening a jasp file.
- font scaling works better.
- `gettext` is used everywhere.
- many quality of life changes in the R side because the .cpp does more work now. E.g., column names in the data set are detected in cpp rather than in R.
- there are unit tests but since Travis and JAGS don't like each other they are skipped for now.


JorisGoosen:
@TimKDJ could you take a look at the R-code?
The cpp and qml looks fine.

TimKDJ (approved):

PR #3750, opened 2020-01-16
Add Bayesian meta analysis

sophieberkhout:


Resources/Meta Analysis/qml/BayesianMetaAnalysis.qml
```diff
@@ -0,0 +1,682 @@
...
+								name:     		"iterBridge"
+								visible:      	bridge.checked
+								defaultValue: 	5000
+								max:			1000000
+								fieldWidth: 	100
+							}
+						}
+					}
+				}
+				CheckBox {
```
vandenman: Any reason you didn't use the [SetSeed{} component](https://github.com/jasp-stats/jasp-desktop/blob/629dab702c3647c74ae3d6a86c52edbe79d79786/JASP-Desktop/components/JASP/Widgets/SetSeed.qml)? That will add an element called `setSeed` to the options list.

JASP-Engine/JASP/R/bayesianmetaanalysis.R
```diff
@@ -0,0 +1,1749 @@
...
+  
+  # Dependencies: basically everything
+  dependencies <- .bmaGetDependencies()
+  
+  # Dataset with effectSize, standardError, and studyLabels
+  # If data is null stuff is missing
+  dataset <- .bmaReadData(jaspResults, options)
+  
+  # Set seed if user wants to
+  if(options[["seedBox"]]) set.seed(options[["seed"]])
```
vandenman: if you change the QML component (see my other comment), this could just become `.setSeedJASP(options)`.

JASP-Engine/JASP/R/bayesianmetaanalysis.R
```diff
@@ -0,0 +1,1749 @@
...
+  # Lower and upper limits without truncation
+  lowerES <- -Inf
+  upperES <- Inf
+  
+  # Cauchy prior
+  if(options$priorES == "cauchy"){
+    familyES <- "t"
+    paramES <- c(options$informativeCauchyLocation, 
+                 options$informativeCauchyScale, 1)
+  }
```
vandenman: can't this be:
```r
if(options$priorES == "cauchy"){
...
} else if(options$priorES == "normal"){
...
} else if(options$priorES == "t"){
...
}
```
? (`else if` rather than multiple if`'s). Or perhaps a switch.

JASP-Engine/JASP/R/bayesianmetaanalysis.R
```diff
@@ -0,0 +1,1749 @@
...
+}
+
+#For state
+.bmaResultsState <- function(jaspResults, dataset, options, dependencies) {
+  
+  if(!is.null(jaspResults[["bmaResults"]])) return(jaspResults[["bmaResults"]]$object)
+
+  results <- .bmaResults(jaspResults, dataset, options)
+  
+  # The results object is too large for .jasp files. Break it up and reassemble only the required components.
```
vandenman, line 185: > \# The results object is too large for .jasp files. Break it up and reassemble only the required components.

Good idea!

JASP-Engine/JASP/R/bayesianmetaanalysis.R
```diff
@@ -0,0 +1,1749 @@
...
+
+# Main function ----
+
+BayesianMetaAnalysis <- function(jaspResults, dataset, options) {
+  
+  # Ready: variables needed for the analysis (confidence interval missing)
+  ready <- options[["effectSize"]] != "" && (options[["standardError"]] != "" || (all(unlist(options$confidenceInterval) != "")  && !is.null(unlist(options[["confidenceInterval"]])))) 
+  
+  # Dependencies: basically everything
+  dependencies <- .bmaGetDependencies()
```
vandenman: rather than calling this function in the body of the main function you could also just call it whenever you need it in a subfunction. Also, rather than
```r
.bmaGetDependencies <- function(){
  dependencies <- c("effectSize", "standardError", "confidenceInterval", "modelSpecification",
                      "allPos", "allNeg", "priorH0FE", "priorH1FE", "priorH0RE", "priorH1RE",
                      "priorES", "informativeCauchyLocation", "informativeCauchyScale",
                      "checkLowerPrior", "checkUpperPrior", "lowerTrunc", "upperTrunc",
                      "informativeNormalMean", "informativeNormalStd",
                      "informativeTLocation", "informativeTScale", "informativeTDf",
                      "priorSE", "inverseGammaShape", "inverseGammaScale",
                      "informativehalfTScale", "informativehalfTDf",
                      "BFComputation", "iterBridge", "iterMCMC", "chainsMCMC", "seed", "seedBox")
  return(dependencies)
}
```
you could just do
```r
.bmaDependencies <- c("effectSize", "standardError", "confidenceInterval", "modelSpecification",
                      "allPos", "allNeg", "priorH0FE", "priorH1FE", "priorH0RE", "priorH1RE",
                      "priorES", "informativeCauchyLocation", "informativeCauchyScale",
                      "checkLowerPrior", "checkUpperPrior", "lowerTrunc", "upperTrunc",
                      "informativeNormalMean", "informativeNormalStd",
                      "informativeTLocation", "informativeTScale", "informativeTDf",
                      "priorSE", "inverseGammaShape", "inverseGammaScale",
                      "informativehalfTScale", "informativehalfTDf",
                      "BFComputation", "iterBridge", "iterMCMC", "chainsMCMC", "seed", "seedBox")
```
and then it'll be available in every subfunction of the module, so everywhere you add `dependencies` you could just add `.bmaDependencies`.

Resources/Meta Analysis/qml/BayesianMetaAnalysis.qml
```diff
@@ -0,0 +1,682 @@
...
+
+			CheckBox
+			{
+				name: 	"esTable";
+				label: 	qsTr("Effect sizes per study")
+			}
+		}
+
+	//// BF ////
+		RadioButtonGroup
```
vandenman: For this one we also have a custom component called `BayesFactorType{}`.

vandenman (commented):
Overall this looks great! I couldn't find any major issues. It would be great if you could change/ fix a couple of things though:

1. The unit tests fail.
2. No translations. For every string that gets shown to the user and is hardcoded add `gettext()`. For example:
```r
# old
bmaTable$addColumnInfo(name = "ES", title = "Mean", type = "number")
# should be
bmaTable$addColumnInfo(name = "ES", title = gettext("Mean"), type = "number")
```
If strings are more complicated and paste things together use `gettextf`, basically the translation analogue of `sprintf`.

3. There is some code duplication in the qml files. Perhaps this was necessary when the analysis is run as a dynamic module, but if the analysis is added to JASP in the way this PR proposes then we can tidy some things a bit.

JASP-Engine/JASP/R/bayesianmetaanalysis.R
```diff
@@ -0,0 +1,1760 @@
...
+  valuesLine <- c("solid", "dotted")
+  
+  if(type == "ES"){
+    xlab <- bquote(paste(.(gettext("Effect size")), ~mu))
+    xlim <- c(-4, 4)
+    if(options[["modelSpecification"]] == "BMA"){
+      int <- c(bmaResults[["bma"]]$estimates["averaged", "2.5%"], bmaResults[["bma"]]$estimates["averaged", "97.5%"])
+      postName <- "Averaged"
+      if(options[["addLines"]]){
+        labelsModel <- c(bquote(paste(.(gettext("Fixed H")), [1])), bquote(paste(.(gettext("Random H")), [1])), bquote(paste(.(gettext("Averaged H")), [1])), bquote(paste(.(gettext("Prior H")), [1])))
```
TimKDJ: This line is not valid R code, according to travis: `Error: unexpected '[' in "c(bquote(paste(.(gettext("Fixed H")), ["`

sophieberkhout:
Hi Tim,

Ik had het gezien inderdaad!
De 1 moet een subscript zijn. Weet jij hoe dat moet toevallig? Ik snap die
bquote nog niet zo goed om eerlijk te zijn..

Groetjes,
Sophie

Op do 23 jan. 2020 om 12:38 schreef Tim de Jong <notifications@github.com>:

TimKDJ:
It would appear that besides the model probabilities table your analysis is not unit-test-able. The results are unique to all systems they were tested on (my MacBook (MacOS), Travis (MacOS), my home pc (Windows)). So while the seed seems to ensure results remain constant within one device, the seed does nothing to generate reproduceable results across devices.
I'm not too familiar with rstan, but given that we pass it a specific seed and also fix a seed in R, I'm not sure there's much more we can do.

I could investigate at some point, but for now I'd advice putting `skip("rstan results differ across different devices")` as the first line of each failing unit test (right under the `test_that("...`'s).

vandenman:
the Stan documentation has a chapter on [Reproducibility](https://mc-stan.org/docs/2_21/reference-manual/reproducibility-chapter.html) and mentions the following:


> Stan results will only be exactly reproducible if all of the following components are identical:
> 
> - Stan version
> - Stan interface (RStan, PyStan, CmdStan) and version, plus version of interface language (R, Python, shell)
> - versions of included libraries (Boost and Eigen)
 - operating system version
> - computer hardware including CPU, motherboard and memory
> - C++ compiler, including version, compiler flags, and linked libraries
> - same configuration of call to Stan, including random seed, chain ID, initialization and data

The key problem is highlighted. I know of two workarounds:

1. Use very few samples. If differences are due to floating point rounding errors related to hardware then these won't accumulate using e.g., 5 samples.

2. Use loads of samples, and only test *aggregate* results (e.g., mean, variance, etc.). The different runs should sample from the same distribution, so estimates of the properties of those distributions should converge to the same value, given enough samples.

If the models are relatively simple and sampling doesn't take too long, I'd try option 2 with a tolerance of maybe 0.1 (0.1 is used by rstan in their [unit tests](https://github.com/stan-dev/rstan/blob/develop/rstan/tests/unitTests/runit.test.stanmodel.R)). Otherwise I would also skip the tests.

vandenman (approved):

PR #3754, opened 2020-01-16
Remove 'force(col)' from LDA plotting functions

koenderks:
Changes in unit tests are due to letting the limits of the axes depend on the breaks created in JASPgraphs.


koenderks:
@TimKDJ Apparently the unit test fails again...but on my jasptools the plots are fine now. Maybe you can take a quick look when you find the time?

TimKDJ:
The size of the y-axis "Density" is what's different. Is this controlled by JASPgraphs or ggplot?

koenderks:
By ggplot, but after rebasing more plots bail (also bayesian correlation matrix etc...)

TimKDJ:
Yeah that's my bad, I merged Lexi's rewrite of the Correlation and Don's change of JASPgraphs today, but Don's thing negates Lexi's changes to the plot

TimKDJ:
If you rebase now, then you should be able to remake just the lda classification plots

koenderks:
Worked, this can be merged.

PR #3755, opened 2020-01-17
Some translation errors fixed

FransMeerhoff:
Unable to generate new R-nl.po, Some fixes were needed.


TimKDJ:
I don't think the .po/.mo files should be included. There are no translations in them; the .pot file suffices

PR #3756, opened 2020-01-20
gettext should have been gettextf

vandenman:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/659

### DIFF
--- a/JASP-Engine/JASP/R/commonAnovaBayesian.R
+++ b/JASP-Engine/JASP/R/commonAnovaBayesian.R
@@ -2562,7 +2562,7 @@
   if (!is.null(jaspResults[["tableSMICRI"]]) || !options[["singleModelCriTable"]])
     return()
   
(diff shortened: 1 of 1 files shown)


AlexanderLyNL:
- Bayesian RM ANOVA: The first time I did this, it seems to work. 
- However, when I do this again, I hit the error of https://github.com/jasp-stats/INTERNAL-jasp/issues/660
- RM ANOVA: Still broken, but more related to https://github.com/jasp-stats/INTERNAL-jasp/issues/660
- In principle this PR should be okay I guess, but I can't be sure without the other fix.

vandenman:
Right, but if you (re)name the repeated measures factors appropriately it shouldn't crash right? Anyway I think (and hope) this suffices for jasp-stats/INTERNAL-jasp#660.

vandenman:
actually, it looks like #3755 also fixes this issue. That one also fixes a couple of other mistakes in `commonAnovaBayesian.R`. I'm closing this PR in favor of that one.

PR #3757, opened 2020-01-20
Takes both "extra encodings" and normal columns into account when .v and .unv

JorisGoosen:
fixes issue https://github.com/jasp-stats/INTERNAL-jasp/issues/660

I had to do some static magic though


JorisGoosen:
@boutinb the Travis errors seem to have nothing to do with my changes as far as I can see.

PR #3759, opened 2020-01-21
Set FormulaField in JAGS

boutinb:


JASP-Desktop/components/JASP/Controls/TableView.qml
```diff
@@ -249,4 +225,60 @@ JASPControl
...
+
+	Component
+	{
+		id: formulaFieldComponent
+		Item
+		{
+			FormulaField
+			{
+				id:						formlaInput
+				inputType:				columnIndex === 1 ? "formula" : "string"
```
JorisGoosen: Could you not reuse this component entirely (poss. with an extra check here) also for `!tableView.modelType === "JAGSDataInputModel`?

FormulaField with `inputType == "formula"` is the same as TextField right?

PR #3772, opened 2020-01-22
Fix cluster means plot bars not showing in ML

koenderks:
As mentioned in https://github.com/jasp-stats/jasp-issues/issues/563, the bars from the cluster means plot would disappear when using whole numbers as data.


App-State-Prof:
Thank you for for the excellent updates and fixes!

PR #3775, opened 2020-01-23
Fix a number of interface issues

TimKDJ:
1. Transposed tables look a little nicer:

before
Screenshot 2020-01-23 at 10 32 08
after
Screenshot 2020-01-23 at 10 27 28

2. The progressbar container is always included in the DOM to ensure results don't shift when a progressbar is added (regression)

3. If an analysis wide error occurs the results show this visually by greying out (regression)
before <-> after
Screenshot 2020-01-23 at 10 38 27

4. The separator in all analyses that showed variable pairs was very misaligned (regression; fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/586):

5. Paired samples ttest table now shows all variable pairs (regression; fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/668, https://github.com/jasp-stats/jasp-issues/issues/594)

6. Bayesian Correlation and frequentist correlation tables are more similar now

### DIFF
--- a/JASP-Engine/JASP/R/correlation.R
+++ b/JASP-Engine/JASP/R/correlation.R
@@ -134,7 +134,7 @@ Correlation <- function(jaspResults, dataset, options){
   pairTitles <- combn(variables, 2, simplify = FALSE)
   
   mainTable$addColumnInfo(name = "variable1", title = "", type = "string")
-  mainTable$addColumnInfo(name = "separator", title = "", type = "string")
+  mainTable$addColumnInfo(name = "separator", title = "", type = "separator")
   mainTable$addColumnInfo(name = "variable2", title = "", type = "string")
 
   mainTable[['variable1']] <- sapply(pairTitles, function(x) x[[1]])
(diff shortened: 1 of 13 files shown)


JorisGoosen:
Aah, you couldn't get the "rowheaders" to be not-striped? I think that might improve the look actually.
Screen Shot 2019-06-26 at 16 23 17
And as the screenshot @Kucharssim originally sent me probably shows the situation before the jaspResult-rewrite of correlations this would be a regression, no?

TimKDJ:
I didn't try to be honest, I thought it looked a little weird. It's also not a style we use anywhere else, e.g., the paired samples ttest looks like this:
Screenshot 2020-01-23 at 13 43 00

JorisGoosen:
Alright, as long as it was done consciously ;)

(I do think it would look a bit better to emphasis the "header"-nature of them, but whatever)

JASP-Engine/JASP/R/correlation.R
```diff
@@ -513,7 +513,7 @@ Correlation <- function(jaspResults, dataset, options){
   shapiroTable$showSpecifiedColumnsOnly <- TRUE
   
   shapiroTable$addColumnInfo(name = "var1",      title = "",                      type = "string")
-  shapiroTable$addColumnInfo(name = "separator", title = "",                      type = "string")
+  shapiroTable$addColumnInfo(name = "separator", title = "",                      type = "separator")
```
JorisGoosen, line 516: Hmm, would it maybe be nice to add a function to jaspTable called addColumnSeparator or something? It can simply do this and be in zzzWrappers for instance.

JASP-R-Interface/jaspResults/DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Package: jaspResults
 Type: Package
 Title: Easy results for your JASP analysis
-Version: 1.12
+Version: 1.13
 Date: 2018-12-07
```
JorisGoosen, line 5: Maybe the date should also be incremented occasionally?

JASP-Engine/JASP/R/correlation.R
TimKDJ: Yes, standardizing this would probably be a good idea!

JASP-R-Interface/jaspResults/DESCRIPTION
TimKDJ: Possibly.........

PR #3784, opened 2020-01-23
Custom Contrasts AN(C)OVA

JohnnyDoorn:
Tested with Andy field one-way ANOVA example and debug data set on SPSS (only one-way, as SPSS does not do two+-way custom contrasts)


boutinb:
The ContrastsList widget is used in ANOVA, ANCOVA, Repeated Measure ANOVA (and Linear Mixed Models, but it is apparently not yet added in JASP). Do your changes affect all these analyses?

JohnnyDoorn:
This is only for the ANOVA and ANCOVA, I;ll look into the RM ANOVA still =_

PR #3789, opened 2020-01-27
Users cannot select a different executable manually on flatpak

JorisGoosen:
Related to https://github.com/jasp-stats/jasp-issues/issues/591

Problem here is that flatpak sandboxes the application so that we aren't allowed to start a random executable. The only way to open a file properly is through a file portal, so we let flatpak + linux handle all that stuff.


boutinb:
Hoi Joris, je kunt in QML de platform vragen: Qt.platform.os == "osx" bv.

JorisGoosen:
Aah!
Ok, dat kan ik nog wel aanpassen dan.

PR #3790, opened 2020-01-27
Fix some problems with CustomContrast widget

boutinb:


JASP-Desktop/components/JASP/Controls/ComponentsList.qml
```diff
@@ -25,6 +25,7 @@ JASPListControl
 	id				: componentsList
 	controlType		: JASPControlBase.ComponentsList
 	itemComponent	: components
+	listRectangle.color: jaspTheme.analysisBackgroundColor
```
JorisGoosen: Not very nicely lined up... :(

JASP-Desktop/components/JASP/Controls/TableView.qml
```diff
@@ -196,20 +200,18 @@ JASPControl
 				leftTopCornerItem: Rectangle
 				{
 					color: jaspTheme.analysisBackgroundColor
+					width: cornerText.width
```
JorisGoosen: This is overwritten by `DataSetView::createleftTopCorner()` and because you turned eliding off this might make things *very* ugly.

JASP-Desktop/widgets/rowcontrols.cpp
```diff
@@ -46,22 +46,33 @@ void RowControls::init(int row, const QString& key)
...
+			Log::log() << "Could not create control in ListView " << listView->name() << std::endl;
 
 		col++;
 	}
 }
 
 void RowControls::setContext(int row, const QString &key)
 {
-	for (JASPControlWrapper* controlWrapper : _rowJASPWrapperMap.values())
+	QMapIterator<QObject*, QQmlContext*> it(_contextMap);
```
JorisGoosen: Why not use: `for(auto & itemCtxt : _contextMap)`
And then refer to item as `itemCtxt.first` and to context as `itemCtxt.second`.
That would make this all a but less messy.

JorisGoosen (changes requested):
Mostly looks fine, but I would like to see the two changerequests in the PR fixed.

JASP-Desktop/widgets/rowcontrols.cpp
boutinb: It is QMap, not std::map... your code does not work with QMap.

JorisGoosen: No, then why not use std::map instead as well?

boutinb: OK

PR #3800, opened 2020-01-29
seeds part 2

FBartos:
seeds for bayesian abtest, contingency tables, and correlations

### DIFF
--- a/JASP-Engine/JASP/R/correlationbayesian.R
+++ b/JASP-Engine/JASP/R/correlationbayesian.R
@@ -768,9 +768,10 @@ CorrelationBayesian <- function(jaspResults, dataset=NULL, options, ...) {
     return(jaspResults[[indexName]]$object)
 
   results <- list()
-  .setSeedJASP(options)
-  for (method in names(bfObject))
+  for (method in names(bfObject)){
+    .setSeedJASP(options)
     results[[method]] <- bstats::computeCorPosteriorLine(bfObject = bfObject[[method]], method = method, alternative = options[["alternative"]])
+  }
   
   jaspResults[[indexName]] <- createJaspState(results)
   jaspResults[[indexName]]$dependOn(c("missingValues", "alternative", "kappa", "setSeed", "seed"))
--- a/JASP-Engine/JASP/R/correlationbayesian.R
+++ b/JASP-Engine/JASP/R/correlationbayesian.R
@@ -84,6 +84,7 @@ CorrelationBayesian <- function(jaspResults, dataset=NULL, options, ...) {
         v2 <- dataset[[.v(var2)]]
       }
       
+      .setSeedJASP(options)
       bfObject <- bstats::bcor.test("x"=v1, "y"=v2, "kappa"=options[["kappa"]],
                                     "method"=method, "ciValue"=options[["ciValue"]])
       
@@ -100,7 +101,7 @@ CorrelationBayesian <- function(jaspResults, dataset=NULL, options, ...) {
   }
   
   jaspResults[["corModel"]] <- createJaspState(result)
-  jaspResults[["corModel"]]$dependOn(c("missingValues", "variables", "kappa", "ciValue"))
+  jaspResults[["corModel"]]$dependOn(c("missingValues", "variables", "kappa", "ciValue", "setSeed", "seed"))
   
   return(result)
 }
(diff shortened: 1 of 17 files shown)


TimKDJ:
The unit tests for some summary stats analyses fail, because they use shared code with the common analyses and don't have the setseed qml component. 
@AlexanderLyNL I guess it would be nice for the summary stats to have the seed option, too?

TimKDJ:
EJ was ok with postponing the summary stats until later; @FBartos you'll just need to add some logic to the joint R functions to prevent summary stats analyses from setting seeds

FBartos:
ok, will do

FBartos:
@TimKDJ 

does the options list carry any indication from which analysis it came from?

I wanted to add a switch to the `.setSeedJASP` function itself. When I generate the options list using 
`options <- jasptools::analysisOptions("SummaryStatsCorrelationBayesianPairs")`, it creates an atribute with the name of analysis. Then, calling `if( grepl("SummaryStats", attr(options, "analysisName")) )return()` should've skipped the rest of the function in case of Summary Stats. 

However, it doesn't work when the options are passed from JASP or testAnalysis, because the attribute doesn't exist (don't really understand why the ` jasptools::analysisOptions` function creates it then... So, I was thinking whether it wouldn't be better to add an element to the options list on the beginning of each of the Summary Stats script that would tell the `.setSeedJASP` to not set any seed. We wouldn't have to go back to the analyses where the seed is implemented and only change the Summary Stats in this regard. 

What do you think?

TimKDJ:
> does the options list carry any indication from which analysis it came from?

No

> I wanted to add a switch to the .setSeedJASP function itself. When I generate the options list using
options <- jasptools::analysisOptions("SummaryStatsCorrelationBayesianPairs"), it creates an atribute with the name of analysis. Then, calling if( grepl("SummaryStats", attr(options, "analysisName")) )return() should've skipped the rest of the function in case of Summary Stats.
> However, it doesn't work when the options are passed from JASP or testAnalysis, because the attribute doesn't exist (don't really understand why the jasptools::analysisOptions function creates it then... 

So you can do:
```
opts <- analysisOptions("correlationBayesian")
run(options=opts, dataset="debug.csv")
```
(Note that the analysis name doesn't need to be given to `run`)

> So, I was thinking whether it wouldn't be better to add an element to the options list on the beginning of each of the Summary Stats script that would tell the .setSeedJASP to not set any seed. We wouldn't have to go back to the analyses where the seed is implemented and only change the Summary Stats in this regard.

I think there are 3 options:
1) Add a `type` argument to the function used in multiple analyses and send in the type from these analyses, e.g., `type = "correlationBayesian"` (this can then be checked before calling `setSeed()`).
2) Check to see if the named item `setSeed` exists in the `options` list and if so call `setSeed()`.
3) In `setSeed()` only set the seed if the options list actually contains the proper items instead of throwing an error.

My preference would be 2 or 3.

FBartos:
ok, I commented the error message out for now. I think that the idea was to have a notification if someone adds seed in the analysis but forgets to add the menu. Once I add seeds to the summary-stats, I will add the error message back.

JASP-Engine/JASP/R/correlationbayesian.R
```diff
@@ -766,11 +768,12 @@ CorrelationBayesian <- function(jaspResults, dataset=NULL, options, ...) {
     return(jaspResults[[indexName]]$object)
 
   results <- list()
+  .setSeedJASP(options)
   for (method in names(bfObject))
     results[[method]] <- bstats::computeCorPosteriorLine(bfObject = bfObject[[method]], method = method, alternative = options[["alternative"]])
```
TimKDJ: If you do this in the for loop then the outcome of a test will be the same regardless of the number of tests

FBartos: Thanks, I missed that before.

TimKDJ:
The code looks good to me. Can you add a small entry explaining the seed option in the `input` section of the help files?

FBartos:
I added the description to the help files. The following analyses do not have help files yet, so I didn't add it there
Bayesian Contingency Tables
Bayesian Log-linear Regression

the text:
#### Repeatability
- Set seed: Value used to set the random number generator for exact reproducibility.

Resources/Help/analyses/ttestindependentsamples.md
```diff
@@ -47,6 +47,9 @@ The independent samples t-test allows the user to estimate the effect size and t
  - Exclude cases analysis by analysis: In case of multiple t-tests within a single analysis, each test will be conducted using all cases with valid data for the dependent variable for the particular t-test. Sample sizes may therefore vary across the tests. This options is selected by default.
  - Exclude cases listwise: In case of multiple t-tests within a single analysis, each t-test will be conducted using only cases with valid data for all dependent variables. Sample size is therefore constant across the tests.
 
+#### Repeatability
```
TimKDJ: This is the wrong file, it should be the bayesian independent samples ttest ;)

PR #3811, opened 2020-02-03
JAGS: bugfixes in parser and small cleanup

vandenman:
This PR does the following:

- Uses `"%1$s"` instead of `"%s"` + add one forgotten `gettext`.
- Catches crashes when computing Rhat.
- Fixes a bug in detecting data set columns in a model.
- Encodes columns names properly encoded on the C++ side.
- Removes old base64 encoding from R.
- "Deviance" can now be toggled instead of being always present.
- Fixes a visual bug where some numbers on the x-axis would fall of the plot (fixed by the ` + ggplot2::theme(plot.margin = ggplot2::margin(0, 10, 0, 0))` part).

### DIFF
--- a/JASP-Desktop/widgets/boundqmljagstextarea.cpp
+++ b/JASP-Desktop/widgets/boundqmljagstextarea.cpp
@@ -55,7 +55,7 @@ void BoundQMLJAGSTextArea::bindTo(Option *option)
 		{
 			std::vector<std::string> variables = columnsOption->variables();
 			for (const std::string& variable : variables)
-				_usedColumnNames.insert(QString::fromStdString(variable));
+				_usedColumnNames.insert(variable);
 		}
 		OptionVariables* parametersOption = dynamic_cast<OptionVariables*>(_options->get("parameters"));
 		if (parametersOption)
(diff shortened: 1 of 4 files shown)


JASP-Desktop/widgets/boundqmljagstextarea.cpp
```diff
@@ -92,17 +100,20 @@ void BoundQMLJAGSTextArea::checkSyntax()
 	// google: jags_user_manual (4.3.0) for documentation on JAGS symbols
 
 	// get the column names of the data set
-	std::vector<std::string> colnms = DataSetPackage::pkg()->getColumnNames();
-	std::set<std::string> columnNames(std::make_move_iterator(colnms.begin()), std::make_move_iterator(colnms.end()));
+	std::set<std::string> columnNames;
+	_textEncoded = tq(ColumnEncoder::columnEncoder()->encodeRScript(stringUtils::stripRComments(fq(_text)), &columnNames));
+	_usedColumnNames.clear();
```
JorisGoosen, line 103: Why not just change _usedColumnNames to `std::set<std::string>` and using it directly?

JorisGoosen: (If you want too, it is a blocker or anythin)

JorisGoosen (approved):

JASP-Engine/JASP/R/jagsModule.R
```diff
@@ -720,7 +674,7 @@ JAGS <- function(jaspResults, dataset, options, state = NULL) {
   )
 
   if (length(toAdd) > 0L)
-    errorMessage <- paste0(errorMessage, "\n\nIn addition:\n", toAdd)
+    errorMessage <- paste0(errorMessage, "\n\n", gettext("In addition:"), "\n", toAdd)
```
TimKDJ: This whole thing should probably be a `gettextf`  for those languages that do not go left->right

vandenman: This is not so much left to right, but rather top to bottom. I'll still change it though. Shall I escape the newlines?

TimKDJ: Yeah you can put those outside the gettext

JASP-Desktop/widgets/boundqmljagstextarea.cpp
vandenman: ok I changed it to `std::set<std::string>`

TimKDJ (approved):

PR #3821, opened 2020-02-05
Bayesian multinomial bugfix

vandenman:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/384

~~This approach, using a QML widget to determine the label orders, is a very unusual way to determine the order in which levels are shown. Basically, every analysis that handles nominal text then needs an invisible QML widget to display things in the correct order? To me, this looks really prone to error. 
Perhaps we can deprecate the reordering when we have plot editing?~~


JorisGoosen:
As an aside here: the above complaint is not valid, as the columns in the data are given as a (possibly ordered) factor with the needed order.

PR #3825, opened 2020-02-06
Alignment of TextField in a Group may lead to strange behaviour

boutinb:
Fix https://github.com/jasp-stats/INTERNAL-jasp/issues/703


JorisGoosen (changes requested):
Please use a Connection to preferencesModel instead of a real copy of uiScale.

PR #3827, opened 2020-02-06
Classical meta analysis bugfixes

vandenman:
Fixes:

jasp-stats/jasp-test-release/issues/473
jasp-stats/jasp-test-release/issues/476
jasp-stats/jasp-test-release/issues/477

### DIFF
--- a/JASP-Engine/JASP/R/classicalmetaanalysis.R
+++ b/JASP-Engine/JASP/R/classicalmetaanalysis.R
@@ -730,9 +730,7 @@ ClassicalMetaAnalysis <- function(jaspResults, dataset = NULL, options, ...) {
   ci.ub    <- rma.fit$yi + qnorm(rma.fit$level/2, lower.tail = FALSE) * sqrt(rma.fit$vi)
   xlims    <- c(-1, rma.fit$k+1)
   ylims    <- c(min(ci.lb), max(ci.ub))
-  ci.lb    <- round(ci.lb, digits = 2)
-  ci.ub    <- round(ci.ub, digits = 2)
-  ci.int   <- paste0(round(rma.fit$yi, 2), "[", ci.lb, ", ", ci.ub, "]")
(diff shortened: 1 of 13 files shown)


TimKDJ:
Very nice; much better!

My minor nitpicks would be (note that these things were already present in Anthony's changes, but absent in 0.11):
- Forest plot:
  - [ ] The median[interval] text does not have a space (for the model, bottom one)
  - [ ] The horizontal line is black on the top of the graph, but light gray on the bottom (between the studies and the model)
  - [ ] The x-axis looks non-jaspy with those tiny tick marks and little space between the tick labels and tick marks
- Q-Q Plot:
  - [ ] The gray area of the confidence interval is bounded to only about 40% of the entire horizontal width

They are minor, so I'll leave it up to you which you want to implement.

vandenman:
Okay that should fix all 4 of your comments. Let me know if there's anything else!

TimKDJ:
Ah yes it is getting there now, nice nice, the top and bottom line are still not the same though (although better than before old<->new:)
Screenshot 2020-02-06 at 15 56 48

And it randomly drops the top line when I resize it..

vandenman:
I managed to draw the lines in the margin and hopefully also fixed the issue where the lines would disappear.

image

TimKDJ:
The plot by itself looks great now!

... only the lines are intruding a little in the neighboring plot when you ask for the collection:
meta-analysis

vandenman:
aah I forgot that those are made into one plot instead of a collection.. (why though?).

vandenman:
Okay, I reverted the clipping behavior. I think that after the release we should remove the matrix plot and just make a collection of two plots (then we can re-enable the clipping). I don't see any reason to show the forest plot and funnel plot twice, but maybe that's just me.

vandenman:
Remotes can't be installed, thus Travis crashes...

TimKDJ:
I guess they updated remotes and the binaries are lagging behind?

TimKDJ (approved):
Nice work!

PR #3829, opened 2020-02-06
PlotRobustnessSequential fixed incorrect evidence label

vandenman:
Fixes jasp-stats/jasp-test-release/issues/505

There's something fundamentally broken in the API of PlotRobustnessSequential (all analyses pass as BF a BF10 regardless of the bfType), but I think it works now.


JASP-Engine/JASPgraphs/R/PlotRobustnessSequential.R
```diff
@@ -286,7 +286,8 @@ PlotRobustnessSequential <- function(
       idx <- findInterval(val, c(1, 3, 10, 30, 100), rightmost.closed = FALSE)
       evidenceLevel <- c("Anecdotal", "Moderate", "Strong", "Very~Strong", "Extreme")[idx]
       
-      if (BF > 1 && bfType == "BF01" || BF < 1 && bfType != "BF01")
+      BF01 <- 1 / BF
```
TimKDJ: Would this break if someone supplied `BF = 1 / BF10, bfType = "BF01"`?
If so, it might be good to remove `bfType` and rename `BF` to `BF10` -- that way it becomes a little more likely that people will continue to supply only BF10's

vandenman:
@TimKDJ I think this is ready for another round of reviews.

TimKDJ:
I tested two analyses: Bayesian One Sample TTest and Bayesian Correlation.
The ttest seemed fine now, but in the sequential plot of the Bayesian Correlation both problems still seem present (piechart flipping around and the evidence text changing)

vandenman:
Good catch! Should be fixed now.

PR #3837, opened 2020-02-06
Network analysis - remove layout from data set if checkbox is unchecked

vandenman:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/513

Removes the columns from the data set when the checkbox is unchecked.

Just setting the value to `""` wasn't enough, I also had to trigger some signal so this actually got processed, which I did using `doEditingFinished()`. Let me know if I should do this differently!


JorisGoosen:
Yeah, maybe a `clear()` function should be added or something, for now your solution is perfectly fine though!

PR #3840, opened 2020-02-07
Fix gettextf bug in bayesian linear regression

vandenman:
see title. If `generalNote = NULL`, then `gettextf("%s", generalNote) = NULL`, which causes the footnote to complain.


boutinb:
This solution generates 2 entries in the translation files... Is it not better to do something like this:
generalNote <- gettextf("%s Table displays only a subset of models; to see all models, select \"No\" under \"Limit No. Models Shown\".", is.null(generalNote) ? "", generalNote)

vandenman:
@boutinb Sure, I was a bit in doubt what to do with the additional space. Now I just assume that all languages want a space after a period.

JohnnyDoorn (approved):

PR #3831, opened 2020-02-06
Fix standardized residual plot in meta analysis

TimKDJ:
If you unticked "Q-Q plot standardized residuals" then the plot collection crashed. The plot was also not completely correct after I fixed the error.

@vandenman: I don't think we should get merge conflicts, but I'll merge your PR first. I have to wait with the creation of the reference plot anyway.

PR #3847, opened 2020-02-10
Backward compatibility with FormulaType input

boutinb:
Fix https://github.com/jasp-stats/jasp-test-release/issues/551

My branch was not up-to-date, and has still other commits. This has been fixed.

### DIFF
--- a/JASP-Desktop/analysis/options/optionstring.cpp
+++ b/JASP-Desktop/analysis/options/optionstring.cpp
@@ -16,6 +16,7 @@
 //
 
 #include "optionstring.h"
+#include "utils.h"
 
 using namespace std;
 
@@ -33,7 +34,11 @@ Json::Value OptionString::asJSON() const
 
 void OptionString::set(const Json::Value &value)
 {
-	setValue(value.asString());
+	if (value.isInt() || value.isUInt() || value.isDouble())
+		// Why the json library cannot convert it to string???
+		setValue(Utils::doubleToString(value.asDouble()));
+	else
+		setValue(value.asString());
 }
 
 Option *OptionString::clone() const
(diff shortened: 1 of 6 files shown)


JorisGoosen:
Is there a reason you are including #3846 and #3831 in this PR?

JASP-Desktop/analysis/options/optionstring.cpp
```diff
@@ -33,7 +34,11 @@ Json::Value OptionString::asJSON() const
 
 void OptionString::set(const Json::Value &value)
 {
-	setValue(value.asString());
+	if (value.isInt() || value.isUInt() || value.isDouble())
+		// Why the json library cannot convert it to string???
```
JorisGoosen, line 38: Because it wants you to be aware of the datatype in there ;)

JASP-Engine/JASP/R/binomialtest.R
```diff
@@ -187,13 +198,13 @@ BinomialTest <- function(jaspResults, dataset = NULL, options, ...) {
     
   # Add footnote: Alternative hypothesis
   if (options$hypothesis == "lessThanTestValue")
-    note <- gettext("For all tests, the alternative hypothesis specifies that the proportion is less than ")
+    note <- gettextf("For all tests, the alternative hypothesis specifies that the proportion is less than %s.", options$testValueUnparsed)
```
JorisGoosen, line 194: Shouldnt the option be called `testFormula` or something?

JorisGoosen (changes requested):

JASP-Engine/JASP/R/binomialtest.R
boutinb: The analysis should not have to parse the formula at the beginning of the analysis: this should be done automatically in the next JASP version. The analysis should just use the name testValue and should not worry about the kind of field the interface uses.

PR #3853, opened 2020-02-11
Bayesian t-tests, many bugfixes

vandenman:
Fixes:

https://github.com/jasp-stats/jasp-test-release/issues/571
https://github.com/jasp-stats/jasp-test-release/issues/572
https://github.com/jasp-stats/jasp-test-release/issues/576
https://github.com/jasp-stats/jasp-test-release/issues/577

Adds progress bars for both the robustness and the sequential plots.

I can't replicate https://github.com/jasp-stats/jasp-test-release/issues/573 so I didn't fix that one.

https://github.com/jasp-stats/jasp-test-release/issues/575 should be fixed by #3829


koenderks (approved):

JASP-Engine/JASP/R/commonbayesianttest.R
```diff
@@ -1376,7 +1389,15 @@
     k <- idData + 1
 
 
-    while ((i <= length(x) | j <= length(y)) & k <= length(BF10)) {
+    nTicks <- length(BF10) - i
+    if (plotDifferentPriors)
+      nTicks <- 3L * nTicks
+    startProgressbar(nTicks, sprintf(ngettext(totalPlots,
+                                              "Sequential analysis",
```
TimKDJ: Maybe "Performing sequential analysis"?

TimKDJ (commented):
Screenshot 2020-02-13 at 15 44 18
I think dots > NaN when the table is not ready (note the lack of grouping variable)

JASP-Engine/JASP/R/commonbayesianttest.R
vandenman: `nchar("Performing sequential analysis 10 / 11")` has 38 characters. I can see this introducing bugs in specific translations. For example, `nchar("uitvoeren van sequentiële analyse 10 / 11") = 41`, and 41 > 40 so the progress bar will throw an error, but only for a Dutch translation.

Preferably, I'd stay way below the limit of 40 so that translators can freely make longer translations without having to worry about things like these. I'm fine with "Performing sequential analysis", but then maybe we should increase the limit of 40 characters.

The same goes for the robustness check of course.

vandenman:
@TimKDJ There should now be empty cells when a variable is added but the analysis is not ready, rather than NaN. Also, I found another bug in the interaction of hypothesis type and BF type. I've hopefully fixed that, but please check :). To avoid future issues with this, I've added unit tests for all possible combinations of the hypothesis and BF type for the prior posterior plot, robustness plot, and sequential plot. Hence, the 9 * (4 + 3 + 3) = 90 additional SVG files.

TimKDJ (changes requested):
Just so this doesn't get merged yet. I'll check out your changes

JASP-Engine/JASP/R/commonbayesianttest.R
TimKDJ: After https://github.com/jasp-stats/jasp-desktop/pull/3863 you get more characters and never a crash

vandenman: This comment is obsolete btw, https://github.com/jasp-stats/jasp-desktop/pull/3853/commits/e9b4f85950aed6004cd13a477f9996f7946c1578 shows I changed it to your suggestion.

JASP-Engine/JASP/R/commonbayesianttest.R
```diff
@@ -1586,6 +1595,14 @@
     #   }
     # }
 
+    nTicks <- nrow(subDataSet)
+    if (plotDifferentPriors)
+      nTicks <- 3L * nTicks
+    startProgressbar(nTicks, sprintf(ngettext(totalPlots,
+                                              "Sequential analysis",
```
TimKDJ: Can you add the "running" here as well?

TimKDJ: Other than this it looks good!

vandenman: good catch!

PR #3855, opened 2020-02-11
Make .hasErrors() catch columns with only NA's

TimKDJ:
Fixes
https://github.com/jasp-stats/jasp-test-release/issues/568 (partially)
https://github.com/jasp-stats/jasp-test-release/issues/562
https://github.com/jasp-stats/jasp-test-release/issues/574
https://github.com/jasp-stats/jasp-test-release/issues/570
https://github.com/jasp-stats/jasp-test-release/issues/559
https://github.com/jasp-stats/jasp-test-release/issues/558

The reason this wasn't changed earlier, is because analyses still used`.readDataSetHeader()` which returned a named zero-row data.frame. And this dataset is identical to what would be provided after calling `na.omit()` on a data.frame with `debNaN` in it. Considering  `readDataSetHeader` is no longer needed after the jaspResults rewrites, I can finally make this change.

I still needed to make some changes to analyses that relied on `.hasErrors()` to deal with  situations in which the analyses didn't actually want to perform any error checking.

### DIFF
--- a/JASP-Engine/JASP/R/commonerrorcheck.R
+++ b/JASP-Engine/JASP/R/commonerrorcheck.R
@@ -115,11 +115,10 @@
 }
 
 
-.hasErrors <- function(dataset, perform="run", type=NULL, custom=NULL, message='default', exitAnalysisIfErrors=FALSE, allowEmptyDataset = TRUE, ...) {
+.hasErrors <- function(dataset=NULL, type=NULL, custom=NULL, message='default', exitAnalysisIfErrors=FALSE, ...) {
   # Generic error checking function.
   # Args:
   #   dataset: Normal JASP dataset.
-  #   perform: 'run' or 'init'.
   #   type: List/vector of strings containing check types.
   #   message: 'short' or 'default' should only the first failure of a check be reported in footnote style ('short'), or should every check failure be mentioned in multi-line form.
   #   exitAnalysisIfErrors: Boolean, should the function simply return its results (FALSE), or abort the entire analysis when a failing check is encountered (TRUE).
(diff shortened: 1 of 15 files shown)


JASP-Engine/JASP/R/commonerrorcheck.R
```diff
@@ -115,11 +115,10 @@
 }
 
 
-.hasErrors <- function(dataset, perform="run", type=NULL, custom=NULL, message='default', exitAnalysisIfErrors=FALSE, allowEmptyDataset = TRUE, ...) {
```
vandenman, line 118: really happy to see this disappear :smile:

vandenman (commented):
So, the bug in `.hasErrors` is fixed but the error message are a bit suboptimal. Maybe we should make the errors thrown more informative somehow, see below for some examples. We should probably do that in a different PR though since I guess it needs to be handled analysis by analysis. 

jasp-stats/jasp-test-release#558

image

jasp-stats/jasp-test-release#562

image

TimKDJ:
The above would be fixed (except for the custom one) if `na.omit()` is called after `.hasErrors()`, because all necessary information is lost if you provide a zero-row data.frame to `.hasErrors()` (mainly: what column caused the rows to be empty?)

TimKDJ:
I guess a nice moment to solve this properly (because I agree this is not optimal), is when the dataset is directly provided to the analysis in the dataset param. Then we can simply move the error check to be the first thing an analysis does, before it performs its data cleaning.

For now I'll merge this though, as this situation is kinda niche to begin with.

PR #3856, opened 2020-02-12
Bug fixes

koenderks:
In this pull request:

**Audit**
* https://github.com/jasp-stats/jasp-test-release/issues/566
* https://github.com/jasp-stats/jasp-test-release/issues/514
* https://github.com/jasp-stats/jasp-test-release/issues/499

**Bain**
* https://github.com/jasp-stats/jasp-test-release/issues/585
* https://github.com/jasp-stats/jasp-test-release/issues/580
* https://github.com/jasp-stats/jasp-test-release/issues/563

**Machine Learning**
* https://github.com/jasp-stats/jasp-test-release/issues/583
* https://github.com/jasp-stats/jasp-test-release/issues/561
* https://github.com/jasp-stats/jasp-test-release/issues/497
* https://github.com/jasp-stats/jasp-test-release/issues/495
* https://github.com/jasp-stats/jasp-test-release/issues/491
* https://github.com/jasp-stats/jasp-test-release/issues/493

For most Machine Learning analyses I just added a try for the errors. Should we extend that?

### DIFF
--- a/JASP-Engine/JASP/R/bainttestbayesianonesample.R
+++ b/JASP-Engine/JASP/R/bainttestbayesianonesample.R
@@ -133,7 +133,7 @@ BainTTestBayesianOneSample <- function(jaspResults, dataset, options, ...) {
                             "equalNotEqual"     = paste0(note, options[["testValue"]], ".", gettext(" The posterior probabilities are based on equal prior probabilities.")),
                             "equalBigger"       = paste0(note, options[["testValue"]], ".", gettext(" The posterior probabilities are based on equal prior probabilities.")),
                             "equalSmaller"      = paste0(note, options[["testValue"]], ".", gettext(" The posterior probabilities are based on equal prior probabilities.")),
-                            "biggerSmaller"     = gettextf("The hypothesis H1 specifies that the mean is bigger than %1$s and the hypothesis H2 specifies that the mean is smaller than %2$s. The posterior probabilities are based on equal prior probabilities.", options[["testValue"]], options[["testValue"]]),
+                            "biggerSmaller"     = gettextf("The hypothesis H1 specifies that the mean is bigger than %1$s and the hypothesis H2 specifies that the mean is smaller than %1$s. The posterior probabilities are based on equal prior probabilities.", options[["testValue"]]),
                             "equalBiggerSmaller"= gettextf("%1$s %2$s is tested against the other hypotheses. H1 states that the mean is bigger than %2$s and H2 states that the mean is smaller than %2$s. The posterior probabilities are based on equal prior probabilities.", note, options[["testValue"]]))
   bainTable$addFootnote(message=message)
 
--- a/JASP-Engine/JASP/R/bainttestbayesianonesample.R
+++ b/JASP-Engine/JASP/R/bainttestbayesianonesample.R
@@ -133,8 +133,8 @@ BainTTestBayesianOneSample <- function(jaspResults, dataset, options, ...) {
                             "equalNotEqual"     = paste0(note, options[["testValue"]], ".", gettext(" The posterior probabilities are based on equal prior probabilities.")),
                             "equalBigger"       = paste0(note, options[["testValue"]], ".", gettext(" The posterior probabilities are based on equal prior probabilities.")),
                             "equalSmaller"      = paste0(note, options[["testValue"]], ".", gettext(" The posterior probabilities are based on equal prior probabilities.")),
-                            "biggerSmaller"     = gettextf("The hypothesis H1 specifies that the mean is bigger than %s and the hypothesis H2 specifies that the mean is smaller than %s. The posterior probabilities are based on equal prior probabilities.", options[["testValue"]]),
-                            "equalBiggerSmaller"= gettextf("%1$s %2$s is tested against the other hypotheses. H1 states that the mean is bigger than %2$s and H2 states that the mean is smaller than %2$sn. The posterior probabilities are based on equal prior probabilities.", note, options[["testValue"]]))
+                            "biggerSmaller"     = gettextf("The hypothesis H1 specifies that the mean is bigger than %1$s and the hypothesis H2 specifies that the mean is smaller than %2$s. The posterior probabilities are based on equal prior probabilities.", options[["testValue"]], options[["testValue"]]),
+                            "equalBiggerSmaller"= gettextf("%1$s %2$s is tested against the other hypotheses. H1 states that the mean is bigger than %2$s and H2 states that the mean is smaller than %2$s. The posterior probabilities are based on equal prior probabilities.", note, options[["testValue"]]))
   bainTable$addFootnote(message=message)
 
   bainTable$addCitation(.bainGetCitations())
--- a/JASP-Engine/JASP/R/commonMachineLearningClassification.R
+++ b/JASP-Engine/JASP/R/commonMachineLearningClassification.R
@@ -623,7 +623,7 @@
   if(!ready) return()
 
   if(length(options[["predictors"]]) < 2){
-    andrewsCurve$setError("Andrews curves require a minimum of 2 predictor variables.")
+    andrewsCurve$setError(gettext("Andrews curves require a minimum of 2 predictor variables."))
     return()
   }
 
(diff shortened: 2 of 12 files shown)


JASP-Engine/JASP/R/bainttestbayesianonesample.R
```diff
@@ -133,8 +133,8 @@ BainTTestBayesianOneSample <- function(jaspResults, dataset, options, ...) {
...
                             "equalSmaller"      = paste0(note, options[["testValue"]], ".", gettext(" The posterior probabilities are based on equal prior probabilities.")),
-                            "biggerSmaller"     = gettextf("The hypothesis H1 specifies that the mean is bigger than %s and the hypothesis H2 specifies that the mean is smaller than %s. The posterior probabilities are based on equal prior probabilities.", options[["testValue"]]),
-                            "equalBiggerSmaller"= gettextf("%1$s %2$s is tested against the other hypotheses. H1 states that the mean is bigger than %2$s and H2 states that the mean is smaller than %2$sn. The posterior probabilities are based on equal prior probabilities.", note, options[["testValue"]]))
+                            "biggerSmaller"     = gettextf("The hypothesis H1 specifies that the mean is bigger than %1$s and the hypothesis H2 specifies that the mean is smaller than %2$s. The posterior probabilities are based on equal prior probabilities.", options[["testValue"]], options[["testValue"]]),
```
TimKDJ: You don't need to repeat the testValue argument twice, just put `%1$s` for both cases

JASP-Engine/JASP/R/commonMachineLearningClassification.R
```diff
@@ -613,6 +622,11 @@
 
   if(!ready) return()
 
+  if(length(options[["predictors"]]) < 2){
+    andrewsCurve$setError("Andrews curves require a minimum of 2 predictor variables.")
```
TimKDJ: `gettext()`

JASP-Engine/JASP/R/commonMachineLearningClustering.R
```diff
@@ -107,9 +109,14 @@
     } else if(type == "randomForest"){
       clusterResult <- .randomForestClustering(dataset, options, jaspResults)
     }
-    jaspResults[["clusterResult"]] <- createJaspState(clusterResult)
-    jaspResults[["clusterResult"]]$dependOn(options = c("predictors", "noOfClusters","noOfRandomSets", "noOfIterations", "algorithm", "modelOpt", "seed",
-                                                      "maxClusters", "seedBox", "scaleEqualSD", "m", "distance", "linkage", "eps", "minPts", "noOfTrees", "maxTrees", "optimizationCriterion"))
+  })
+
+  if(isTryError(p))
+   JASP:::.quitAnalysis(gettextf("An error occurred in the analysis: %s", .extractErrorMessage(p)))
```
TimKDJ: I don't think this will add much to the user experience if we're being honest ;)
All that changes for users is that it goes from an unexpected error with advice to report it, to an error that gives them less information on what to do. 

Let's not implement this, and instead try to figure out on a bug-by-bug basis if we can find out why the bug occurs. If you can't find the cause for some problems, we'll move them to internal and deal with them when we have more time.

TimKDJ (changes requested):
**Bain**
---
> * https://github.com/jasp-stats/jasp-test-release/issues/585

In 0.11 bain had no trouble dealing with the computation:
0.11<->0.12
Screenshot 2020-02-13 at 14 20 07
Especially given how uninformative the message is, this doesn't seem good. What changed?

---

> * https://github.com/jasp-stats/jasp-test-release/issues/580

Would be good to make this a bit more uniform (note the difference between the first variable slot being empty or the second):
Screenshot 2020-02-13 at 14 30 10

---

**Machine Learning**
---
> * https://github.com/jasp-stats/jasp-test-release/issues/493

If you make it a tiny bit wider it doesn't cut off the label by default:
Screenshot 2020-02-13 at 15 08 37

---

koenderks:
> ## **Bain**
> > * [jasp-stats/jasp-test-release#585](https://github.com/jasp-stats/jasp-test-release/issues/585)
> 
> In 0.11 bain had no trouble dealing with the computation:
> 0.11<->0.12
> Screenshot 2020-02-13 at 14 20 07
> Especially given how uninformative the message is, this doesn't seem good. What changed?
> 
So what happends is that JASP apparently reads in the value "100000" as 1+e5 and not as 100000, crashing all kinds of stuff in bain. Can we read in the value "10000" as 10000 by default? I made https://github.com/jasp-stats/jasp-test-release/issues/612 for this that would solve this problem.
> > * [jasp-stats/jasp-test-release#580](https://github.com/jasp-stats/jasp-test-release/issues/580)
> 
> Would be good to make this a bit more uniform (note the difference between the first variable slot being empty or the second):
> Screenshot 2020-02-13 at 14 30 10
> 
I don't nessecarily get what you mean? What are you suggesting as a change here?
> ## **Machine Learning**
> > * [jasp-stats/jasp-test-release#493](https://github.com/jasp-stats/jasp-test-release/issues/493)
> 
> If you make it a tiny bit wider it doesn't cut off the label by default:
> Screenshot 2020-02-13 at 15 08 37

I believe it depends on computers, on my computer the label is not cut off by default. I'll make it bigger?

JASP-Engine/JASP/R/commonMachineLearningClustering.R
koenderks: Should we not keep it in general? Always better to have a try system than none?

TimKDJ: We have a try system: `try(analysis)`. That's also why JASP doesn't crash hard if an analysis error is not caught. By doing `try(package) .quitAnalysis(`) the only thing that is accomplished is that the `try(analysis)` does nothing

TimKDJ:
Apologies for my formatting, it makes this discussion a little tricky...

> I don't nessecarily get what you mean? What are you suggesting as a change here?

The top row has a footnote and is empty 
The bottom row has no footnote but instead does have the hypothesis specification

I just figured it'd be good if both rows showed the same behaviour for an incomplete variable pair. Regardless if the first variable or second is the missing one

TimKDJ:
> I believe it depends on computers, on my computer the label is not cut off by default. I'll make it bigger?

Yeah not sure if there's a more universal way to ensure that hard coded values at least always fit (@vandenman?)

vandenman:
Two possible solutions:

```r
library(ggplot2)
# very long label
mtcars$cyl <- paste0("hhhhhooasdoooooooiiiiiiii", mtcars$cyl)

g <- ggplot(mtcars, aes(x = mpg, y = disp, group = factor(cyl), color = factor(cyl))) + 
  geom_point() + theme(legend.position = "top")
g
```
In the plot below, the last legend key has it's "8" cut off:
image

1. Add a row to the legend
```r
g2 <- g + scale_color_discrete(guide = ggplot2::guide_legend(nrow = 2)) 
g2
```
The legend title is a bit ugly here, but you don't have one so you wouldn't get this I think. 
image

2. Tweak the margins:
```r
# push the legend left by adjusting the margin
g3 <- g + theme(legend.margin = ggplot2::margin(0, 20, 0, 0))
g3
```
Now, the legend title dropped off, but I think that might work because your legend has a lot of space on the left.
image

koenderks:
> Apologies for my formatting, it makes this discussion a little tricky...
> 
> > I don't nessecarily get what you mean? What are you suggesting as a change here?
> 
> The top row has a footnote and is empty
> The bottom row has no footnote but instead does have the hypothesis specification
> 
> I just figured it'd be good if both rows showed the same behavior for an incomplete variable pair. Regardless of the first variable or second is the missing one

You are right, I didn't see at first. I've implemented it so that now it shows the same behavior:
image

With respect to the plots, maybe we should fix the spacing of legends somewhere centralized (JASPgraphs?) instead individually.

JASP-Engine/JASP/R/commonMachineLearningClustering.R
koenderks: I will remove it

vandenman:
> With respect to the plots, maybe we should fix the spacing of legends somewhere centralized (JASPgraphs?) instead individually.

Well, you're doing something with the legends that nobody else does. So we could make the plot area of all plots smaller, (meaning smaller plot relative to the png size) just so it works out here. But then if somebody else puts the legend right of the plot instead of above then they'll need completely different settings..

TimKDJ:
Would be good to have a better general solution, but this can be done later

PR #3860, opened 2020-02-13
Make plot behaviour more consistent in descriptives and add error handling

TimKDJ:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/379, fixes https://github.com/jasp-stats/jasp-test-release/issues/494, fixes https://github.com/jasp-stats/jasp-test-release/issues/498

- Adds error handling where it was either incomplete or absent
- Performs error handling before omitting NA’s
- Plots always omit variables of types that they do not support (rather than sometimes showing an error, sometimes omitting)
- Scatterplots now also work with nominal int data (to be consistent with other plots)
- Removes NA’s from scatterplot data instead of throwing an error
- Adds missing dependency to correlation matrix

### DIFF
--- a/JASP-Engine/JASP/R/descriptives.R
+++ b/JASP-Engine/JASP/R/descriptives.R
@@ -585,10 +585,10 @@ Descriptives <- function(jaspResults, dataset, options) {
   }
 
   plotMat <- matrix(list(), l, l)
-	axisBreaks <- vector("list", l)
+  axisBreaks <- vector("list", l)
 
-	# minor adjustments to plot margin to avoid cutting off the x-axis labels
-	adjMargin <- ggplot2::theme(plot.margin = ggplot2::unit(c(.25, .40, .25, .25), "cm"))
+  # minor adjustments to plot margin to avoid cutting off the x-axis labels
+  adjMargin <- ggplot2::theme(plot.margin = ggplot2::unit(c(.25, .40, .25, .25), "cm"))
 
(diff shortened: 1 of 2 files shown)


JASP-Engine/JASP/R/descriptives.R
```diff
@@ -612,19 +597,29 @@ Descriptives <- function(jaspResults, dataset, options) {
...
   # now do off-diagonal and use the same breaks
   for (row in seq_len(l-1)) {
   	for (col in seq(row+1, l)) {
-  		plotMat[[row, col]] <- .plotScatterDescriptives(
-  			xVar    = dataset[[variables[[col]]]],
-  			yVar    = dataset[[variables[[row]]]],
-  			xBreaks = axisBreaks[[col]]$x,
-  			yBreaks = axisBreaks[[row]]$x
-  		) + adjMargin
+	    if (variable.statuses[[row]]$error != "") {
```
vandenman: Mixing tabs and spaces eh?

vandenman (approved):
Bugs are fixed! The comment about mixing tabs & spaces is not that important but it would be nice to clean that up before merging.

JASP-Engine/JASP/R/descriptives.R
TimKDJ: In my defence the tabs were already there.. I'll make everything spaces though

vandenman: I know you didn't cause it but it's one of the easy things to improve :)

PR #3861, opened 2020-02-13
bayesian t-test seed rewrite

FBartos:
The prior & posterior plot function used sampling for non-informative Cauchy prior even though we have an implementation with numerical integration for more general cases and doesn't require sampling. Thus the seeds for t-test are now needed only for the Mann-Whitney test in (independent samples) - I removed the seeds from the one-sample and paired-samples t-test.

It will fail the tests because the prior-and-posterior plots are apparently slightly different, but it _should_ work fine.

addresses: https://github.com/jasp-stats/jasp-test-release/issues/525


EJWagenmakers:
Also no sampling required for the one-sided tests?

FBartos:
no, it seems to be done by numerical integration

EJWagenmakers:
Can you run the one-sided test a few times with a different seed, to confirm it gives the same results? It used to vary noticeably, maybe the code has been updated..

FBartos:
Just did. It seems to be the same.

vandenman:
The code looks fine. 

Perhaps `SetSeed` element should only be enabled in QML when the Mann-Whitney is selected?

However, the unit tests do fail correctly. I reran them locally and some credible intervals have changed, in addition to more things. If you run `manageTestPlots("TTestBayesianIndependentSamples")` you'll notice that there are many changes. Probably something to look at.

FBartos:
hmm, I look into the changes and the differences didn't seem that big to me. I think that the differences between sampling and numerical integration are responsible for that. I'm adding all of the charts here for comparison:

image
image

another difference:
CI = [-0.303, 0.439] vs [-0.303, 0.436] 

image
image

median =- 0.189 vs -0.191

this is the biggest difference, however, it's entirely due to different plotting methods used for the sampled data
image
image

image
image

vandenman:
After rebasing on your branch, the changes look a lot more minor. So you have to:

1. Rebase (there have been some changes to the Bayesian t-tests since you opened this PR).
2. Call `manageTestPlots(analysis name)` for the analyses with failing tests (or call `manageTestPlots()` without argument if you don't mind the wait).
3. Update the test figures using the Shiny app.
4. Push.

Then Travis should be happy.

TimKDJ:
Why did the summary stats plots change (that's what travis fails on)? They should be unaffected by the seed changes right?

PR #3863, opened 2020-02-14
Allow longer progressbar labels, truncate if necessary

TimKDJ:
This should be a little less strict. 80 is also a completely random value. I just don't want people to write essays in the label, because the progressbar usually only appears for a short while


vandenman:
A limit of 80 characters is much better! I can't judge the .js, but it looks good to me.

PR #3866, opened 2020-02-14
Quill notes - on focusout show/hide toolbar

akashrajkn:
Fixes jasp-stats/jasp-test-release#501

### DIFF
--- a/JASP-Desktop/html/css/darkTheme-jasp.css
+++ b/JASP-Desktop/html/css/darkTheme-jasp.css
@@ -559,7 +559,6 @@ td.squash-left
 
 
 .jasp-notes {
-  border-top:     1px solid #5c5c5c;
   border-bottom:  1px solid #5c5c5c;
   border-radius:  0;
   color:          #000;
@@ -570,6 +569,11 @@ td.squash-left
   margin-bottom:  0.6em ;
   filter:         invert(1);
 }
+
+.jasp-notes-border {
+  border-top:     1px solid #5c5c5c;
+}
+
 
 .jasp-firstNote-note {
   margin-bottom: 0.5em;
@@ -652,4 +656,11 @@ pre {
 
 ::-webkit-scrollbar-thumb:hover {
  box-shadow: inset 0 0 0 10px #4784B2;
+}
+
+
+/* Custom values */
+.ql-toolbar.ql-snow {
+  /* light theme uses the default value */
+  border: 1px solid #5c5c5c;
 }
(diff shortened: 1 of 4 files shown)


TimKDJ:
Ah, this is much better. There is still one flaw I found though:
If I have two noteboxes open and I click from the top notebox in the bottom one, then I do not get a toolbar in the bottom notebox. I attached a .jasp file that hopefully clarifies what I mean.

[notes.zip](https://github.com/jasp-stats/jasp-desktop/files/4204725/notes.zip)

JorisGoosen:
Looks good!
Only bug I saw was the one mentioned by Tim.

JorisGoosen:
@akashrajkn did you have a change to look at the last bug?

akashrajkn:
Yes, I am taking a look at it right now :)

akashrajkn:
@TimKDJ @JorisGoosen  this is not a bug. I am using the `focusout` and `click` events to hide/display the toolbar. On `focusout`, the toolbar is removed. Then (for short notes), the click event is registered on a different element. If you click the bottom note first and then the top note, we get the expected behavior. (This is annoying - I will search for an alternative. If you have suggestions, let me know)

JorisGoosen:
Can't you just leave the click event registered on all noteboxes all the time?

akashrajkn:
Commit 5dc3d1b fixes the issue. I used `mousedown` instead of `click` event ;)

TimKDJ:
This one is only if you have time, if not I'll make an issue for it for the future:
The click after editing a note should never be registered as a "select-analysis-click". This still works fine in development, so this PR breaks it.

Steps to reproduce:
1. Add a random analysis
2. Add a note via the "Results" menu (top most element in the output)
3. Edit the note and click on the analysis (the note is saved and the analysis is not selected: good behaviour)
4. Edit the note again and click on the analysis once more (the note is saved, but the analysis is also selected. This is not intended behaviour and it also causes the note to be scrolled out of view)

akashrajkn:
I'll check this out tomorrow

akashrajkn:
I am not able to reproduce this (of course, when I click the analysis twice after editing the main note, the analysis is focused - I think this is the intended behaviour)

[EDIT] I see that clicking the analysis header, focuses the analysis. Do you mean this?

TimKDJ:
[Screen Recording 2020-02-26 at 16.05.18.zip](https://github.com/jasp-stats/jasp-desktop/files/4256433/Screen.Recording.2020-02-26.at.16.05.18.zip)
Ignore the last 10 seconds ^.^

This time around it didn't even require two clicks. Just clicking from the note on an analysis serves as a select-analysis-click, which it shouldn't. You can also try this in e.g., 0.11.1

PR #3868, opened 2020-02-18
Sort out some small interface, gettext and footnote issues

TimKDJ:
Fixes jasp-stats/jasp-test-release#554, fixes jasp-stats/jasp-test-release#534, fixes jasp-stats/jasp-test-release#539, fixes jasp-stats/jasp-test-release#605, fixes jasp-stats/jasp-test-release#604

### DIFF
--- a/JASP-Engine/JASP/R/ttestonesample.R
+++ b/JASP-Engine/JASP/R/ttestonesample.R
@@ -154,26 +154,33 @@ TTestOneSample <- function(jaspResults, dataset = NULL, options, ...) {
   
   
   ### check the directionality
-  
+  tMessage <- NULL
+  wMessage <- NULL
+  sep <- "."
+  if (optionsList$wantsStudents && optionsList$wantsWilcox)
+    sep <- ";"
   if (options$hypothesis == "greaterThanTestValue") {
     direction <- "greater"
-    tMessage <- gettext("For the Student t-test, the alternative hypothesis specifies that the mean is greater than ")
-    wMessage <- gettext("For the Wilcoxon test, the alternative hypothesis specifies that the median is greater than ")
+    if (optionsList$wantsStudents)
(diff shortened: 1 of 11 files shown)


JASP-Engine/JASP/R/ttestonesample.R
```diff
@@ -154,26 +154,33 @@ TTestOneSample <- function(jaspResults, dataset = NULL, options, ...) {
...
+  wMessage <- NULL
+  sep <- "."
+  if (optionsList$wantsStudents && optionsList$wantsWilcox)
+    sep <- ";"
   if (options$hypothesis == "greaterThanTestValue") {
     direction <- "greater"
-    tMessage <- gettext("For the Student t-test, the alternative hypothesis specifies that the mean is greater than ")
-    wMessage <- gettext("For the Wilcoxon test, the alternative hypothesis specifies that the median is greater than ")
+    if (optionsList$wantsStudents)
+      tMessage <- gettextf("For the Student t-test, the alternative hypothesis specifies that the mean is greater than %1$s%2$s", options$testValue, sep)
```
vandenman: Soo... this is going to create 6 things to translate. But can't it be:

```r
gettextf("For the %1$s, the alternative the alternative hypothesis specifies 
          that the %2$s is %3$s %4$s.",

  "Student t-test"|"Wilcoxon test",
  "mean"|"median",
  "greater than"|"less than"|"different from",
  options$testValue
)
```

Maybe that becomes too messy though. Also, to ease translation, I think it's a bad idea to use semicolons. That may create sentence structures that can't be translated easily. For instance, I don't think Japanese uses semicolons at all ([ref](https://www.livinglanguage.com/community/discussion/1339/japanese-punctuation-marks)), and now there's one hardcoded.

TimKDJ: I reduced it down to 2 (but not 1, because having 5 %'s in there seems hellish to translate) and improved the hardcoded ;

vandenman (approved):

PR #3869, opened 2020-02-18
Bayesian Meta Analysis small changes & Meta-analysis hyphens added

sophieberkhout:
Bayesian meta-analysis
- Changed error message
- Fixed some footnotes
- Added help file

Meta-analysis module
- Added hyphens to meta-analysis module

Will add a dataset & make a JASP file for Bayesian meta-analysis after I can download a new nightly.

### DIFF
--- a/JASP-Engine/JASP/R/bayesianmetaanalysis.R
+++ b/JASP-Engine/JASP/R/bayesianmetaanalysis.R
@@ -386,7 +386,7 @@ BayesianMetaAnalysis <- function(jaspResults, dataset, options) {
   }
   
   if(isTryError(p)){
-    .quitAnalysis(gettextf("The model could not fit. Error while running R code from the metaBMA package: %s", .extractErrorMessage(p))) 
+    .quitAnalysis(gettextf("The model could not be fit. Please try different settings.")) 
   }
   
   return(results)
@@ -610,7 +610,8 @@ BayesianMetaAnalysis <- function(jaspResults, dataset, options) {
   
   footnoteRandomBFtau <- gettextf("Bayes factor of the random effects H%1$s over the fixed effects H%1$s.", "\u2081")
   
-  footnoteAverage <- gettextf("Posterior estimates are based on the models that assume an effect to be present. The Bayes factor is based on all four models: random effects H%1$s & H%2$s over the fixed effects H%1$s & H%2$s.", "\u2080", "\u2081")
+  footnoteAverage <- gettextf("Posterior estimates are based on the models that assume an effect to be present. The Bayes factor is based on all four models: fixed effects H%2$s & random effects H%2$s over the fixed effects H%1$s & random effects H%1$s.", "\u2080", "\u2081")
+  footnoteAverageBFtau <- gettextf("Model averaged posterior estimates for %3$s are not yet available, but will be added in the future. The Bayes factor is based on all four models: random effects H%1$s & H%2$s over the fixed effects H%1$s & H%2$s.", "\u2080", "\u2081", "\u03C4")
   footnoteOrderedBFtau <- gettextf("Bayes factor of the (unconstrained/constrained) random effects H%1$s over the fixed effects H%1$s.", "\u2081")
   
   if(options[["modelSpecification"]] == "CRE") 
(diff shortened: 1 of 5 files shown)


JASP-Engine/JASP/R/bayesianmetaanalysis.R
```diff
@@ -386,7 +386,7 @@ BayesianMetaAnalysis <- function(jaspResults, dataset, options) {
   }
   
   if(isTryError(p)){
-    .quitAnalysis(gettextf("The model could not fit. Error while running R code from the metaBMA package: %s", .extractErrorMessage(p)))
```
TimKDJ, line 389: Does this error message ever provide some useful information? Because "Try different settings" is a very general suggestion

Resources/Help/analyses/bayesianmetaanalysis.md
```diff
@@ -0,0 +1,158 @@
+Bayesian Meta-Analysis
```
TimKDJ, line 1: Any chance you could add the Dutch translation of this as well?
You can add the file in the same place, only name it bayesianmetaanalysis_nl.md

sophieberkhout: Honestly, I am not sure what the right Dutch translations are for most terms. I could try to figure it out, but I really don't have time for that this week.

JASP-Engine/JASP/R/bayesianmetaanalysis.R
sophieberkhout: So far, I have only seen the error messages "no parameter specified (pars is empty)" and "not enough unique values". It seems to give the first one when the analysis does not work (i.e. the model doesn't fit) and the second one appeared with the Mac/callr issue (no idea why). 

The model does not fit, for example, when the data is mostly negative/positive while the prior only contains positive/negative values. I believe there were other situations as well, but I cannot think of them right now.

I don't really know how to deal with this elegantly. If you have any suggestions, I am happy to hear them!

TimKDJ: Yeah those error messages aren't great. But maybe you could expand a tiny bit upon "try different settings"; do you mean certain checkboxes? Interval sizes? Variables?

Resources/Help/analyses/bayesianmetaanalysis.md
TimKDJ: I'll merge this before the Dutch translation, but it would be good if you could at least make a start with it and make a PR later. We don't have any dedicated translators to do this anymore unfortunately

JASP-Engine/JASP/R/bayesianmetaanalysis.R
boutinb: @sophieberkhout Could you answer/implement Tim's suggestion today, than we can push your PR, and start a second testing round

sophieberkhout: How is this?

TimKDJ:
I didn't think a title change would cause this:
`Desktop:	Path C:/JASP/build-debug-64-qt5.12/Resources/Meta-Analysis/ does not exist!`
The folder `Meta-Analysis` does indeed not exist, but shouldn't it still look for `Meta Analysis`?

Is the title used in the folder instead of the name, or am I doing something wrong @boutinb, @JorisGoosen?

JASP-Engine/JASP/R/bayesianmetaanalysis.R
TimKDJ: Yes that's a much more informative error message :)

Resources/Meta Analysis/description.json
```diff
@@ -1,8 +1,8 @@
 {
 	"moduleDescription" :
 	{
-		"title" : 		"Meta Analysis",
-		"description" : "Meta Analysis module.",
+		"title" : 		"Meta-Analysis",
```
JorisGoosen, line 5: You can add "name" here and let that be "Meta Analysis". That solves the problem you have now. Otherwise rename the folder.

If name isn't entered title is used.

TimKDJ: Ah... The Dutch description file has a name, but the English one does not.
Thanks for having a look!

@sophieberkhout can you add `"name": "Meta Analysis"` to the moduleDescription of this file?

sophieberkhout: Done!

TimKDJ (approved):

PR #3870, opened 2020-02-18
Scatterplot function inside descriptives shouldn't always use optionContainsValue

vandenman:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/617

The plots disappeared because the ANOVAs don't have something called "variables" in their options list.


TimKDJ:
Can you check if this plays nice with https://github.com/jasp-stats/jasp-desktop/pull/3860

vandenman:
Just checked, the state of the scatter plots in descriptives is unaffected.

JohnnyDoorn (approved):

PR #3874, opened 2020-02-18
Refresh audit JASP files

koenderks:
Because some minor typos were fixed in the meantime


TimKDJ:
Did you perform any new analyses in these refreshed JASP files?
If not, then this refresh is performed automatically after we're finished testing for all JASP files.
(We keep the old JASP files during testing so we can compare output of 0.11.1 with 0.12)

koenderks:
No, I basically just hit refresh and for all three jasp files and made this pull request. However, the jasp files currently in development feature new analyses when compared to those in the last release. Does this matter?

TimKDJ:
Alright, if you've just hit refresh without changing any input options or adding analyses, then we'll keep the old files for a bit. That way it doesn't become a mix of old and new JASP files and it's more consistent with regards to testing. 

Once we release version 0.12, we'll definitely refresh the files though :)

PR #3880, opened 2020-02-19
Distributions fixes

Kucharssim:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/516
Fixes https://github.com/jasp-stats/jasp-test-release/issues/532
Fixes https://github.com/jasp-stats/jasp-test-release/issues/533
Fixes https://github.com/jasp-stats/jasp-test-release/issues/530 (point 2 and 3, for 1 we will wait for the next release https://github.com/jasp-stats/INTERNAL-jasp/issues/710#issuecomment-582851972)

Also changes some texts based on E.J. feedback and updates some unit tests.


TimKDJ:
If estimation fails, should the error only be on the "Estimated Parameters" table or on the whole container? I ask because everything else in the container is still being computed, despite there being an error overlay:
Screenshot 2020-02-20 at 12 34 15

TimKDJ:
The P-P plot is also being cut-off (related to your fix for https://github.com/jasp-stats/jasp-test-release/issues/532):
Screenshot 2020-02-20 at 12 38 24

[distributions-pp-plot.jasp.zip](https://github.com/jasp-stats/jasp-desktop/files/4230478/distributions-pp-plot.jasp.zip)

TimKDJ:
Everything looks good otherwise; the bugs are indeed fixed (creative thinking on the error box one -- this is something that should probably be fixed more generally in the js)

Kucharssim:
> If estimation fails, should the error only be on the "Estimated Parameters" table or on the whole container? I ask because everything else in the container is still being computed, despite there being an error overlay

Yeah, I thought about this, but I think it should be the whole container. I originally intended to throw an error/footnote for the parameter estimates table; The rest of the container could potentially show you how terribly the distribution misfits. But then again, the point estimates cannot be trusted (these estimates are used in the other output in that container), so I don’t think it’s good to show the rest of the results as if nothing is wrong with it.

TimKDJ:
> thought about this, but I think it should be the whole container. I originally intended to throw an error/footnote for the parameter estimates table; The rest of the container could potentially show you how terribly the distribution misfits. But then again, the point estimates cannot be trusted (these estimates are used in the other output in that container), so I don’t think it’s good to show the rest of the results as if nothing is wrong with it.

Fair enough! You could consider adding `|| mleContainer$getError()` to your `if (!ready) return()` statements so you don't compute things you shouldn't be computing. Up to you though.

Kucharssim:
I fixed the p-plots and added the `$getError()` statements as @TimKDJ suggested, but need to fix the conflicts

TimKDJ:
Conflicts are caused by https://github.com/jasp-stats/jasp-desktop/commit/9edd6dcaa88f37733ad3a2514e2cdca8ab7ae90d; it added a bunch of missing `qsTr()`

Kucharssim:
> Conflicts are caused by [9edd6dc](https://github.com/jasp-stats/jasp-desktop/commit/9edd6dcaa88f37733ad3a2514e2cdca8ab7ae90d); it added a bunch of missing `qsTr()`

Got it, thanks! It should be combined now (preserved both `qsTr()` and my changes)

PR #3881, opened 2020-02-19
Add footnote when covariates are never included in regression model

TimKDJ:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/540


TimKDJ:
@EJWagenmakers do you have an opinion about the text in the footnote:
Screenshot 2020-02-19 at 17 47 06

This footnote intends to clear up confusion with respect to covariates that do not appear in the coefficients table at all, see https://github.com/jasp-stats/jasp-test-release/issues/540

vandenman:
The code looks good to me. However, logistic regression has the same situation, you can do stepwise or forward and it may seem like some predictors are never considered. Should we also add this footnote there?

vandenman (approved):

TimKDJ:
Yes, that's a good point. I'll do it during the second round of testing

PR #3891, opened 2020-02-21
Add tr function to Messages

boutinb:


JorisGoosen:
@boutinb maybe next time you can assign the issue to yourself so that we do not duplicate our efforts. It is a bit frustrating to see this.

PR #3899, opened 2020-02-24
Fixes several problems with language and refresh

JorisGoosen:
- Keep notes and changed title after language change
-- fixes https://github.com/jasp-stats/jasp-test-release/issues/500
- Removes QEventloop calls from resultsJsInterface because they really ought not be there
- Makes sure "Results" in results is translatable but, only when the user hasn't changed it.
-- fixes https://github.com/jasp-stats/jasp-test-release/issues/626
- Makes GENERATE_LANGUAGE_FILES a boolean instead of an integer because it is nicer


akashrajkn:
@JorisGoosen The mainNote no longer gets deleted. However, notes for some of the tables get deleted. 

jasp file: Data Library > Frequencies > Dancing Cats and Dogs.

The note under Chi-Squared Tests table (Analysis: Contingency Tables) gets deleted

JorisGoosen:
Aah yes, too bad about that. I will take a more in-depth look at that tomorrow, something fishy is going on there.

PR #3873, opened 2020-02-18
Bayesian AB Test - Robustness plot enhancements

akashrajkn:
Fixes jasp-stats/jasp-test-release#524

PR #3905, opened 2020-02-26
Implement restrictions on the abtest robustness parameters, update docs

TimKDJ:
This PR:

1. Adds Dutch documentation for the features added in https://github.com/jasp-stats/jasp-desktop/pull/3873
2. Imposes some restrictions on the lower <-> upper relationship (otherwise the analysis can crash)
3. Implements the third bullet point in https://github.com/jasp-stats/jasp-test-release/issues/524#issuecomment-585470878
4. Makes it slightly more pretty:

Before:
Screenshot 2020-02-26 at 11 39 14
After:
Screenshot 2020-02-26 at 11 37 38

### DIFF
--- a/JASP-Desktop/components/JASP/Controls/DoubleField.qml
+++ b/JASP-Desktop/components/JASP/Controls/DoubleField.qml
@@ -24,6 +24,7 @@ TextField
 {
 	id:									doubleField
 	property double defaultValue:		0
+	property double _prevDefaultValue:	0
 	property alias	doubleValidator:	doubleValidator
 	property bool	negativeValues:		false
 	property double	min:				negativeValues ? -Infinity : 0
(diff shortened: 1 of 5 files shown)


akashrajkn:
@TimKDJ Looks good.

Also, maybe it's just me, but restricting the mu_low value to be smaller than mu_high and vice versa is a bit inconvenient (the input boxes don't allow typing). 

With regard to point 3 - after giving custom values (say mu_range 6 - 10), if the user changes BF type, it is reset to default values. Is this the acceptable behaviour?

TimKDJ:
> Also, maybe it's just me, but restricting the mu_low value to be smaller than mu_high and vice versa is a bit inconvenient (the input boxes don't allow typing).

Yeah you'll have to start typing in the correct boxes. The alternative would be to do this in the analysis itself. But considering equivalence testing also uses this way of validation I figured it'd be ok. What do you think?

> With regard to point 3 - after giving custom values (say mu_range 6 - 10), if the user changes BF type, it is reset to default values. Is this the acceptable behaviour?

~~Because it's under advanced options (and we want to enforce sensible defaults): I think so. I noticed this too, but @boutinb didn't see an easy solution for letting changes override the default. I will poke Bruno some more about this one.~~

TimKDJ:
Alright I fixed the overwriting of changed values with default values

akashrajkn (approved):

PR #3916, opened 2020-02-27
fix https://github.com/jasp-stats/jasp-test-release/issues/585

koenderks:
But still.... something is up related to: https://github.com/jasp-stats/jasp-test-release/issues/612

Because if I add `format(options[["testValue"]], scientific = FALSE)` then it does work... @boutinb

### DIFF
--- a/JASP-Engine/JASP/R/bainttestbayesianonesample.R
+++ b/JASP-Engine/JASP/R/bainttestbayesianonesample.R
@@ -73,10 +73,11 @@ BainTTestBayesianOneSample <- function(jaspResults, dataset, options, ...) {
                     "equalBiggerSmaller"  = 5)
     
   variableData <- dataset[[ .v(variable) ]]
+  testValue <- format(options[["testValue"]], scientific = FALSE)
 
   p <- try({
     # Call bain from package
-    bain:::bain_ttest_cran(x = variableData, nu = options[["testValue"]], type = type, seed = options[["seed"]])
+    bain:::bain_ttest_cran(x = variableData, nu = testValue, type = type, seed = options[["seed"]])
   })
   bainContainer[[variable]] <- createJaspState(p, dependencies = c("testValue", "hypothesis", "seed"))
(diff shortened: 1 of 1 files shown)


boutinb:
@TimKDJ Als 100000 intypt, wordt wel 100000 gestuurd in de options, maar als je browse toevoegd in die R functie voor de try, dan wordt het geprint met scientific notation, en werkt het niet in de bain functie. Vreemd toch? Het is niet de bedoeling dat iedereen z'n code moet toevoegen om te zorgen dat de waard niet in scientific format komt?

TimKDJ:
I had a look and the problem lies with bain, not JASP. Basically what bain does is a `paste` of `"x >"` with whatever `options$testValue` holds. The "problem" is that pasting a number that is larger than `1e5` results in a character in scientific notation:
```
> paste(100000)
[1] "1e+05"
```
 So when bain does `paste("x >", 100000)` they get `"x > 1e+05"`.
This value is then after a few calls passed on to `bain:::parse_hypothesis()` which cannot deal with the scientific notation:
```
> bain:::parse_hypothesis(varnames = "x", "x>1e+05;x<1e+05")
Error in -+1 * e : non-numeric argument to binary operator
```

So why does `format(options[["testValue"]], scientific = FALSE)` work? Because it returns a string: `"100000"`. Which is `paste`'d to be `"x > 100000"`.
And the parse function in bain does like this:
```
> bain:::parse_hypothesis(varnames = "x", "x>100000;x<100000")
$hyp_mat
$hyp_mat[[1]]
         x      
x>100000 1 1e+05

$hyp_mat[[2]]
          x       
x<100000 -1 -1e+05


$n_constraints
[1] 0 1 0 1

$n_ec
[1] 0 0

$original_hypothesis
[1] "x>100000" "x<100000"
```

The solution by @koenderks is fine in this case, but we should report this problem to the maintainer of the bain R package.

TimKDJ (approved):

boutinb:
@koenderks Do you report this error to the BAIN Maintainer?

PR #3918, opened 2020-02-27
Juggling Notes!

JorisGoosen:
- QML now knows whether the analysis needs a refresh or not (and this indication now also takes dynamic modules into account)
- It also knows if it has volatile notes (aka coupled with possibly changed names of tables or collections)
- when an analysis status goes from Complete -> something else it stores its userdata
- when it then later gets status Complete again it runs the algorithm in Analysis::fitOldUserDataEtc to try to match the changed names to the old notes
- this changed userData is passed tot the js. And that sometimes handles it correctly
- There is something fishy going on with quill though
-- See JASPWidgets.NoteBox.isTextboxEmpty in jaspWidgets.js
- Also from now on any analysis that was loaded from an old jasp version or dynamic module will state so to the user in polite terms on top of the form

This commit fixes https://github.com/jasp-stats/jasp-test-release/issues/469 (mostly anyway)

### DIFF
--- a/JASP-Desktop/components/JASP/Controls/Form.qml
+++ b/JASP-Desktop/components/JASP/Controls/Form.qml
@@ -89,11 +89,36 @@ AnalysisForm
 	{
 		id:				formContent
 		width:			parent.width
-		height:			errorMessagesBox.height + contentArea.implicitHeight
+		height:			oldFileMessagesBox.height + errorMessagesBox.height + contentArea.implicitHeight
 		anchors
 		{
 			top:		form.top
 			left:		form.left
+		}
+
+		Rectangle
+		{
+			id:				oldFileMessagesBox
+			visible:		myAnalysis !== null && myAnalysis.needsRefresh
+			color:			jaspTheme.controlWarningBackgroundColor
+			width:			parent.width
+			height:			visible ? oldAnalysisText.height : 0
+			anchors.top:	parent.top
+
+			Text
+			{
+				id:					oldAnalysisText
+				color:				jaspTheme.controlWarningTextColor
+				anchors.centerIn:	parent
+				padding:			5 * jaspTheme.uiScale
+				wrapMode:			Text.Wrap
+				width:				parent.width - 10 * jaspTheme.uiScale
+				verticalAlignment:	Text.AlignVCenter
+				text:				qsTr("This analysis was created with an older version of JASP (or a dynamic module)") + //I do not want to bother with formatting strings here to be honest
+									( myAnalysis !== null && !myAnalysis.hasVolatileNotes ? qsTr(", refreshing could give a slightly different result.") :
+																	 qsTr(", to keep your notes where they are it is highly recommended to first refresh your analyses!"))
+
+			}
 		}
 				
 		Rectangle
@@ -106,15 +131,17 @@ AnalysisForm
 			color:			jaspTheme.errorMessagesBackgroundColor
 			width:			parent.width
 			height:			visible ? errorMessagesText.height : 0
+			anchors.top:	oldFileMessagesBox.bottom
 
 			Text
 			{
 				id:					errorMessagesText
 				anchors.centerIn:	parent
-				padding:			5
+				padding:			5 * jaspTheme.uiScale
 				wrapMode:			Text.Wrap
-				width:				parent.width - 10
+				width:				parent.width - 10 * jaspTheme.uiScale
 				verticalAlignment:	Text.AlignVCenter
+				//Should we maybe set a color here?
 			}
 		}
 		
--- a/Dynamic Modules/TestModule/DESCRIPTION
+++ b/Dynamic Modules/TestModule/DESCRIPTION
@@ -2,7 +2,7 @@ Package: DevelopmentModule
 Type: Package
 Title: Test Module Module for JASP
 Version: 1.0
-Date: 2020-02-04
+Date: 2020-02-27
 Author: Joris Goosen
 Website: 
 Maintainer: Joris Goosen <Joris@JorisGoosen.nl>
--- a/Dynamic Modules/TestModule/NAMESPACE
+++ b/Dynamic Modules/TestModule/NAMESPACE
@@ -2,5 +2,6 @@ export(testTransposedTableFunc)
 export(testFootnotesTableFunc)
 export(testStateFunc)
 export(testHtmlFunc)
+export(testContainerFunc)
 export(customContrasts)
   import('svglite');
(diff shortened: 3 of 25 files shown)


JorisGoosen:
So this is a pretty big change all in all but probably does what it is supposed to do.

Only problem is that if something weird is going on with Quill and I hope that @akashrajkn can figure out what is going on.

Let's say you open frequencies - dancing cats.
You see the comments under the tables and a message above the qml form saying its old and you should refresh. All good. 
You refresh, but at least one comment (Under contingency table table) disappears anyway...
You add some other comment under contingency thing and then you "Close" the jasp-file through the file-menu.

Now you reload the exact same Frequencies-dancing cats file from the data library and refresh.

Now all comments remain! So I think something is kept in the webengine or some initialization goes wrong the first time. I dont know. But @akashrajkn should have a look at that.

TimKDJ:
I did some initial testing and found some problems unfortunately (for the .jasp files I only added ".zip" to the name, so you only need to rename, not extract):

1. Refreshing Regression->"Adam Sandler.jasp" crashes JASP for me because of an assertion fail (also tested this on rebased development and I did not run into an issue then):
bcorAssertFail

2. Refreshing the T-Tests->"Directed reading activities.jasp" file causes the Descriptives analysis options form to "detach" from its output (nothing you do in the Descriptives analysis does anything in the output). The Descriptives also gets this eternal spinning icon:
[descriptivestable-loading.jasp.zip](https://github.com/jasp-stats/jasp-desktop/files/4267780/descriptivestable-loading.jasp.zip)

3. Refreshing the Regression->"Big Five Personality Traits.jasp" file removes the toolbar (title etc) from the analysis and makes its output unresponsive (possibly the same as 3, but not sure):
[webengine-js-problem.jasp.zip](https://github.com/jasp-stats/jasp-desktop/files/4267800/webengine-js-problem.jasp.zip)

4. The yellow warning in the options form doesn't always go away when refreshing. This is from ANOVA->"Beer goggles.jasp":
[warning-doesnt-go-away.jasp.zip](https://github.com/jasp-stats/jasp-desktop/files/4267845/warning-doesnt-go-away.jasp.zip)

Probably not necessary but this file seems to contain all of the last 3 bullets (ANOVA->"Heart Rate.jasp"):
[all-problems-in-one.jasp.zip](https://github.com/jasp-stats/jasp-desktop/files/4267886/all-problems-in-one.jasp.zip)

JorisGoosen:
Thanks for testing. I fixed the bugs you found and also added support for notes under plots, because we should actually support that to avoid a regression.

Related issue: https://github.com/jasp-stats/jasp-test-release/issues/647

TimKDJ:
Refreshing T-Tests->"Directed Reading Activities".jasp loses the note `The assumption checks are not statistically significant. ` that is supposed to be under "Assumption Checks":

refreshed <-> not refreshed
Schermafbeelding 2020-03-03 om 14 05 33

It's not in the `orphanedNotes` either:

```
Desktop:	New userData after attempt to fix is: {
   "children" : {
      "ttest" : {
         "note" : {
            "delta" : {
               "ops" : [
                  {
                     "insert" : "Since the variance appears to be larger in the control group (see also the violin plot below), we consider the Welch version of the t-test; the Welch version does not assume that the variances in the two groups are equal.\n"
                  }
               ]
            },
            "deltaAvailable" : true,
            "format" : "html",
            "text" : "<p>Since the variance appears to be larger in the control group (see also the violin plot below), we consider the Welch version of the t-test; the Welch version does not assume that the variances in the two groups are equal.</p>",
            "visible" : true
         }
      }
   },
   "firstNote" : {
      "delta" : {
         "ops" : [
            {
               "insert" : "As we can see, the difference between the groups is statistically significant at the .05 level. The Treatment group has on average about ten points more on the DPR test than the Control group ( "
            },
            {
               "attributes" : {
                  "italic" : true
               },
               "insert" : "t"
            },
            {
               "insert" : "(37.86) = 2.311, "
            },
            {
               "attributes" : {
                  "italic" : true
               },
               "insert" : "p"
            },
            {
               "insert" : " = 0.013), translating to medium effect size: Cohen's "
            },
            {
               "attributes" : {
                  "italic" : true
               },
               "insert" : "d"
            },
            {
               "insert" : " = 0.691. \n"
            }
         ]
      },
      "deltaAvailable" : true,
      "format" : "html",
      "text" : "<p>As we can see, the difference between the groups is statistically significant at the .05 level. The Treatment group has on average about ten points more on the DPR test than the Control group ( <em>t</em>(37.86) = 2.311, <em>p</em> = 0.013), translating to medium effect size: Cohen's <em>d</em> = 0.691.&nbsp;</p>",
      "visible" : true
   },
   "lastNote" : {
      "delta" : {
         "ops" : [
            {
               "insert" : "\n"
            }
         ]
      },
      "deltaAvailable" : true,
      "format" : "html",
      "text" : "<p><br></p>",
      "visible" : true
   },
   "orphanedNotes" : {
      "descriptives" : {
         "note" : {
            "format" : "markdown",
            "text" : "",
            "visible" : true
         }
      }
   }
}
```

JorisGoosen:
Ok, so Bayesian Corellation Pairs goes wrong still, but I think this is related to https://github.com/jasp-stats/jasp-test-release/issues/647

BCG Vaccine misses a note, but the algorithm does seem to place it in it's proper place. However nothign shows up in the results. I've asked @akashrajkn in [this comment](https://github.com/jasp-stats/jasp-test-release/issues/469#issuecomment-594420132) if he has time to take a look at that issue. Basically the same problem as for Dancing Cats.

JASP-Desktop/components/JASP/Controls/Form.qml
```diff
@@ -89,12 +89,37 @@ AnalysisForm
...
+				padding:			5 * jaspTheme.uiScale
+				wrapMode:			Text.Wrap
+				width:				parent.width - 10 * jaspTheme.uiScale
+				verticalAlignment:	Text.AlignVCenter
+				text:				"This analysis was created with an older version of JASP (or a dynamic module)" +
+									( myAnalysis !== null && !myAnalysis.hasVolatileNotes ? ", refreshing could give a slightly different result." :
+																	 ", to keep your notes where they are it is highly recommended to first refresh your analyses!")
+
+			}
+		}
```
boutinb, line 122: needs qsTr

boutinb (commented):
Why did you update the JASP files of the library?

JorisGoosen:
Ive undone the data library files, keep in mind though that when we resave them we should make sure they are done properly. Aka take them from 0.11.1 directly. Because at least one file already lost some notes between 0.11.1 and now.

PR #3919, opened 2020-02-28
Adding and then removing an Empty value does not update the analyses

boutinb:
If you add an empty value, and an analysis that uses a column that is
changed by this new empty value, is updated.
When you remove afterwards this empty value, the analysis is not
changed anymore

### DIFF
--- a/JASP-Common/dataset.cpp
+++ b/JASP-Common/dataset.cpp
@@ -76,21 +76,19 @@ string DataSet::toString()
 	return ss.str();
 }
 
-vector<string> DataSet::resetEmptyValues(std::map<std::string, std::map<int, std::string> > emptyValuesPerColumnMap)
+std::map<string, std::map<int, string> > DataSet::resetEmptyValues(const std::map<std::string, std::map<int, std::string> >& emptyValuesPerColumnMap)
 {
-	std::vector<std::string> colChanged;
+	std::map<string, std::map<int, string> > colChanged;
 
 	for (Column& col : _columns)
 	{
-		std::map<int, std::string> emptyValuesMap;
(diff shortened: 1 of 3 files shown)


JASP-Desktop/data/datasetpackage.h
```diff
@@ -107,7 +107,7 @@ class DataSetPackage : public QAbstractItemModel //Not QAbstractTableModel becau
 		const	Json::Value		&	analysesData()						const	{ return _analysesData;							 }
 		const	std::string		&	warningMessage()					const	{ return _warningMessage;						  }
 		const	Version			&	archiveVersion()					const	{ return _archiveVersion;						   }
-		const	emptyValsType	&	emptyValuesMap()					const	{ return _emptyValuesMap;   						}
+				emptyValsType	&	emptyValuesMap()							{ return _emptyValuesMap;   						}
```
JorisGoosen: @boutinb this kind of breaks the whole "datasetpackage does everything"-paradigm, doesn't it?

Because this way the emptyValuesMap could be changed at any time anywhere right? And nothing would be triggered, not packageModified, not modelChanged, nothing...

JorisGoosen (changes requested):
Can you not change it by finding the right place to update it instead of passing non-const references around?

JASP-Desktop/data/datasetpackage.h
boutinb: OK, I have set back the const, and change the emptyValuesMap in DataSetPackage, not in DataSet anymore.

JorisGoosen:
Code looks nice now, I'll just check if it actually works now and after that Ill merge it.

PR #3923, opened 2020-03-02
Makes synching an spss file work

### LINKED ISSUE #619: Reloading data file does not work (any more)  0.11.1
* JASP version: 0.11.1

* Bug description: Im editing a file in SPSS and doing an analysis in JASP. After I edit and save the file in SPSS, JASP (earlier versions) used to refresh the data/synchronize with the new file. This does not work anymore. Nor does the manual Ctrl+Y option work. I need to manually re open the file "Recent files" -> "Open data file" and this opens the updated file in a new JASP window. Closing SPSS after editing the file does not change this behavior.

JorisGoosen:
- Added a small sleep to give the os some time to handle stuff, necessary because the sync file is opened twice in rapid succesion clobbering the fopen (on windows anyway).
- Also let a column dictionary be built for spss, not needed for opening a file, but it *is* for synching
- And compare two strings in readstat label with operator< instead of compare because that breaks the std::map...
Fixes https://github.com/jasp-stats/jasp-issues/issues/619

### DIFF
--- a/JASP-Desktop/data/datasetloader.cpp
+++ b/JASP-Desktop/data/datasetloader.cpp
@@ -31,6 +31,7 @@
 #include <QFileInfo>
 
 #include "timers.h"
+#include "utils.h"
 
 using namespace std;
 using namespace ods;
@@ -77,6 +78,8 @@ void DataSetLoader::loadPackage(const string &locator, const string &extension, 
 
 void DataSetLoader::syncPackage(const string &locator, const string &extension, boost::function<void(const string &, int)> progress)
 {
+	Utils::sleep(100);
+
 	Importer* importer = getImporter(locator, extension);
 
 	if (importer)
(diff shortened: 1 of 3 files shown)


JASP-Desktop/data/datasetloader.cpp
```diff
@@ -77,6 +78,8 @@ void DataSetLoader::loadPackage(const string &locator, const string &extension,
 
 void DataSetLoader::syncPackage(const string &locator, const string &extension, boost::function<void(const string &, int)> progress)
 {
+	Utils::sleep(100);
+
```
boutinb, line 82: Fortunately I did not have to write this code!

JorisGoosen: I really wished it wasn't necessary, but I can't think of anything better...

PR #3932, opened 2020-03-04
Descriptives position conditional

JohnnyDoorn:
new attempt to fix https://github.com/jasp-stats/jasp-test-release/issues/649
only for Bayesian anova is it hardcoded now (per Don's request)


JorisGoosen:
Hmm, so thanks for taking a look, but actually the notes are now still in the wrong place after refreshing:
Schermafbeelding 2020-03-04 om 16 46 59
So @JohnnyDoorn thank you very much for taking a look but I am afraid this PR won't help. So we'll just leave it as is.

PR #3935, opened 2020-03-05
different df type contrasts

JohnnyDoorn:
very small fix I just noticed
Also the position in descriptives should be updated now, so that it is still in line with how Don wants the descriptives

### DIFF
--- a/JASP-Engine/JASP/R/commonAnovaBayesian.R
+++ b/JASP-Engine/JASP/R/commonAnovaBayesian.R
@@ -1159,7 +1159,10 @@
   if (is.null(descriptivesContainer)) {
     descriptivesContainer <- createJaspContainer(title = gettext("Descriptives"))
     descriptivesContainer$dependOn(c("dependent", "repeatedMeasuresCells"))
-    # descriptivesContainer$position <- 9001 # always last
+    
+    if (is.null(options$confidenceIntervalInterval)) # TRUE implies Bayesian
+      descriptivesContainer$position <- 9001 # always last
+    
     jaspResults[["descriptivesContainer"]] <- descriptivesContainer
   }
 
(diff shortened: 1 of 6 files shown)


JASP-Engine/JASP/R/commonAnovaBayesian.R
```diff
@@ -1159,7 +1159,10 @@
   if (is.null(descriptivesContainer)) {
     descriptivesContainer <- createJaspContainer(title = gettext("Descriptives"))
     descriptivesContainer$dependOn(c("dependent", "repeatedMeasuresCells"))
-    descriptivesContainer$position <- 9001 # always last
+    
+    if (is.null(options$confidenceIntervalInterval)) # TRUE implies Bayesian
```
vandenman: So this works fine but I'd prefer something like:

```r
.BANOVAdescriptives <- function(jaspResults, dataset, options, errors, analysisType, ready = TRUE, position = 9001) {

  ...

  descriptivesContainer$position <- position
}
```
That way, other/ future analyses could also reuse this function and still easily change the position.

JohnnyDoorn: OK, made it like that - also dont need to specify it anymore for the freq anova, because it didn't seem to fix the notes issue anyway

vandenman (approved):

vandenman:
If Travis is happy then I'm happy.

JohnnyDoorn:
> If Travis is happy then I'm happy.

truer words have never been spoken...

JohnnyDoorn:
I also added some updates to the dutch helpfiles, since these were apparently based on the old gui layout (where we had additional options as a menu, now it's just called marginal means, and the other options were moved up).

PR #3939, opened 2020-03-06
fix to custom contrast check

### LINKED ISSUE #627: Custom contrasts
* JASP version: Nightly 0.12 Mar 6 2020
* OS name and version: Windows 7

On a classic ANOVA with one factor 
[04 ViagraContrasteInter.txt](https://github.com/jasp-stats/jasp-issues/files/4297573/04.ViagraContrasteInter.txt)

When I try to use custom contrasts in the last nightly such as "-2 1 1"
then I get "Nonconforming number of contrasts coefficients".
So I try to add another contrast such as "0 -1 1".
And then I get "Please specify non-zero contrasts weights". Then I tried to specify another contrast without a 0 coefficient and it does not work either.
I think that It would be very restrictive to just allow orthogonal contrasts.

Another problem is that in an ANOVA or RM ANOVA using 2 factors  it is not possible to determine contrasts using two factors
For instance in Music.Simple.txt
[04 Music.Simple.txt](https://github.com/jasp-stats/jasp-issues/files/4297570/04.Music.Simple.txt)
, it could be interesting to test if for inf40 if there is a difference between ABBA and Kenji.

For this, one possibility would be in the "Factors" window to select a factor "music*age" and then to have a table with the possible coefficients for the two factors.

Best

JohnnyDoorn:
fix https://github.com/jasp-stats/jasp-issues/issues/627#event-3105326571

### DIFF
--- a/JASP-Engine/JASP/R/ancova.R
+++ b/JASP-Engine/JASP/R/ancova.R
@@ -638,12 +638,12 @@ Ancova <- function(jaspResults, dataset = NULL, options) {
       if (contrast$contrast == "custom") {
         if (isTryError(contrastResult)) {
           if (grepl(contrastResult[1], pattern = "Nonconforming number")) {
-            contrastContainer$setError(gettext("Nonconforming number of contrast coefficients."))
+            contrastContainer$setError(gettext("Please specify an additional contrast."))
           } else if (grepl(contrastResult[1], pattern = "number of contrast matrix rows")) {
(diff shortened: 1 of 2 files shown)


vandenman:
So this looks fine, but why is the code identical in two different files? That's maybe something to clean up later.

Furthermore, you're `grepl`ing on the error message returned by the package. If the package is translated, then these error messages won't be the same when you switch version.

vandenman (approved):

PR #3941, opened 2020-03-09
Layout problem in Distribution module in Dutch

boutinb:
Solves jasp-stats/jasp-test-release#669


Kucharssim (changes requested):
So the Layout seems corrected, thanks for implementing this! I like how you solved the annoying pasting of the “generate new variable from ..." :)

The test fail hard though, should I check that out or are you fixing that yourself? It seems that the function .simulateData gets called during the tests, which is weird because it should be run only when you click the “Draw samples” button, but not otherwise.

TimKDJ:
> The test fail hard though, should I check that out or are you fixing that yourself? It seems that the function .simulateData gets called during the tests, which is weird because it should be run only when you click the “Draw samples” button, but not otherwise.

I will fix this. It's something in jasptools :)

PR #3944, opened 2020-03-10
IntegerField & DoubleField set default value back when language is changed

boutinb:
Solves also the layout problem

Solves jasp-stats/jasp-test-release#670

### DIFF
--- a/JASP-Desktop/widgets/boundqmltextinput.cpp
+++ b/JASP-Desktop/widgets/boundqmltextinput.cpp
@@ -274,6 +274,19 @@ bool BoundQMLTextInput::isJsonValid(const Json::Value &optionValue)
 	return valid;
 }
 
+void BoundQMLTextInput::setUp()
+{
+	if (form())
+		// For unknown reason, when the language is changed, QML reset the default value.
+		// We have then to set back the value from the option
+		connect(form(), &AnalysisForm::languageChanged, this, &BoundQMLTextInput::resetValue);
+}
+
+void BoundQMLTextInput::resetValue()
+{
+	setItemProperty("value", _value);
+}
+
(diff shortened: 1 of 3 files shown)


JASP-Desktop/widgets/boundqmltextinput.cpp
```diff
@@ -274,6 +274,19 @@ bool BoundQMLTextInput::isJsonValid(const Json::Value &optionValue)
 	return valid;
 }
 
+void BoundQMLTextInput::setUp()
+{
+	if (form())
+		// For unknown reason, when the language is changed, QML reset the default value.
```
JorisGoosen, line 280: QML resets all bindings on language changed, this is not unknown, it is an implementation detail. The weird thing is that the reverting to the default value (to zero anyway) only happens with these few elements...

JorisGoosen:
Well, your fix works... But, I still think there must be a different explanation why the values get reset for these two fields and not the percentage field next to it. I hope you do not mind that I will first take a look at that before I merge this?

JorisGoosen:
Ok, I guess your solution is good enough :)

But you know what would've been better here? A clean separation of view and data. So not storing defaultValue and value in QML but in the boundqmltextinput. And putting the logic there as well.

Instead of what we have now...

boutinb:
What strange is, is that this problem depends on the defaultValue I gave in the DoubleField or  IntegerField... with a 0 or 10, I got the error, but not with 20... that's why I gave up to try to understand what was the real problem. But your suggestion to set the defaultValue from the model is anyway a good idea.

JorisGoosen:
Aah yeah, I thought it was weird as well that the `PercentField`'s default value works, despite it also secretly being a `DoubleField`...

PR #3955, opened 2020-03-13
T-Test fix ups

TimKDJ:
This got a little out of hand, but the ttest code was in need of some refactoring. At least now functions are a little more broken up and error handling is more robust. There is still redundancy, but let's just call it work in progress.

- Fixed crashes when the analyses tried to put try-errors inside of $addRows() (fixes https://github.com/jasp-stats/jasp-test-release/issues/661, fixes https://github.com/jasp-stats/jasp-test-release/issues/692)
- Fixed the independent samples t-test descriptives plot that always used listwise NA deletion
- Changed the Levene's is violated footnote to only appear if both Welch and Mann-Whitney aren't ticked (previously it was only Welch)
- Added isNewGroup
- Fixes unicode copy-paste problem (fixes https://github.com/jasp-stats/jasp-test-release/issues/694)
- Fixes https://github.com/jasp-stats/jasp-test-release/issues/695

Only thing that changed in the unit tests is the .IsNewGroup.


JASP-Engine/JASP/R/ttestindependentsamples.R
```diff
@@ -196,294 +199,295 @@ TTestIndependentSamples <- function(jaspResults, dataset = NULL, options, ...) {
...
-    for (variable in options$variables) {
-      errors <- .hasErrors(dataset, 
-                           message = 'short', 
-                           type = c('observations', 'variance', 'infinity'),
-                           all.target = variable, 
-                           all.grouping = options$groupingVariable,
-                           observations.amount = '< 2')
-      variableData <- dataset[[ .v(variable) ]]
+    ## test is a number, indicating which tests should be run
+    for (test in seq_len(length(optionsList$whichTests))) {
```
Kucharssim: If you do not want to fix this indication by number now, then I would at least change this
```suggestion
    for (test in seq_along(optionsList$whichTests)) {
```

JASP-Engine/JASP/R/ttestindependentsamples.R
```diff
@@ -196,294 +199,295 @@ TTestIndependentSamples <- function(jaspResults, dataset = NULL, options, ...) {
...
-      ## test is a number, indicating which tests should be run
-      for (test in seq_len(length(optionsList$whichTests))) {
-        
-        ## don't run a test the user doesn't want
-        currentTest <- optionsList$whichTests[[test]]
-        if (!currentTest)
-          next
+      # TODO Tim: whichTests should just be a character vector of test names requested. Then you lose the confusion of figuring out which test is which int and you don't need this check
+      ## don't run a test the user doesn't want
+      currentTest <- optionsList$whichTests[[test]]
```
Kucharssim: (Unsolicited comment as you probably know best how to do this :D)
Perhaps a lot of this can be simplified if inside of `.ttestOptionsList()`, you would define something like:

```
    optionsList$whichTests  <- list("1" = optionsList$wantsStudents, 
                                    "2" = optionsList$wantsWelchs, 
                                    "3" = optionsList$wantsWilcox)

    optionsList$requestedTests <- c(“Student”, “Welch”, “Mann-Whitney”)[unlist(optionsList$whichTests)]
```

Then you can just loop over the requested tests when appropriate, enabling you to skip all of these `next` statements and defining `whatTest`.

JASP-Engine/JASP/R/ttestpairedsamples.R
```diff
@@ -199,150 +191,160 @@ TTestPairedSamples <- function(jaspResults, dataset = NULL, options, ...) {
...
-    if (ready && p1 != p2) {
-      c1   <- dataset[[ .v(p1) ]]
-      c2   <- dataset[[ .v(p2) ]]
-      data <- na.omit(c1 - c2)
+    if (p1 != "" && p2 != "") {
+      errors <- .hasErrors(dataset,
+                           message = 'short',
+                           type = c('observations', 'variance', 'infinity'),
+                           all.target = c(p1, p2), 
+                           observations.amount  = c('< 2', '> 5000'))
```
Kucharssim: Just out of curiosity, why do we check `> 5000`?

Kucharssim (changes requested):
Looking good! Did not find a bug, left some minor comments (whether or not you decide to act on them is up to you).

Btw. also fixes https://github.com/jasp-stats/jasp-test-release/issues/695

JASP-Engine/JASP/R/ttestindependentsamples.R
TimKDJ: Yeah, totally. I just got lazy, because my changes incited more changes, which incited changes. I figured I needed to draw the line somewhere. But alas! You have convinced me :-) I will get rid of the TODO and actually do it.

JASP-Engine/JASP/R/ttestpairedsamples.R
TimKDJ: It's in the shapiro wilks test:
```
> stats::shapiro.test
function (x) 
{
    DNAME <- deparse(substitute(x))
    stopifnot(is.numeric(x))
    x <- sort(x[complete.cases(x)])
    n <- length(x)
    if (is.na(n) || n < 3L || n > 5000L) 
        stop("sample size must be between 3 and 5000")
    rng <- x[n] - x[1L]
    if (rng == 0) 
        stop("all 'x' values are identical")
    if (rng < 1e-10) 
        x <- x/rng
    res <- .Call(C_SWilk, x)
    RVAL <- list(statistic = c(W = res[1]), p.value = res[2], 
        method = "Shapiro-Wilk normality test", data.name = DNAME)
    class(RVAL) <- "htest"
    return(RVAL)
}
```

Although I guess it should be `< 3` instead of `< 2`

TimKDJ:
> Btw. also fixes jasp-stats/jasp-test-release#695

That's a lucky coincidence, I wasn't trying

TimKDJ:
@Kucharssim if you could have one more look that'd be greatly appreciated

Kucharssim (approved):

PR #3957, opened 2020-03-13
Sync problem with Excel

boutinb:
With some version of Excel, the file is either temporarily not
existing, or empty.
Fixes jasp-stats/jasp-test-release#674

### DIFF
--- a/JASP-Desktop/widgets/filemenu/filemenu.cpp
+++ b/JASP-Desktop/widgets/filemenu/filemenu.cpp
@@ -399,7 +399,7 @@ void FileMenu::setSyncRequest(const QString& path)
 	if (path.isEmpty())
 		return;
 
-	if (checkSyncFileExists(path))
+	if (checkSyncFileExists(path, true))
 	{
 		FileEvent *event = new FileEvent(this, FileEvent::FileSyncData);
 		event->setPath(path);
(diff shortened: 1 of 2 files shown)


JASP-Desktop/widgets/filemenu/filemenu.cpp
```diff
@@ -408,12 +408,22 @@ void FileMenu::setSyncRequest(const QString& path)
...
+bool FileMenu::checkSyncFileExists(const QString &path, bool waitForExistence)
 {
 	if (path.startsWith("http"))
 		return true;
 
 	bool exist = false;
+	int maxExistTrials = waitForExistence ? 5 : 1;
+	int existTrials = 1;
+
+	while (!QFileInfo::exists(path) && existTrials < maxExistTrials)
```
JorisGoosen, line 420: You could put the `existTrials++` in the `while` check directly right?

boutinb: That's right.
But I had to think really carefully about the condition 
> while (!QFileInfo::exists(path) && existTrials < maxExistTrials)
that I was already happy to understand that it was the right one...

JorisGoosen: I can imagine :p 
while loops are scary!

PR #3958, opened 2020-03-13
VariablesForm gets wrong height when hidden

boutinb:
Fixes jasp-stats/jasp-test-release#709

### DIFF
--- a/JASP-Desktop/components/JASP/Controls/VariablesForm.qml
+++ b/JASP-Desktop/components/JASP/Controls/VariablesForm.qml
@@ -23,15 +23,15 @@ import JASP				1.0
 Item
 {
 	id					: variablesForm
-	implicitWidth		: form.width
 	implicitHeight		: jaspTheme.defaultVariablesFormHeight
+	implicitWidth		: form.width
 	height				: implicitHeight
 	width				: implicitWidth
 	Layout.columnSpan	: parent.columns
 	visible				: !debug || DEBUG_MODE
 
 	default property alias	content:						items.children
-			property int	listWidth:						implicitWidth * 2 / 5
+			property int	listWidth:						width * 2 / 5
 			property var	availableVariablesList
 			property var	allAssignedVariablesList:		[]
 			property var	allJASPControls:				[]
@@ -58,7 +58,7 @@ Item
 			setControlsSize();
 		}
 	}
-	
+
 	Repeater
 	{
 		id: assignButtonRepeater
@@ -252,11 +252,11 @@ Item
 	
 	function setControlsSize()
 	{
-		availableVariablesList.height = variablesForm.height
+		availableVariablesList.height = Qt.binding(function() { return variablesForm.height; })
 		// Set the width of the VariablesList to listWidth only if it is not set explicitely
 		// Implicitely, the width is set to the parent width.
 		if (availableVariablesList.setWidthInForm) 
-			availableVariablesList.width = variablesForm.listWidth
+			availableVariablesList.width = Qt.binding(function() { return variablesForm.listWidth; })
 		
 		var firstControl				= true;
 		var minHeightOfAssignedControls = 0;
@@ -270,7 +270,7 @@ Item
 
 			if (isControlList && control.setWidthInForm)
 				// Change the width of the VariablesList only if was not set explicitely
-				control.width = variablesForm.listWidth
+				control.width = Qt.binding(function() {return variablesForm.listWidth; })
 			else if (isControlComboBox && control.setWidthInForm)
 			{
 				control.setLabelAbove	= true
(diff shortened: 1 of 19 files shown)


JASP-Desktop/components/JASP/Controls/VariablesForm.qml
```diff
@@ -42,8 +40,8 @@ Item
 
 			property double	_lastListWidth:					0
 	
-	property int preferredHeight:	implicitHeight
-	property int preferredWidth:	implicitWidth
+	property int preferredHeight:	jaspTheme.defaultVariablesFormHeight
```
JorisGoosen: So, what is implicitHeight and width set to now?

Shouldn't it be:
```
property int preferredHeight:	implicitHeight
property int preferredWidth:	implicitWidth
implicitHeight:	jaspTheme.defaultVariablesFormHeight
implicitWidth:	form.width
```

JASP-Desktop/components/JASP/Controls/VariablesForm.qml
```diff
@@ -23,15 +23,13 @@ import JASP				1.0
...
-	height				: implicitHeight
-	width				: implicitWidth
+	height				: preferredHeight
+	width				: preferredWidth
 	Layout.columnSpan	: parent.columns
 	visible				: !debug || DEBUG_MODE
 
 	default property alias	content:						items.children
-			property int	listWidth:						implicitWidth * 2 / 5
+			property int	listWidth:						preferredWidth * 2 / 5
```
JorisGoosen: And this I am less sure about, but shouldn't the listWidth be based on the actual width?

JorisGoosen (changes requested):

JASP-Desktop/components/JASP/Controls/VariablesForm.qml
boutinb: A VariablesForm is normally used in a GridLayout (in the Form). In this case the preferredHeight & preferredWidth should be enough. But a VariablesForm might be used in  a sub-item that has no GridLayout. In that case, what count are the height and width properties.
I don't see why we should set the implicitWidth & implicitHeight

JASP-Desktop/components/JASP/Controls/VariablesForm.qml
boutinb: Yes, that's true it should be width in place of preferredWidth: inside a GrigLayout width is set to preferredWidth, but without GridLayout preferredWidth has no sense, and only width should be used.

JorisGoosen: Because you never know where it might be used in the future and [the documentation](https://doc.qt.io/qt-5/qml-qtquick-item.html#implicitWidth-prop) states that:
>Defines the natural width or height of the Item if no width or height is specified.
>The default implicit size for most items is 0x0, however some items have an inherent implicit size which cannot be overridden, for example, Image and Text.
>Setting the implicit size is useful for defining components that have a preferred size based on their content,...

I can see situations where you would want to be able to select several variables in something that isn't a form. For defining computed columns from several columns for instance.
Or maybe we could replace the columnlist in the filterconstructor by this or who knows.

boutinb: Ik zet nu ook de implicitWidth/Height

boutinb: Hi Joris, I did a push force, you cannot see the changes. What I did is just to set the implicitHeight/Width.

JorisGoosen: I can actually see the changes!
But...
Why are you setting implicitWidth/Height from preferredWidth/Height and not the other way round? Like I suggested...

boutinb: Because if the QML form changes the preferredHeight property, then the height and implicitHeight won't be changed.

JorisGoosen: Yes, that is exactly right. It shouldn't!
implicitWidth and height must be set by the component itself, so the code we are debating right now.
It is the "natural size" of the element.

If the user wants a different size then that's fine, but then they should change width/height.
However, we want to work with layout and *just to make it easier for the users* we added preferredWidth/Height as a replacement for Layout.preferredWidth etc.

JorisGoosen: And height will be changed by Layout.

boutinb: OK, if I set height to be preferredHeight, it solves also this issue.
Is this what you meant?

boutinb: I think I start to understand what you mean...
So the height and preferredHeight is set to implicitHeight (which has a default height). And the user has to set the preferredHeight if in Layout (that should set the right height), or directly the height if not in a layout.
I solve the issue by adding a binding.
Maybe my slow brain is by-effect of the corona virus...

PR #3962, opened 2020-03-16
Fixes logistic regression

Kucharssim:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/705
Fixes https://github.com/jasp-stats/jasp-test-release/issues/706

The bootstrapping error was a little bit nasty, in the end I went for a simple fix, but I think the whole bootstrap would deserve a slight rewrite.

### DIFF
--- a/JASP-Engine/JASP/R/regressionlogistic.R
+++ b/JASP-Engine/JASP/R/regressionlogistic.R
@@ -221,7 +221,6 @@ RegressionLogistic <- function(jaspResults, dataset = NULL, options, ...) {
   
   .reglogisticSetError(res, estimatesTableBootstrap)
   
-  # estimatesTableBootstrap$addFootnote(gettextf("Bootstrapping based on %i replicates.", options$coeffEstimatesBootstrappingReplicates))
   if (options$robustSEOpt)
     estimatesTableBootstrap$addFootnote(gettext("Coefficient estimate and robust standard error are based on the median of the bootstrap distribution."))
   else
--- a/JASP-Engine/JASP/R/regressionlogistic.R
+++ b/JASP-Engine/JASP/R/regressionlogistic.R
@@ -618,9 +618,10 @@ RegressionLogistic <- function(jaspResults, dataset = NULL, options, ...) {
         
         if (is.null(bootstrapResults[[bootname]])) {
 
-          # we compute additional statistics while bootstrapping, but we can't do this using boot
-          # (this is not true, you can boot whatever statistics you want, but ok... the bootstrapping should get some review anyway at some point)
+          # vandenman: we compute additional statistics while bootstrapping, but we can't do this using boot
           # so we hack it in there using an environment
+          # kucharssim: this is not true, you can boot whatever statistics you want, but ok... the bootstrapping should get some review anyway at some point
+          # discussion here: https://github.com/jasp-stats/jasp-desktop/pull/3962#discussion_r394348052
           envir <- new.env()
           envir$idx <- envir$idx_rse <- 0L
           envir$stdEst <- envir$robustSE <-
--- a/JASP-Engine/JASP/R/regressionlogistic.R
+++ b/JASP-Engine/JASP/R/regressionlogistic.R
@@ -221,7 +221,7 @@ RegressionLogistic <- function(jaspResults, dataset = NULL, options, ...) {
   
   .reglogisticSetError(res, estimatesTableBootstrap)
   
-  estimatesTableBootstrap$addFootnote(gettextf("Bootstrapping based on %i replicates.", options$coeffEstimatesBootstrappingReplicates))
+  # estimatesTableBootstrap$addFootnote(gettextf("Bootstrapping based on %i replicates.", options$coeffEstimatesBootstrappingReplicates))
   if (options$robustSEOpt)
     estimatesTableBootstrap$addFootnote(gettext("Coefficient estimate and robust standard error are based on the median of the bootstrap distribution."))
   else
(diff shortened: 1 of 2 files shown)


JASP-Engine/JASP/R/regressionlogistic.R
```diff
@@ -221,7 +221,7 @@ RegressionLogistic <- function(jaspResults, dataset = NULL, options, ...) {
   
   .reglogisticSetError(res, estimatesTableBootstrap)
   
-  estimatesTableBootstrap$addFootnote(gettextf("Bootstrapping based on %i replicates.", options$coeffEstimatesBootstrappingReplicates))
+  # estimatesTableBootstrap$addFootnote(gettextf("Bootstrapping based on %i replicates.", options$coeffEstimatesBootstrappingReplicates))
```
vandenman:
```suggestion
```
You can just delete this, since you do it more properly a couple lines down.

JASP-Engine/JASP/R/regressionlogistic.R
```diff
@@ -616,27 +616,30 @@ RegressionLogistic <- function(jaspResults, dataset = NULL, options, ...) {
         rn <- rownames(summary(glmObj[[i]])[["coefficients"]])
         rn[which(rn == "(Intercept)")] <- .v("(Intercept)")
         bootname <- paste(rn, collapse = "-")
-
+        
         if (is.null(bootstrapResults[[bootname]])) {
 
           # we compute additional statistics while bootstrapping, but we can't do this using boot
+          # (this is not true, you can boot whatever statistics you want, but ok... the bootstrapping should get some review anyway at some point)
```
vandenman: Just to clarify my comment, the issue was that `boot` isn't really suited to bootstrap both the standardized and unstandardized coefficients at the same time. Mainly because the standardized results can fail without the normal results failing. Rather than bootstrapping twice, I resorted to this approach with an environment. 

I get and agree with your comment, but maybe it gets a little bit confusing now for people that read it in the future. Perhaps:
```r
# vandenman: we compute additional statistics while bootstrapping, but we can't do this using boot so we hack it in there using an environment
# Kucharssim (this is not true, you can boot whatever statistics you want, but ok... the bootstrapping should get some review anyway at some point)
```
to clarify this is not a crazy person talking to themself?

vandenman (changes requested):
Looks good to me, two minor comments.

JASP-Engine/JASP/R/regressionlogistic.R
Kucharssim: Yeah, thanks! I am going with the following to make it more clear

```
# vandenman: we compute additional statistics while bootstrapping, but we can't do this using boot
# so we hack it in there using an environment
# kucharssim: this is not true, you can boot whatever statistics you want, but ok... the bootstrapping should get some review anyway at some point
# discussion here: https://github.com/jasp-stats/jasp-desktop/pull/3962#discussion_r394348052
```

vandenman (approved):

Resources/Regression/qml/RegressionLogistic.qml
```diff
@@ -107,16 +107,16 @@ Form
                 }
 
             }
-			CheckBox { name: "stdCoeff";		label: qsTr("Standardized coefficients")	}
-			CheckBox { name: "oddsRatios";		label: qsTr("Odds ratios")					}
+			CheckBox { name: "stdCoeff";		label: qsTr("Standardized coefficients"); enabled: coeffEstimates.checked	}
```
TimKDJ: Shouldn't all of these just be children of `coeffEstimates`? So we're inline with other analyses

Kucharssim: Hm, yeah you are right, although we will not be really consistent in across analyses regardless what I do here (e.g., confidence interval in linear regression is also not a child of estimates checkbox, but it also does not get enabled/disables depending on the value of it). 

I also find it more visually pleasing as it is now compared to how it looks when I make them children of the estimates button (see screenshots). For example, it seems to me logical then to remove the embedding of these buttons under the “Regression coefficients” group and rename the button “Estimates” to “Regression coefficients”, but then again that would totally go against the convention in other analyses. 

I don’t really mind any of the solutions, but I am not sure what is the best course of action. Any thoughts?

Current status
Screen Shot 2020-03-23 at 13 02 08

Proposed as children
Screen Shot 2020-03-23 at 13 02 45

vandenman: I agree that we should use the `QML` syntax to specify them as children. That's nicer and less error-prone than repeating the id many times. However, if everything is indented then nothing is. So that looks a bit redundant. Perhaps we can add a `QML` property to enable/ disable the indenting of children?

TimKDJ: > Hm, yeah you are right, although we will not be really consistent in across analyses regardless what I do here (e.g., confidence interval in linear regression is also not a child of estimates checkbox, but it also does not get enabled/disables depending on the value of it).

I think it should also be indented in Linear Regression to be honest.

> I also find it more visually pleasing as it is now compared to how it looks when I make them children of the estimates button (see screenshots). For example, it seems to me logical then to remove the embedding of these buttons under the “Regression coefficients” group and rename the button “Estimates” to “Regression coefficients”, but then again that would totally go against the convention in other analyses.

The group is probably called "Regression Coefficients" because that's where they go in Linear Regression as well, only in Logistic Regression it's the only option available for it. So it's a decision between across-analysis-consistency vs within-analysis-clarity I guess... In which case I'd say it's better to be consistent across analyses

> Proposed as children

To my eyes it becomes immediately clear that those checkboxes are meant to tweak estimates further. So I think that's actually really nice. Just like Descriptives, it just isn't as clear when it's not indented:
descriptives_indents

Clarity should be key in my opinion.

Kucharssim: Ok, ok! You convinced me.

TimKDJ (approved):

PR #3963, opened 2020-03-16
Use currentValue in place of currentText to check when custom is selected in CustomContrast

boutinb:
In Dutch, selecting handmatig in CustomContrat did not add a
CustomContrast: the condtionValriables in the customContrast
ComponentsList used the currentText, and it should use the currentValue
(which is not translated).

Clean up naming convention between value, label and key, this was too
confusing

### DIFF
--- a/JASP-Desktop/components/JASP/Controls/ComboBox.qml
+++ b/JASP-Desktop/components/JASP/Controls/ComboBox.qml
@@ -16,8 +16,11 @@ JASPControl
 	property alias	control:				control
 	property alias	controlLabel:			controlLabel
 	property alias	label:					controlLabel.text
-	property string	currentText				//Am i empty or what?
-	property string	currentColumnType		//Same here
+	// The 4 following properties should be set only from the backend. Unfortunately they cannot be set readonly
+	property string currentText:			""	// This is the current label displayed. currentText is the official name for a ComboBox in QML
+	property alias	currentLabel:			comboBox.currentText
+	property string currentValue:			"" // This is the current value (what is used by R)
+	property string currentColumnType:		"" // When the values come from column names, this property gives the column type of the current selected column
 	property alias	currentIndex:			control.currentIndex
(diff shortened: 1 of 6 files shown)


JASP-Desktop/components/JASP/Controls/ComboBox.qml
```diff
@@ -16,8 +16,10 @@ JASPControl
 	property alias	control:				control
 	property alias	controlLabel:			controlLabel
 	property alias	label:					controlLabel.text
-	property string	currentText				//Am i empty or what?
```
JorisGoosen, line 19: Would've been nice if you added an explanation here, because Im pretty sure I added those comments. Because I could not figure out (fast enough) what was the purpose of these properties and where/how they were filled/used.

I would really appreciate if you do add some explanation there.

JorisGoosen: Because, currentText means currentLabel? Or what?

JorisGoosen (changes requested):
The code in general looks good, the only thing I am still a bit iffy about is the "currentText" property. 
I was already confused about it before and still am.

Is it the shown label? 
Maybe a rename is in order? Or just a comment next to it explaining what it is.

I did actually put that comment  "Am i empty or what?" there to get a response, just throwing it away doesn't help at all.

JASP-Desktop/components/JASP/Controls/ComboBox.qml
```diff
@@ -16,8 +16,11 @@ JASPControl
...
-	property string	currentColumnType		//Same here
+	// The 4 following properties should be set only from the backend. Unfortunately they cannot set readonly
+	property string currentText:			""
+	property string currentLabel:			currentText
```
JorisGoosen: Shouldn't this be an alias?

JorisGoosen (approved):

PR #3964, opened 2020-03-16
Call MessageForwarder directly from QML

JorisGoosen:
- Fixes https://github.com/jasp-stats/jasp-test-release/issues/712

### DIFF
--- a/JASP-Desktop/components/JASP/Widgets/ModuleInstaller.qml
+++ b/JASP-Desktop/components/JASP/Widgets/ModuleInstaller.qml
@@ -70,7 +70,7 @@ Popup
 					margins:	jaspTheme.generalAnchorMargin
 				}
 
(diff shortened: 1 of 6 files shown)


JASP-Desktop/components/JASP/Widgets/ModuleInstaller.qml
```diff
@@ -70,7 +70,7 @@ Popup
 					margins:	jaspTheme.generalAnchorMargin
 				}
```
boutinb, line 71: This file if full of strings without qsTr. I suppose you don't think this is necessary for this version?

JorisGoosen: Well, no not really it is going to be replaced entirely anyway right?
By the module library?

JorisGoosen: And anyway, I only changed it to change the functioncall, I didn't even look at the other lines of code.

PR #3966, opened 2020-03-17
Create the R-translations files during the build

FransMeerhoff:
- Create the R-JASP.mo and R-JASPgraphs.mo translation files for translated R-code

### DIFF
--- a/JASP-Desktop/JASP-Desktop.pro
+++ b/JASP-Desktop/JASP-Desktop.pro
@@ -48,7 +48,6 @@ LIBS += -L.. -lJASP-Common
 windows:	LIBS += -llibboost_filesystem$$BOOST_POSTFIX -llibboost_system$$BOOST_POSTFIX -llibboost_date_time$$BOOST_POSTFIX -larchive.dll -llibreadstat -lole32 -loleaut32
 macx:   	LIBS += -lboost_filesystem-mt -lboost_system-mt -larchive -lz -lreadstat -liconv
 
-
 linux {
     LIBS += -larchive
     exists(/app/lib/*)	{ LIBS += -L/app/lib }
@@ -60,8 +59,6 @@ $$JASPTIMER_USED {
   linux:    LIBS += -lboost_timer -lboost_chrono
   macx:     LIBS += -lboost_timer-mt -lboost_chrono-mt
 }
-
-
 
 INCLUDEPATH += $$PWD/../JASP-Common/
 
@@ -86,7 +83,6 @@ exists(/app/lib/*) {
   #flatpak_appinfo_xml.files = ../Tools/flatpak.org.jaspstats.JASP.appdata.xml
   #flatpak_appinfo_xml.path = /app/share/appdata
   #INSTALLS += flatpak_appinfo_xml
-
 
   flatpak_appinfo_icon.files = ../Tools/flatpak/org.jaspstats.JASP.svg
   flatpak_appinfo_icon.path = /app/share/icons/hicolor/scalable/apps/
@@ -138,12 +134,10 @@ COPY_CRYPT = $$[ENVIRONMENT_CRYPTKEY] # We need to copy it to make sure the equa
   message("Key now set!")
 }
 
-
    macx:ICON = macOS/icon.icns
 windows:RC_FILE = icon.rc
 
 HELP_PATH = $${PWD}/../Docs/help
-
 
 SOURCES_TRANSLATIONS = $${PWD}/po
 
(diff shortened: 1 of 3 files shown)


FransMeerhoff:
@JorisGoosen 
Of course I'm open for a better solution but I found very annoying problems in the .pro file with using different path environments.

JASP-Desktop/JASP-Desktop.pro
```diff
@@ -210,8 +211,17 @@ unix {
...
 
+    #Create jasp_nl.qm
     maketranslations.commands += lrelease $$SOURCES_TRANSLATIONS/jasp_nl.po -qm $$RESOURCES_TRANSLATIONS/jasp_nl.qm ;
     maketranslations.commands += cp $$RESOURCES_TRANSLATIONS/*.qm $$RESOURCES_DESTINATION_TRANSLATIONS/ ;
+
+    #Create R-JASP.mo translation file. (Need to add GETTEXT location to PATH environment.)
+    maketranslations.commands += export SP=$PATH; export PATH=$PATH:$$GETTEXT_LOCATION;$$_R_HOME/Rscript $$PWD/../Tools/translate.R $$PWD/../JASP-Engine/JASP;export PATH=$SP ;
+
+    #Create R-JASPgraphs.mo translation file. (Need to add GETTEXT location to PATH environment.)
+    maketranslations.commands += export SP=$PATH; export PATH=$PATH:$$GETTEXT_LOCATION;$$_R_HOME/Rscript $$PWD/../Tools/translate.R $$PWD/../JASP-Engine/JASPgraphs;export PATH=$SP
```
JorisGoosen: Do you really need to restore PATH? 
You could just do `PATH=$PATH:$$GETTEXT_LOCATION $$_R_HOME/Rscript $$PWD/../Tools/translate.R $$PWD/../JASP-Engine/JASPgraphs` and then it just uses it for that command.

And it would be even better if you first stored the `PATH=$PATH:$$GETTEXT_LOCATION` in a var or something instead of duplicating it.

Tools/translate.R
```diff
@@ -1,4 +1,12 @@
...
-tools::update_pkg_po("../JASP-Engine/JASP")
+args <- commandArgs(trailingOnly = TRUE)
+
+if (length(args)==0) {
+  cat("Give the root folder of the package or the root of the  R-code of JASP")
+} else if (length(args) == 1) {
+  rootfolder<- args[1]
+}
+
+tools::update_pkg_po(rootfolder)
```
JorisGoosen, line 11: So if `length(args) == 0` then `tools::update_pkg_po(rootfolder)` is called, but what is `rootfolder` here?

And why can't you just call `$$_R_HOME/Rscript -e "tools::update_pkg_po('.')"` or something?

Tools/translate.cmd
```diff
@@ -0,0 +1,9 @@
+rem Create the R-nl language .mo file
+
+rem %1 is Rscript location
+rem %2 is get gettext location (stripped "")
+rem %3 Tools location
+rem %4 R source folder
+
+set PATH=%PATH%;%~2
+%1\Rscript %3\translate.R %4
```
JorisGoosen, line 9: So why is it a bat file on windows but a call on unix?
This is exactly the same commands right?

Or well, I guess the temporary env trick I alluded to earlier doesn't work on Windows...

JorisGoosen (changes requested):

JASP-Desktop/JASP-Desktop.pro
FransMeerhoff: Yes, but I was afraid the PATH would grow every time you extend it. It uses a variable now and that works fine too.

Tools/translate.R
FransMeerhoff: Yes I changed that in a stop when not 1 argument is given.

Tools/translate.cmd
FransMeerhoff: Yes the reason is  the env annoying thing windows/Qt

JASP-Desktop/JASP-Desktop.pro
JorisGoosen: Yeah I understand, but that is why I am telling you.
If you change this to not include `export` it is temporary.
Like, the exact code I've shown you

JorisGoosen: Just makes it simpler!

JorisGoosen: And that way you do not actually ever touch PATH which I appreciate a lot, what if there is some mistake sometime later and you try to build JASP with a messed up PATH? Would be bad

Tools/translate.R
JorisGoosen: Wouldn't `$$_R_HOME/Rscript -e "tools::update_pkg_po('$$SOME_BEAUTIFUL_QMAKE_VAR_HERE')"` also break when you do not enter a directory there? And we always run it on the exact same place right?

I think having an extra R file just for this is overkill.

Tools/translate.cmd
JorisGoosen: Or you want to have the R file because then you can use it from here? Because that isn't a bad reason actually, Windows is annoying indeed.

JorisGoosen (changes requested):
I guess the only change I would request now is changing the `export PATH=blabla; CMD` to just `PATH=blabla CMD` (make sure you remove the `;`)

JorisGoosen:
The translate.R does make sense to make sure you run the same command on Windows and Mac, but again, seeing as how it is a single command anyway I still think it is overkill.

Tools/translate.R
FransMeerhoff: It works fine now and I do not think it is an overkill. We can use it in the future for other R translated things.

Tools/translate.cmd
FransMeerhoff: You think I can merge this one after last changes? Then I can try to find the SIMPLE_CRYPT issue  :-)

PR #3967, opened 2020-03-17
Bayesian paired t-test: handle duplicate pairs

vandenman:
fixes https://github.com/jasp-stats/jasp-test-release/issues/678

Multiple duplicates:
image

One duplicate:
image

This behavior differs from the frequentist t-test. That analysis is less computationally demanding, so I suppose it's fine that the analysis is (needlessly) repeated. However, I'd rather not repeat the robustness/ sequential plots for identical variables.

### DIFF
--- a/JASP-Engine/JASP/R/commonbayesianttest.R
+++ b/JASP-Engine/JASP/R/commonbayesianttest.R
@@ -288,6 +288,21 @@
     derivedOptions[["wilcoxTest"]] <- FALSE
 
     dependents <- sapply(options[["pairs"]], paste, collapse = " - ")
+    duplicatedDependents <- duplicated(dependents)
+    if (any(duplicatedDependents)) {
+
+      derivedOptions[["footnotes"]] <- sprintf(
+        ngettext(sum(duplicatedDependents),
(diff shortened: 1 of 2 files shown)


TimKDJ:
While this solution seems fine, I'm wondering if it wouldn't be better to do this in qml instead. @boutinb would it be easy to allow only unique pairs? Hell we could even allow people to specify duplicate pairs and only send the unique ones

vandenman:
I think that doing this in qml would be cleaner. However, if we have R syntax and people call this in R the checks made by qml wouldn't be done.

TimKDJ:
> 
> 
> I think that doing this in qml would be cleaner. However, if we have R syntax and people call this in R the checks made by qml wouldn't be done.

I want every check qml does to be included in an R6 options class; hopefully this'll be doable! We'll include your fix just in case. But this should also be solved in qml, as it fixes the issue for 3 other analyses (2x equivalence, 1x frequentist)

boutinb:
I can easily add a constraint in the Paired VariablesList to forbid duplicate pairs.
This fix is still OK, but I add an issue to add this property.

PR #3969, opened 2020-03-17
Bayesian RM ANOVA: properly match names in old state and current run

vandenman:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/665

Something probably changed in the way variable names are generated, because the names differed between consecutive runs of the analysis. Maybe this is related to the column encoding, but I'm not entirely sure. To summarize the bug:

In run 1, the interaction effect is called `a:b`, while in run 2, the interaction effect is called `b:a`. In run 2, matching on `a:b` caused an error.


Kucharssim (approved):
Looks good, bug seems to be fixed, the code is clean.

PR #3970, opened 2020-03-17
Linear Discriminant Classification: Properly recreate MASS design matrix

vandenman:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/696

the problem was that `classificationResult[["train"]]` did not correspond to the model matrix used by `MASS:::lda.formula` when (ordered) factors were present.


FBartos:
I don't really understand the ML code to evaluate the change but I build it and it fixes the issue. It also doesn't seem to introduce new ones.

TimKDJ (approved):

PR #3972, opened 2020-03-18
Test Value should accepts formula

boutinb:
Fixes jasp-stats/jasp-test-release#682


boutinb:
@TimKDJ I have changed the test case so that the testValue is now a string, but the unit test is still failing. Do you have an idea?

TimKDJ:
Can you add `options$testValue <- "0.5"` after line 23 of test-summarystatsbinomialtestbayesian.R?

TimKDJ (approved):

PR #3973, opened 2020-03-18
Windows gonna be windows

### LINKED ISSUE #633: Using Filters as factor levels for Mann-Whitney
Hi, 

When doing a Mann-Whitney Independent Samples Test, I obviously can't use grouping variables with more than two factors. Unfortunately JASP also does not recognise filtering in this case. Thus you can't use a variable with several groups/factors as the grouping variable. In my case, there is a variable with several age groups and I'm only interested in comparing the youngest against the oldest group.  
Would be nice, if this could be fixed. 

Thank you, and keep up the good work!

JorisGoosen:
- should allow for local characters in windows
-- Fixes https://github.com/jasp-stats/jasp-issues/issues/633
- fixes small mistake with version of module not being correct
- Adds a test to testmodule for encoding of string in data
- Also changed unicode characters in common.R to \unicode escape chars notation
Double check for COPY_CRYPT
- fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/780


boutinb (approved):
I think this is ok... I have tested it only on Mac, and it seems to work.
But this is a quite a risky change just before the release. I will ask some people on mac/windows/linux tomorrow to test the nightly -built version with the 'strange characters' file.

JorisGoosen:
Yeah I agree it is a bit risky, but I've tested it myself on windows and linux and the application works better then before. Aka, you can actually filter on unicode.

I'll just revert the whitespace change in rbridge.cpp and then Ill merge it.

PR #3975, opened 2020-03-18
RM ANOVA - Make emmeans handle integer factor level names 

JohnnyDoorn:
Also no more unpooling error terms unbalanced design (series of paired t-tests is not with marginal means so would not be OK)

### DIFF
--- a/JASP-Engine/JASP/R/anovarepeatedmeasures.R
+++ b/JASP-Engine/JASP/R/anovarepeatedmeasures.R
@@ -917,9 +917,9 @@ AnovaRepeatedMeasures <- function(jaspResults, dataset = NULL, options) {
           levelBNoDots <- gsub(.unv(comparisons[[compIndex]][2]), pattern = "\\.", replacement = " ")
           facLevelNoDots <- gsub(longData[[var]], pattern = "\\.", replacement = " ")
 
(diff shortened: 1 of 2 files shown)


JASP-Engine/JASP/R/anovarepeatedmeasures.R
```diff
@@ -917,12 +919,12 @@ AnovaRepeatedMeasures <- function(jaspResults, dataset = NULL, options) {
           levelBNoDots <- gsub(.unv(comparisons[[compIndex]][2]), pattern = "\\.", replacement = " ")
           facLevelNoDots <- gsub(longData[[var]], pattern = "\\.", replacement = " ")
 
-          x <- subset(longData, facLevelNoDots == levelANoDots)
+          x <- subset(longData, gsub("X", "", facLevelNoDots) == gsub("X", "", levelANoDots))
```
vandenman, line 923: so why is there a `gsub` here? Because integer factors are automatically converted to X1, X2, etc? Maybe add a comment so that in the future we still understand what's going on.

JohnnyDoorn: Yes they get introduced by emmeans

PR #3976, opened 2020-03-18
Only the result column should be editable in a FilteredDataEntryModel TableView

boutinb:
Fixes jasp-stats/jasp-test-release#663


boutinb:
@koen is it only the last column that should be editable?

koenderks:
> @koen is it only the last column that should be editable?

Yes, only the column containing the audit result should be editable.

PR #3977, opened 2020-03-19
Add allowDuplicatesInMultipleColumns to VariablesList

boutinb:
Fixes jasp-stats/jasp-test-release#717

### DIFF
--- a/JASP-Desktop/widgets/listmodelmultitermsassigned.cpp
+++ b/JASP-Desktop/widgets/listmodelmultitermsassigned.cpp
@@ -27,6 +27,7 @@ ListModelMultiTermsAssigned::ListModelMultiTermsAssigned(QMLListView* listView, 
 	, _columns(columns)
 {
 	_copyTermsWhenDropped = true;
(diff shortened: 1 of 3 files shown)


JASP-Desktop/widgets/listmodelmultitermsassigned.cpp
```diff
@@ -27,6 +27,7 @@ ListModelMultiTermsAssigned::ListModelMultiTermsAssigned(QMLListView* listView,
 	, _columns(columns)
 {
 	_copyTermsWhenDropped = true;
+	_allowDuplicatesInMultipleColumns = listView->getItemProperty("allowDuplicatesInMultipleColumns").toBool();
```
JorisGoosen, line 30: So the value will not be updated afterwards if it changes right?
Maybe we should put this kind of info next to the properties in qml. with a comment or something.

boutinb: Hey Joris, you are still working?
I have added a comment

PR #3979, opened 2020-03-20
Add  bottom and top margin for paragraphs

boutinb:
Fixes jasp-stats/jasp-test-release#648

### DIFF
--- a/JASP-Desktop/html/css/darkTheme-jasp.css
+++ b/JASP-Desktop/html/css/darkTheme-jasp.css
@@ -670,3 +670,12 @@ pre {
   /* light theme uses the default value */
   border: 1px solid #5c5c5c;
 }
+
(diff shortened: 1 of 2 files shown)


TimKDJ:
Was this what was happening? A missing css styling? Or were those actual linebreaks?

Edit: I had a look at an older .jasp file and it is indeed styling that's missing. This fixes that problem. Could you perhaps have a quick glance if we can restore the shift+enter functionality (open JASP 0.11 write some text in a note and then press shift+enter)? If we merge this it'll become impossible to create a paragraph with linebreaks in it.

boutinb:
The shift-enter inserted a br not a p, my fix does not influences this.
That this function does not exist anymore should be solved in another issue I think

TimKDJ (approved):

PR #3981, opened 2020-03-20
Cleaned up some unused signals and fixed options ignores

JorisGoosen:
Instead of only increasing revision whenever analysis is sent to engine, always increment for each change of options
- fixes https://github.com/jasp-stats/jasp-test-release/issues/688

### DIFF
--- a/JASP-Desktop/engine/enginerepresentation.cpp
+++ b/JASP-Desktop/engine/enginerepresentation.cpp
@@ -113,7 +113,6 @@ void EngineRepresentation::clearAnalysisInProgress()
 
 void EngineRepresentation::setAnalysisInProgress(Analysis* analysis)
 {
-	analysis->incrementRevision(); // Increment revision only when the analysis request is about to be sent
 	if(_engineState == engineState::analysis)
 	{
 		if(_analysisInProgress == analysis)	return; //we are already busy with this analysis so everything is fine
--- a/JASP-Desktop/analysis/analyses.cpp
+++ b/JASP-Desktop/analysis/analyses.cpp
@@ -149,15 +149,10 @@ void Analyses::storeAnalysis(Analysis* analysis, size_t id, bool notifyAll)
 
 void Analyses::bindAnalysisHandler(Analysis* analysis)
 {
-	connect(analysis, &Analysis::optionsChanged,					this, &Analyses::analysisOptionsChanged				);
 	connect(analysis, &Analysis::statusChanged,						this, &Analyses::analysisStatusChanged				);
 	connect(analysis, &Analysis::sendRScript,						this, &Analyses::sendRScriptHandler					);
-	connect(analysis, &Analysis::toRefreshSignal,					this, &Analyses::analysisToRefresh					);
 	connect(analysis, &Analysis::titleChanged,						this, &Analyses::setChangedAnalysisTitle			);
-	connect(analysis, &Analysis::saveImageSignal,					this, &Analyses::analysisSaveImage					);
-	connect(analysis, &Analysis::editImageSignal,					this, &Analyses::analysisEditImage					);
 	connect(analysis, &Analysis::imageSavedSignal,					this, &Analyses::analysisImageSaved					);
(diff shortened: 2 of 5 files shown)


JASP-Desktop/engine/enginerepresentation.cpp
```diff
@@ -113,7 +113,6 @@ void EngineRepresentation::clearAnalysisInProgress()
 
 void EngineRepresentation::setAnalysisInProgress(Analysis* analysis)
 {
-	analysis->incrementRevision(); // Increment revision only when the analysis request is about to be sent
```
TimKDJ, line 116: I remember this fixed a nasty bug a while ago @boutinb (https://github.com/jasp-stats/jasp-desktop/commit/82195d175182488587fe4afdfdc0a763eb38d593). I can't remember the exact context though. Do you? Something about the analysis having a different revision status then what was returned from the engine; possibly after a filter? (You need to add more information to your pull requests ;))

boutinb: Yes I have to have a look. Next week.

JorisGoosen: It would be so practical if there was more info in @boutinb commits and PRs ;)

JorisGoosen: Like a link to an issue or something.

JorisGoosen:
@boutinb did you figure out what was up with your older commit and what nasty bug it solved?

boutinb:
I don't understand why you increment the revision when the options are changed. 
If an option is changed, the analysis status is set to 'empty': this is this status that triggers an engine to run the analysis. The options are read form the analysis just before it is send to the engine (in EngineRepresentation::runAnalysisOnProcess), and the status is set to 'running' (in Analysis::createAnalysisRequestJson). So if an option is changed just before runAnalysisOnProcess, the change is set in the options sent to the engine. If the change is set afterwards, the analysis status is set again to empty, and the analysis will be run again.
The revision is used only to ignore the results of a run if the options were in the meantime changed again. It is for this case important to increment the revision only when it is send to the engine.

JorisGoosen:
The problem here is that somewhere in the whole loop one of the options is missed. 

You know what, Ill just call you.

boutinb:
I think the problem I solved in October 2019 is solved in another way now....
The problem took place when an option set a new computed column (or did something with the filtering): in that case the engines are first paused, aborting the running analyses, and afterwards the analysis is refreshed (refresh used also to increment the revision).
The revision was incremented but the analysis was not run (because of the pause), and then it was refreshed, but apparently the wrong revision number was sent.
I cannot reproduce this error. But we have to ask @koenderks to test it carefully with the AUDIT module.

boutinb:
@JorisGoosen It's still a bit tricky to have solved a problem without knowing really why it helps: I still don't know why the increment of the revision during the option change, fixes the problem.

JorisGoosen:
Well, the problem is that apparently when two option-changes happen quickly the status of the analysis is set to empty. But then right in between the engine-sync decides to send the analysis to the actual engine. At that point there is (apparently) a small gap of time where an optionchange is missed. 

However, just trying to look now, to figure out what exactly *could* go wrong is quite confusing... In fact that is the reason I decided to fix it like this. To make it less confusing. Because right now, before this PR is merged, what does revision mean? I have no clue.

After this change: times an option changed. Quite clear right?

boutinb (approved):

koenderks:
If I need to test anything somewhere I am available. I'm not sure however what exactly this entails, but if you have some more information I would gladly test it.

boutinb:
@koenderks Can you test directly on Joris branch? Do you know how to do that? If not we can merge this PR, and you test it tomorrow.
What you have to test is in the AUDIT module, all the functions that create Computed Columns, and the Data Entry tables.

JorisGoosen:
The instructions are right there, in under "view command line instructions".

JorisGoosen:
@koenderks  and @boutinb how are things going with this?

boutinb:
Let's merge it. @koenderks you can test it carefully tomorrow?

PR #3982, opened 2020-03-23
PCA - missing dutch translations

FransMeerhoff:
- Added some missing translations
- Fixes https://github.com/jasp-stats/jasp-test-release/issues/690
- Start and end punctuations must be the same in the translated and original version


koenderks:
The PO syntax is not my best skill, but the punctuations etc. look good.

koenderks (approved):

PR #3983, opened 2020-03-24
Fixes bCor

AlexanderLyNL:
Requires Frans' update on jasp-required-files. Already implemented on mac. Windows is work in progress.

- Changed QML labels:
* Fixes: https://github.com/jasp-stats/jasp-test-release/issues/702
- Changes in '.getPosteriorPlotValuesCorBayes()', so dependencies are added to for pairs and sumStat accordingly
* Fixes: https://github.com/jasp-stats/jasp-test-release/issues/703
- Changed in bstats and getSidedObject to grab the ciValue as well, which is now passed through properly
* Fixes https://github.com/jasp-stats/jasp-test-release/issues/704
- Changed so there's a check on options[["ciValue"]] and postPlotValues[["ciValue"]]. If not, then take ciValue from corModel
* Fixes: https://github.com/jasp-stats/jasp-test-release/issues/711
- Changed in '.drawPosteriorPlotCorBayes()' so that for pairs, the posterior values for options[["pairsMethod"] is chosen instead of postPlotValuesPerMethod[[1]]
* Fixes: https://github.com/jasp-stats/jasp-test-release/issues/716

### DIFF
--- a/JASP-Engine/JASP/R/correlationbayesian.R
+++ b/JASP-Engine/JASP/R/correlationbayesian.R
@@ -783,17 +783,16 @@ CorrelationBayesian <- function(jaspResults, dataset=NULL, options, ...) {
         }
         # Note(Alexander): Not stored
         # 
-        ciValue <- options[["ciValue"]]
         lowerCi <- sidedBfWithoutPost[["lowerCi"]]
         upperCi <- sidedBfWithoutPost[["upperCi"]]
         medianPoint <- sidedBfWithoutPost[["posteriorMedian"]]
       }  else {
-        ciValue <- postPlotValues[["ciValue"]]
         lowerCi <- postPlotValues[["lowerCi"]]
         upperCi <- postPlotValues[["upperCi"]]
         medianPoint <- postPlotValues[["posteriorMedian"]]
       }
       
+      ciValue <- options[["ciValue"]]
       CRI <- c(lowerCi, upperCi)
       CRItxt <- gettextf("%s%% CI:", ciValue * 100)
     }
--- a/JASP-Engine/JASP/R/correlationbayesian.R
+++ b/JASP-Engine/JASP/R/correlationbayesian.R
@@ -39,15 +39,16 @@ CorrelationBayesian <- function(jaspResults, dataset=NULL, options, ...) {
 
   # 5. PairsPlot Container: (Optional) ------
   #
-  if (options[["plotScatter"]] || options[["plotPriorPosterior"]] || options[["plotBfRobustness"]] || options[["plotBfSequential"]]) {
+  if (options[["plotScatter"]] || options[["plotPriorPosterior"]] || options[["plotBfRobustness"]] || 
+      options[["plotBfSequential"]]) {
     pairsPlotCollection <- .getPairPlotsContainerCorBayes(jaspResults, options)
+    
     if (!is.null(pairsPlotCollection)) {
       pairsPlotCollection <- .initPlotContainerSubStructureCorBayes(pairsPlotCollection, options)
   
       .fillPairsPlotsCorBayes(jaspResults, pairsPlotCollection, corModel, dataset, options)
     }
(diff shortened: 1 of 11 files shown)


Resources/Help/analyses/summarystatscorrelationbayesianpairs_nl.md
```diff
@@ -10,18 +10,18 @@ Met Bayesiaanse correlatie analyse kan men de nulhypothese testen dat de populat
 - *n*: Steekproefgrootte (minimaal 2).
 
 #### Correlatiecoëfficiënt
-  - *Pearson's r*: Pearson's product-moment correlatie coëfficiënt.
-  - *Kendall's tau-b*
+  - *Pearson's r*: De geöbserveerde Pearson's product-moment correlatie coëfficiënt.
+  - *Kendall's tau-b*: De geöbserveerde Kendall's tau-b rangorde correlatiecoëfficiënt.
```
TimKDJ: `geobserveerde` -- zonder trema

JASP-Engine/JASP/R/correlationbayesian.R
```diff
@@ -471,12 +502,13 @@ CorrelationBayesian <- function(jaspResults, dataset=NULL, options, ...) {
...
-  if (bf >= 100)
+  if (bf >= 100) {
     symbol <- "***"
-  else if (bf >= 30)
+  } else if (bf >= 30) {
     symbol <- "**"
-  else if (bf >= 10)
+  } else if (bf >= 10) {
     symbol <- "*"
+  }
```
TimKDJ, line 511: I see that a major part of this PR is adding `{ }` around all the single statements. I find code easier to read when braces are not included when not necessary (it's nice to be consistent with our style; what do you think @vandenman, @JorisGoosen)?

TimKDJ (commented):
Seems to fix all the issues without introducing new ones. Nice!
I only have small nitpicks.

JASP-Engine/JASP/R/correlationbayesian.R
AlexanderLyNL: I do this especially whenever there's an 
- else 
- else if
- nested ifs/for/while
- statements of more than one line

Previous experience has shown R code failure due to these ambiguities

Resources/Help/analyses/summarystatscorrelationbayesianpairs_nl.md
AlexanderLyNL: I'm a bit old-fashioned; I also write poeyup with a long u

JASP-Engine/JASP/R/correlationbayesian.R
JorisGoosen: nested if: totally nice if there are braces because otherwise else becomes terribly ambiguous.

All other cases of one line: please no braces.. more typing, more mess and unnecessary.
More than one line: of course braces.

vandenman: I tend to avoid `{` whenever I can. That includes constructions like:

```r
for (x in xs)
  if (x < 0)
    # do thing
  else
    # do other thing
```
This only works if it's a single line though.

AlexanderLyNL: That already introduces ambiguity to me as I don't know where exactly it ends the for loop for me + white space issues.

TimKDJ: Okay so it seems I'm not alone in this. I'll go ahead and merge this, because I do not consider it important enough right before a release. But as you can see this type of change is not considered to be very desired.

JorisGoosen: And @AlexanderLyNL where is the ambiguity? How could the for loop only run the "then" part while ignoring the "else"? If it did that it would wouldn't work...

PR #3986, opened 2020-03-26
Make sure order of analyses is saved properly.

### LINKED ISSUE #639: Analyses order not retained...
JASP version: 0.11.1
OS name and version: WIndows 10

Repro:
1. Open JASP file with mutliple analyses e.g., Album Sales from the data library (Regression)
2. Re-order one or more analyses  e.g., drag Descriptives to the bottom and save again
3. Close and reopen the JASP file
Observed: analyses are not in the saved order e.g., Descriptives is back at the top
Expected: order of analyses should be retained from when last saved

JorisGoosen:
- Fixes https://github.com/jasp-stats/jasp-issues/issues/639
- Also moved all the functions for moving analyses together and into Analyses, Im not sure why part of it was on MainWindow.

### DIFF
--- a/JASP-Desktop/analysis/analyses.cpp
+++ b/JASP-Desktop/analysis/analyses.cpp
@@ -248,8 +248,8 @@ Json::Value Analyses::asJson() const
 	Json::Value analysesJson		= Json::objectValue,
 				analysesDataList	= Json::arrayValue;;
 
-	for(auto idAnalysis : _analysisMap)
-			analysesDataList.append(idAnalysis.second->asJSON());
+	applyToAll([&analysesDataList](const Analysis * analysis)
+		{ analysesDataList.append(analysis->asJSON()); });
 
 	analysesJson["analyses"]	= analysesDataList;
 	analysesJson["meta"]		= resultsMeta();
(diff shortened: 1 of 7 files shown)


TimKDJ:
That issue goes nowhere for me

JorisGoosen:
@TimKDJ what do you mean by that statement?

TimKDJ:
The issue link https://github.com/jasp-stats/jasp-issues/issues/639o doesn't link to an issue for me

JorisGoosen:
Haha, Im not used to this mac keyboard, the fn is on the place of ctrl which means fn+o in nano gives superfluous 'o' :-p 

Ill fix it and push-force it.

JorisGoosen:
So @boutinb how do you like this PR?

boutinb (commented):
By reviewing your code, I still don't see why it was not storing the new order in the JASP file?

JASP-Desktop/analysis/analyses.cpp
```diff
@@ -248,8 +248,8 @@ Json::Value Analyses::asJson() const
 	Json::Value analysesJson		= Json::objectValue,
 				analysesDataList	= Json::arrayValue;;
 
-	for(auto idAnalysis : _analysisMap)
-			analysesDataList.append(idAnalysis.second->asJSON());
+	applyToAll([&analysesDataList](const Analysis * analysis)
```
JorisGoosen, line 251: @boutinb here is the actual fix, the rest is just rearranging the deckchairs on the titanic. 
But the problem was that using an iterator on a std::map does not give the proper order at all. So the analyses weren't stored in the right order in the jasp-file. Which is why order was lost.

The function applyToAll was specifically added to make sure all analyses are processed in the right order. Why we didn't do it here, I don't know.

boutinb (approved):

PR #3990, opened 2020-03-27
The bracketeer

AlexanderLyNL:


JASP-Engine/JASP/R/summarystatscorrelationbayesianpairs.R
```diff
@@ -252,25 +253,18 @@ SummaryStatsCorrelationBayesianPairs <- function(jaspResults, dataset=NULL, opti
         options[["plotPriorPosterior"]] <- options[[currentName]]
       }
       
-      
-      if (currentName=="plotPriorAndPosteriorAdditionalInfo") {
+      if (currentName=="plotPriorAndPosteriorAdditionalInfo")
```
JorisGoosen: This could be done even nicer with a switch is what I wanted to say, but even better is a list:
```R
converter <- list( #I would align this all better if I was in an actual editor, but alas...
  plotPriorAndPosteriorAdditionalInfo="plotPriorPosteriorAddTestingInfo",
  plotBayesFactorRobustness="plotBfRobustness",
  plotBayesFactorRobustnessAdditionalInfo="plotBfRobustnessAddInfo",
  priorWidth="kappa")

options[[converter[[currentName]] ]] <- options[[currentName]]
```

Although I have the feeling that maybe some options ought to be renamed instead of doing this...

JorisGoosen (commented):
Bracketremoval is awesome ^^

Weird optionrenaming thing could be better and even better if it was made unecessary by renaming some stuff (in QML I guess?)

But for the actual approval denial thing I defer to @TimKDJ

AlexanderLyNL:
The QML names are based on the naive names we used in the past. I initially changed the names, but Tim told me to revert back to the old names due to compatibility issues. Instead of rewriting the code, I fixed it in this clunky manner.

JorisGoosen:
But we can just fix those names in the upgrader right? So there is no more problem then?

TimKDJ:
This was a relic from before we had the awesome upgrader. We can definitely use that now :)

AlexanderLyNL:
So can I change all the QML label names to the new sensible ones?

TimKDJ:
Only if you specify them in the upgrade json: https://github.com/jasp-stats/jasp-desktop/blob/development/Docs/development/jasp-upgrade-json.md

JorisGoosen:
The link to the manual was a good one, you will need it. If you get stuck, let me know though and I'll do it for you.

JorisGoosen:
And you should add them to https://github.com/jasp-stats/jasp-desktop/blob/development/JASP-Desktop/modules/upgrader/upgrades.json

AlexanderLyNL:
I just read it. Really nice set up. I'll check and mimic what happened for Bayesian correlation pairs, then adapt some R code so the dependencies work there as well.

AlexanderLyNL:
The upgrade system works amazingly well. However, whenever I save an old jasp file and open it in the new version, it resets the qml form. Did I do something wrong in the upgrade.json?

[oldSumStat102and111.zip](https://github.com/jasp-stats/jasp-desktop/files/4395814/oldSumStat102and111.zip)

AlexanderLyNL:
For some reason I can't get the radio buttons right. I basically copied pasted the correlation Bayesian pairs upgrader and it should work, but it doesn't :( Of course this is really disturbing, but perhaps I should think about it.

AlexanderLyNL:
Okay, the tricks seems to be not to change a radio button under changes first, and then add it again as a changed "conditional", which I didn't get.

AlexanderLyNL:
@JorisGoosen I'm not sure what's wrong here, but after opening

https://www.dropbox.com/s/albrbgnq5mju901/sumStatCor102.jasp?dl=0

it shows Pearson's r being 0.6 and inactive, as it should. When I tick the radio button in front of it to make it active, it doesn't refresh. Not necessary to look at it now though.

JorisGoosen:
I'll take a look tomorrow ^^
And yeah, the whole manually specifying upgrades in json is a bit hard, that's why I hope to move it to qml soon. I also get confused with this system despite having made it myself...

JorisGoosen:
Yeah, so it seems you actually set up the upgrade perfectly fine. There is only a little bug in the radiobuttons... I've fixed this and I'll make a separate pull-request for it.

PR #3989, opened 2020-03-27
Prevent editing wrong label when changing column

boutinb:
Fixes jasp-stats/jasp-test-release#715

PR #3991, opened 2020-03-27
There can be *only one* fix

JorisGoosen:
Just remember whatever was the chosenColumn on start editing and only overwrite when it is the same

-- Fixes jasp-stats/jasp-test-release#715

This PR is in direct opposition to https://github.com/jasp-stats/jasp-desktop/pull/3989


boutinb (approved):
This one is more generic... OK, then!

PR #3992, opened 2020-03-27
Adds a maxWidth property to jaspHtml that is set to 10cm by default

JorisGoosen:
(maybe we should make this 15cm?)

- implements https://github.com/jasp-stats/INTERNAL-jasp/issues/766 partially
- fixes https://github.com/jasp-stats/jasp-test-release/issues/673 partially

image
You see, 10cm is not that big. The Audit Risk thing is set to 20cm.

If this PR is merged all those manual breaks in gettext have to be removed as well..


JASP-R-Interface/jaspResults/R/zzzWrappers.R
```diff
@@ -102,8 +102,8 @@ createJaspTable <- function( title="",       data = NULL, colNames = NULL,     c
   return(jaspTableR$new(      title = title,  data = data, colNames = colNames, colTitles = colTitles,  overtitles = overtitles, colFormats = colFormats, rowNames = rowNames, rowTitles = rowTitles, dependencies = dependencies, position = position, expectedRows = expectedRows,  expectedColumns = expectedColumns))
 
 # if you change "hide me" here then also change it in Common.R and in HtmlNode.js or come up with a way to define it in such a way to make it show EVERYWHERE...
-createJaspHtml <- function( text = "",    elementType = "p",         class = "",    dependencies = NULL,         title = "hide me", position = NULL)
-    return(jaspHtmlR$new(   text = text,  elementType = elementType, class = class, dependencies = dependencies, title = title,     position = position))
+createJaspHtml <- function( text = "",    elementType = "p",         maxWidth = "10cm",   class = "",    dependencies = NULL,         title = "hide me", position = NULL)
```
TimKDJ: Might be better to use pixels? So all widths/heights you can specify in jaspResults are of the same unit. This could help when people try to add a html element under a 300 pixel plot for example. I'd also get rid of the unit and make it an integer argument so you have

```
createJaspPlot(width = 300)
createJaspHtml(maxWidth= 300)
```

JorisGoosen: you could just enter 300px if you like.

JorisGoosen: I specifically did it with cm by default to avoid all the dpi related madness.

TimKDJ: But then you get
```
createJaspPlot(width = 300)
createJaspHtml(maxWidth = "300px")
```
Which people will probably do wrong, because it's not very consistent.

JorisGoosen: I could simply add in the zzzWrappers a check for if numeric then as.character+px. That would make it consistent.

And actually, I am also unhappy with the fact that we specify pixels for the plots. it really doesn't make sense anymore.

TimKDJ: https://www.w3.org/Style/Examples/007/units.en.html has some pointers

Whatever we ultimately do, we should strive for a consistent API.
So if for now we can at least specify all the width's in a similar fashion I'm happy.

JorisGoosen: Ill make that change so it is consistent if needed but I think that despite the fact that that link recommends against using cm on screen I still think it is better. 
Because `em` is a bit better than `px` but still, I want max-width to be visually the same on all systems and while `em` would do that it would be a pretty weird number right?

While a value in cm's well, it should be pretty clear even if it isn't exact.

TimKDJ: I think the overal experience needs to be the same (size of elements relative to each other and sensible for the system it's on), but I disagree that an element should be the same absolute size regardless if it's on a 10" 1024 x 768 or a 50" 3840 x 2160...

JorisGoosen:
afbeelding
Is what it looks like now.

Problem is that now all those manual breaks look pretty bad...
But if those are replaced then also all dutch translations must be fixed?

JASP-R-Interface/jaspResults/R/zzzWrappers.R
JorisGoosen: Yeah, or a 20" 3840x2160... 

The whole dpi thing is meant exactly for this right? And while dpi is talking about inches I am very happy I can just use cm's. And then when you export to pdf/latex you get reasonable results right?

TimKDJ: > The whole dpi thing is meant exactly for this right?

I'm not sure what you mean by this. The dpi does not influence the size of the cm, but this is not a good thing, right? 15 cm on a high resolution screen will allow a lot more text than 15 cm on a low resolution screen. There's a good chance it won't work well on both.

TimKDJ:
Yeah it'll have to be changed in both the R code and also in the R-nl.po file...
I can do that later as well though

JorisGoosen:
You want to just do that in development? I guess the PR that Frans opened should also be processed first.

So this PR can be squashed and merged as far as you are concerned?

JorisGoosen:
Aah wait I missed that other comment...

JASP-R-Interface/jaspResults/R/zzzWrappers.R
JorisGoosen: Well, the font will actually be bigger on the higher res screen. Admittedly using `em` would also fix that. It is just that I find it hard to gauge what size a certain `em` specified size will be.

And `cm * dpi / 2.5` is the amount of pixels needed to get to the requested maxWidth. And if the font is also based on dpi (which I think it is) then that means the same amount of text fits in the same real-world screenspace. Just higher or lower res.

PR #3995, opened 2020-03-30
Fixes for JAGS

vandenman:
A lot of the issues in https://github.com/jasp-stats/jasp-test-release/issues/392 are fixed by this. Not all, so don't close that issue with this PR.

Tested with the model:
```r
model{
  theta ~ dbeta(1, 1)
  k ~ dbinom(theta, 10)
}
```

Below an overview of the fixed things.

Easy issues, all changed:

- In the output table, can we have "Posterior" on top of the "mean, sd, median" group?
- Same table: can we have the order "mean, median, sd"? (mean and median kinda go together)
- In JASP, is "sd" the way we always denote standard deviation? (I changed this to SD, like in Bayesian Linear Regression)
- The histogram is constructed by a series of vertical lines. Is this a histogram? Why not a histogram the way we do our other histograms?
- The option "Show legends" should be active for all plots -- right now it does not work for the ACF plot (I did not check the bivariate plot, but I'm guessing it does not work there either)
- The header of the main output table: should "summary" be "Summary", to be consistent with other JASP output (I checked one other analysis and there it was caps)
- The ACF plot -- maybe worthwhile to plot a horizontal dotted line on a correlation of zero? This is what people usually have as a comparison standard

More complicated issues that need some explanation:

> The option "Show Deviance" under "Advanced" does not appear to do anything.

My guess is that you tried a model without data? Because then the deviance cannot be computed. I've added a footnote to inform the user of this.

> Right now, clicking several display options (that merely control the cosmetics) cause the model to be rerun. This is completely unnecessary and prevents users from freely exploring the options. Imagine running the model takes 5 minutes. 

> ticking "Show Deviations" should not make the model rerun. Only options that require a rerun should trigger that.

Deviance is explictly not sampled unless someone asks for it. We could change that of course.

> So changing "thinning" should not cause a rerun, but entering a new seed should cause a rerun. I think this is a general concern for other analyses also.

There are two choices here.
1. We set the thinning in JAGS to 1 and then do the thinning ourselves. In that case, we don't have to resample whenever we change the thinning. However, if someone runs a large model with many samples and high thinning, this will be rather memory inefficient. 
2. We let JAGS do the thinning (current situation). This is more memory efficient because intermediate samples are never stored. At the same time, it does mean that we need to resample whenever the thinning value changes.

> The the binomial model, I am monitoring theta as well as k (so it is a predictive value). I was doing this to explore the bivariate plots. Then I discovered that the histogram of k looks funky, as the values of k=0 and k=1 seem to be added to one another, making the histogram there twice as high. Is this true?

This was a numerical artefact in the way `hist` computes breaks. It disappears when the number of mcmc samples increases. Nevertheless I've improved the logic a bit for things that can be detected as discrete (parameters with 25 or less unique values across all chains). Now this artefact should be gone. I've also prevented the x-axis labels from taking on non integer labels when there is a discrete parameter.

> The bivariate scatter also shows the anomaly near k=0. 

Same as the previous issue, the code to make these plots is identical.

### DIFF
--- a/JASP-Engine/JASP/R/jagsModule.R
+++ b/JASP-Engine/JASP/R/jagsModule.R
@@ -267,24 +267,28 @@ JAGS <- function(jaspResults, dataset, options, state = NULL) {
 # Tables ----
 .JAGSoutputTable <- function(jaspResults, options, mcmcResult) {
 
-  tb <- createJaspTable("MCMC summary")
+  tb <- createJaspTable("MCMC Summary")
   tb$position <- 1L
-  ovt  <- gettextf("95%% Credible Interval")
+  ovt0 <- gettext("Posterior")
+  ovt1 <- gettextf("95%% Credible Interval")
   ovt2 <- gettext("Rhat")
   tb$addColumnInfo(name = "parameter", title = gettext("Parameter"),            type = "string")
-  tb$addColumnInfo(name = "Mean",      title = gettext("mean"),                 type = "number")
-  tb$addColumnInfo(name = "SD",        title = gettext("sd"),                   type = "number")
-  tb$addColumnInfo(name = "50%",       title = gettext("median"),               type = "number")
-  tb$addColumnInfo(name = "2.5%",      title = gettext("Lower"),                type = "number", overtitle = ovt)
-  tb$addColumnInfo(name = "97.5%",     title = gettext("Upper"),                type = "number", overtitle = ovt)
+  tb$addColumnInfo(name = "Mean",      title = gettext("Mean"),                 type = "number", overtitle = ovt0)
+  tb$addColumnInfo(name = "50%",       title = gettext("Median"),               type = "number", overtitle = ovt0)
+  tb$addColumnInfo(name = "SD",        title = gettext("SD"),                   type = "number", overtitle = ovt0)
+  tb$addColumnInfo(name = "2.5%",      title = gettext("Lower"),                type = "number", overtitle = ovt1)
+  tb$addColumnInfo(name = "97.5%",     title = gettext("Upper"),                type = "number", overtitle = ovt1)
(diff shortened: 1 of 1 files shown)


AlexanderLyNL:
Could you please rebase? I've corrected my unit tests, and travis should be happy after you rebase.

boutinb:
@vandenman Tim cannot reviewen. I had a look to the code, it does not look dangerous... Can someone else review it? Or should we merge it?

vandenman:
@AlexanderLyNL or @koenderks do you have time? @boutinb if they don't have time I'd suggest merging it.

koenderks:
I can look at it, it does not look much right? However, when using the current Build of JASP I get this error message:

image

vandenman:
@koenderks see the [building guide](https://github.com/jasp-stats/jasp-desktop/blob/development/Docs/development/jasp-building-guide.md#Windows).

There should be a folder `JAGS` in your clone of jasp-required-files (e.g., [this one](https://github.com/jasp-stats/jasp-required-files/tree/Windows/64)). If you copy it to your `build-debug` folder,  everything should works.

koenderks:
The code looks fine. I tested it with your branch and that also seems to work fine. 

- Just thinking: Is the 95% credible interval not also actually part of the "Posterior" part of the table? This might be something for a future PR. 

- Also saw that the marginal density plot has a legend sometimes regardless of whether "show legends" is enabled, indicating something with the dependencies. When you click the plot "aggregate chains..." and then disable "show legends" the legend still appears next to the marginal density plot. I think this is best fixed as a small change in this PR.

- Also, when I click "selected parameters" all monitored parameters disappear...? This might be something for a future PR.

koenderks (approved):

boutinb:
@vandenman I suppose you have implemented the comment of Koen? We can then merge this PR.

vandenman:
@boutinb Koen's comment is now implemented.

PR #4002, opened 2020-03-31
Fix for Note compatibility issue

boutinb:
Fixes jasp-stats/jasp-test-release#653

There were 2 problems:
. when the userdata is changed in fitOldUserDataEtc, it's read in analysis render function, but the noteViews object must be then refreshed to use the new userdata.
. the analyses:userDataChanged event triggers the setResultsMetaFromJavascript function, resetting the userdata in cpp. This event is triggered each time a noteBox is created, that is when the analysis is rendered, but when the analysis is rendered, it reads the userdata from cpp : this explain the concurrency problem between the thread in JavaScript in the main thread in cpp.
That's why I have added a boolean in the userdata, so that it reads it from cpp only when the userdata comes from fitOldUserDataEtc

### DIFF
--- a/JASP-Desktop/analysis/analysis.cpp
+++ b/JASP-Desktop/analysis/analysis.cpp
@@ -856,6 +856,7 @@ void Analysis::fitOldUserDataEtc()
 		if(orphans.size() > 0)
 			_userData["orphanedNotes"] = orphans;
 
+		emit userDataChangedSignal(this);
 		Log::log() << "New userData after attempt to fix is: " << _userData.toStyledString() << std::endl;
 
 	} catch (...) {
--- a/JASP-Desktop/analysis/analyses.cpp
+++ b/JASP-Desktop/analysis/analyses.cpp
@@ -159,6 +159,7 @@ void Analyses::bindAnalysisHandler(Analysis* analysis)
 	connect(analysis, &Analysis::requestComputedColumnCreation,		this, &Analyses::requestComputedColumnCreation,		Qt::DirectConnection);
 	connect(analysis, &Analysis::requestComputedColumnDestruction,	this, &Analyses::requestComputedColumnDestruction,	Qt::DirectConnection);
 	connect(analysis, &Analysis::titleChanged,						this, &Analyses::somethingModified					);
+	connect(analysis, &Analysis::userDataChangedSignal,				this, &Analyses::analysisOverwriteUserdata			);
 
 	
 	if (Settings::value(Settings::DEVELOPER_MODE).toBool())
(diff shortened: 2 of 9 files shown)


JorisGoosen:
Wouldn't it be easier to only trigger analyses:userDataChanged when the user enters actual data in a notebox?

And then always process the userdata as it comes from Analysis?
That way the userData is also always stored there properly?

JorisGoosen:
Because your fix looks fine, don't get me wrong, I just think it is always a bit confusing if some random property is added somewhere in CPP that influences some small piece of code somewhere in JS.

And later on, maybe we would want to add some other cpp code that changes userData, which would then be ignored.

And that wouldn't be transparent, because it *would* be sent, just ignored without this random property...

JASP-Desktop/analysis/analysis.cpp
```diff
@@ -856,6 +856,7 @@ void Analysis::fitOldUserDataEtc()
 		if(orphans.size() > 0)
 			_userData["orphanedNotes"] = orphans;
 
+		_userData["fitOldUserDataEtc"] = true;
```
JorisGoosen: And please add comments when you do non-obvious things like this. If we look at this code in a year, would you expect to know what is going on here?

JorisGoosen: If you want to keep it likes this (as opposed to my suggestion) then maybe rename this property to "processThisUserdata" or something. Because "fitOldUserDataEtc" is not very descriptive.

JorisGoosen: And then we could reuse it later as well in other places.

boutinb:
The analyses:userDataChanged is I think needed when the noteBox is created to add the default value of the noteBox in the userdata.

This code
`var userdataCPP = this.model.get("userdata");`
is only added here because of this compatibility problem. The userdata should not be read again from the model. This code is already strange here.
Maybe a beter solution is to call from fitOldUserDataEtc a javascript function that reset the userdata of the analysis. What do you think?

JorisGoosen:
Well, I was thinking that I would like all the data to be in Analysis and *not* in javascript.
Because then we can keep all the eventloops out and all that. And there is no double administration and all that jazz.

Right now with this fix this is the case.

But, what I propose is a bigger change, so not a good plan for now anyway.

Calling a javscript-function "overwriteUserdata(json userdata)" would be better yeah.
Then we do not need this extra strange property and if later on somebody wants to change the userdata from C++ we can actually do it easily by calling that function.
It is still a bit confusing that userdata is still included in Analysis::asJSON() but well. We can always see about that later.

boutinb:
OK, I have changed it so that fitOldUserDataEtc calls an ovewriteUserData javascript function.

PR #4007, opened 2020-04-02
Post 0.12 - Handling column changes

JorisGoosen:
Solves https://github.com/jasp-stats/INTERNAL-jasp/issues/793 by removing or renaming variables used in computedcolumns or filters. This is done through the renamed or new signal/slot `datasetChanged`, previously `packageSynchronized`. 
This also triggered when a computed column is removed.

I also made some changes to AnalysisForm and Term(s) to make sure the variables/columnnames get replaced there and in the ListModels as well. (Because changing a columnname used in variableslist just removed it from there).

It is a rather large change though, so maybe it isn't a bad idea to wait with merging this until after we release 0.12.


JorisGoosen:
@boutinb I think we can merge this now, should give us enough time to test it before releasing. And Im pretty sure this qualifies as a bugfix ^^

JorisGoosen:
But Im not sure if you reviewed it or not.

boutinb (approved):

JorisGoosen:
Aah, it wouldve been nice to have this switched to stable branch, but we can do that later on.

PR #4010, opened 2020-04-03
bayesian meta analysis prettier axes posterior (Quentins request)

sophieberkhout:
@quentingronau noted that the prior and posterior plot axes are wider then they used to be. So with this, the axes should be closer to the spike again. 
Note that the prior distribution becomes a little difficult to see now (just looks flat).

@quentingronau could you maybe take my branch and see if this is better?


boutinb:
@vandenman If you can review it today, and merge it, then can Quentin test it tomorrow.

JorisGoosen:
@quentingronau mentioned in the issue that he can't build at the moment and well, if @vandenman approves we can just merge it and then @quentingronau can (double-)check if it works in a nightly.

vandenman (approved):

PR #4012, opened 2020-04-06
Added logs to and simplified LanguageModel

JorisGoosen:
- implements https://github.com/jasp-stats/jasp-test-release/issues/742

### DIFF
--- a/JASP-Desktop/utilities/languagemodel.cpp
+++ b/JASP-Desktop/utilities/languagemodel.cpp
@@ -17,23 +17,34 @@
 #include <QDirIterator>
 
 
-using namespace std;
+LanguageInfo::LanguageInfo()
+{
+	Log::log() << "new LanguageInfo created: " << toString() << std::endl;
+}
+
+LanguageInfo::LanguageInfo(QLocale::Language language, QString languageName, QString nativeLanguageName, QString localName, QString qmFilename, QString qmFolder)
+	: language(language), languageName(languageName), nativeLanguageName(nativeLanguageName), localName(localName), qmFolder(qmFolder), qmFilenames({qmFilename})
+{
+	Log::log() << "new LanguageInfo created: " << toString() << std::endl;
+}
+
+QString LanguageInfo::toString()
+{
+	return "Language: '" + languageName + "' qmFolder: '" + qmFolder + "' and qmFilenames: '" + qmFilenames.toList().join(", ") + "'";
+}
 
 LanguageModel * LanguageModel::_singleton = nullptr;
 
 LanguageModel::LanguageModel(QString qmresourcepath, QApplication *app, QQmlApplicationEngine *qml, QObject *parent)
-	: QAbstractListModel(parent)
+	: QAbstractListModel(parent),
+	  _mApp(app),
+	  _mTranslator(new QTranslator(this)),
+	  _qml(qml)
 {
 	assert(!_singleton);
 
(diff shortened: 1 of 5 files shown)


JASP-Desktop/utilities/languagemodel.cpp
```diff
@@ -157,59 +159,31 @@ void LanguageModel::resultsPageLoaded()
...
-	//Checks if the filename has a proper localename suffix for a valid CLocale::Language
-	//https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes
-	//We use the two lettecode in the qm filename
-
-	QString localname = getLocalNameFromQmFileName(filename);
-	QLocale loc(localname);
-	QString languagename = "";
-	languagename = loc.nativeLanguageName();
-	if (languagename == "")
+	if(!isValidLocalName(filename, loc))
```
boutinb, line 184: According to my c++ knowledge is the method
`bool LanguageModel::isValidLocalName(QString filename, QLocale & local)
{
	//Checks if the filename has a proper localename suffix for a valid CLocale::Language
	//https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes
	//We use the two lettecode in the qm filename

	QLocale loc			 = getLocalNameFromQmFileName(filename);
	QString languagename = loc.nativeLanguageName();

	if (languagename == "")
		return false;

	local = loc;
	return true;

}`

really wrong: local get a local value. I don't know how it can work

JorisGoosen: I do not understand what you are trying to say here, and whatever you copied there is a bit unclear?

Maybe it helps to look at [the entire file](https://github.com/jasp-stats/jasp-desktop/blob/bd35adce2dd8c6a184a16022fd90459ed39c8e7a/JASP-Desktop/utilities/languagemodel.cpp)?

Then we see:
```C++
bool LanguageModel::isValidLocalName(QString filename, QLocale::Language & cl)
{
	QLocale loc;

	if(!isValidLocalName(filename, loc))
		return false;

	cl = loc.language();
	return true;

}

bool LanguageModel::isValidLocalName(QString filename, QLocale & local)
{
	//Checks if the filename has a proper localename suffix for a valid CLocale::Language
	//https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes
	//We use the two lettecode in the qm filename

	QLocale loc			 = getLocalNameFromQmFileName(filename);
	QString languagename = loc.nativeLanguageName();

	if (languagename == "")
		return false;

	local = loc;
	return true;

}
```

JorisGoosen: First the code (and comment) were duplicated, which I thought a bit ugly. And then I noticed that it can be reused.

If you then look at `local = loc;` then that means that local gets a copy of loc right?
Because it is already a reference to whatever was outside the function?

JorisGoosen: See also https://isocpp.org/wiki/faq/references

How could you ever assign anything to a reference if it would change the address the reference points to?

boutinb: Sorry, it is something I would never do (local = loc) with an object reference, I thought 'local' would take the reference of 'loc' local variable. But it is a copy indeed. Usually if I want to copy a local variable, I return it, but it can be done this way also.

boutinb (approved):

PR #4019, opened 2020-04-08
Changed equivalence help files

Jillderon:


boutinb:
This is a lot of changes, just before the release tomorrow. I cannot review it completely. But as it is a beta module, we can push this PR.

Jillderon:
Ah, that's annoying! I was not aware that today is release, but if possible I would like to add these to the module before release yes

PR #4020, opened 2020-04-08
Model terms issue

boutinb:


boutinb:
If you have a JASP file with ModelTerms (Bayesian Anova/AncovaRepeated Measures, or Regression Linear, Regression Logistic/ Bayesian Regression Linear), where a CheckBox 'Add to null model' was checked, then opening the JASP file does not set the checkbox, making error when you refresh the analysis.

PR #4024, opened 2020-04-09
Add special software render options to startup

JorisGoosen:
- If safeGraphics is on "--disable-gpu" is added to startup, this should (could) turn of gpu rendering for chromium/webengine
- If safeGraphics is on the envvar LIBGL_ALWAYS_SOFTWARE=1 is set during startup, this should turn off all gpu rendering?
- Let's just try it all
- Hopefully fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/720
Changed the dutch message on turning on/off safe graphics
Made one of the functions (for language code) of languagemodel a bit bigger with a check and workaround, because I got some weird vector out of bounds errors from qt but not very reliably.
- In that scenario at least this would try to salvage it a bit and adds something to the logs for if this is a recurring thing.


JorisGoosen:
@boutinb the change to main with the arguments is the most important one. I am quite sure that it works properly this way, but if I messed it up JASP doesn't really work anymore so would be good if you check it well ^^

boutinb (approved):

PR #4029, opened 2020-04-14
Changing language let the VariablesList small

boutinb:
Fixes jasp-stats/INTERNAL-jasp#819

### DIFF
--- a/JASP-Desktop/components/JASP/Controls/VariablesForm.qml
+++ b/JASP-Desktop/components/JASP/Controls/VariablesForm.qml
@@ -57,7 +57,9 @@ Item
 		{
 			// Apparently a Qt bug: the height is not always recalculated by the GridLayout when the language is changed.
 			// Force this by changing temporarily the Layout.preferredHeight
-			variablesForm.Layout.preferredHeight = 0
+			// This fixes jasp-stats/jasp-test-release#731, but with Qt5.14.2, if preferredHeight is temporarly set to 0, it does not set it back to the original value afterwards.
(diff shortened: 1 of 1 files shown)


boutinb:
I know this fix is awkward... It might be worth a hotfix release

JorisGoosen:
Could this be towards `stable` instead of development?

JorisGoosen:
Also, maybe it would be worthwhile to check the warning in the output of Qt. I noticed a new warning about bindings or values so check out this:
https://doc.qt.io/qt-5/qml-qtqml-binding.html#restoreMode-prop
The warning I see on language change is `Warning: The default value is Binding.RestoreBinding. This will change in Qt 6.0 to Binding.RestoreBindingOrValue.` So maybe setting that in the original definition of the Binding helps?
Would be a bit less ugly and dependent on the current version behaving weirdly

boutinb:
The warning comes from the ScrollBar (and the restoreMode should be set indeed), but it does not seem to be related to this problem.

JorisGoosen:
No, but I thought if you replace the binding on `Layout.preferredHeight` (the original one) with a full `Binding{}` maybe it works better?

But If this solution works...

boutinb:
You pushed it already.... Your hint (with the Binding) did not work.

JorisGoosen:
Yes I know I pushed it already, I figured that your fix was as good as any and that you would notice the email that I pushed it as easily as a comment stating the fact that I was going to push it and then pushing it ;)

PR #4032, opened 2020-04-15
Remove warnings

boutinb:
With Qt5.14, new warnings are generated.
Repair also a crash when a discard model does not exist.

### DIFF
--- a/JASP-Desktop/components/JASP/Controls/JASPControl.qml
+++ b/JASP-Desktop/components/JASP/Controls/JASPControl.qml
@@ -97,7 +97,7 @@ JASPControlBase
 		State
 		{
 			name: "hasError"
-			when: focusIndicator && jaspControl.hasError
+			when: focusIndicator !== null  && jaspControl.hasError
 			PropertyChanges
 			{
 				target:			focusIndicator
@@ -108,7 +108,7 @@ JASPControlBase
 		State
 		{
 			name: "hasWarning"
-			when: focusIndicator && jaspControl.hasWarning && !jaspControl.hasError
+			when: focusIndicator !== null  && jaspControl.hasWarning && !jaspControl.hasError
 			PropertyChanges
 			{
 				target:			focusIndicator
@@ -119,7 +119,7 @@ JASPControlBase
 		State
 		{
 			name: "hasFocus"
-			when: focusIndicator && jaspControl.shouldShowFocus
+			when: focusIndicator !== null  && jaspControl.shouldShowFocus
 			PropertyChanges
 			{
 				target:			focusIndicator
@@ -130,7 +130,7 @@ JASPControlBase
 		State
 		{
 			name: "isDependency"
-			when: focusIndicator && jaspControl.isDependency && !jaspControl.shouldShowFocus
+			when: focusIndicator !== null && jaspControl.isDependency && !jaspControl.shouldShowFocus
 			PropertyChanges
 			{
 				target:			focusIndicator
--- a/JASP-Desktop/widgets/jaspcontrolwrapper.cpp
+++ b/JASP-Desktop/widgets/jaspcontrolwrapper.cpp
@@ -137,6 +137,7 @@ AnalysisForm *JASPControlWrapper::form() const
 
 void JASPControlWrapper::addControlError(const QString &error)
 {
+	Log::log() << "JASPControlWrapper::addControlError: " << error << std::endl;
 	if (form())
 		form()->addFormError(error);
 }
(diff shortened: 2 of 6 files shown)


JASP-Desktop/components/JASP/Controls/JASPControl.qml
```diff
@@ -97,7 +97,7 @@ JASPControlBase
 		State
 		{
 			name: "hasError"
-			when: focusIndicator && jaspControl.hasError
+			when: (typeof focusIndicator !== 'undefined')  && jaspControl.hasError
```
JorisGoosen: Im guessing `focusIndicator !== null` didn't work?

JASP-Desktop/widgets/jaspcontrolwrapper.cpp
```diff
@@ -137,6 +137,7 @@ AnalysisForm *JASPControlWrapper::form() const
 
 void JASPControlWrapper::addControlError(const QString &error)
 {
+	Log::log() << "Error: " << error << std::endl;
```
JorisGoosen: Could you change "Error: " to "JASPControlWrapper::addControlError: ", otherwise it will be pretty much impossible to guess what kind of error it is from the log only...

JASP-Desktop/widgets/listmodel.cpp
```diff
@@ -122,7 +122,8 @@ Terms ListModel::getSourceTerms()
 			Terms sourceTerms = sourceModel->terms(sourceItem->modelUse);
 
 			for (const QMLListView::SourceType& discardModel : sourceItem->discardModels)
-				sourceTerms.discardWhatDoesContainTheseComponents(discardModel.model->terms(discardModel.modelUse));
+				if (discardModel.model)
```
JorisGoosen: does it make sense to have a `discardModel` in a property called `discardModels` that does not actually have a model?

If so: this is a good fix
If not: why is that discardModel in the list at all?

boutinb: Don't understand your remark...
SourceType is a struct with a model property (of ListModel type) and a discardModels property, which is self a vector of SourceType. There is no discardModel property, it is in the code above just a variable name.

JorisGoosen: What I mean is that `sourceItem->discardModels` is a property, a vector right?

So why is there something in there without a model? Could it just be left out of that vector instead of adding those `if`s everywhere maybe?
Because that seems like something that will be forgotten at some point, or overlooked.

boutinb: A SourceType is structure with the definition of a model (the source property in QML). So it is first set with this definition. Afterwards, when the form is completely read, the real model is searched in the form. If it finds it, it is set to the SourceType.model property, if not, an error message is generated in the form.
The function above tries to read the terms of a source: if a discard model does not exist, this function crashes, that's why I have added this if statement.
It is indeed a quite peculiar situation, but JASP should not crash.

JorisGoosen: And `SourceType` is your own struct right?

Wouldn't it make more sense then to make a `getModelComponents()` function that does the check and returns something useful. 

I really think that needing to check if some property is null in quite a different place then where the struct is defined is bad programming practice to be honest. It is easily forgotten later on, when somebody needs to change the code.

And now if you would need components from the SourceType you can just ask for it instead of needing to check random stuff that is only clear because it crashed somewhere sometime.

JASP-Desktop/widgets/qmllistview.cpp
```diff
@@ -370,7 +370,8 @@ void QMLListView::sourceChangedHandler()
 			removeDependency(sourceModel->listView());
 			disconnect(sourceModel, &ListModel::modelChanged, listModel, &ListModel::sourceTermsChanged);
 			for (SourceType& discardModel : sourceItem->discardModels)
-				disconnect(discardModel.model, &ListModel::modelChanged, listModel, &ListModel::sourceTermsChanged);
+				if (discardModel.model)
+					disconnect(discardModel.model, &ListModel::modelChanged, listModel, &ListModel::sourceTermsChanged);
```
JorisGoosen: Here it is a bit reasonable to check the model for being null, but you could also let the connecting be done by `SourceType` itself and simply passing the listModel as a pointer. And then telling `discardModel.connectSourceTermsChanged(listModel)`. 

Hmm.

JorisGoosen:
So, honestly I don't like the way these models interact but never mind all my complaints about it, it is what it is.

JorisGoosen:
So the only thing I really would like to see changed is:
> Could you change "Error: " to "JASPControlWrapper::addControlError: ", otherwise it will be pretty much impossible to guess what kind of error it is from the log only..

boutinb:
I have added most of the changes you suggested.

JorisGoosen:
This looks nice :)

PR #4036, opened 2020-04-17
use same logic as c++ for missing qml files

vandenman:
@JohnnyDoorn Can you check if you can now directly load qml files without specifying them in `description.json`?

@JorisGoosen please verify if this is exactly the same logic as in `c++`.


JorisGoosen (commented):
qml from function if missing looks fine and the same as c++

JohnnyDoorn (approved):

PR #4037, opened 2020-04-17
Bayesian one-sample & paired Wilcoxon Tests

JohnnyDoorn:
Did not include  changes to inst/description.json (which is in my dynamic module PR:
https://github.com/jasp-stats/TTests/pull/1 )

### DIFF
--- a/JASP-Engine/JASP/R/commonbayesianttest.R
+++ b/JASP-Engine/JASP/R/commonbayesianttest.R
@@ -84,7 +84,7 @@
       }
     }
     ttestContainer$dependOn(c(
-      "effectSizeStandardized", "groupingVariable", "hypothesis",
+      "effectSizeStandardized", "groupingVariable", 
       "informativeCauchyLocation", "informativeCauchyScale", "informativeNormalMean",
       "informativeNormalStd", "informativeStandardizedEffectSize",
       "informativeTDf", "informativeTLocation", "informativeTScale",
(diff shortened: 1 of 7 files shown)


JohnnyDoorn:
Oh I guess that JASP::: does not work with Travis..

vandenman:
@JorisGoosen An easy way to deal with this is to create `JASP-Engine/JASP/DESCRIPTION` so that Travis can actually install JASP as a package. That way, we can copy the code from a dynamic module without having to fix annoying issues with Travis.

Shall I open a PR to stable to implement that?

JorisGoosen:
@vandenman there is already a [DESCRIPTION](https://github.com/jasp-stats/jasp-desktop/blob/stable/JASP-Engine/JASP/DESCRIPTION) and [NAMESPACE](https://github.com/jasp-stats/jasp-desktop/blob/stable/JASP-Engine/JASP/NAMESPACE) there right? You want to change it then? 

And whatever is the case, having it work in Travis would be great.

vandenman:
Ah I completely missed that they exist already. Well then adding a simple `devtools::install("~/github/jasp-desktop/JASP-Engine/JASP")` should make the package available on Travis (and locally). I'll see if I can add that.

vandenman:
@JohnnyDoorn I've created a pull request to this branch at https://github.com/JohnnyDoorn/jasp-desktop/pull/1 that should fix Travis.

vandenman:
@JohnnyDoorn Travis fails because `derivedOptions[["wilcoxTest"]] <- options[["testStatistic"]] == "Wilcoxon"` is only done for the paired-samples ttest. I'd put this before the if else chain or before the `if (derivedOptions[["wilcoxTest"]]) {` (line 309). Right now Travis fails because `derivedOptions[["wilcoxTest"]]` is still NULL when the if statement is reached.

JohnnyDoorn:
> @JohnnyDoorn Travis fails because `derivedOptions[["wilcoxTest"]] <- options[["testStatistic"]] == "Wilcoxon"` is only done for the paired-samples ttest. I'd put this before the if else chain or before the `if (derivedOptions[["wilcoxTest"]]) {` (line 309). Right now Travis fails because `derivedOptions[["wilcoxTest"]]` is still NULL when the if statement is reached.

haha yeah i noticed.. hope it's fixed now =)

vandenman (approved):

PR #4042, opened 2020-04-22
Custom R Library on Linux

vandenman:
Now you can set `JASP_R_Library` to a custom location under projects and use that folder instead. This means that

1. `/usr/R/library` is no longer contaminated by JASP (this folder is available to all R instances, which can make it unclear which package is loaded from where).

2. R packages no longer need to be installed in a folder with root permissions (this is more a convenience thing than strictly necessary).

The only thing you then still need on Linux is Rcpp in the folders with root permission.

I didn't make this work for Windows or Mac because (1) I can't test on those platforms and (2) they didn't have the two problems mentioned above to begin with. 

I also made some small changes to `gather-r-package-info.R`, that should make it easier to install the entire library from source to a custom directory.

### DIFF
--- a/JASP-Engine/JASP-Engine.pro
+++ b/JASP-Engine/JASP-Engine.pro
@@ -57,12 +57,12 @@ exists(/app/lib/*) {
    # InstallJASPgraphsRPackage.commands	= \"$$R_EXE\" CMD INSTALL --no-multiarch $$PWD/JASPgraphs
 } else {
     win32:RemoveJASPRPkgLock.commands   = IF exist \"$$OUT_PWD/../R/library/00LOCK-JASP/\" (rd /s /q \"$$OUT_PWD/../R/library/00LOCK-JASP/\" && echo Lock removed!) ELSE (echo No lock found!);
-    win32:InstallJASPRPackage.commands  = \"$$R_EXE\" -e \".libPaths(\'$$OUT_PWD/../R/library\'); install.packages(\'$$PWD/JASP\', lib=\'$$OUT_PWD/../R/library\', repos=NULL, type=\'source\', INSTALL_opts=\'--no-multiarch\')\"
-    unix: InstallJASPRPackage.commands  = export JASP_R_HOME=\"$$_R_HOME\" ; \"$$R_EXE\" -e \".libPaths(\'$$_R_HOME/library\'); install.packages(\'$$PWD/JASP\', lib=\'$$OUT_PWD/../R/library\', repos=NULL, type=\'source\', INSTALL_opts=\'--no-multiarch\')\"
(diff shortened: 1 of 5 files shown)


JASP-Engine/JASP-Engine.pro
```diff
@@ -57,12 +57,12 @@ exists(/app/lib/*) {
...
 
     win32:RemoveJASPgraphsRPkgLock.commands   = IF exist \"$$OUT_PWD/../R/library/00LOCK-JASPgraphs/\" (rd /s /q \"$$OUT_PWD/../R/library/00LOCK-JASPgraphs/\" && echo Lock removed!) ELSE (echo No lock found!);
-    win32:InstallJASPgraphsRPackage.commands  = \"$$R_EXE\" -e \".libPaths(\'$$OUT_PWD/../R/library\'); install.packages(\'$$PWD/JASPgraphs\', lib=\'$$OUT_PWD/../R/library\', repos=NULL, type=\'source\', INSTALL_opts=\'--no-multiarch\')\"
-    unix: InstallJASPgraphsRPackage.commands  = export JASP_R_HOME=\"$$_R_HOME\" ; \"$$R_EXE\" -e \".libPaths(\'$$_R_HOME/library\'); install.packages(\'$$PWD/JASPgraphs\', lib=\'$$OUT_PWD/../R/library\', repos=NULL, type=\'source\', INSTALL_opts=\'--no-multiarch\')\"
+    win32:InstallJASPgraphsRPackage.commands  = \"$$R_EXE\" -e \".libPaths(\'$$_R_Library\'); install.packages(\'$$PWD/JASPgraphs\', lib=\'$$OUT_PWD/../R/library\', repos=NULL, type=\'source\', INSTALL_opts=\'--no-multiarch\')\"
```
JorisGoosen, line 64: using _R_Library here changes the way mac works. `$OUTPWD/../R` is a folder called `R` in the buildfolder. While _R_Library is presumable `$_R_HOME/library` which is in `jasp-required-files` right?

JorisGoosen:
It mostly looks fine but I do think it will break Mac. Maybe @boutinb can test it out? 
I am going to chill in the sunshine now.

JASP-Engine/JASP-Engine.pro
vandenman: Hmm then maybe `.libPaths` works differently on mac? On Linux calling `.libPaths(new)` *replaces* the first entry of `.libPaths()`. If it works the same on mac then setting `.libPaths(buildfolder)` would crash the installation of `JASPgraphs` because the dependencies are missing. Note that the packages are still installed in `$$OUT_PWD/../R/library`.

JorisGoosen: Aah right, ok then this all looks fine.

JorisGoosen (approved):

PR #4050, opened 2020-04-22
fix encoding RM ANOVA Friedman

JohnnyDoorn:


boutinb:
@JohnnyDoorn Can you make another PR to jasp-stats:v.0.12.2.0 ?

JohnnyDoorn:
@boutinb  Hmm it seems that that branch is missing a lot that "stable" does have - should I ignore those? as in, clone V0.12.2.0 and than include my fix there?

boutinb:
Yes stable miss 2 big fixes from Joris & Koen. That's on purpose.
You should not need to rebase to v.0.12.2.0, just make a new PR with the same commit to v0.12.2.0. That should work

JohnnyDoorn:
If I try to do that then I get this, which does not seem trivial:
image

boutinb:
OK I will do it

JohnnyDoorn:
thanks! it's really only removing the .v in front of all instances of "subject"

PR #4059, opened 2020-04-24
Prevent crash when closing JASP

boutinb:


boutinb:
Fixes jasp-stats/jasp-issues#667

PR #4063, opened 2020-04-29
more tweaks to custom R library

vandenman:
- removed unnecessary DEFINES in `R_HOME.pri`
- custom r library now works for dynamic modules

after installing a dynamic module I now get the following libPaths in the analysis:
```
[1] "/home/dvdb/.local/share/JASP/JASP/Modules/DevelopmentModule"             
[2] "/home/dvdb/github/build-JASP-Desktop_Qt_5_14_2_GCC_64bit-Debug/R/library"
[3] "/home/dvdb/R/x86_64-pc-linux-gnu-library/3.6_JASP"                       
[4] "/usr/lib/R/library"                                                      
[5] "/usr/lib/R/site-library"                                                 
[6] "/usr/local/lib/R/site-library"                                           
```
for a normal analysis I get the same except for the first path:
```
[1] "/home/dvdb/github/build-JASP-Desktop_Qt_5_14_2_GCC_64bit-Debug/R/library"
[2] "/home/dvdb/R/x86_64-pc-linux-gnu-library/3.6_JASP"                       
[3] "/usr/lib/R/library"                                                      
[4] "/usr/lib/R/site-library"                                                 
[5] "/usr/local/lib/R/site-library"                                           
```

### DIFF
--- a/JASP-Desktop/modules/dynamicmodule.cpp
+++ b/JASP-Desktop/modules/dynamicmodule.cpp
@@ -358,7 +358,7 @@ std::string DynamicModule::generateModuleInstallingR(bool onlyModPkg)
 	std::string typeInstall = "'source'";
 //<< ".runSeparateR(\"{"
 
-	std::string libPathsToUse = "c('" + moduleRLibrary().toStdString()	+ "', .libPaths(.Library))";
+	std::string libPathsToUse = "c('" + moduleRLibrary().toStdString()	+ "', .libPaths())";
 
 	const char * pkgType =
 #ifdef __linux__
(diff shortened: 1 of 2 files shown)


JorisGoosen:
And the custom R libraries are added somewhere in your local environment then?

vandenman:
yes, `"/home/dvdb/R/x86_64-pc-linux-gnu-library/3.6_JASP"` is the one I add using an environment variable, the other ones are system defaults and the QT build folder.

JorisGoosen:
And that envvar is builtin to R?

vandenman:
nope builtin to QT, you specify it by setting by `JASP_R_LIBRARY`. The code above is oblivious to that though. it just ensures that no existing paths are lost from the `.libPaths`

JorisGoosen:
Ok, but then you should still make some changes. Because If I check R_HOME.pri then it seems you set that value into "_RLibrary" right?

And then that is used during install of JASP and JASPgraphs.

But how would that ever affect installing a dynamic module?

vandenman:
inside `enginesync.cpp` we do:
```c++
	if (env.contains("JASP_R_Library"))
		custom_R_library = ":" + env.value("JASP_R_Library");
```
and later
```c++
env.insert("R_LIBS",			programDir.absoluteFilePath("R/library") + custom_R_library + ":" + rHome.absoluteFilePath("library") + ":" + rHome.absoluteFilePath("site-library"));
```
At that point, the contents of `env.value("JASP_R_Library")` are added to the `.libPaths()`. The changes above result in appending the location of a dynamic module to the .`libPaths` rather than overwriting them. That's all.

JorisGoosen:
Aah yes ok, good. I have actually seen that before yeah. Thanks for clarifying it again for me :)

PR #4064, opened 2020-04-29
Add runAnalysisWhenOptionChanged property to QML and form

boutinb:
. Add also optionChanged signal and make refreshAnalysis invokable:
this makes now possible to control from QML when the analysis should be
run when some options are changed.
. Clean up code that set the debug & focus properties: transfer some
handlings from javascript to c++

### DIFF
--- a/JASP-Desktop/analysis/analysis.cpp
+++ b/JASP-Desktop/analysis/analysis.cpp
@@ -380,11 +380,8 @@ void Analysis::optionsChangedHandler(Option *option)
 	if (_refreshBlocked)
 		return;
 
-	if (form())
-	{
-		if (form()->hasError() || !form()->runAnalysisWhenThisOptionIsChanged(option))
-			return;
-	}
+	if (form() && (form()->hasError() || !form()->runWhenThisOptionIsChanged(option)))
+		return;
 
 	setStatus(Empty);
 }
--- a/JASP-Desktop/analysis/analysis.cpp
+++ b/JASP-Desktop/analysis/analysis.cpp
@@ -380,8 +380,11 @@ void Analysis::optionsChangedHandler(Option *option)
 	if (_refreshBlocked)
 		return;
 
-	if (form() && form()->hasError())
-		return;
+	if (form())
+	{
+		if (form()->hasError() || !form()->runAnalysisWhenThisOptionIsChanged(option))
+			return;
+	}
 
 	setStatus(Empty);
 }
--- a/JASP-Desktop/analysis/analysisform.h
+++ b/JASP-Desktop/analysis/analysisform.h
@@ -52,7 +52,7 @@ class AnalysisForm : public QQuickItem, public VariableInfoProvider
 	Q_PROPERTY(QQuickItem * errorMessagesItem	READ errorMessagesItem	WRITE setErrorMessagesItem	NOTIFY errorMessagesItemChanged	)
 	Q_PROPERTY(bool			needsRefresh		READ needsRefresh									NOTIFY needsRefreshChanged		)
 	Q_PROPERTY(bool			hasVolatileNotes	READ hasVolatileNotes								NOTIFY hasVolatileNotesChanged	)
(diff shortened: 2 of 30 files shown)


JASP-Desktop/analysis/analysis.cpp
```diff
@@ -380,8 +380,11 @@ void Analysis::optionsChangedHandler(Option *option)
 	if (_refreshBlocked)
 		return;
 
-	if (form() && form()->hasError())
-		return;
+	if (form())
```
JorisGoosen: Why not `if(form() && (form()->hasError() ||  !form()->runAnalysisWhenThisOptionIsChanged(option))) `, it saves like three lines

JASP-Desktop/analysis/analysisform.h
```diff
@@ -52,19 +52,24 @@ class AnalysisForm : public QQuickItem, public VariableInfoProvider
 	Q_PROPERTY(QQuickItem * errorMessagesItem	READ errorMessagesItem	WRITE setErrorMessagesItem	NOTIFY errorMessagesItemChanged	)
 	Q_PROPERTY(bool			needsRefresh		READ needsRefresh									NOTIFY needsRefreshChanged		)
 	Q_PROPERTY(bool			hasVolatileNotes	READ hasVolatileNotes								NOTIFY hasVolatileNotesChanged	)
+	Q_PROPERTY(bool			runAnalysisWhenOptionChanged	READ runAnalysisWhenOptionChanged	WRITE setRunAnalysisWhenOptionChanged NOTIFY runAnalysisWhenOptionChangedChanged )
```
JorisGoosen: Wouldn't `runOnChange` be a little bit more succint?
I mean, it is obviously about an analysis, because it is a form and that it is an option should also be obvious.

JASP-Desktop/analysis/analysisform.h
```diff
@@ -149,6 +159,8 @@ private slots:
 	QMap<QString, JASPControlWrapper* >			_controls;
 
 	QVector<JASPControlWrapper*>				_orderedControls;
```
JorisGoosen, line 161: Is this then a list *with* names? 
And it would be cleaner to avoid duplicating the type all the time, so:
```
QVector<JASPControlWrapper*>				_orderedControls,
																	_allItems;
...
QSet<JASPControlBase*>						_jaspControlsWithErrorSet,
																_jaspControlsWithWarningSet;
```

JASP-Desktop/analysis/analysis.cpp
boutinb: Done

JASP-Desktop/analysis/analysisform.h
boutinb: Done

JASP-Desktop/analysis/jaspcontrolbase.cpp
```diff
@@ -49,6 +49,28 @@ void JASPControlBase::setHasWarning(bool hasWarning)
 	}
 }
 
+void JASPControlBase::setRunOnChangeToChildren(bool change)
```
JorisGoosen, line 52: Is it hard to get the children to simply check their parent? Instead of having to go through the children and changing it manually?
Would be a bit cleaner right? And you could simply connect the runOnChangeChanged signals from parent->child.

JorisGoosen: Although I guess this isn't so bad either and we could reuse the code you've written now (getChildJASPControls) for the QML help file generation as well. So nvm.

JorisGoosen:
Looks good!

Is there an easy to actually test if it works? Because as far as Im concerned it could be merged already, but testing it might be nice.

boutinb:
You change for example Descriptives.qml (directly in the build/Resources directory:
. Add `runOnChange: false` in a section for example
. Add a button like this:
```
	Button
	{
		id: runAnalysis
		name: "runAnalysis"
		label: "Run Analysis"
		enabled: false
		Connections
		{
			target: form
			onOptionChanged: runAnalysis.enabled = true
		}
		onClicked:
		{
			form.refreshAnalysis()
			enabled = false;
		}
	} 
```
Check then the changes made onder the section does not update the results, but that the button 'Run Analysis' is then enabled.
Click the button: the analysis in run, and the button is disabled.

JorisGoosen:
Ill do that tomorrow

PR #4065, opened 2020-04-29
p-value should be set with the 'Number of decimals' setting

### LINKED ISSUE #706: p value shown as 0.00 instead of <.001
* JASP version: 0.12.2
* OS name and version: macOS Catalina 10.15.4
* Analysis: Classical ANOVA
* Bug description: In the Assumption Checks, the _p_ value is shown as 0.00.
ANOVA_teachers
* Expected behaviour: The _p_ value is expected to be shown as <.001 just like in the ANOVA table and the Post Hoc Tests.

#### Steps to reproduce:
1. Click on ANOVA > Classical ANOVA.
2. Under 'Assumption Checks', check 'Homogeneity tests'.

[ANOVA_teachers.jasp.zip](https://github.com/jasp-stats/jasp-issues/files/4531121/ANOVA_teachers.jasp.zip)

boutinb:
Fixes jasp-stats/jasp-issues#706

### DIFF
--- a/JASP-Desktop/html/js/utils.js
+++ b/JASP-Desktop/html/js/utils.js
@@ -78,8 +78,8 @@ function formatColumn(column, type, format, alignNumbers, combine, modelFootnote
             // override APA style if exact p-values wanted
             if (window.globSet.pExact) {
                 sf = 4;
-            } else {
-                p = f.substring(2);
+			} else {
+				p = 1/(Math.pow(10,dp));
             }
         }
 


vandenman:
I don't know much about js, what is this supposed to do?

boutinb:
I thought you knew a little bit. 
The problem was that if a table uses a p-value, it is automatically associated with the format 'p:0.001', that means that it takes maximum 3 decimals, and if it is smaller than 0.001, then is displays '< .001'. This was done without taking into account the 'Number of Decimals' setting in Preferences. I don't know why it was done like that, but now I just change the code to use always the 'Number of Decimals'

vandenman:
Well, I know what it's *supposed* to do.. but I've got no clue about the actual implementation on the javascript end!

boutinb:
OK, then I ask Joris

JorisGoosen:
Haha, well I guess your solution looks ok. 

But to be honest, that is one horrific function you've got there... I remember trying to understand it to get the formatting supported in jaspTable as well and going crazy because it is so convoluted.

You tested it right? Because I don't really want to go and have to double check it by building and running JASP just for this one-liner. 

Your change definitely looks a lot better then the weird substring thing there was before.

boutinb:
Yes it is tested.

PR #4067, opened 2020-04-30
Add different features/fixes for Mixed Models

boutinb:
. addNewAvailableTermsToAssignedModel property should work for all
VariablesList
. Alignment of TextField in a Group should work even if some of the
TextField becomes visible/invisible
. Add “noInteraction” in the source property to discard interaction
terms.
. Add “type=scale|ordinale|ordinal| in the source property to select
only some column types
. Add itemTypes property in VariablesList that contain all column types
. MarginalMeansContrast does not update if value is set back to 0
(default value)

### DIFF
--- a/JASP-Common/enumutilities.h
+++ b/JASP-Common/enumutilities.h
@@ -33,6 +33,7 @@
 	std::string		&operator+=(std::string &str, E enumTmp);													\
 	E				operator++(E &enumTmp);																		\
 	E				E##FromString(std::string enumName);														\
+	E				E##FromString(std::string enumName, E devaultValue);										\
 	std::string		E##ToString(E enumVal);																		\
 	bool			valid##E(T value);
 
@@ -71,6 +72,12 @@
 			throw std::runtime_error(#E" enum from string misses requested value \""+enumName+"\"");			\
 		return (E)E##FromNameMap.at(enumName); 																	\
 	}																											\
+	E E##FromString(std::string enumName, E defaultValue)														\
+	{																											\
+		if(E##FromNameMap.count(enumName) == 0) 																\
+			return defaultValue;																				\
+		return (E)E##FromNameMap.at(enumName); 																	\
+	}																											\
 	std::string E##ToString(E enumVal)		{ return ~enumVal; }												\
 	bool		valid##E(T value) { return (E##MapName.find(value) != E##MapName.end()); }
 
@@ -78,6 +85,7 @@
 	#define DECLARE_ENUM_WITH_TYPE_HEADER(E, T, ...)															\
 	DECLARE_ENUM_WITH_TYPE_BASE(E, T, __VA_ARGS__)																\
 	inline E		E##FromQString(QString enumName)	{ return (E)E##FromString(enumName.toStdString()); }	\
+	inline E		E##FromQString(QString enumName, E defaultValue)	{ return (E)E##FromString(enumName.toStdString(), defaultValue); }	\
 	inline QString	E##ToQString(E enumVal)		{ return QString::fromStdString(~enumVal); }					\
 	inline QString	operator+(QString &&str, E enumTmp) { return str + E##ToQString(enumTmp); }					\
 	inline QString	operator+(E enumTmp, QString &&str) { return E##ToQString(enumTmp) + str;	}				\
--- a/JASP-Desktop/analysis/analysisform.cpp
+++ b/JASP-Desktop/analysis/analysisform.cpp
@@ -798,7 +798,9 @@ bool AnalysisForm::runWhenThisOptionIsChanged(Option *option)
 	BoundQMLItem* control = getBoundItem(option);
 	JASPControlBase* item = control ? control->item() : nullptr;
 
-	emit optionChanged(item);
+	emit valueChanged(item);
+	if (item)
+		emit item->valueChanged();
 
 	if (!_runOnChange)
 		return false;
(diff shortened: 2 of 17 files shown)


JASP-Desktop/analysis/analysisform.cpp
```diff
@@ -798,7 +798,9 @@ bool AnalysisForm::runWhenThisOptionIsChanged(Option *option)
 	BoundQMLItem* control = getBoundItem(option);
 	JASPControlBase* item = control ? control->item() : nullptr;
 
-	emit optionChanged(item);
+	emit valueChanged(item);
```
JorisGoosen, line 801: Why would you want to send `valueChanged(nullptr)`, does that make sense?

JASP-Desktop/widgets/listmodel.cpp
```diff
@@ -465,6 +479,37 @@ const QString &ListModel::name() const
 	return _listView->name();
 }
 
+const Terms &ListModel::terms(const QString &what) const
+{
+	if (what.startsWith("type="))
+	{
+		static Terms terms;
+
+		QStringList typesStr = what.right(what.length() - 5).toLower().split("|");
```
JorisGoosen: I would be more happy with:
```
const QString typeIs = "type=";
if (what.startsWith(typeIs))
	{
		static Terms terms;

		QStringList typesStr = what.right(what.length() - typeIs.length()).toLower().split("|");
```
This way we don't get weird bugs if someone changes it (or worse, copies it and reuses it somewhere else without thinking properly)

JASP-Desktop/widgets/listmodel.cpp
```diff
@@ -465,6 +479,37 @@ const QString &ListModel::name() const
...
+		QList<columnType> types;
+
+		for (const QString& typeStr : typesStr)
+		{
+			columnType type = columnType::unknown;
+			try { type = columnTypeFromQString(typeStr); } catch (...) {}
```
JorisGoosen: Do you think it would be useful to create a second version of `enumFromQString` that gives back a default enumval if the one the user gives is crap?
Would make this code prettier!
And it wouldn't be that hard either.

JASP-Desktop/widgets/listmodelinteractionassigned.cpp
```diff
@@ -40,6 +39,23 @@ void ListModelInteractionAssigned::initTerms(const Terms &terms, const RowContro
 	ListModelAssignedInterface::initTerms(interactionTerms(), allOptionsMap);
 }
 
+const Terms &ListModelInteractionAssigned::terms(const QString &what) const
+{
+	if (what == "noInteraction")
+	{
+		static Terms terms;
+
+		terms.clear();
```
JorisGoosen, line 48: Why not do: `terms = { _fixedFactors, _randomFactors, _covariates };` ?

JorisGoosen (changes requested):

JASP-Desktop/analysis/analysisform.cpp
boutinb: The Form valueChanged signal does not need per se to know which item is changed (in the code I gave you for the last PR for example). So even if the item is not found (it could be the case with controls made dynamically (like rowControls), the signal should be emit.

JASP-Desktop/widgets/listmodel.cpp
boutinb: Good idee!

JASP-Desktop/widgets/listmodelinteractionassigned.cpp
boutinb: Terms does not support this construction

JASP-Desktop/widgets/listmodel.cpp
boutinb: Done

JASP-Desktop/analysis/analysisform.cpp
JorisGoosen: Ok!

JASP-Desktop/widgets/listmodelinteractionassigned.cpp
JorisGoosen: Aah, wouldn't it be nice to add that? Well, I guess it isn't so important because we will remove/change Terms anyway right?

JASP-Desktop/widgets/listmodel.cpp
JorisGoosen: Will you try it? Then I won't be the only one who has a clue about `enumutilities.h` ;)

JorisGoosen: You already did it! Haha

JASP-Common/enumutilities.h
```diff
@@ -71,13 +72,20 @@
...
+		return (E)E##FromNameMap.at(enumName); 																	\
+	}																											\
 	std::string E##ToString(E enumVal)		{ return ~enumVal; }												\
 	bool		valid##E(T value) { return (E##MapName.find(value) != E##MapName.end()); }
 
 #ifdef JASP_USES_QT_HERE
 	#define DECLARE_ENUM_WITH_TYPE_HEADER(E, T, ...)															\
 	DECLARE_ENUM_WITH_TYPE_BASE(E, T, __VA_ARGS__)																\
 	inline E		E##FromQString(QString enumName)	{ return (E)E##FromString(enumName.toStdString()); }	\
+	inline E		E##FromQString(QString enumName, E defaultValue)	{ return (E)E##FromString(enumName.toStdString(), defaultValue); }	\
```
JorisGoosen, line 88: Im a little bit sad you broke the nice alignment here, but well... That is life I guess?

PR #4069, opened 2020-05-04
Independent samples T-Test: fix "result not found" bug

### LINKED ISSUE #724: Error in an independent sample T-test
* JASP version: 0.12.2

* OS name and version: Windows
* Analysis: ?
* Bug description:
This analysis terminated unexpectedly.

Error in .ttestIndependentMainFill(ttest, dataset, options, testStat, : object 'result' not found

Stack trace
tryCatchOne(expr, names, parentenv, handlers[[1]])

doTryCatch(return(expr), name, parentenv, handler)

withCallingHandlers(expr = analysis(jaspResults = jaspResults, dataset = dataset, options = options), error = .addStackTrace)

.ttestIndependentMainTable(jaspResults, dataset, options, ready, type)

.ttestIndependentMainFill(ttest, dataset, options, testStat, optionsList)

* Expected behaviour: a table

#### Steps to reproduce:
1. Go to '.T-test..'
2. Click on '..independent samples t test..'
3. Scroll down to '..fill in a variable and a grouping variable..'

vandenman:
Fixes https://github.com/jasp-stats/jasp-issues/issues/724

The bug is the following (the diff doesn't show all relevant code). Starting from https://github.com/jasp-stats/jasp-desktop/blob/stable/JASP-Engine/JASP/R/ttestindependentsamples.R#L239, we have:

```r
if (identical(errors, FALSE)) {
  result <- try(ttestIndependentMainTableRow(variable, dataset, test, testStat, effSize, optionsList, options))

  # more stuff 
}

if (result[["leveneViolated"]])
  table$addFootnote(gettext("Levene's test is significant (p < .05), suggesting a violation of the equal variance assumption"), colNames = "p", rowNames = rowName)
```
The bug is that when `identical(errors, FALSE)` returns `FALSE`, `result` is never created and `if (result[["leveneViolated"]])` causes an error. 

The fix is to put the `if (result[["leveneViolated"]])` part inside of the `if (identical(errors, FALSE))`. 

To replicate the error, run an independent samples t-test with `debInf` as variable and `contBinom` as grouping variable. 
Stable returns the same error as the linked issue, this PR shows what's wrong with `debInf` as a footnote in the main table (the intended behavior).

### DIFF
--- a/JASP-Engine/JASP/R/ttestindependentsamples.R
+++ b/JASP-Engine/JASP/R/ttestindependentsamples.R
@@ -241,8 +241,13 @@ TTestIndependentSamples <- function(jaspResults, dataset = NULL, options, ...) {
         
         if (!isTryError(result))
           row <- c(row, result[["row"]])
-        else
+        else {
           errorMessage <- .extractErrorMessage(result)
+
+          if (result[["leveneViolated"]])
+            table$addFootnote(gettext("Levene's test is significant (p < .05), suggesting a violation of the equal variance assumption"), colNames = "p", rowNames = rowName)
+
+        }
         
       } else {
         errorMessage <- errors$message
@@ -252,9 +257,6 @@ TTestIndependentSamples <- function(jaspResults, dataset = NULL, options, ...) {
         row[[testStat]] <- NaN
         table$addFootnote(errorMessage, colNames = testStat, rowNames = rowName)
       }
-      
-      if (result[["leveneViolated"]])
-        table$addFootnote(gettext("Levene's test is significant (p < .05), suggesting a violation of the equal variance assumption"), colNames = "p", rowNames = rowName)
       
       table$addRows(row, rowNames = rowName)
     }


boutinb (approved):

JASP-Engine/JASP/R/ttestindependentsamples.R
```diff
@@ -241,8 +241,13 @@ TTestIndependentSamples <- function(jaspResults, dataset = NULL, options, ...) {
         
         if (!isTryError(result))
           row <- c(row, result[["row"]])
-        else
+        else {
           errorMessage <- .extractErrorMessage(result)
+
+          if (result[["leveneViolated"]])
```
TimKDJ, line 247: @vandenman: Shouldn't this go in the `if (!isTryError(result))` block?
Otherwise `result[["LeveneViolated"]]` won't exist anyway

vandenman: you're right, it should actually be in the try.

PR #4076, opened 2020-05-11
Generating markdown help from QML and R

JorisGoosen:
Info now present on all jaspObjects and all jaspControls.
Implements https://github.com/jasp-stats/INTERNAL-jasp/issues/740
But the formatting can still use some work.

All filled info will be collected when the help is opened and will be updated whenever anything changes (aka translation).
For each qml element that has info but not a manually set "toolTip" the info will also be shown as a tooltip directly to give people easy access to the docs.
The info set in R is transmitted through the "meta" field and is coupled with the title of each object and displayed beneath the qml help.
This info could also be shown in between the results but I don't think that needs to be done right now.

Also set some info on Summary Statistics Bayesian Binomial Test to get a feel for how it will look, so to see an example of this: turn on developers mode and open that analysis.


JorisGoosen:
Example markdown:

JorisGoosen:
S.S. Bayesian Binomial Test
=====================
This function computes the Bayes factor for a binomially distributed observation.
The Bayesian binomial test is described in Jeffreys (1961, p. 256).
This test informs us whether the data support or contradict a value suggested for the parameter (chance) in question.

- Null model: *p = p0*
- Alt  model: *p ~ Beta(a,b)*
*p0* is the suggested value of the rate parameter of Binomial under null hypothesis. 
Data observed: *s* successes and *f* failures, total number of trials, *n = s + f*. 
In Theory of Probability, Jeffreys assumes a uniform prior on the rate parameter under alternative hypothesis. 
The Bayes factor used here is a more general case, assuming a beta prior on the rate parameter. 
*Note*: beta(1,1) corresponds to a uniform prior.

---
# Input


### Radio Buttons - *Alt. Hypothesis*:

- *≠ Test value*: Two-sided alternative hypothesis that the population mean is not equal to the test value.
- *> Test value*: One-sided alternative hypothesis that the population mean is larger than the test value.
- *< Test value*: One sided alternative hypothesis that the population mean is smaller than the test value.




### Group Box - *Plots*:



#### CheckBox - *Prior and posterior*:
Displays the prior (dashed line) and posterior (solid line) density of the effect size under the alternative hypothesis; 
			the gray circles represent the height of the prior and the posterior density at effect size delta = 0. 
			The horizontal solid line represents the width of the 95% credible interval of the posterior.
##### CheckBox - *Additional info*:
Displays the Bayes factor computed with the user-defined prior; displays a probability wheel depicting the odds of the data under the null vs. alternative hypothesis; displays the median and 95% credible interval of the posterior density.








### Radio Buttons - *Bayes Factor*:

- *BF₁₀*: Bayes factor to quantify evidence for the alternative hypothesis relative to the null hypothesis.
- *BF₀₁*: Bayes factor to quantify evidence for the null hypothesis relative to the alternative hypothesis.
- *Log(BF₁₀)*: Natural logarithm of BF10.




### Group Box - *Prior*:
Parameters *a* and *b* are set to '1' each. This corresponds to a uniform prior.


---
# Output


### Table - *Bayesian Binomial Test*:
- *Bayes factor*: If one-sided test is requested:
  - BF+0: Bayes factor that quantifies evidence for the one-sided alternative hypothesis that the population mean is larger than the test value
  - BF-0: Bayes factor that quantifies evidence for the one-sided alternative hypothesis that the population mean is smaller than the test value
  - BF0+: Bayes factor that quantifies evidence for the null hypothesis relative to the one-sided alternative hypothesis that the population mean is larger
   than the test value
  - BF0-: Bayes factor that quantifies evidence for the null hypothesis relative to the one-sided alternative hypothesis that that the population mean is
  smaller than the test value
  - **p**: p-value corresponding to t-statistic.

JorisGoosen:
But we can later tweak all of this of course.

JASP-Desktop/analysis/analysisform.cpp
```diff
@@ -267,40 +267,43 @@ void AnalysisForm::_setUpRelatedModels()
...
-			});
+		[](JASPControlWrapper* a, JASPControlWrapper* b) {
+			return a->depends().size() < b->depends().size();
+		});
 
 	for (JASPControlWrapper* control : controls)
 	{
-		_orderedControls.push_back(control);
+		_dependsOrderedCtrls.push_back(control);
+		connect(control->item(), &JASPControlBase::helpMDChanged, this, &AnalysisForm::helpMDChanged);
```
boutinb, line 303: Why do you connect helpMDChanged here? Should the connection be done in depend related order?

JASP-Desktop/analysis/analysisform.cpp
```diff
@@ -821,3 +832,57 @@ bool AnalysisForm::hasVolatileNotes() const
...
+	QStringList markdown =
+	{
+		_analysis->titleQ(), "\n",
+		"=====================\n",
+		_info, "\n\n",
+		"---\n# ", tr("Input"), "\n"
+	};
+
+	QList<JASPControlBase*> orderedControls = JASPControlBase::getChildJASPControls(this);
+
```
boutinb, line 881: getChildJASPControls gives only children JASP controls, not grand-children, etc...

JASP-Desktop/analysis/jaspcontrolbase.h
```diff
@@ -5,34 +5,39 @@
...
+	Q_PROPERTY( QQuickItem		*	section				READ section			WRITE setSection												)
+	Q_PROPERTY( QQuickItem		*	parentListView		READ parentListView									NOTIFY parentListViewChanged		)
+	Q_PROPERTY( QQuickItem		*	innerControl		READ innerControl		WRITE setInnerControl		NOTIFY innerControlChanged			)
+	Q_PROPERTY( QQuickItem		*	background			READ background			WRITE setBackground			NOTIFY backgroundChanged			)
+	Q_PROPERTY( QQmlComponent	*	rowComponent		READ rowComponent		WRITE setRowComponent		NOTIFY rowComponentChanged			)
+	Q_PROPERTY( QQmlComponents		rowComponents		READ rowComponents)
 
 
 public:
+	// Any addition here should also be added manually to ControlTypeToFriendlyString... I couldnt get this to work with DECLARE_ENUM...
```
boutinb, line 40: Would it not be better to have friendName member in JASPControl?

Resources/Summary Statistics/qml/SummaryStatsBinomialTestBayesian.qml
```diff
@@ -50,7 +63,16 @@ Form
...
-			CheckBox { name: "plotPriorAndPosteriorAdditionalInfo"; label: qsTr("Additional info"); checked: true }
+
+			info: qsTr(`Displays the prior (dashed line) and posterior (solid line) density of the effect size under the alternative hypothesis; 
+			the gray circles represent the height of the prior and the posterior density at effect size delta = 0. 
+			The horizontal solid line represents the width of the 95% credible interval of the posterior.`)
+
+			CheckBox 
+			{ 
+				name: "plotPriorAndPosteriorAdditionalInfo"; label: qsTr("Additional info"); checked: true 
+				info: "Displays the Bayes factor computed with the user-defined prior; displays a probability wheel depicting the odds of the data under the null vs. alternative hypothesis; displays the median and 95% credible interval of the posterior density."
```
TimKDJ: `qsTr()`

Resources/Summary Statistics/qml/SummaryStatsBinomialTestBayesian.qml
```diff
@@ -50,7 +63,16 @@ Form
 		CheckBox
 		{
 			name: "plotPriorAndPosterior";		label: qsTr("Prior and posterior")
-			CheckBox { name: "plotPriorAndPosteriorAdditionalInfo"; label: qsTr("Additional info"); checked: true }
+
+			info: qsTr(`Displays the prior (dashed line) and posterior (solid line) density of the effect size under the alternative hypothesis;
```
TimKDJ, line 67: Backticks work for this?

JASP-R-Interface/jaspResults/R/zzzWrappers.R
```diff
@@ -192,7 +192,8 @@ jaspResultsR <- R6Class(
     getOtherObjectsForState = function()        private$jaspObject$getOtherObjectsForState()
 	),
 	active = list(
-	  status = function(x) { if (missing(x)) private$jaspObject$status else private$jaspObject$status <- x }
+	  status = function(x) { if (missing(x)) private$jaspObject$status else private$jaspObject$status <- x },
+	  info   = function(x) { if (missing(x)) private$jaspObject$info   else private$jaspObject$info   <- x }
```
TimKDJ, line 196: Would you ever want to request `info`? For `status` I can see that being useful.

JASP-Desktop/components/JASP/Widgets/BayesFactorType.qml
```diff
@@ -25,7 +25,7 @@ RadioButtonGroup
 {
     title: qsTr("Bayes Factor")                     ; name: "bayesFactorType"
 
-    RadioButton { label: qsTr("BF\u2081\u2080")      ; name: "BF10"; checked: true }
-    RadioButton { label: qsTr("BF\u2080\u2081")      ; name: "BF01"                }
-    RadioButton { label: qsTr("Log(BF\u2081\u2080)") ; name: "LogBF10"             }
+    RadioButton { label: qsTr("BF\u2081\u2080")      ; name: "BF10"; checked: true; info: "Bayes factor to quantify evidence for the alternative hypothesis relative to the null hypothesis." }
+    RadioButton { label: qsTr("BF\u2080\u2081")      ; name: "BF01"               ; info: "Bayes factor to quantify evidence for the null hypothesis relative to the alternative hypothesis." }
+    RadioButton { label: qsTr("Log(BF\u2081\u2080)") ; name: "LogBF10"            ; info: "Natural logarithm of BF10." }
```
boutinb: Why don't you use qsTr for info?

JASP-Desktop/widgets/jaspcontrolwrapper.h
```diff
@@ -53,11 +55,13 @@ class JASPControlWrapper
...
 	static JASPControlWrapper*	buildJASPControlWrapper(JASPControlBase* control);
+
+	virtual QString				friendlyName() const { return ""; }
 	
 protected:
 	
-	JASPControlBase*				_item;
-	QVector<JASPControlWrapper*>	_depends;
+	JASPControlBase	*	_item = nullptr;
+	Set					_depends; //So Joris changed this to a set instead of a vector because that is what it seemed to be, the order isn't important right?
```
boutinb, line 64: OK

JASP-R-Interface/jaspResults/R/zzzWrappers.R
JorisGoosen: Well, maybe not, but I like to reason from possibilities and not from constraints ;)
Because there is nothing bad about allowing introspection into this info right?
Maybe if you are working in R and are like "what should this container I got from this other module be anyway?". 

But mostly, I see nothing better resulting from not allowing it to be requested.

TimKDJ: I'm mainly concerned about having a consistent API for developers.
As an example:
`$addCitation("Some citation")`
`$info <- "Some info"`

It'd be good to be consistent in what we think ought to be active bindings and what ought to be functions.

JASP-Desktop/analysis/analysisform.cpp
JorisGoosen: That is good! 
In fact, it is exactly what I want :)

Each helpMD function of the controlbases collects their own children ad infitum

JASP-Desktop/analysis/analysisform.cpp
JorisGoosen: Because here the items are setup? So I figured this was a good place?
Where exactly would you suggest is better?

 I had to do quite some guessing where to put places and I was also misdirected for a while because of the original name of `_dependsOrderedCtrls` as I thought it was all `ordered ctrls` but they weren't actually ordered as  I expected :p

It seems to work now so I thought this was fine but Im all ear for any suggestion ^^

JASP-Desktop/analysis/jaspcontrolbase.h
JorisGoosen: Well, I was doubting how to call the property but that is what `title` is right? 
We do this on more objects (like Analysis or jaspObjects) where we have a `name` that is non-translated and simple(ish) and a `title` that is shown to the user. So I figured this would be a good name for it here as well.

JASP-Desktop/components/JASP/Widgets/BayesFactorType.qml
JorisGoosen: Because I forgot, Ill add it.

JASP-Desktop/widgets/jaspcontrolwrapper.h
JorisGoosen: Ok good! I put the comment here to make sure you double-checked me because I figured there might be more going on than I though. (I changed the type of `_depends` several times :p)

JASP-R-Interface/jaspResults/R/zzzWrappers.R
JorisGoosen: But you can add multiple citations right? And only one "info".

TimKDJ: `$citations <- c("cite1", "cite2")`
But it was just an example, `setError("error")` does the same...

What is the rule we can apply to differentiate bewteen an active binding and a function? (Maybe @vandenman has an idea?)

JorisGoosen: I would go more for active bindings I think. Because it doesn't really matter what happens on the background to R anyway and I like the syntax of being able to set and get so easily. 
But you can veto that real hard if you like.

vandenman: > What is the rule we can apply to differentiate between an active binding and a function?

We don't have any, but we probably should.

Right now everything that can be set but not read is done with `$addXXX` right? E.g., `$setError`, `$addCitation`. I'm not really seeing any use case for reading the info text if it's should only be set. So without a use case for `x <- obj$info` I'd make this `obj$addInfo("...")`, and I'd also add a check that its argument is a character of length 1 (although maybe you now do that in C++?).

On the other hand, if I'd use `jaspResults` objects in R I'd prefer `obj$info` to print the info and `obj$citations` to print the citations. So in that use case, I'd prefer active methods. Active methods might make it easier to write bad R code though.

Perhaps a middle ground is to only allow `x <- obj$info` in interactive sessions (for module development) using something like `if (jaspResultsCalledFromJASP) stop("that's illegal") else return(x$...$info)`.

JorisGoosen: >On the other hand, if I'd use jaspResults objects in R I'd prefer obj$info to print the info and obj$citations to print the citations. So in that use case, I'd prefer active methods. Active methods might make it easier to write bad R code though.

But why would having active bindings lead to bad code? 
And I think having more active bindings (or properties) is more in line also with how I would work in QML.

> Perhaps a middle ground is to only allow `x <- obj$info` in interactive sessions (for module development)

I guess we could do this but I would vote against it based:
1. It makes the code more complicated just to break something in a specific scenario (and the unittests would still accept these...) 
2. People are already confused when they realize jaspResults isn't even a package in JASP. If it also behaves differently, that would make it quite annoying.

My argument *for* active bindings is: easier debugging because you can check in the code what was put in there. And I would extend this to `setError` and `addCitations` as well.

But maybe you guys can indicate to me what kind of bad code you are afraid these bindings will lead to? Because if that is scary enough ;)

JASP-Desktop/analysis/jaspcontrolbase.h
boutinb: I think you are mistaken: title is not the same as the controlTypeFriendlyName...
What I actually thought is in fact to have controlTypeFriendlyName as a virtual function of JASPControlWrapper, overriden by each sub-class, that returns a string static class member.

JorisGoosen: Like the one I already made on JASPControlWrapper you mean?
I will change it a bit as you said because I can see it work better. But only if im assured there is always a wrapper for a control. Because a fully static one won't work because of `BoundQMLTextInput::friendlyName()`

JorisGoosen: Better like this?

JorisGoosen: https://github.com/jasp-stats/jasp-desktop/pull/4076/commits/5daffd09c5aa003a16ba2043982385a85a3c11ef

boutinb (approved):

JASP-R-Interface/jaspResults/R/zzzWrappers.R
vandenman: I think we were worried about something like:
```r
if (table$citations != "") # means the table code didn't crash but now imagine the comment isn't there.
```
I remember that @TimKDJ once encountered something similar, where the analysis checked if a particular element of the state existed which meant that some other function had crashed or not.

But actually, I agree with your arguments. For citations, we should just make it into an active binding. For errors, this is a bit complicated. `setError` takes a string, and `hasError` returns a bool. If you translate this to an active binding it looks a bit odd. (We could create a custom class for that which on the R side is just a bool with an attribute that contains an error message but it's still odd that the setter takes a string and the getter returns a bool). Perhaps we should make a concrete list of what methods should be active functions. The whole `getColumnName`, `setColumnName`, `setColumnTitle`, `getColumnTitle`, ... extravaganza could just be done with active bindings (and hopefully `colnames(...)` and `colnames(...)<-` wrappers).

Either way, if we decide to completely switch to one of the two, I agree that active bindings makes the most sense. It does require adjusting all analyses though.

JorisGoosen: Yeah I actually implemented support for all thos `colNames` things long long ago in the first release of jaspTable... They just aren't present in th R6 wrapper. And it kind of seemed that the whole `addColumnInfo` was more workable anyway.

The error thing could be an active binding that sets and gets the errormsg. But sets `hasError` to true as well under water. That would make it nicer no?
Ill make an issue for this point and link to this discussion.

JorisGoosen: The example looks like a pretty bad check... But Im sure we would catch most of these with code review. And you know giving people an easy way to check stuff didn't crash or fail. Which the whole `setError` thing is for right?

PR #4082, opened 2020-05-13
Speed up Correlation [Regression] - version 2

### LINKED ISSUE #679: Very slow estimation of correlation matrix
* JASP version: 0.12
* Analysis: Correlation (Regression -> Classical -> Correlation
* Bug description: Very slow estimation of correlation matrix

I have [not-so-big dataset](https://is.muni.cz/www/hynek.cigler/vaha4.csv) (N=1000). If I run a correlations over all the 28 items (vah.a1-vah.b28), the estimation is extremely slow and the analysis fails soon. It is slow even if I run it over just e.g. three or four variables. If the dataset is loaded, these variables are set as ordinal. I tried to set them as continous, but the estimation is slow too. On the other hand, running e.g. factor analysis (which needs also an estimation of covariance matrix) is quick, so the problem is only in the correlation matrix. I tried Pearson and Spearman correlations, both are similarly slow.

### LINKED ISSUE #712: JASP Goes Blank When Computing Large Tables of Correlations
* JASP version: 12.2
* Analysis: Pearson's correlation (for a large set of variables; in this case, 94 variables, 93 of which are the language features generated by the Linguistic Inquiry and Word Count software). 
* Bug description: The whole output window simply goes blank after a while (see attached). It first looks as if it's computing (i.e., the variable names are shown, with blank entries where the _r_-values and _p_-values will appear), but then it all goes blank.  
* Expected behaviour: The whole table of correlations should appear. 
Screenshot (43)

--->

Kucharssim:
Fixes jasp-stats/jasp-issues#679
Fixes jasp-stats/jasp-issues#712

By "Fixes" I mean makes it better. See also jasp-stats/Regression#1 and jasp-stats/INTERNAL-jasp#869

### DIFF
--- a/JASP-Engine/JASP/R/correlation.R
+++ b/JASP-Engine/JASP/R/correlation.R
@@ -224,7 +224,6 @@ Correlation <- function(jaspResults, dataset, options){
   mainTable$transposeWithOvertitle <- FALSE
   
   mainTable$addColumnInfo(name = "var1", title = "", type = "string", combine = FALSE, overtitle = "Variable")
-  mainTable$addColumns(list(var1 = variables))
   
   whichtests <- c(options$pearson, options$spearman, options$kendallsTauB)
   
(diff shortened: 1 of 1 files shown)


TimKDJ:
To truly fix the second issue you listed, we need to prevent correlation from sending back a gigantic table so often.
This basically means not setting up an empty table (through `setExpectedSize()` et al) that will later be filled and not filling it a row at a time, but in one `$setData()` call.
Do you think you have the time to look at that?

Kucharssim:
> Do you think you have the time to look at that?

Thanks, I will have a look tomorrow.

TimKDJ:
Much quicker! :)

PR #4083, opened 2020-05-13
fix default priorPosterior plot in T-test independent samples - version 2

### LINKED ISSUE #730: Issue with Informed Prior for Bayesian Independent Samples T-Test
* JASP version: Version 0.12
* OS name and version: Windows 10 (Version: 10.0.18362 Build 18362)
* Analysis: Bayesian Independent Samples T-Test
* Bug description: 
First, I changed from the default prior to an informed prior. This change went as expected.
Second, I changed back from the informed prior to the default prior. This change did not go as expected. The 'Prior and Posterior' plot stays the same as before (with the informed prior depicted). The analyses, however, change (the BF changes).
* Expected behaviour:
I hoped that the 'Prior and Posterior' plot would adapt every time the prior specification is changed.

#### Steps to reproduce:
1. Go to 'T-Tests > Bayesian Independent Samples T-Test'
2. Click on and look at 'Prior and Posterior' plot
3. Click on 'Prior > Informed' 
4. Set 'Cauchy location: 1 scale: 0.707'
5. Look at 'Prior and Posterior' plot
6. Click on 'Prior > Default'
7. Look at 'Prior and Posterior' plot

Documents attached.
[Preregistration_Rekentuin_TTest.zip](https://github.com/jasp-stats/jasp-issues/files/4580861/Preregistration_Rekentuin_TTest.zip)

Cheers,
Tabea

Kucharssim:
Fixes jasp-stats/jasp-issues#730

see jasp-stats/TTests#4


vandenman:
@Kucharssim It looks like the summary stats t-tests don't like the changes you made.

Kucharssim:
Hm, so summary stats fail because the tests contained the bug that I now fixed (well, at least that's my interpretation). This begs a question whether I should add some unit tests for this case in the idependent samples (that is, not the summary stats one). I guess that would be nice?

Kucharssim:
> This begs a question whether I should add some unit tests for this case in the idependent samples

Although, the summary stats tests already test this, that's why it failed now. Then perhaps I just need to remember to run summary stats tests (or just `testAll()`) to catch these next time.

Kucharssim:
@vandenman it's ready now

vandenman (approved):

PR #4084, opened 2020-05-14
Show folder in window title

JorisGoosen:
Window Title is now a property of DataSetPackage
- It also shows the folder where the file is stored
- Implements https://github.com/jasp-stats/INTERNAL-jasp/issues/730
- Also moves "current file" to DataSetPackage
- Added a small extra field to FileEvent to store browse path of OSF
-- This works pretty good except for when an OSF file is loaded from recent files, because the info isnt available then. But oh well, it then just says "(OSF)" as folder as fallback.
- Made sure OSF labels for filename and dirname aren't too small
- Some fixes and changes to make title look nice on Windows
- Also made sure that whenever a file is loaded from data library they do not get a folder behind them, because we do not want to make people scared with things like "Program Files" or "/usr/..."
- And finally, when a csv needs to be generated JASP will check if the directory it loaded the file from exists and if so open a save dialog positioned there. Unless it is the Data Library, then it will show documents.

### DIFF
--- a/JASP-Desktop/data/datasetpackage.h
+++ b/JASP-Desktop/data/datasetpackage.h
@@ -47,8 +47,7 @@ class DataSetPackage : public QAbstractItemModel //Not QAbstractTableModel becau
 	Q_PROPERTY(QString		windowTitle				READ windowTitle									NOTIFY windowTitleChanged			)
 	Q_PROPERTY(bool			modified				READ isModified				WRITE setModified		NOTIFY isModifiedChanged			)
 	Q_PROPERTY(bool			loaded					READ isLoaded				WRITE setLoaded			NOTIFY loadedChanged				)
-	Q_PROPERTY(QFileInfo	currentFile				READ currentFile									NOTIFY currentFileChanged			)
-	Q_PROPERTY(QString		currentPath				READ currentPath			WRITE setCurrentPath	NOTIFY currentPathChanged			)
+	Q_PROPERTY(QString		currentFile				READ currentFile			WRITE setCurrentFile	NOTIFY currentFileChanged			)
 
 	typedef std::map<std::string, std::map<int, std::string>> emptyValsType;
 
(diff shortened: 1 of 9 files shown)


JASP-Desktop/data/datasetpackage.h
```diff
@@ -39,7 +41,14 @@ class EngineSync;
...
 	Q_OBJECT
-	Q_PROPERTY(int columnsFilteredCount READ columnsFilteredCount		NOTIFY columnsFilteredCountChanged)
+	Q_PROPERTY(int			columnsFilteredCount	READ columnsFilteredCount							NOTIFY columnsFilteredCountChanged	)
+	Q_PROPERTY(QString		name					READ name											NOTIFY nameChanged					)
+	Q_PROPERTY(QString		folder					READ folder					WRITE setFolder			NOTIFY folderChanged				)
+	Q_PROPERTY(QString		windowTitle				READ windowTitle									NOTIFY windowTitleChanged			)
+	Q_PROPERTY(bool			modified				READ isModified				WRITE setModified		NOTIFY isModifiedChanged			)
+	Q_PROPERTY(bool			loaded					READ isLoaded				WRITE setLoaded			NOTIFY loadedChanged				)
+	Q_PROPERTY(QFileInfo	currentFile				READ currentFile									NOTIFY currentFileChanged			)
+	Q_PROPERTY(QString		currentPath				READ currentPath			WRITE setCurrentPath	NOTIFY currentPathChanged			)
```
boutinb: I'm not sure what is the difference between folder and currentPath, and whether it is necessary to have 2 properties for this. Or could you give a better name?

JorisGoosen: currentPath is the path to where the file is stored. I guess it could be merged back into curentFile. But at some point I thought I needed a QUrl for OSF files (currentPath) and a QFileInfo for normal files (currentFile).

Ill merge them!

JorisGoosen: Done

boutinb (approved):

PR #4086, opened 2020-05-15
RM (B)ANOVA: fix names in long format data set

### LINKED ISSUE #737: bug on Bayesian Repeated Measures ANOVA
* JASP version: 0.12.2
* OS name and version: Windows 10 Enterprise
* Analysis: Bayesian Repeated Measures ANOVA
* Bug description: When I enter three columns as levels for a repeated measures factor, I get the following:

image

* Expected behaviour: I expected it to run a Bayesian Repeated Measures ANOVA :D

#### Steps to reproduce:
1. Open attached data file (csv). [jasp_bug_report_data.zip](https://github.com/jasp-stats/jasp-issues/files/4601966/jasp_bug_report_data.zip)
2. Go to 'ANOVA', 'Bayesian Repeated Measures ANOVA'
3. Move dependent variables (level1, level2, level3) to the Repeated Measures Cells window
3. See error

vandenman:
Fixes https://github.com/jasp-stats/jasp-issues/issues/737

Both the repeated measures ANOVAs created a data set in long format with columns `subject` and `dependent`. Both analyses weren't very careful with encoding or decoding the column names and also encoded these names from time to time. The problem: if there's a column in the dataset called `subject` then the new encoding would happily replace this with `JaspColumn.xxx.Encoded`. However, the data set in long format does not have that as a column name. That's the summary of 737.

In addition, bad things happen when you added a covariate called `subject` or `dependent` to the analyses. The `.shortToLong` had the names `subject` and `dependent` hard coded and R "fixes" any duplicate names to become e.g., `subject.1`.

A summary of the fix:

- Made two constants, `.BANOVAdependentName <- "JaspColumn_.dependent._Encoded"` and `.BANOVAsubjectName <- "JaspColumn_.subject._Encoded"`. These particular names can be changed of course.

- adjust `.shortToLong` so that the names for `subject` and `dependent` can be specified through arguments. Rather than `dependent` and `subject` these names are now set to the previously introduced constants in the ANOVAs. Remark: why oh why does `.shortToLong` exists and do we not use a package for this?

- Replaced every instance of `subject` by `.BANOVAsubjectName`, same for `dependent` and `.BANOVAdependentName`. These new names should make it less likely that there are name clashes with the names in the data set.

- Remove any encoding done on these two variables. Note that if the data set contains a variable called `subject`, then `.v("JaspColumn_.subject._Encoded")` returns `"JaspColumn_.JaspColumn.0.Encoded._Encoded"`, i.e., the subject part is decoded. Removing this encoding is necessary to avoid partial encoding of the new subject and dependent names.

@JorisGoosen Are the names `JaspColumn_.dependent._Encoded` and `JaspColumn_.subject._Encoded` ok? These could be customized in any way, I just want to ensure they won't interfere with any column encoding.

### DIFF
--- a/JASP-Engine/JASP/R/anovarepeatedmeasures.R
+++ b/JASP-Engine/JASP/R/anovarepeatedmeasures.R
@@ -228,16 +228,16 @@ AnovaRepeatedMeasures <- function(jaspResults, dataset = NULL, options) {
   
   main <- paste("(",paste(unlist(terms.base64), collapse=" + "),")", sep="")
   termsBS <- paste("(",paste(termsBS.base64, collapse=" + "),")", sep="")
-  errorRM <- paste("Error(",paste("subject/(", termsRM.base64, ")",sep="", collapse=" + "),")",sep="")
+  errorRM <- paste("Error(",paste0(.BANOVAsubjectName, "/(", termsRM.base64, ")", collapse=" + "),")",sep="")
   
   if (is.null(termsBS.base64) && is.null(termsRM.base64)) {
     model.def <- dependent ~ 1
   } else if (is.null(termsBS.base64)) {
-    model.def <- paste(.v("dependent"), "~", paste(main, errorRM, sep=" + "))
+    model.def <- paste(.BANOVAdependentName, "~", paste(main, errorRM, sep=" + "))
   } else if (is.null(termsRM.base64)) {
-    model.def <- paste(.v("dependent"), "~", main)
+    model.def <- paste(.BANOVAdependentName, "~", main)
   } else {
-    model.def <- paste(.v("dependent"), "~", paste(main, errorRM, termsBS, sep=" + "))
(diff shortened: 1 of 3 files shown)


JorisGoosen:
@vandenman those names are fine, I see you already noticed that the decoding thing happening if subject is available in the dataset. It would probably be better if you could add `subject` and `dependent` to the extra encoding. 

On the other hand, if this works it is fine like this too. If the code now does the encoding and decoding logically things should work out fine.

JohnnyDoorn:
@vandenman  Can/should this be reviewed now?

vandenman:
@JohnnyDoorn yes please :)

JohnnyDoorn (approved):

TimKDJ:
I guess this needs to be added to https://github.com/jasp-stats/ANOVA as well?

vandenman:
@TimKDJ Yes, it should be added there too

PR #4087, opened 2020-05-19
Gettextf fixes

FransMeerhoff:
Cleanup gettextf - Phase 1 : Remove unicode characters as parameters


JASP-Engine/JASP/po/R-nl.po
```diff
@@ -754,17 +754,19 @@ msgstr ""
...
+msgstr ""
+"De puntschatting van de ware populatiewaarde <i>W&#770 = N × w&#772</i>"
 
 msgid "The uncertainty of the estimator %1$s%2$s%3$s"
 msgstr "De onzekerheid van de schatter %1$s%2$s%3$s"
 
-msgid "The confidence interval around the point estimate %1$s"
-msgstr "Het betrouwbaarheidsinterval rond de puntschatting %1$s"
+msgid "The confidence interval around the point estimate <i>W&#770 ± U</i>"
+msgstr "Het betrouwbaarheidsinterval rond de puntschatting <i>W&#770 ± U</i>"
```
TimKDJ: Are these correct?

FransMeerhoff: The confidence interval around the point estimate Ŵ ± U
Het betrouwbaarheidsinterval rond de puntschatting Ŵ ± U

The point estimate of the true population value Ŵ = N × w̄
De puntschatting van de ware populatiewaarde Ŵ = N × w̄
 
I do see the italic in the real result:

Screen Shot 2020-05-20 at 09 45 57

So I think it is correct

TimKDJ: Might be good to change the html characters to unicode too
`W&#770` is `\U0174` for example

JASP-Engine/JASP/R/anovarepeatedmeasures.R
```diff
@@ -1391,7 +1391,7 @@ AnovaRepeatedMeasures <- function(jaspResults, dataset = NULL, options) {
     marginalMeansTable$addFootnote(symbol = "\u2020", message = gettext("Bias corrected accelerated."))
     
   } else {
-    thisOverTitle <- gettextf("95%% CI for Mean Difference")
+    thisOverTitle <- gettext("95% CI for Mean Difference")
```
TimKDJ: Double `%` are still a necessity as well as the `gettextf()`

FransMeerhoff: Yes, I just did that yesterday. I was believing in a miracle i think because in 
gettext("95% bca\u002A CI") anovarepeatedmeasure.R 1387 it works fine. I though that once you found that part of parsing that went wrong. Because this is very annoying too I think.

TimKDJ: See my explanation here: https://github.com/jasp-stats/INTERNAL-jasp/issues/777#issuecomment-597740225
It will only work when the translation and original text share the same "percent parameters". So it's just luck of the draw if it works.

PR #4088, opened 2020-05-19
Add combineWithOtherModels attribute to source property

boutinb:
The mixedModelTerms property is specific for Available VariablesList. But in case of the ContrastList of the Repeated Anova where an assigned VariablesList is used, such a feature is also needed. 
This property affects in fact the terms given by the source of a VariablesList. So this property is removed and the combineWithOtherModels source attribute is added, making it available for all kind if VariablesList.

### DIFF
--- a/JASP-Desktop/components/JASP/Controls/AvailableVariablesList.qml
+++ b/JASP-Desktop/components/JASP/Controls/AvailableVariablesList.qml
@@ -22,6 +22,4 @@ VariablesList
 {
 	listViewType:	JASP.AvailableVariables
 	showSortMenu:	true
-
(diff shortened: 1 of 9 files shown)


JorisGoosen:
Is there some issue related to this or some reason you are doing this?

JorisGoosen:
So I can see in the code changes that this is for instance for Mixed Models and that you generalized some functionality in ContrastsList so you no longer need dedicated code for the mixedmodels-model. Because you now allow VariablesLists to combine several available vars lists?

Anyway, that is my interpretation, but really, it would be nice if you typed some more of a description than just a title and *No description provided.* 

For the rest: this PR looks fine!

PR #4090, opened 2020-05-21
Kill! Kill! Kill!

JorisGoosen:
BuschemiTheKiller

If the engine takes too long to pause just kill it instead of crashing the entire GUI

- implements https://github.com/jasp-stats/INTERNAL-jasp/issues/875
- Small tweak to also kill the engine if it doesnt respond to a stopRequest
- Implemented a "abort analysis if no response" on option changes
- Make sure engines pause gracefully and do not kill the desktop for each little unexpected reply.
-- Some of them are unexpected but not that bad, so it is nice to be able to differentiate those errors from others (new exception added)
- Added an exception to enumutilities to be able to catch the "missing enum val problem" specifically (aka entered a string that isnt an enumval)
- Resume engine now also sends the settings, this avoids another send->reply cycle afterwards
- Split the filter pause-resume from running the filter so that a process loop can be made in between and let the desktop possibly respond to something
- Engine now also reads in columnnames on startup, because maybe it just started after being killed (or just plain crashing ofc)
- We might be on a killing spree but it would be nice if the desktop also responds well to an engine that *does* respond properly to option changes.
-- For this Ive added a KeepStatus Analysis::Status that tells the Analysis it should, well duh, keep it's current status instead of changing it to something new.
- Taken some code for ~MainWindow from stablish branch that destroys all the loaded qml in reverse order
-- this should avoid a bunch of qml warnings/errors were the qml was dependent on other objects that are already deleted by MainWindow

### DIFF
--- a/JASP-Common/enginedefinitions.cpp
+++ b/JASP-Common/enginedefinitions.cpp
@@ -1,2 +1,10 @@
 #define ENUM_DECLARATION_CPP
 #include "enginedefinitions.h"
+
+const char * unexpectedEngineReply::what() const noexcept{ return std::runtime_error::what(); }
+
+void unexpectedEngineReply::checkIfExpected(engineState expectedReplyState, engineState currentState, int channelNo)
+{
+	if(expectedReplyState != currentState)
+		throw unexpectedEngineReply(expectedReplyState, channelNo, ", in state: " + engineStateToString(currentState));
+}
--- a/JASP-Common/enginedefinitions.h
+++ b/JASP-Common/enginedefinitions.h
@@ -3,10 +3,22 @@
 
 #include "enumutilities.h"
 
-DECLARE_ENUM(engineState,			initializing, idle, analysis, filter, rCode, computeColumn, moduleRequest, pauseRequested, paused, resuming, stopRequested, stopped, logCfg, settings);
+DECLARE_ENUM(engineState,			initializing, idle, analysis, filter, rCode, computeColumn, moduleRequest, pauseRequested, paused, resuming, stopRequested, stopped, logCfg, settings, killed);
 DECLARE_ENUM(performType,			init, run, abort, saveImg, editImg, rewriteImgs);
 DECLARE_ENUM(analysisResultStatus,	validationError, fatalError, imageSaved, imageEdited, imagesRewritten, complete, inited, running, changed, waiting);
 DECLARE_ENUM(moduleStatus,			initializing, installNeeded, installModPkgNeeded, loadingNeeded, unloadingNeeded, readyForUse, error);
 DECLARE_ENUM(engineAnalysisStatus,	empty, toInit, initing, inited, toRun, running, changed, complete, error, exception, aborted, stopped, saveImg, editImg, rewriteImgs, synchingData);
+
+struct unexpectedEngineReply  : public std::runtime_error
+{
+	unexpectedEngineReply(std::string msg)															: std::runtime_error(msg) {}
+	unexpectedEngineReply(engineState unexpectedState, std::string msg = "Engine got unexpected")	: std::runtime_error(msg + ": " + engineStateToString(unexpectedState)) {}
+	unexpectedEngineReply(engineState unexpectedState, int engineNo, std::string extra = "")		: std::runtime_error("Engine " + std::to_string(engineNo) + " got unexpected reply: " + engineStateToString(unexpectedState) + extra) {}
+	const char* what() const noexcept override; //Put that
+
+	static void checkIfExpected(engineState expectedReplyState, engineState currentState, int channelNo);
+};
+
+
 
 #endif // ENGINEDEFINITIONS_H
--- a/JASP-Common/enumutilities.h
+++ b/JASP-Common/enumutilities.h
@@ -17,6 +17,12 @@
 #include <QString>
 #endif
 
+struct missingEnumVal  : public std::runtime_error
+{
+	missingEnumVal(std::string enumName, std::string missingValue)	: std::runtime_error("Enum " + enumName + " does not contain value \"" + missingValue + "\"!") {}
+};
+
+
 #define STRING_REMOVE_CHAR(str, ch) str.erase(std::remove(str.begin(), str.end(), ch), str.end())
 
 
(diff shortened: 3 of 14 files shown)


JorisGoosen:
So, Ive tested this quite a bit on several systems and I think it is stable now. It is barely noticable for me that the engine was restarted dozens of time for a single run.

I think it would be quite useful if besides the code review by @boutinb this code is tested out actually. By someone else than me, because this could break things. And the running stuff is sort of the corething that JASP does so it would be nice if it keeps working.

@vandenman maybe you can test this out with some slow JAGS model? And @TimKDJ maybe you can test this as well? It could be nice to combine that with the MixedModels by @FBartos but then without that button and the `runOnChange: false` bits.

vandenman:
A slow JAGS model for the debug data set:
```
model{
  theta ~ dbeta(1, 1)
  for (i in 1:100) {
    contNormal[i] ~ dnorm(theta, 1)
    contGamma[i] ~ dnorm(1, theta)
  }
  qqq <- sum(contNormal) * theta
}
```
Under `Advanced` set `No. samples` to 20000 (or more). Under `Parameters in model`, quickly add or remove a parameter to see the effect. 

> It is barely noticeable for me that the engine was restarted dozens of times for a single run.


I tried to time the massacre by counting the seconds manually. If the analysis is running, the spinning wheel disappears almost immediately after I change an option. I took that as a sign that Joris murdered the engine. About 6 seconds later the spinner reappears and the engine appears reincarnated. I think that's a bit on the long side but at the same time, it's a very acceptable delay. I didn't experience any crashes despite pressing buttons like a crazy person.

I'm not sure why Travis failed though.

vandenman (approved):

JorisGoosen:
> I tried to time the massacre by counting the seconds manually. If the analysis is running, the spinning wheel disappears almost immediately after I change an option. I took that as a sign that Joris murdered the engine. About 6 seconds later the spinner reappears and the engine appears reincarnated. I think that's a bit on the long side but at the same time, it's a very acceptable delay. I didn't experience any crashes despite pressing buttons like a crazy person.

Aah yes, on my system it seems to start a bit quicker, but not much of a difference I guess. Im not sure if the spinner disappearing is related to the status change or to the engine being killed. It coming back is definitely a sign that the engine is back on. Of course if you are running release there should be multiple engines anyway, so it might be faster if it just switches engine. (Which I guess it does do?)

And the delay would only be there for analyses that get stuck in internal loops for more then 1 or 2 seconds, so previously the delay would've been much worse :p

> I'm not sure why Travis failed though.

It seemed it couldn't download some macos pkg, maybe it is temporary? Or do we specify exactly what kind of mac we want to run on and did they drop support for it maybe?

TimKDJ:
I can't say I notice the engine killing if I'm being honest. I tested it with the Bayesian Log Linear Regression on a release build. 

I added two factors ("facFive" and "facGender") and set samples to 100k, then I set the number of samples to 100, however the engine doesn't seem to update to the latest options, because 100 samples should take only a few seconds to compute.

[does-not-finish.zip](https://github.com/jasp-stats/jasp-desktop/files/4680878/does-not-finish.zip) (rename to .jasp)

JorisGoosen:
> I can't say I notice the engine killing if I'm being honest. I tested it with the Bayesian Log Linear Regression on a release build.
> 
> I added two factors ("facFive" and "facGender") and set samples to 100k, then I set the number of samples to 100, however the engine doesn't seem to update to the latest options, because 100 samples should take only a few seconds to compute.
> 
> [does-not-finish.zip](https://github.com/jasp-stats/jasp-desktop/files/4680878/does-not-finish.zip) (rename to .jasp)

Ok, so when I load your file it seems that the results are indeed stuck in "running" mode. However, if I refresh the analysis I get results. Then I up the samples to 10e8 an it starts running again, clears results, I switch back to 100 and it stops, runs fast, shows results. And if I repeat it, it uhm shows results again.

I will build a nightly and test that again on mac, to see if it works differently, but I doubt it. Maybe you are doing something else differently?

JorisGoosen:
Aah yes, I get the same problem on mac with the nightly...

TimKDJ:
> Aah yes, I get the same problem on mac with the nightly...

I did my testing on Windows, so it isn't platform specific (in case you'll try to find the solution there).

JorisGoosen:
Yeah, I think I know what is going on. Im building a testversion now. But I think an engine should also be counted as "paused" when it is "initializng" because if they take a bit longer to startup they are killed before they can do anything :p

And because I set the killtime on 2 seconds and the engine startup can be slower than that, so if you open your jasp file quickly enough... Then it gets stuck, I think, Im building a new nightly now to check ^^

TimKDJ (approved):

boutinb (approved):
I could not see mistakes in the code.
Just 2 remarks: some settings are hard-coded like 2 seconds KILLTIME or 10 seconds for waiting for the filtering. There are surely such internal settings you don't really want to be displayed to the user, but maybe settings that could be set  in a hidden way, by starting JASP with a special argument for example.... Not sure if it is a good idea...
Also could be convenient to have such hidden setting to prevent killing engines in such pittyless way.
What do you think?

JorisGoosen:
I guess we could add those settings but I didn't want to complicate things.
Maybe we can do this some other time? If these settings turn out to not be perfect for everyone?

boutinb (approved):

FBartos:
I've been testing the kill feature in Mixed Models and it's working great, apart from one side effect.

I fit a model that can take a long time to fit and then I want to create a plot using the fitted model (Note that the plot can also take 5-20sec to be generated). I drag one variable to the plotting box, the plot starts being created. I add another variable(s) to the plotting box, the engine is killed -> the whole model is lost. This is gonna be even more problematic for the Robust Bayesian Meta-Analysis module, where fitting a model takes time in tens of minutes and some plotting functions generate multiple figures (which also take a lot of time).

Could it be maybe possible to specify which qml elements will have the authority to kill the engine if updated? For example, I would set it only for the model fitting elements, while the plots related elements wouldn't be able to kill the engine.

TimKDJ:
I think the only thing we could improve for this scenario is to try to include the state in the new engine if it was stored successfully before the previous engine was killed.

PR #4093, opened 2020-05-25
Add TabView control

boutinb:
This control reuse the BoundQMLComponentList
ComponentList, TabView & InputListView should be better integrated
together. This will be done after 0.13 release. For the time being I
have just added new functionality so that this does not interfere with
existing
controls.

### DIFF
--- a/JASP-Desktop/widgets/boundqmlcomponentslist.cpp
+++ b/JASP-Desktop/widgets/boundqmlcomponentslist.cpp
@@ -34,9 +34,10 @@ BoundQMLComponentsList::BoundQMLComponentsList(JASPControlBase *item)
 	setTermsAreNotVariables();
 	readModelProperty();
 
+	QQuickItem::connect(_item, SIGNAL(nameChanged(int, QString)), this, SLOT(nameChangedHandler(int, QString)));
 	QQuickItem::connect(_item, SIGNAL(valuesChanged()), this, SLOT(valuesChangedHandler()));
-	QQuickItem::connect(_item, SIGNAL(addRow()), this, SLOT(addRowHandler()));
-	QQuickItem::connect(_item, SIGNAL(removeRow(int)), this, SLOT(removeRowHandler(int)));
+	QQuickItem::connect(_item, SIGNAL(addItem()), this, SLOT(addItemHandler()));
+	QQuickItem::connect(_item, SIGNAL(removeItem(int)), this, SLOT(removeItemHandler(int)));
 }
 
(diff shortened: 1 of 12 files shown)


JASP-Desktop/widgets/boundqmlcomponentslist.cpp
```diff
@@ -133,27 +134,46 @@ Option* BoundQMLComponentsList::createOption()
...
-				QVariant value = it.value();
 
-				Option* option = row->get(name.toStdString());
-				if (option)		option->set(value.toJsonValue().toString().toStdString());
-				else			Log::log() << "Default value with name " << name << " is unknown in the component" << std::endl;
+				if (name != keyName)
+				{
+					QVariant value = it.value();
+					Option* option = row->get(name.toStdString());
+					if (option)		option->set(value.toJsonValue().toString().toStdString());
```
JorisGoosen, line 174: `tq()` is shorter you know.

JASP-Desktop/widgets/boundqmlcomponentslist.cpp
```diff
@@ -133,27 +134,46 @@ Option* BoundQMLComponentsList::createOption()
...
+					else			Log::log() << "Default value with name " << name << " is unknown in the component" << std::endl;
```
JorisGoosen: I wouldn't know what this errormsg means... Not important enough to block this PR but, maybe it could be clarified a bit?

JASP-Desktop/widgets/boundqmlcomponentslist.cpp
```diff
@@ -228,27 +248,99 @@ void BoundQMLComponentsList::valuesChangedHandler()
...
+}
+
+QString BoundQMLComponentsList::_makeUnique(const QString &val, const QList<QString> &values, int index)
+{
+	QString result = val;
+
+	bool isUnique = true;
+	do
+	{
+		int i = 0;
```
JorisGoosen, line 329:
```
do
{
	isUnique = true;
	for(int i=0; i<values.size(); i++)
		if( i != index && values[i] == result )
		{
			isUnique = false;
			result = _changeLastNumber(result);
		}
}
```

Is a bit shorter and cleaner don't you think?

JorisGoosen:
So some points, overall looks fine I guess.

You can decide if you want to make the changes or not, as far as I am concerned this is ready to be squash&merged ;)

JorisGoosen (approved):
As mentioned already, it looks fine but I think some things can be cleanner.

PR #4096, opened 2020-05-25
Fixed Frantisek's problem

JorisGoosen:
I do not see why we need to bind implicit to preferred, it caused binding loops, in unpredictable situations.

Why did you put it there @boutinb?
Without it @FBartos 's module no longer has the weird height glitch and if I look at JaspControl, JaspGridControl etc I don't see the need for this binding trick.


boutinb (approved):
There is no need to bind the implicitHeight with the preferredHeight.
I remember that the goal of this code was to prevent a binding loop, but apparently after some other changes, this code is not necessary anymore, and even worse, it generates a binding loop...

PR #4098, opened 2020-05-26
Tjur R-squared calculation

### LINKED ISSUE #764: Tjur R^2 wrongly calculated?
* JASP version: 0.12.2
* Analysis: during logistic regression Tjur's R^2 is shown in the model summary
* Bug description: value seemed unrealistic, very low. I calculated with R and got a much more reasonable value
* Expected behaviour: see below

#### Steps to reproduce:
If you import the attached data into JASP and perform a logistic regression the Tjur R^2 is evaluated as 0.081, compared to a value of 0.5318 which I figured in R. The higher value is much more realistic, given 1. the scatterplot and 2. the other pseudo measures.

>>> In case attachment didn't work: Here's the data (some made up treatment data for logistic regression):

"","Dosis","Wirkung"
"1",0.59,1
"2",0.01,0
"3",0.29,0
"4",0.28,0
"5",0.81,1
"6",0.26,1
"7",0.72,1
"8",0.91,1
"9",0.95,1
"10",0.07,0
"11",0.75,1
"12",0.29,0
"13",0.1,0
"14",0.95,1
"15",0.42,1
"16",0.46,0
"17",0.97,1
"18",0.58,0
"19",0.96,1
"20",0.76,1
"21",0.71,1
"22",1,1
"23",0.51,0
"24",0.49,1
"25",0.65,1
"26",0.83,1
"27",0.48,0
"28",0.84,1
"29",0.51,0
"30",0.53,0

Kucharssim:
Fixes https://github.com/jasp-stats/jasp-issues/issues/764


vandenman (approved):
The changes look fine to me. 

Maybe we should use some data set where the result is known for testing? Because now we update these values based on what the user mentioned, but we're still not testing for correctness as you nicely pointed out elsewhere.

Kucharssim:
True, I wanted to take the dataset from the original paper from Tjur, but gave up after a while searching for where is the dataset actually stored. If you have an idea for an example, it would be great (afaik the textbooks we have in the library do not report Tjur).

I could also just use the data set that the user kindly shared with us, as I cross checked the new JASP results with `performance` package, but granted that will again just become a “result taken from somewhere”, and easily rewritten if it breaks again.

Kucharssim:
Ah, I think I got it: https://github.com/lbraglia/aplore3/blob/master/rawdata/LOWBWT/LOWBWT.txt

Well then let me make the unit tests! :)

JASP-Tests/R/tests/testthat/test-regressionlogistic.R
```diff
@@ -299,4 +299,44 @@ test_that("Error Handling", {
...
\ No newline at end of file
+})
+
+test_that("Pseudo R-squared are correct", {
+  # Specifically, tests McFadden, Nagelkerke, Tjur, and Cox & Snell coefficients
+  # Example from Tjur (2008) http://dx.doi.org/10.1198/tast.2009.08210, page 370
+  # Tjur's results for Tjur R2: 0.096231
+  #.
+  # Check against code from performance package
+  # library(aplore3)
```
vandenman: Great! A minor comment, could you add the package versions as a comment? If they ever change and mess up then we can retrace the package versions where it did work.

Kucharssim:
I'll do a separate PR to the Regression module once this is merged (i.e., once it's certain that I don't need to make additional changes)

PR #4100, opened 2020-05-27
jaspResults now makes a "Write Seal" that tells it last run wasnt killed before writing

JorisGoosen:
- Implements https://github.com/jasp-stats/INTERNAL-jasp/issues/884
- This removes the need to refresh an analysis after an abort because either nothing was changed on disk (because it was killed before writing the new state/plot/jaspresults.json) or it worked
- And if it was somehow killed during writing the write-seal would be missing so it knows to start afresh
- Also figured we do not support old school analyses anymore so engine will now crash if you do try that


JorisGoosen:
This should fix the problems @FBartos was having with the model, I tested it with his mixed models analysis and it seemed to behave now.

TimKDJ (approved):
Seems to work like a charm. I also tested it with a `Sys.sleep()` in common.R to target an option change in the "write window" and it refreshed correctly.

Edit: I did notice that when I made multiple copies of an analysis that was running, the new analyses did not run and instead appeared to use the state. However, I could not replicate this behaviour in a new instance, but I figured I'd mention it anyway:
copies-dont-run

JorisGoosen:
> Seems to work like a charm. I also tested it with a `Sys.sleep()` in common.R to target an option change in the "write window" and it refreshed correctly.

Aah nice, I didn't think of testing it like that, good to hear it actually works!

>
> Edit: I did notice that when I made multiple copies of an analysis that was running, the new analyses did not run and instead appeared to use the state. However, I could not replicate this behaviour in a new instance, but I figured I'd mention it anyway:

Well, a copy of an analysis simply takes the results that were already there. 
Although I guess I would've expected a duplicated analysis to also run if its parent-state was running. But that seems a problem for a different issue.

JorisGoosen:
Also, I asked @FBartos to test it with Mixed Models and he said it basically works fine unless he mashing his keyboard like crazy, and then it refits the model. But that could be reasonable.

TimKDJ:
> Well, a copy of an analysis simply takes the results that were already there.
> Although I guess I would've expected a duplicated analysis to also run if its parent-state was running. But that seems a problem for a different issue.

I changed the number of samples in the first analysis, which then started to run, then I duplicated it. These duplicates should run also, because otherwise the QML options are not correct with the output you're seeing (output belongs to the fewer samples run of the first analysis). But like I said when I tried to replicate this, I couldn't and the duplicates did refresh. So I hope this can be ignored ;)

JorisGoosen:
Well, I guess if you copy it right at the moment it is killed/aborted it might copy that status of the analysis. And then it won't run because the engineRepresentation makes sure that it restarts/reruns the analysis again after killing the engine.

I guess we could make it so that when an analysis is duplicated it copies everything. But then if the analysis status is one of: `Aborting, Aborted, ValidationError, FatalError` we set it to `Empty`?
Then it would rerun.

JASP-R-Interface/jaspResults/src/jaspResults.cpp
```diff
@@ -67,19 +79,24 @@ jaspResults::jaspResults(std::string title, Rcpp::RObject oldState)
...
+		if(_writeSealRoot + _writeSealRelative == "")
+			throw std::runtime_error("Write seal location not given and we are running in JASP, this should never happen!");
 	}
 	else
 		_RStorageEnv = new Rcpp::Environment(Rcpp::as<Rcpp::Environment>(Rcpp::Environment::namespace_env("jaspResults")[".plotStateStorage"]));
 
-	if(!oldState.isNULL() && Rcpp::is<Rcpp::List>(oldState))
+	bool imNotReincarnatedAfterBeingMurdered = lastWriteWorked();
+
+	if(imNotReincarnatedAfterBeingMurdered && !oldState.isNULL() && Rcpp::is<Rcpp::List>(oldState))
```
boutinb, line 91: if lastWriteWorked is false and there is an old state file, should this file not be deleted?

boutinb (approved):
So if i understand well, when an analysis is run, in the engine occurs the following steps:
. Create JaspResult: it checks whether the writeSeal exists, if yes it reads the state file (if exists)
. Run the analysis
. Remove the writeSeal file
. Save the state file
. Write a new writeSeal file

I have to say I cannot see a scenario when it goed wrong...
The only remark is it seems to me neater to delete the state when the writeSeal does not exist.

JASP-R-Interface/jaspResults/src/jaspResults.cpp
JorisGoosen: It will happen automatically because it won't be in the "KeepList" and the tempfiles delete everything not in it every time anyway.
And because the check happens inside jaspResults and the state is still loaded through older code from `common.R` I figured just doing it like this would be fine.

JorisGoosen:
So you understand well but miss some point, the engine does the following things during the run of an analysis:

- Create jaspResults: it checks whether the writeSeal exists, if yes it reads from loaded state (from `common.R .retrieveState` and loads the previous run's `jaspResults.json` file if it exists
 - Run the analysis
 - Remove the writeSeal file
 - Save the state file and write the `jaspResults.json`
 - Write a new writeSeal file

PR #4101, opened 2020-05-28
Adds a FileSelector qml element

JorisGoosen:
- implements https://github.com/jasp-stats/jasp-issues/issues/731
- Also tweaks engine crash handler to not crash on changes in R code
- Adds an example and test to TestModule


FBartos:
Just tested the load version and it works great.

How does the save version works? I would like to save a list as .RDS. How do I connect it to a piece of r-code so it's called when the button selects a path?

JorisGoosen:
Well in QML you do:
```qml
FileSelector
{
    name: "rdsLocationSave"
    filter: "*.rds"
    save: true
    //label: "Text in front"
    //caption: "Text in titlebar dialog"
}
```

And then in R you just get `options$rdsLocationSave` and you have to manage saving loading yourself there.

boutinb (approved):

PR #4106, opened 2020-05-30
Save figures as *.pptx

vandenman:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/888

[Diff without whitespace changes](https://github.com/jasp-stats/jasp-desktop/pull/4106/files?w=1)

New packages:

| Package       | Version       |
| ------------- | ------------- |
| uuid          | 0.1-4         |
| rvg           | 0.2.4         |
| officer       | 0.3.11        |

all from CRAN.

To use, click on a plot, select `Save Image As`. A window opens. Under `Files of type`, select `PowerPoint (*.pptx)`:

image

This works well for both `ggplot` and `qgraph` objects, but less well for `JASPgraphsPlot`. There the rastering fails and the `.pptx` file just contains an image, which is okay I guess.

I only tested this on Linux but the package authors say it works everywhere and doesn't require any dependencies. Still, I think we should test this on multiple systems to be safe.

@AlexanderLyNL can you test MacOS? 

@TimKDJ can you test Windows?

### DIFF
--- a/JASP-Desktop/mainwindow.cpp
+++ b/JASP-Desktop/mainwindow.cpp
@@ -713,7 +713,7 @@ void MainWindow::_analysisSaveImageHandler(Analysis* analysis, QString options)
 	parser.parse(utf8, root);
 
 	QString selectedFilter;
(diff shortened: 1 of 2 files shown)


EJWagenmakers:
Awesome.

AlexanderLyNL:
@FransMeerhoff could you please update jasp-required-files/macOS and add the three packages mentioned by don here https://github.com/jasp-stats/jasp-desktop/pull/4106#issue-425431379 ?

TimKDJ (approved):
Looks good to me. We just need the required packages in jasp-required-files

PR #4109, opened 2020-06-02
Experimental R access from JASP

JorisGoosen:
- Implements https://github.com/jasp-stats/INTERNAL-jasp/issues/886
- Adds some knowledge to EngineRepresentation, aka can it run an analyis or not
- Looks gorgeous

afbeelding

And to test it: 
- [Win64 msi](https://static.jasp-stats.org/Nightlies/Windows/JASP-nightly-R-b11631488a1fa38664797a75bcbfc71630574426.msi)
- [Win64 zip](https://static.jasp-stats.org/Nightlies/WindowsZip/JASP-nightly-R-b11631488a1fa38664797a75bcbfc71630574426.zip)
- [Mac](http://static.jasp-stats.org/Nightlies/MacOS/JASP-nightly-R-b11631488a1fa38664797a75bcbfc71630574426.dmg)


TimKDJ:
It is impressive how quick you managed to implement this!

Some thoughts with respect to the design:
- [x] I think it would be nicer if this was a ribbon bar item and we can treat it as a sort of module for barebone statistical analysis.
- [ ] If it moves to the ribbon bar I think it makes sense to have it behave more like a module. This would mean using the input <-> output design we use for other modules. I think this would also make it integrate better and give users some additional space for input. However, I can imagine that this might be too much work?

Some thoughts with respect to functionality:
- [x] Can we rename `jaspData` to `data` (or `dataset`) and `jaspFiltered` to `filteredData`? Given that we're inside JASP and all (`jaspFiltered` also does not really convey that it is a data object).
- [x] It feels strange that you can manually delete parts of the output, but this might be a feature? If we're going to ship the R input/output with the .jasp file then I don't think we should do this, because it'd be confusing if what you see isn't necessarily what was output.
- [x] Maybe disallow running when the input is empty as there is nothing to do anyway.
- [x] It feels odd that the columns are available as separate variables. If the point is to introduce users to R, then going `data$V1` makes more sense. Also, if you name a column `mean` then you can no longer use the function `mean()`.

Some (pretty minor) buggy behaviour:
- [ ] If you plot and try to close the plotting window the jaspEngine has to be forced to shut down. Maybe we can have a look at preventing graphics devices from being opened.
- [x] The output seems to jump to the top and down again if you run something when there's already output.
- [x] The scrollwheel doesn't work to scroll in the output window, I can only scroll by dragging the scrollbar.

vandenman:
This is probably a better spot than https://github.com/jasp-stats/INTERNAL-jasp/issues/886

@JorisGoosen Looks good! Some issues below that may or may not need some adjustments.

1. It sort of breaks when you call `x11()`. That opens a new window to plot in, which is good, but pressing the close button on that window causes it to freeze / not respond. Windows will then suggest killing JASP IIRC. Running `dev.off()` will close it but I doubt that many people who'll use this know that. In addition, for me the default graphics device was `pdf`, so nothing happens when I do e.g., `plot(rnorm(10))`.

2. Text output is not shown immediately. For example `install.packages("DstarM", repos = "https://cloud.r-project.org")` will install this package + dependencies (just look at your R folder in the build directory) but all information about this is only shown once the function finishes. In contrast, R studio would show a progress bar for each package that is being downloaded.

3. Warnings are not shown. For example, `testit <- function() warning("testit"); testit()` gives no output but it does in an R terminal. However, `stop("is shown")`. Is `stderr()` setup correctly? It looks a bit different from `stdout()`:

image

4. if `install.packages` works, like the example above, then you can change the results of analyses in JASP by installing newer packages. This can cause a fresh installation of JASP to give different answers for analyses than an installation where someone called install.packages. I suggest *not* fixing this, but rather adding a big warning message

5. Interactive commands cause the interface to become non-responsive. For example:
```r
foo <- function() {
  while(TRUE) {
    num <- readline("How many number do you want to enter? > ") # expects user input
    num <- as.numeric(num)
    if (!is.na(num)) {
      num2 <- readline(paste0("Enter ",num, " numbers > ")) # expects user input
      print(num2)
      break
    }
  }
}
foo()
```
this works in Rstudio, but the response from `readline` is never read in JASP so you basically wait forever. There is no way to exit this once it happens without restarting.

JorisGoosen:
> It is impressive how quick you managed to implement this!

:)

> Some thoughts with respect to the design:
> 
>     * [ ]  I think it would be nicer if this was a ribbon bar item and we can treat it as a sort of module for barebone statistical analysis.

Making a ribbonbar entry could be nice to, it would certainly look better. @EJWagenmakers also mentioned something about adding it there. I have no preference on the whole.

>     * [ ]  If it moves to the ribbon bar I think it makes sense to have it behave more like a module. This would mean using the input <-> output design we use for other modules. I think this would also make it integrate better and give users some additional space for input. However, I can imagine that this might be too much work?

Yes totally, although I in that case I would want it to be fleshed out a bit more. What I would like in the future is that there is a jaspResults object available somehow (maybe invisibly or so) and have proper output and all that. In that case maybe the R prompt could also be integrated in the actual interface. Maybe in the same place as the analyses or something? I dont know though. And this is definitely something for another time.


> Some thoughts with respect to functionality:
> 
>     * [ ]  Can we rename `jaspData` to `data` (or `dataset`) and `jaspFiltered` to `filteredData`? Given that we're inside JASP and all (`jaspFiltered` also does not really convey that it is a data object).

Yeah sure, it's just that there seemed to be a `utils::data` that that clashed with, but I guess it doesn't matter if nobody can get there. I'll change them to `data` and `filteredData`.

>     * [ ]  It feels strange that you can manually delete parts of the output, but this might be a feature? If we're going to ship the R input/output with the .jasp file then I don't think we should do this, because it'd be confusing if what you see isn't necessarily what was output.

I think it probably isn't a good idea to include the input output in R to be honest. As in, maybe at some point but then there would need to be some R file support, syntax highlighting and having the R code disabled if you are not the author. But I agree that editing the output isn't very practical. I'll make it readonly.

>     * [ ]  Maybe disallow running when the input is empty as there is nothing to do anyway.

Good one

>     * [ ]  It feels odd that the columns are available as separate variables. If the point is to introduce users to R, then going `data$V1` makes more sense. Also, if you name a column `mean` then you can no longer use the function `mean()`.

I was doubting if I should `attach` the data also because are you getting `data` or `filteredData` ?
I'll take it out.

> 
> 
> Some (pretty minor) buggy behaviour:
> 
>     * [ ]  If you plot and try to close the plotting window the jaspEngine has to be forced to shut down. Maybe we can have a look at preventing graphics devices from being opened.

Sure, but easily avoidable as well.
 
>     * [ ]  The output seems to jump to the top and down again if you run something when there's already output.

Yeah I can fix that.

>     * [ ]  The scrollwheel doesn't work to scroll in the output window, I can only scroll by dragging the scrollbar.

Oops! Let's see what went wrong there. But can you scroll if your mouse is above the code (literally)?

EJWagenmakers:
Awesome work.
1. "Welcome to JASP's builtin R!" -> Welcome to R-in-JASP! (open to alternatives; "builtin" seemed redundant and I don't think it's written this way).
2. "Welcome to JASP's builtin R!" -> the "W" is slightly cut off for me:
image
3. Should we indicate the R version number somewhere? 
4.  "Enter you R code here" -> youR, not you
5.  Suppose I enter this:
x <- c(1,23)
mean(x)
bla

then I get an error (because "bla" is silly), without getting "mean(x)" first. In contrast, pasting the code in the regular R console gives this:
> x<-c(1,2,3,4)
> mean(x)
[1] 2.5
> bla
Error: object 'bla' not found

This is different behavior, and I like the one from R best (also, generally good to keep the experience similar)

6. Yes, plotting opens a window with the plot, but closing it demands that the engine is shut down.

E.J.

EJWagenmakers:
1. OK, I think it should be "Welcome to R in JASP!" (without the hyphens)
2. the font size is smaller than the font size in regular JASP. So I have to go control+ to increase the font size, and going back to regular JASP then confronts me with font sizes that are too large.
3. Should the input box be at the bottom? It is then easily cut off. (screenshot). Anybody with an opinion on this?
image

4. The font type is different from that used in JASP, right? Do we want that?
5. If I press "run code" the result is presented on top of the output panel, but then it is removed very quickly and presented again at the bottom of the output panel.

vandenman:
A possible icon for the ribbon could be the R icon itself: https://www.r-project.org/logo/

JorisGoosen:
> 1. OK, I think it should be "Welcome to R in JASP!" (without the hyphens)

I'll leave the hyphens because otherwise the QML parser will complain ;)

> 2. the font size is smaller than the font size in regular JASP. So I have to go control+ to increase the font size, and going back to regular JASP then confronts me with font sizes that are too large.

Well, technically it **is** the same fontsize but because it is a different font this looks smaller I guess? Ill increase it a bit.
 
> 3. Should the input box be at the bottom? It is then easily cut off. (screenshot). Anybody with an opinion on this?

If you just put your window in the screen completely the box isn't cut-off right? I don't think we should be designing our GUI's based on "possibly being invisible due to being pushed out of screenspace".
Of course normally you would be working in a terminal or something (well, I would normally be doing that) and then the prompt is topside initially but that is a bit hard to emulate in QML.

> 4. The font type is different from that used in JASP, right? Do we want that?

It is the same font already used in the model entry of lavaan, jags etc.

> 5. If I press "run code" the result is presented on top of the output panel, but then it is removed very quickly and presented again at the bottom of the output panel.

The same issue as Tim reported, I was already contemplating a fix for it but was wondering if it would be annoying for users as-is/was. It was! So I fixed it.

JorisGoosen:
> Awesome work.

Thank you!

> 2. "Welcome to JASP's builtin R!" -> the "W" is slightly cut off for me:

Should be fixed now.

> 3. Should we indicate the R version number somewhere?

We could I guess, it is also not very hard to request in R right? Would it be useful?

>     4. "Enter you R code here" -> youR, not you

Ahum, I think my brain had a little glitch there...

> 5. Suppose I enter this:
>        x <- c(1,23)
>        mean(x)
>        bla
> 
> 
> then I get an error (because "bla" is silly), without getting "mean(x)" first. In contrast, pasting the code in the regular R console gives this:
> 
> > x<-c(1,2,3,4)
> > mean(x)
> > [1] 2.5
> > bla
> > Error: object 'bla' not found
> 
> This is different behavior, and I like the one from R best (also, generally good to keep the experience similar)

This is caused by the fact that I already paste the code you enter to the output log before it is sent to the engine. And the engine then sends the entire resulting log of output back. If somebody knows a way to make R "echo" each statement it runs then this could be fixed but I can't find a way to do so. And I agree that it would be nice to make our prompt behave as actual R as much as possible but there are quite some restrictions here and this is not such a big problem right?

> 1. Yes, plotting opens a window with the plot, but closing it demands that the engine is shut down.

Yeah, some jaspPlot support and catching any devices being opened by the user would be nice. But we can do that later right?

JorisGoosen:
So I think most of the things that were mentioned are done now.

What I didn't do is:
- R logo in ribbonbar that opens window (that is something for tomorrow)
- make [builtin](https://en.wikipedia.org/wiki/Built-in) R library readonly or catch calls to installed.packages(). (But see: https://github.com/jasp-stats/INTERNAL-jasp/issues/896 )
- plotting related stuff
- interactive stuff

JorisGoosen:
Aah and there should be new nightlies for testing!

JASP-Desktop/components/JASP/Widgets/RCommanderWindow.qml
```diff
@@ -0,0 +1,294 @@
...
+import QtQuick.Window	2.12
+import QtQuick.Controls 2.12
+import JASP				1.0
+import JASP.Widgets		1.0 as JW
+import JASP.Controls	1.0 as JC
+
+Window
+{
+    id:					rcmdRoot
+	title:				"R in JASP"
```
boutinb: qsTr for the title

JASP-Desktop/engine/enginerepresentation.h
```diff
@@ -22,6 +22,9 @@
 class EngineRepresentation : public QObject
 {
 	Q_OBJECT
+	Q_PROPERTY(bool		runsAnalysis		READ runsAnalysis		WRITE setRunsAnalysis		NOTIFY runsAnalysisChanged		)
+	Q_PROPERTY(bool		runsUtility			READ runsUtility		WRITE setRunsUtility		NOTIFY runsUtilityChanged		)
+	Q_PROPERTY(bool		runsRCmd			READ runsRCmd			WRITE setRunsRCmd			NOTIFY runsRCmdChanged			)
```
boutinb, line 28: I don't understand why you use Q_PROPERTY here. I don't see where in the code the signal of their changes is used.

JASP-Desktop/mainwindow.cpp
```diff
@@ -474,6 +486,12 @@ void MainWindow::loadQML()
 	_upgrader->loadOldSchoolUpgrades();
 }
 
+void MainWindow::showRCommander()
+{
+	Log::log() << "Loading RCmdWindow"  << std::endl;
+	_qml->load(QUrl("qrc:///components/JASP/Widgets/RCommanderWindow.qml"));
+}
+
```
boutinb, line 502: Why should we need several R windows?

JASP-Desktop/engine/enginerepresentation.h
```diff
@@ -137,6 +153,9 @@ public slots:
 	std::string		_lastCompColName	= "???";
 
 
+	bool _runsAnalysis;
+	bool _runsUtility;
+	bool _runsRCmd;
```
boutinb: The state of the engine is quite difficult to catch, between the _engineState, _pauseRequested, _stopRequested, _slaveCrashed, and now these 3 new booleans.
Can't the engineState be extended?

JASP-Desktop/components/JASP/Widgets/RCommanderWindow.qml
JorisGoosen: Oops!

JASP-Desktop/engine/enginerepresentation.h
JorisGoosen: I mostly used used the code generation feature of it. And what is the harm?

JASP-Desktop/mainwindow.cpp
JorisGoosen: Why not?

JASP-Desktop/engine/enginerepresentation.h
JorisGoosen: No the state of the engine is always stored in `_engineState`, that hasn't changed. But it describes what `JASP-Engine` is actually doing at the moment.

`_pauseRequested` and `_stopRequested` can not be *in* the state because there the engine is actually in another state (possibly anyway). Next process-cycle these are checked and then once that is done it sends a `pauseRequested` or `stopRequested` msg and sets *that* in the `_engineState`. Because then the engine is doing that and once the reply comes the state is set to `paused` or `stopped`. This is a bit complicated but the result of not having a msgbuffer. 

These 3 new booleans are not the state at all by the way, they simply define what this engine is used for. You remember the loop where `engine #0` is used for some things and the other engines for other things? I have simply made explicit what they are allowed to do. Because this way you can add or remove engines as desired. This will also be used in running engines more dynamically (for instance per module).

I agree that the "state of the engine" is difficult to catch but, it needs to do many things... A good solution could be to implement a msgbuffer over http or something. Then a lot of things here can be simplified. But that would be a big big change.And then those `runsAnalysis` or `runsRCmd` booleans are still useful.

boutinb: These boolean should at least have a default value.
I never see in the code setRunsCmd(true), or setRunsUtility(true)...

JorisGoosen: Yes they should! I forgot to add that...

JorisGoosen: Oof, I dropped a bunch of balls there

JASP-Desktop/mainwindow.cpp
TimKDJ: My $0.02:
I wanted to retrieve the R popup after it went behind some window and did it by clicking on the menu item again. Which promptly opened a new popup window -- behaviour I was not expecting. Additionally, during all my use of RStudio I've only ever had two instances of an R session open if I was working in different RStudio projects, which we do not support. So I'd be in favour of a single instance.

vandenman: I agree, until JASP I've never had a reason for multiple R windows.

JorisGoosen: Ok!
But what should the behaviour of the R-ribbonbutton be then?
Click it first time -> Open window
Click it second time? Show window? Close window?

TimKDJ: Clicking it a second time should make it the active window

JorisGoosen:
Ok, so I think all commentary has now been processed.

I also added a small thing that is easily disabled/modified: a loading/running indication in the outputwindow showing the user that R is running and that they have to be patient. It is a bit large if you open it fullscreen but I like having the feedback. It is the same loader as OSF has for connecting but then slightly transparent.

JorisGoosen:
Aah and adding a SplitView is a bit annoying so I would like to postpone that, untill we go to Qt >5.13 so that I can use a newer (and better) one from QtQuick.Controls 2

boutinb (approved):

AlexanderLyNL:
Can we by the way add a big fat disclaimer stating that 
- the R console is in beta, 
- it shouldn't be used by inexperience users,
- most likely break jasp
- as with jasp comes with no guarantee

Example disclaimer in chrome:

chrome://flags/

WARNING: EXPERIMENTAL FEATURE AHEAD! The R console is in beta, and could lead JASP to break, or you could lose data or compromise your analyses.

EJWagenmakers:
The words "Beta" should be underneath the R icon in the ribbon, right?
We could add some text following "Enter your R code here", or elsewhere. Maybe "[NB. Experimental version]"

TimKDJ:
> I also added a small thing that is easily disabled/modified: a loading/running indication in the outputwindow showing the user that R is running and that they have to be patient. It is a bit large if you open it fullscreen but I like having the feedback. It is the same loader as OSF has for connecting but then slightly transparent.

Maybe we can use the normal analysis-is-running-spinner (bottom left of the output window or in the `Run Code` button)? I agree that it's nice to know when it's running, but I feel like this is a little much :P

R-commander

And something very minor: can `Run Code` be the top button? As run is the action you'll be using most often.

JorisGoosen:
> Can we by the way add a big fat disclaimer stating that
> 
>     * the R console is in beta,
> 
>     * it shouldn't be used by inexperience users,
> 
>     * most likely break jasp
> 
>     * as with jasp comes with no guarantee
> 
> 
> Example disclaimer in chrome:
> 
> chrome://flags/
> 
> WARNING: EXPERIMENTAL FEATURE AHEAD! The R console is in beta, and could lead JASP to break, or you could lose data or compromise your analyses.

Sure!
Should this be in the welcome text?
I think the risk of compromising your analyses or data is rather low to be honest. What could you do? It runs in a separate engine and you cannot change the data from R.

Only thinkg you could mess up is your library and that is what this issue is for: https://github.com/jasp-stats/INTERNAL-jasp/issues/896

Having a good disclaimer couldn't hurt though. Although it doesn't necessarily need to be part of this PR. Maybe we can come up with something nice?

JorisGoosen:
And haha I will see if I can get the running icon in there, but I was happy with the big moving thing ^^

And Ill also change the subtitle of the ribbonbutton to just "Beta" then? @EJWagenmakers ?
Although I thought it would be a bit weird to have that as the title.
Edit: Ah no I will have to keep it like "R (Beta)" because otherwise the modulemenu where you enable it will be rather confusing. (It will just say "Beta" there without an icon...)

EJWagenmakers:
Yes,  "R (Beta)" with a capital B

AlexanderLyNL:
Is there by the way a possibility to kill the R process?

JorisGoosen:
Killing the window ought to kill the r-process

JorisGoosen:
Ah, but Im guessing you want to keep the window open?

PR #4110, opened 2020-06-03
[Audit] Add sample indicator and histograms to selection analysis

koenderks:
This PR implements some additions to the Audit module, and in specific to the selection analysis.
Concretely:
- There is now a computed column that adds the result of the selection analysis to the data set
- I have added plots that compare the distribution of values in the population to those in the resulting sample.

**I haven't tested the unit tests (I downloaded R4.0 something), but we'll see from Travis**

@FransMeerhoff Would you be able to review this pull request and https://github.com/jasp-stats/Audit/pull/4?


Resources/Audit/qml/auditSelection.qml
```diff
@@ -256,9 +256,38 @@ Form {
...
+					name: 			"sampleIndicatorColumn"
+					text: 			qsTr("Column name: ")
+					fieldWidth: 120
+					visible:    addSampleIndicator.checked
+			}
+	}
+
 	Item
 	{
 		Layout.preferredHeight: 			downloadReportSelection.height
```
FransMeerhoff, line 293: You need to update the jasp_nl.po for these changes also. (I can do this for you or tell you how if you don't know)

Resources/Help/analyses/auditselection_nl.md
```diff
@@ -11,6 +11,9 @@ Standaard invoeropties
 #### Steekproefgrootte
 Het benodigde aantal steekproefelementen dat geselecteerd wordt uit de populatie.
 
+#### Voeg selectie uitkomst toe aan data
```
FransMeerhoff: onduidelijkheid counter/result in english en uitkomst in Ned. voor mij

FransMeerhoff (commented):
Some translations are missing (in the result of the selection.jasp, and in the qml)

Resources/Audit/qml/auditSelection.qml
koenderks: Yes, would be nice if you could tell me so that I know for future changes!

PR #4111, opened 2020-06-03
Changing language gives wrong values in Dropdown

boutinb:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/891


JorisGoosen:
This solution looks like it will work. But it would be nicer to have a bona fide model in C++ that is coupled to the combobox. But we can do that some other time.

PR #4116, opened 2020-06-05
Initial some pointers to null.

boutinb:
This might help the crash that Frantic encounters with Mixed Models.

I saw in the call stack that it uses a _boundTo pointer which was not initialized


FBartos:
It seems to fix the crashes caused by closing one analysis and opening another 👍

PR #4117, opened 2020-06-06
[Audit] Add separate evaluation analyses

koenderks:
This PR implements two new analyses in the audit module:

- Evaluation
- Bayesian evaluation

They are both essentially an extracted stage from the audit workflow.

These are a lot of R code changes to the common file. I mainly tried to have the stages share as many functionalities as possible, so that's why. 

I am probably missing some (translation?) stuff still here, I will get to work on creating some help files.
**Edit:** Implemented the help files and some fixes in the qml

This PR accompanies https://github.com/jasp-stats/Audit/pull/5


JASP-Engine/JASP/R/auditCommonFunctionsBayesian.R
```diff
@@ -1620,6 +1615,15 @@
 
     }
 
+    priorAndPosteriorStatisticsTable$addFootnote(message = gettextf("%1$s: The population misstatement is lower than materiality (%2$s %3$s). %4$s: The population misstatement is equal to, or higher than, materiality (%5$s %6$s).",
```
FransMeerhoff, line 1618: Should now become:
priorAndPosteriorStatisticsTable$addFootnote(message = gettextf("H\u208B: The population misstatement is lower than materiality (\u03B8 < %1$s). H\u208A: The population misstatement is equal to, or higher than, materiality (\u03B8 \u2265 %2$s).",
                                                            round(evaluationState[["materiality"]], 3),
                                                            round(evaluationState[["materiality"]], 3)))

So no more literals, even unicode, as parameters in gettexf.
Sorry dit is voortschrijdend inzicht en mede door andere unicode settings.

JASP-Engine/JASP/R/auditCommonFunctions.R
```diff
@@ -3241,7 +3358,11 @@
                                   title = gettextf("BF%1$s", "\u208B\u208A"),
```
FransMeerhoff, line 3358: Now use:  gettext("BF\u208B\u208A")
The Unicode characters will now appear in the .po file. Makes it easier to translate and moreover possible to change these characters in other languages (Parameters are not shown in .po)

JASP-Engine/JASP/R/auditCommonFunctionsBayesian.R
koenderks: This shows up in JASP as :

image

It doesn't seem to work

boutinb:
@koenderks Must this PR be in 0.13? It's a lot of changes in the Audit analyses, but we want now to stabilise the stable branch.

koenderks:
Preferably yes. I believe we can get any remaining bugs (if any) out during testing?

PR #4119, opened 2020-06-08
Add footnote for paired t-tests

### LINKED ISSUE #758: Feature request, Paired Samples T-Test: Clarify correspondence between variables and measures
* Enhancement: The interface should state if a variable is *Measure 1* or *Measure 2*.
* Purpose: Simplify choice of *Alt. Hypothesis* for one-tailed paired t-tests, and its interpretation.

It is visually not 100% clear what variable in *Variable pairs* corresponds to *Measure 1* and *Measure 2*.

1. The interface should make unambiguously clear if a variable is considered *Measure 1* or *Measure 2*.

For example, the *Variable pairs* could indicate which variable will become *Measure 1* and *Measure 2*, like this:

jasp_paired_proposal

2. The *Note.* sentence in the results should use the same wording as in the interface: The formulation should not talk about *measure one* and *measure two* but instead of *Measure 1* and *Measure 2* again, or even better, dynamically use the variable names themselves.

**Additional context**  
Screenshot from version 0.12.2:

jasp_paired_example

In this example, there is no unambiguous visual indication if variable *Pre* is indeed *Measure 1* or rather *Measure 2*.

vandenman:
Fixes https://github.com/jasp-stats/jasp-issues/issues/758

For both paired t-tests, adds a footnote if the alternative hypothesis is directional. See screenshot below.

image

Only gives an example for the first complete pair (if there is any).

### DIFF
--- a/JASP-Engine/JASP/R/ttestpairedsamples.R
+++ b/JASP-Engine/JASP/R/ttestpairedsamples.R
@@ -473,9 +473,9 @@ TTestPairedSamples <- function(jaspResults, dataset = NULL, options, ...) {
     pair2 <- pairs[[idx]][[2L]]
     if (onePair) { # one pair, only give example
       ans <- if (isLess) {
-        gettextf("The alternative hypothesis specifies that %1s is less than %2s.", pair1, pair2)
+        gettextf("For all tests, the alternative hypothesis specifies that %1s is less than %2s.", pair1, pair2)
       } else {
-        gettextf("The alternative hypothesis specifies that %1s is greater than %2s.", pair1, pair2)
+        gettextf("For all tests, the alternative hypothesis specifies that %1s is greater than %2s.", pair1, pair2)
       }
     } else { # multiple pairs, general + example
       ans <- if (isLess) {
--- a/JASP-Engine/JASP/R/ttestpairedsamples.R
+++ b/JASP-Engine/JASP/R/ttestpairedsamples.R
@@ -457,15 +457,35 @@ TTestPairedSamples <- function(jaspResults, dataset = NULL, options, ...) {
 }
 
 .ttestPairedGetHypothesisFootnote <- function(hypothesis, pairs) {
-  directionNote <- if (hypothesis == "groupTwoGreater") gettext("less") else gettext("greater")
+
   idx <- .ttestPairedGetIndexOfFirstNonEmptyPair(pairs)
-  return(
-    if (idx != 0L)
-      gettextf("For all tests, the alternative hypothesis specifies that Measure 1 is %1s than Measure 2. For example, %2s is %3s than %4s.",
-               directionNote, pairs[[idx]][[1L]], directionNote, pairs[[idx]][[2L]])
-    else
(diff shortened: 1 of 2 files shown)


JASP-Engine/JASP/R/ttestpairedsamples.R
```diff
@@ -459,4 +456,21 @@ TTestPairedSamples <- function(jaspResults, dataset = NULL, options, ...) {
   return(p)
 }
 
+.ttestPairedGetHypothesisFootnote <- function(hypothesis, pairs) {
+  directionNote <- if (hypothesis == "groupTwoGreater") gettext("less") else gettext("greater")
```
TimKDJ: It's better not to translate `"less"` and `" greater"`  separately; it may well be possible that these words are translated differently as standalone words than when they are translated within the context of this sentence. This is not something we've done from the start, but it's something we should start doing.

So basically complete sentences > sentence with word substitutions, see also https://www.gnu.org/software/gettext/manual/gettext.html#Entire-sentences

(@FransMeerhoff you probably will want to update your translation guide and maybe make a post about the nuisances related to translations)

TimKDJ (commented):
So a massive nitpick and I'd also be fine with the current implementation, but IMO it'd be better if there was a distinction between one pair and multiple pairs

- For one single pair:
`Note. For all tests, the alternative hypothesis specifies that contNormal is less than contGamma.`
- For multiple pairs:
`Note. For all tests, the alternative hypothesis specifies that Measure 1 is less than Measure 2. For example, contNormal is less than contGamma.`

vandenman:
Good feedback @TimKDJ! I changed the first on to:
`Note. The alternative hypothesis specifies that contNormal is less than contGamma.` without the `For all tests,`.

TimKDJ:
> Good feedback @TimKDJ! I changed the first on to:
> `Note. The alternative hypothesis specifies that contNormal is less than contGamma.` without the `For all tests,`.

I think the `for all tests` refers to "Student" and "Wilcoxon signed-rank" though, not separate pairs

vandenman:
> I think the for all tests refers to "Student" and "Wilcoxon signed-rank" though, not separate pairs

But you can only do one of those? As in, you can't do a Student and Wilcoxon signed-rank at the same time. Regardless, if you think it's better I can also add it back in.

TimKDJ:
Not entirely sure what you mean.. You can tick both tests, right?

Schermafbeelding 2020-06-11 om 16 17 07

I interpreted `tests` as referring to the column `Test` meaning the following two footnotes are equal: 
1. `Note. For all tests, the alternative hypothesis specifies that contNormal is less than contGamma.`
1. `Note. For both Student and Wilcoxon, the alternative hypothesis specifies that contNormal is less than contGamma.`

And if that is the case then it should be present (or absent, it doesn't seem like a very necessary piece of information) in the case of a single pair and multiple pairs.

vandenman:
you're right. I was looking at the Bayesian one, where the tests are radio button rather than a checkbox. I'll change it back to `For all tests,` then.

image

TimKDJ (approved):

TimKDJ:
Ah, that explains the confusion :)

PR #4121, opened 2020-06-09
Give a bit more time to check whether the data file exist or not

boutinb:
In https://github.com/jasp-stats/jasp-issues/issues/786, a
synchronisation problem was again encountered.
I have suppressed the check of whether the data file exists during the
loading of a JASP file, and give up to 1 second to check whether the
file exists and is not empty when the watcher signals a change in the data file.

### DIFF
--- a/JASP-Desktop/widgets/filemenu/filemenu.cpp
+++ b/JASP-Desktop/widgets/filemenu/filemenu.cpp
@@ -291,11 +291,11 @@ void FileMenu::dataSetIOCompleted(FileEvent *event)
 	}
 }
 
-void FileMenu::syncDataFile(const QString& path)
+void FileMenu::syncDataFile(const QString& path, bool waitForExistence)
 {
 	bool autoSync = Settings::value(Settings::DATA_AUTO_SYNCHRONIZATION).toBool();
 	if (autoSync)
-		setSyncRequest(path);
+		setSyncRequest(path, waitForExistence);
 }
 
 void FileMenu::refresh()
@@ -308,7 +308,7 @@ void FileMenu::refresh()
 void FileMenu::dataFileModifiedHandler(QString path)
 {
 	_mainWindow->setCheckAutomaticSync(false);
-	syncDataFile(path);
+	syncDataFile(path, true);
 }
 
 void FileMenu::dataSetIORequestHandler(FileEvent *event)
@@ -392,12 +392,12 @@ void FileMenu::showAboutRequest()
 	emit showAbout();
 }
 
-void FileMenu::setSyncRequest(const QString& path)
+void FileMenu::setSyncRequest(const QString& path, bool waitForExistence)
 {
 	if (path.isEmpty())
 		return;
 
-	if (checkSyncFileExists(path, true))
+	if (checkSyncFileExists(path, waitForExistence))
 	{
 		FileEvent *event = new FileEvent(this, FileEvent::FileSyncData);
 		event->setPath(path);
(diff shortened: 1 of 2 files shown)


JASP-Desktop/widgets/filemenu/filemenu.cpp
```diff
@@ -411,39 +411,33 @@ bool FileMenu::checkSyncFileExists(const QString &path, bool waitForExistence)
...
 		return true;
 
-	bool exist = false;
-	int maxExistTrials = waitForExistence ? 5 : 1;
-	int existTrials = 1;
+	auto checkExists = [](const QString& path) { return QFileInfo::exists(path); };
+	auto checkNotEmpty = [](const QString& path) { return Utils::getFileSize(path.toStdString()) > 0; };
 
-	while (!QFileInfo::exists(path) && existTrials < maxExistTrials)
+	auto checkFn = [&](bool (*f)(const QString&))
```
JorisGoosen: :)

JASP-Desktop/widgets/filemenu/filemenu.cpp
```diff
@@ -411,39 +411,33 @@ bool FileMenu::checkSyncFileExists(const QString &path, bool waitForExistence)
...
-		existTrials++;
-		Utils::sleep(100);
-	}
-
-	if (QFileInfo::exists(path))
-	{
-		int trials = 0;
-		while (Utils::getFileSize(path.toStdString()) == 0 && trials < 5)
+		bool check = f(path);
+		if (!check && waitForExistence)
```
JorisGoosen: Checking for `!check` is superfluous right? 
The condition is always checked in the for loop before executing the body: https://en.cppreference.com/w/cpp/language/for
And you condition already includes `!check`.

I would write it as:
```C++
if(!waitForExistence) return f(path);

for (int i = 0; i < 10; i++)
{
	if(f(path))
       return true;
   	Utils::sleep(100);
}
return false;
```

JorisGoosen (changes requested):

JASP-Desktop/widgets/filemenu/filemenu.cpp
JorisGoosen: I don't really like that programming pattern where stuff is put in booleans or whatever and updated in random loops and you have to read an entire function and understand what is going on. This function is short but you really don't need `bool check` at all.

I think this is much cleaner and easier to understand.

JorisGoosen: (Cleaner except for the messed up tabs but that is because Im typing here in github comment field...)

boutinb: During my study I was told to use only one return at the end of a function. But I see your point. Maybe I'm a bit old-fashioned.
I was happy with my lambda function as parameter of a lambda function! Apparently it does not work if the parameter lambda function captures external variables.

JASP-Desktop/widgets/filemenu/filemenu.cpp
JorisGoosen: Ooh yeah, but to make this also c++ instead of C:
`auto checkFn = [&](std::function<bool>(const QString &))`
Should also work right?

JASP-Desktop/widgets/filemenu/filemenu.cpp
JorisGoosen: Well, that makes a lot of sense in C, in C++ less.
And yeah now that you mention that, I have suggested a small extra change to that using the `std::function`. Maybe you can try that?

And what exactly doesn't work with the lambda as a parameter?

boutinb: Compile error. But maybe std::function works beter indeed

JorisGoosen: So does this code compile or not? Because it looks nice and clear!

PR #4127, opened 2020-06-12
When changing language, dropdown controls give sometimes wrong values

boutinb:
Fixes jasp-stats/INTERNAL-jasp#906

ComboBox object used to keep its values with some map properties. The
values should be stored in its model. So a specialized model
(ListModelLabelValueTerms) is added that keeps the value-label pairs.


JASP-Desktop/widgets/listmodellabelvalueterms.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+ListModelLabelValueTerms::ListModelLabelValueTerms(QMLListView* listView)
+	: ListModelTermsAvailable(listView)
+{
+	readModelProperty(listView);
+}
+
+QVariant ListModelLabelValueTerms::data(const QModelIndex &index, int role) const
+{
+	int row = index.row();
```
JorisGoosen, line 30: You should probably check for `< 0` in case bogus indices are sent. Right now this would lead to an error

JASP-Desktop/widgets/listmodellabelvalueterms.cpp
```diff
@@ -0,0 +1,148 @@
...
+			if (srcModel == this)
+				return;
+
+			QMap<QString, int> roleMap;
+			QHash<int, QByteArray> roles = srcModel->roleNames();
+			QHashIterator<int, QByteArray> i(roles);
+			while (i.hasNext())
+			{
+				i.next();
+				QString valueStr = QString::fromStdString(i.value().toStdString());
```
JorisGoosen: toStdString -> fromStdString is a bit ridiculous don't you think?

JorisGoosen (changes requested):

PR #4130, opened 2020-06-15
Add Robust BMA

JorisGoosen:
- https://github.com/jasp-stats/MetaAnalysis/commit/b33e9a2f74ac7587bf6ade5866ad4e1c9423e346
- Implements https://github.com/jasp-stats/INTERNAL-jasp/issues/910

### DIFF
--- a/JASP-Tests/R/tests/testthat/test-RobustBayesianMetaAnalysis.R
+++ b/JASP-Tests/R/tests/testthat/test-RobustBayesianMetaAnalysis.R
@@ -1,0 +1,1455 @@
+context("Robust Bayesian Meta-Analysis")
+
+# path to the pre-fitted RoBMA model
+fitted_path <- "RoBMA_testfit.RDS"
+
+### prior distibutions plots 
+{
+  options <- jasptools::analysisOptions("RobustBayesianMetaAnalysis")
+  options$.meta <- list(input_CI = list(containsColumn = TRUE), input_ES = list(
+    containsColumn = TRUE), input_N = list(containsColumn = TRUE), 
+    input_N1 = list(containsColumn = TRUE), input_N2 = list(containsColumn = TRUE), 
+    input_SE = list(containsColumn = TRUE), input_labels = list(
+      containsColumn = TRUE), input_t = list(containsColumn = TRUE), 
+    priors_mu = list(list(name = list(containsColumn = TRUE)), 
+                     list(name = list(containsColumn = TRUE)), list(name = list(
+                       containsColumn = TRUE)), list(name = list(containsColumn = TRUE)), 
+                     list(name = list(containsColumn = TRUE)), list(name = list(
+                       containsColumn = TRUE)), list(name = list(containsColumn = TRUE)), 
+                     list(name = list(containsColumn = TRUE))), priors_omega = list(
+                       list(name = list(containsColumn = TRUE)), list(name = list(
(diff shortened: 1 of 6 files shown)


JASP-Tests/R/tests/testthat/test-RobustBayesianMetaAnalysis.R
```diff
@@ -0,0 +1,1455 @@
+context("Robust Bayesian Meta-Analysis")
+
+# path to the pre-fitted RoBMA model
+fitted_path <- "RoBMA_testfit.RDS"
```
FBartos, line 4: this needs to link to the pre-fitted model.

FBartos:
The unit tests will fail because they need a newer version of the package & set-up the path to the pre-fitted model.

JASP-Tests/R/tests/testthat/test-RobustBayesianMetaAnalysis.R
TimKDJ: Is this something Joris should do?

FBartos: anyone who knows how to make it work. The analysis needs to load the .RDS object and I have no idea how are the paths in the unit tests specified.

JorisGoosen: Cant you just do `".\RoBMA_testfit.RDS"` ? It is right next to the unittest R file right?

FBartos: that's how I set it up in the module, but I didn't know whether the location would be the same inside of jasp-desktop

FBartos: and it's loaded from inside the `jasptools::run()` command. I'm just stating that I have no idea how this should be set-up.

JorisGoosen: Me neither, but I figured either @TimKDJ or @vandenman would have a better guess.

vandenman: Something like `file <- file.path(jasptools:::.pkgOptions$tests.dir, "RoBMA_testfit.RDS")` should work if you place the `.RDS` file in the same directory as all the unit tests.

Resources/Meta Analysis/description.json
```diff
@@ -1,13 +1,13 @@
 {
 	"moduleDescription" :
 	{
-		"name": 		"Meta Analysis",
```
TimKDJ, line 4: Then the folder name in Resources & `RibbonModel` in mainwindow.cpp also need to be changed right?

Schermafbeelding 2020-06-17 om 09 42 47

JorisGoosen: Yes, they should, wait, Ill fix this either by reverting or renaming

TimKDJ (approved):

JASP-Tests/R/tests/testthat/test-RobustBayesianMetaAnalysis.R
FBartos: it did, thanks!

PR #4136, opened 2020-06-16
Added upgrade options for Reliability

JorisGoosen:
- Implements https://github.com/jasp-stats/INTERNAL-jasp/issues/915
- Also made the Reliability conform to the qml styleguide because they were a mess
- Updated JASPs version to 0.13

### DIFF
--- a/JASP-Desktop/mainwindow.cpp
+++ b/JASP-Desktop/mainwindow.cpp
@@ -303,8 +303,6 @@ void MainWindow::makeConnections()
 	connect(_resultsJsInterface,	&ResultsJsInterface::allUserDataChanged,			_analyses,				&Analyses::allUserDataChanged								);
 	connect(_resultsJsInterface,	&ResultsJsInterface::resultsPageLoadedSignal,		_languageModel,			&LanguageModel::resultsPageLoaded							);
 
-
-
 	connect(_analyses,				&Analyses::countChanged,							this,					&MainWindow::analysesCountChangedHandler					);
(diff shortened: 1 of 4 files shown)


vandenman (approved):
So I can't really judge upgrades.json but the rest looks good to me

vandenman:
unit tests fail because of MixedModelsBGLMM btw

TimKDJ:
> So I can't really judge upgrades.json but the rest looks good to me

You can test it by loading an old Reliability .jasp file into JASP. If that works then it should be good (maybe you already did this)

boutinb:
> > So I can't really judge upgrades.json but the rest looks good to me
> 
> You can test it by loading an old Reliability .jasp file into JASP. If that works then it should be good (maybe you already did this)

I just test it. It seems to work.

JorisGoosen:
Yeah, so I had kind of hoped that @vandenman would take a look at the renames in the upgrades.
Every `"blabla": { "_renamed": "oldBlabla" }` denoted an instance of oldBlabla being renamed to blabla.

Because just loading an old Reliability will work fine, but maybe some options are not correct... Which I think is important.

vandenman:
@JorisGoosen Ok I'll give that a try

vandenman:
Everything seems to work as intended.

JorisGoosen:
Ok good!

JorisGoosen:
Then Ill just add the last few changes by @juliuspf in https://github.com/jasp-stats/INTERNAL-jasp/issues/916

PR #4139, opened 2020-06-17
RoBMA unit tests fix

FBartos:


vandenman:
@FBartos can you rebase and enable the tests? JAGS should work now.

FBartos:
@vandenman This should be it. I also tried to rename the file with unit-tests to be in lower case, but git doesn't seem to register the change.

vandenman:
@FBartos https://stackoverflow.com/questions/17683458/how-do-i-commit-case-sensitive-only-filename-changes-in-git
`git mv -f OldFileNameCase newfilenamecase`
or 
`git config core.ignorecase false`
might be handy too (I remember doing that on windows).

Lowercase would be nice otherwise `jasptools::testAnalysis` doesn't work becasue it converts filenames to lowercase for some reason.

FBartos:
thanks, done (y)

vandenman:
The tests seem to pass, great!

One last thing, on Travis the tests prints:
```
full precision may not have been achieved in 'pnt{final}'
```
a couple of hundred or thousand times. Can you suppress that message?

image

FBartos:
I wish I could - I spent several days trying to do that - but couldn't pull it off. It's printed by a C function from R math library called thru JAGS. It isn't using a standard output and nothing seems to work. I already filled an issue on JAGS page, but they are not that active there.

vandenman:
It's a warning message in distributions right? `dt(1.424781, 1486, -5)` gives it as a warning. Can you link me to the issue? Because this kinda makes the Travis log hard to read..

FBartos:
yes, it's exactly this type of warning. However, it's handled differently when called from JAGS

https://sourceforge.net/p/mcmc-jags/feature-requests/23/

a much better description (to which I link) is here:
https://stackoverflow.com/questions/50635735/suppress-jags-value-out-of-range-warnings-in-r

vandenman:
nvm that didn't work at all.

FBartos:
I would be really happy if you stumbled upon any other way how to disable those. They pollute the R output when using the package and they are not really meaningful for my use-case. I really tried looking for that, but I might not know what to search for.

vandenman:
Well, I'm looking around a bit. But to be honest, the whole output things looks like a design flaw in R (whether output can be suppressed or not shouldn't depend on how other people program, it should depend on what R wants). Rstan people also had difficulties suppressing their output.

vandenman:
Does this work for you?
```r
cat('
model {
    K <- 1.1

    K.mvhypgeom <- exp( logfact(sum(n[])) - logfact(nMissing) - logfact( sum(n[]) - nMissing))

    p ~ dunif(0,1)

    for (t in 1:N) {
    X.missing[t] ~ dpois( missRate )
    }     

    nMissing ~ dsum(X.missing[1],X.missing[2],X.missing[3],X.missing[4],X.missing[5],X.missing[6],X.missing[7],X.missing[8],X.missing[9],X.missing[10])

    for (t in 1:N) {
    pX.missing[t] <- exp(logfact(n[t]) - logfact( X.missing[t]) - logfact( n[t] - X.missing[t]))
    ones2[t] ~ dbern(pX.missing[t]/K.mvhypgeom)
    }

    for (t in 1:N) {
    X[t] <- X.obs[t] + X.missing[t]

    likX[t] <- dbin( X[t], p, n[t])
    ones1[t] ~ dbern( likX[t] / K)    
    }
    }
  ',
    file = {example.model <- tempfile()},
    sep = ''
)

simBinTS <- function(n, p , nMissing) {
  X.full <- X <- rbinom(N, size = n, prob = p)
  for (i in seq_len(nMissing)) {
    idx <- sample(1:N, size = 1, prob = X)
    X[idx] <- X[idx] - 1
  }
  return(data.frame(n = n, X = X, X.full = X.full))
}


N <- 10
p <- 0.3
set.seed(123)
n <- rpois(N, lambda = 30)
nMissing <- 10
missRate <- 1/10
ts <- simBinTS(p = p, n = n, nMissing = nMissing)
X.obs <- ts$X
n <- ts$n
X.full <- ts$X.full
ones1 <- rep(1,N)
ones2 <- rep(1,N)

jags.inits <- function(){
  list(X.missing = X.full-X.obs)
}

library("R2jags")

foo <- function(data, inits, parameters.to.save, model.file, n.chains, n.iter, n.burnin, n.thin, progress.bar,
                X.obs, n, N, nMissing, ones1, ones2, missRate, X.full2) {
  X.full <- X.full2
  jags.inits <- function() {
    list(X.missing = X.full-X.obs)
  }
  
  R2jags::jags(
    data = data,
    # inits = inits,
    parameters.to.save = parameters.to.save,
    model.file = model.file,
    n.chains = n.chains,
    n.iter = n.iter,
    n.burnin = n.burnin,
    n.thin = n.thin,
    progress.bar = progress.bar)
}

ans <- callr::r(
  foo, 
  args = list(
    data = list("X.obs", "n", "N", "nMissing", "ones1", "ones2", "missRate"),
    inits = jags.inits,
    parameters.to.save = "p",
    model.file = example.model,
    n.chains = 3,
    n.iter = 1000,
    n.burnin = 500,
    n.thin = 1,
    progress.bar = "none",
    X.obs, n, N, nMissing, ones1, ones2, missRate, X.full
  )
)
ans
```
I don't like this solution at all but it does work for me. If it also works for you we use at minimum wrap it in `if (identical(Sys.getenv("TRAVIS"), "true"))` so it isn't done inside JASP.

FBartos:
Great, it works. I will add it directly to the package as a silent option.

vandenman:
Note that it's equivalent to opening a new R, running the code there, and then loading the object back into the old R. Which is anything but pretty.

FBartos:
The RoBMA package update is required (the unit tests will fail until then). Afterward, there should be no messages produced by the fitting function.

I also found and fixed a bug with the seed not being passed properly.

boutinb:
@FBartos The RoBMA package was updated. Can you rebase and solve the conflicts of this PR, then we can merge it.

FBartos:
ok, this got way behind the current implementation. I will rewrite the tests and create a new PR

Issue #830, opened 2015-07-25
Loading Indicator Descriptives Frequency Table (Nominal/Ordinal) does not disappear


quentingronau:
![loadingindicatordescriptivestable](https://cloud.githubusercontent.com/assets/8697436/8890257/7a6a6c0a-32f9-11e5-8427-fac6524020ac.png)

jonathon-love:
good catch!

jonathon-love:
hey, i'm having difficulty reproducing this; can you give more detailed instructions to reproduce?

jonathon-love:
hey, i'm having difficulty reproducing this; can you give more detailed instructions to reproduce?

quentingronau:
Hi, maybe this is something that just happens on my laptop (e.g., I also still don't get a loading indicator for the plots). For me, it happens, for example, if I load the 'ssgo-fred' data, click on 'Descriptives', select 'ssd' as variable and then click on 'Display Frequency Tables (nominal and ordinal)'.

PR #4141, opened 2020-06-17
Reliability fix test issues

juliuspfadt:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/798
Fixes https://github.com/jasp-stats/jasp-test-release/issues/799
Fixes https://github.com/jasp-stats/jasp-test-release/issues/813
Fixes https://github.com/jasp-stats/jasp-test-release/issues/814
Fixes https://github.com/jasp-stats/jasp-test-release/issues/812
Fixes https://github.com/jasp-stats/jasp-test-release/issues/811
Fixes https://github.com/jasp-stats/jasp-test-release/issues/830

### DIFF
--- a/JASP-Engine/JASP/R/reliabilityAnalysisBayesian.R
+++ b/JASP-Engine/JASP/R/reliabilityAnalysisBayesian.R
@@ -85,7 +85,10 @@ reliabilityBayesian <- function(jaspResults, dataset, options) {
     if (sidx == 0) {
       return("")
     } else {
-      footnote <- .footnoteNegativeCorrelation(variables, idx)
+      footnote <- sprintf(ngettext(length(variables[idx]),
+                                   "The following item correlated negatively with the scale: %s. ",
+                                   "The following items correlated negatively with the scale: %s. "),
+                          paste(variables[idx], collapse = ", "))      
       return(footnote)
     }
   } else {
@@ -970,10 +973,4 @@ reliabilityBayesian <- function(jaspResults, dataset, options) {
   return(dataset)
 }
 
-.footnoteNegativeCorrelation <- function(variables, idx) {
-  footnote <- sprintf(ngettext(length(variables[idx]),
-                               "The following item correlated negatively with the scale: %s. ",
-                               "The following items correlated negatively with the scale: %s. "),
-                      paste(variables[idx], collapse = ", "))
-}
   
--- a/JASP-Engine/JASP/R/reliabilityAnalysisBayesian.R
+++ b/JASP-Engine/JASP/R/reliabilityAnalysisBayesian.R
@@ -1,8 +1,8 @@
 reliabilityBayesian <- function(jaspResults, dataset, options) {
 
-	dataset <- .BayesianReliabilityReadData(dataset, options)
+	dataset <- .reliabilityReadData(dataset, options)
 
-	.BayesianReliabilityCheckErrors(dataset, options)
+	.reliabilityCheckErrors(dataset, options)
   
 	model <- .BayesianReliabilityMainResults(jaspResults, dataset, options)
 
@@ -52,7 +52,7 @@ reliabilityBayesian <- function(jaspResults, dataset, options) {
   return(derivedOptions)
 }
 
-.BayesianReliabilityReadData <- function(dataset, options) {
+.reliabilityReadData <- function(dataset, options) {
 
   variables <- unlist(options[["variables"]])
 	if (is.null(dataset)) {
@@ -61,10 +61,10 @@ reliabilityBayesian <- function(jaspResults, dataset, options) {
   return(dataset)
 }
 
-.BayesianReliabilityCheckErrors <- function(dataset, options) {
+.reliabilityCheckErrors <- function(dataset, options) {
 
   .hasErrors(dataset = dataset, perform = "run",
-             type = c("infinity", "variance", "observations"),
+             type = c("infinity", "variance", "observations", "varCovData"),
              observations.amount = " < 3",
              custom = .checkEigen,
              exitAnalysisIfErrors = TRUE)
@@ -77,6 +77,23 @@ reliabilityBayesian <- function(jaspResults, dataset, options) {
   }
 }
 
+.reliabilityCheckLoadings <- function(dataset, variables) {
+  if (ncol(dataset > 2)) {
+    prin <- psych::principal(dataset)
+    idx <- prin[["loadings"]] < 0
+    sidx <- sum(idx)
+    if (sidx == 0) {
+      return("")
+    } else {
+      footnote <- .footnoteNegativeCorrelation(variables, idx)
+      return(footnote)
+    }
+  } else {
+    return("")  
+  }
+}
+
+
 # estimate reliability ----
 .BayesianReliabilityMainResults <- function(jaspResults, dataset, options) {
   if (!options[["mcDonaldScale"]] && !options[["alphaScale"]] && !options[["guttman2Scale"]]
@@ -87,7 +104,7 @@ reliabilityBayesian <- function(jaspResults, dataset, options) {
       dataset <- .reverseScoreItems(dataset, options)
     }
     model <- list()
-    model[["footnote"]] <- .BayesianReliabilityCheckLoadings(dataset, variables)
+    model[["footnote"]] <- .reliabilityCheckLoadings(dataset, variables)
     return(model)
   }
   
@@ -116,7 +133,7 @@ reliabilityBayesian <- function(jaspResults, dataset, options) {
           options[["missings"]] <- "complete.obs"
           }
       }
-      model[["footnote"]] <- .BayesianReliabilityCheckLoadings(dataset, variables)
+      model[["footnote"]] <- .reliabilityCheckLoadings(dataset, variables)
       
       chains <- options[["noChains"]]
       samples <- options[["noSamples"]]
@@ -245,25 +262,7 @@ reliabilityBayesian <- function(jaspResults, dataset, options) {
 }
 
 
-.BayesianReliabilityCheckLoadings <- function(dataset, variables) {
-  if (ncol(dataset > 2)) {
-    prin <- psych::principal(dataset)
-    idx <- prin[["loadings"]] < 0
-    sidx <- sum(idx)
-    if (sidx == 0) {
-      return("")
-    } else {
-      hasSchar <- if (sidx == 1L) "" else "s"
-      footnote <- sprintf(ngettext(length(variables[idx]),
-                                   "The following item correlated negatively with the scale: %s. ",
-                                   "The following items correlated negatively with the scale: %s. "),
-                          paste(variables[idx], collapse = ", "))
-      return(footnote)
-    }
-  } else {
-    return("")  
-    }
-}
+
 
 
 # ----------------------------- tables ------------------------------------
@@ -632,7 +631,7 @@ reliabilityBayesian <- function(jaspResults, dataset, options) {
 
 	
 	if (!is.null(shade)) {
-	  datFilter <- datDens[datDens[["x"]] >= shade[1] && datDens[["x"]] <= shade[2], ]
+	  datFilter <- datDens[datDens[["x"]] >= shade[1] & datDens[["x"]] <= shade[2], ]
 	  g <- g + ggplot2::geom_ribbon(data = datFilter, mapping = ggplot2::aes(ymin = 0, ymax = y), 
 	                                fill = "grey", alpha = 0.95) +
 	           ggplot2::geom_line(size = .85)
@@ -971,3 +970,10 @@ reliabilityBayesian <- function(jaspResults, dataset, options) {
   return(dataset)
 }
 
+.footnoteNegativeCorrelation <- function(variables, idx) {
+  footnote <- sprintf(ngettext(length(variables[idx]),
+                               "The following item correlated negatively with the scale: %s. ",
+                               "The following items correlated negatively with the scale: %s. "),
+                      paste(variables[idx], collapse = ", "))
+}
+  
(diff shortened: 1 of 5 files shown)


JASP-Engine/JASP/R/reliabilityAnalysisBayesian.R
```diff
@@ -80,6 +79,20 @@ reliabilityBayesian <- function(jaspResults, dataset, options) {
 
 # estimate reliability ----
 .BayesianReliabilityMainResults <- function(jaspResults, dataset, options) {
+  if (!options[["mcDonaldScale"]] & !options[["alphaScale"]] & !options[["guttman2Scale"]]
```
TimKDJ: Please use double `&&` in `if` statements

JASP-Engine/JASP/R/reliabilityAnalysisBayesian.R
```diff
@@ -80,6 +79,20 @@ reliabilityBayesian <- function(jaspResults, dataset, options) {
...
+      & !options[["guttman6Scale"]] & !options[["glbScale"]] & !options[["averageInterItemCor"]]
+      & !options[["meanScale"]] & !options[["sdScale"]]) {
+    variables <- options[["variables"]]
+    dataset <- as.matrix(dataset) # fails for string factors!
+    if (length(options[["reverseScaledItems"]]) > 0L) {
+      cols <- match(unlist(options[["reverseScaledItems"]]), .unv(colnames(dataset)))
+      total <- apply(as.matrix(dataset[, cols]), 2, min) + apply(as.matrix(dataset[, cols]), 2, max)
+      dataset[ ,cols] = matrix(rep(total, nrow(dataset)), nrow(dataset), length(cols), byrow=T) - dataset[ ,cols]
+    }
```
TimKDJ: As far as I can tell these lines are identical [to these](https://github.com/jasp-stats/jasp-desktop/blob/859eefde7141571d7ed0cf6704df46b7dd35fa4e/JASP-Engine/JASP/R/reliabilityAnalysisBayesian.R#L105-L110). It's good practice to turn this into a little function which you call in both places; it'll ensure that we don't need to make the same changes in multiple places in the future.

JASP-Engine/JASP/R/reliabilityAnalysisBayesian.R
```diff
@@ -238,26 +252,54 @@ reliabilityBayesian <- function(jaspResults, dataset, options) {
...
+  } else {
+    return("")  
+    }
 }
 
 
 # ----------------------------- tables ------------------------------------
 .BayesianReliabilityScaleTable <- function(jaspResults, model, options) {
   
-  if (!is.null(jaspResults[["scaleTable"]])) {
```
TimKDJ, line 260: If none of the dependencies change then we don't need to create this table again right? I think this `if` can stay

JASP-Engine/JASP/R/reliabilityAnalysisBayesian.R
```diff
@@ -238,26 +252,54 @@ reliabilityBayesian <- function(jaspResults, dataset, options) {
...
+    scaleTable <- createJaspTable(gettext("Bayesian Scale Reliability Statistics"))
+    scaleTable$dependOn(options = c("variables", "reverseScaledItems", "credibleIntervalValueScale"))
+    
+    scaleTable$addColumnInfo(name = "estimate", title = "Estimate", type = "string")
+    intervalLow <- gettextf("%s%% CI",
+                            format(100*options[["credibleIntervalValueScale"]], digits = 3, drop0trailing = TRUE))
+    intervalUp <- gettextf("%s%% CI",
+                           format(100*options[["credibleIntervalValueScale"]], digits = 3, drop0trailing = TRUE))
+    intervalLow <- gettextf("%s lower bound", intervalLow)
+    intervalUp <- gettextf("%s upper bound", intervalUp)
```
TimKDJ: If you lift this stuff above the `if` statement then you don't have to do the same thing again [under the `if` statement](https://github.com/jasp-stats/jasp-desktop/blob/859eefde7141571d7ed0cf6704df46b7dd35fa4e/JASP-Engine/JASP/R/reliabilityAnalysisBayesian.R#L306-L319).

JASP-Engine/JASP/R/reliabilityAnalysisFrequentist.R
```diff
@@ -70,7 +70,20 @@ reliabilityFrequentist <- function(jaspResults, dataset, options) {
...
+  if (!options[["mcDonaldScalef"]] & !options[["alphaScalef"]] & !options[["guttman2Scalef"]]
+      & !options[["guttman6Scalef"]] & !options[["glbScalef"]] & !options[["averageInterItemCorf"]]
+      & !options[["meanScalef"]] & !options[["sdScalef"]]) {
+    variables <- options[["variables"]]
+    dataset <- as.matrix(dataset) # fails for string factors!
+    if (length(options[["reverseScaledItems"]]) > 0L) {
+      cols <- match(unlist(options[["reverseScaledItems"]]), .unv(colnames(dataset)))
+      total <- apply(as.matrix(dataset[, cols]), 2, min) + apply(as.matrix(dataset[, cols]), 2, max)
+      dataset[ ,cols] = matrix(rep(total, nrow(dataset)), nrow(dataset), length(cols), byrow=T) - dataset[ ,cols]
+    }
```
TimKDJ: This looks familiar ;)

JASP-Engine/JASP/R/reliabilityAnalysisFrequentist.R
```diff
@@ -385,16 +397,20 @@ reliabilityFrequentist <- function(jaspResults, dataset, options) {
...
+  if (ncol(dataset > 2)) {
+    prin <- psych::principal(dataset)
+    idx <- prin[["loadings"]] < 0
+    sidx <- sum(idx)
+    if (sidx == 0) {
+      return("")
+    } else {
+      hasSchar <- if (sidx == 1L) "" else "s"
+      footnote <- gettextf("The following item%s correlated negatively with the scale: %s. ",
+                           hasSchar, paste0(variables[idx], collapse = ", "))
```
TimKDJ: You should use `sprintf(ngettext())` for this (also in the Bayesian setting), your solution is quite specific to the English language.

JASP-Engine/JASP/R/reliabilityAnalysisFrequentist.R
```diff
@@ -404,14 +420,40 @@ reliabilityFrequentist <- function(jaspResults, dataset, options) {
 
 .frequentistReliabilityScaleTable <- function(jaspResults, model, options) {
   
-  if (!is.null(jaspResults[["scaleTableF"]])) {
+  if ((!options[["mcDonaldScalef"]] & !options[["alphaScalef"]] & !options[["guttman2Scalef"]] 
+       & !options[["guttman6Scalef"]] & !options[["glbScalef"]] & !options[["averageInterItemCorf"]]
+       & !options[["meanScalef"]] & !options[["sdScalef"]])) {
+    scaleTableF <- createJaspTable(gettext("Frequentist Scale Reliability Statistics"))
```
TimKDJ: Same suggestion here as for the Bayesian one

TimKDJ (approved):
All issues are fixed, which is great! 
I do have a few code comments though.

TimKDJ (changes requested):
Hit the wrong button there.

JASP-Engine/JASP/R/reliabilityAnalysisBayesian.R
juliuspfadt: Well, this causes an error for me, because when no coefficients are checked an empty ScaleTable is displayed, when I then check omega as a coefficient, the Table stays empty. Do you have an idea how to resolve this?

juliuspfadt: Ah I found it, the dependencies for the empty table were incomplete.

JASP-Engine/JASP/R/reliabilityAnalysisFrequentist.R
juliuspfadt: okay, I am done with it. Can you take a look again?

JASP-Engine/JASP/R/reliabilityAnalysisBayesian.R
```diff
@@ -647,7 +632,7 @@ reliabilityBayesian <- function(jaspResults, dataset, options) {
 
 	
 	if (!is.null(shade)) {
-	  datFilter <- datDens[datDens[["x"]] >= shade[1] & datDens[["x"]] <= shade[2], ]
+	  datFilter <- datDens[datDens[["x"]] >= shade[1] && datDens[["x"]] <= shade[2], ]
```
TimKDJ: I think here you probably want a single `&`

`&&`:
- expects a single boolean on either side => `T && F`
- returns a single boolean
- stops evaluation of the statement if the left hand side returns `FALSE` => `if (FALSE && stop("we reached the stop")) {}`
- is useful in if statements

`&`:
- is vectorized and can work with multiple boolean values on either side => `c(T, T, F) & c(F, F, T)`
- returns a vector of booleans
- always evaluates both sides => `if (FALSE & stop("we reached the stop")) {}`
- is useful in indexing to select multiple elements

JASP-Engine/JASP/R/reliabilityAnalysisFrequentist.R
```diff
@@ -405,8 +399,10 @@ reliabilityFrequentist <- function(jaspResults, dataset, options) {
       return("")
     } else {
       hasSchar <- if (sidx == 1L) "" else "s"
```
TimKDJ: This variable can go

JASP-Engine/JASP/R/reliabilityAnalysisFrequentist.R
```diff
@@ -405,8 +399,10 @@ reliabilityFrequentist <- function(jaspResults, dataset, options) {
...
-      footnote <- gettextf("The following item%s correlated negatively with the scale: %s. ",
-                           hasSchar, paste0(variables[idx], collapse = ", "))
+      footnote <- sprintf(ngettext(length(variables[idx]),
+                                   "The following item correlated negatively with the scale: %s. ",
+                                   "The following items correlated negatively with the scale: %s. "),
+                          paste(variables[idx], collapse = ", "))
```
TimKDJ: This footnote is shared between frequentist and bayesian right? Then you can again make a function and call the function from both files.

JASP-Engine/JASP/R/reliabilityAnalysisFrequentist.R
```diff
@@ -639,3 +632,11 @@ reliabilityFrequentist <- function(jaspResults, dataset, options) {
   om <- out$omega
   return(om)
 }
+
+.reverseScoreItems <- function(dataset, options) {
```
TimKDJ: I *think* this code is the same as the other `.reverseScoreItems()` you defined? You only need one of them. All functions in JASP are available in all files. You could even call any function in `anova.R` file if you wanted to. When you define two functions with the same name one will simply overwrite the other.

TimKDJ (changes requested):
Good changes! Just little things now

JASP-Engine/JASP/R/reliabilityAnalysisBayesian.R
juliuspfadt: thanks for the clarification :)

JASP-Engine/JASP/R/reliabilityAnalysisFrequentist.R
juliuspfadt: well I just defined one function that checks the loadings now, and call it from both modules

TimKDJ: ~~Yep and that's great! But the same principle applies to other code that is duplicated too (like this footnote): it's better to add it in a function and to call that function from different places. Otherwise there's a good chance that in the future when we change this footnote we forget to adjust it in both files.~~

Never mind, I think I misunderstood what you wrote (you merged the two `ReliabilityCheckLoadings()`? -- that's even better!)

juliuspfadt:
I think I fixed the latest stuff. But, you need to use the latest commit of the package Bayesrel from GitHub. Let me know if anything else crashes

TimKDJ:
> I think I fixed the latest stuff. But, you need to use the latest commit of the package Bayesrel from GitHub. Let me know if anything else crashes

If we merge this before we update the Bayesrel package do bad things happen? Or will it only mean that https://github.com/jasp-stats/jasp-test-release/issues/830 is not fixed yet?

juliuspfadt:
> > I think I fixed the latest stuff. But, you need to use the latest commit of the package Bayesrel from GitHub. Let me know if anything else crashes
> 
> If we merge this before we update the Bayesrel package do bad things happen? Or will it only mean that [jasp-stats/jasp-test-release#830](https://github.com/jasp-stats/jasp-test-release/issues/830) is not fixed yet?

I dont think bad things are going to happen, the updated package fixes another bug that I found. Some analysis may crash. But other than that it should work. Oh, and the #830 will be fixed without the updated package.

JASP-Engine/JASP/R/reliabilityAnalysisFrequentist.R
juliuspfadt: well, there is only one function which is .reliabilityCheckLoadings. But I made another function for the footnote. Which seems redundant to me, but I am here to obey. Also I got rid of other duplicate functions such as , checkErrors, and readData

TimKDJ: Yes, that is indeed redundant, hence my edit in [my previous reply](https://github.com/jasp-stats/jasp-desktop/pull/4141#discussion_r442744804). I didn't realize the whole checkloading functions overlapped completely and could be merged. The footnote only needed to be in its own function when it was used in multiple locations, but that is no longer the case.

Very nice that you also merged the other duplicates!

TimKDJ (approved):
Good stuff

JASP-Engine/JASP/R/reliabilityAnalysisFrequentist.R
juliuspfadt: OK, I deleted the redundant function. All should be well

PR #4142, opened 2020-06-18
Refining interface of Robust Bayesian Meta Analysis

boutinb:
. 4 radio buttons at the top if the interface were in 1 row, making
them slightly too wide: I put the 4 radio buttons in 2 columns of 2
rows.
. The tables of distributions priors did not grow when adding new
distributions: now it grows with the number of distributions.

### DIFF
--- a/JASP-Desktop/components/JASP/Controls/AssignButton.qml
+++ b/JASP-Desktop/components/JASP/Controls/AssignButton.qml
@@ -38,6 +38,8 @@ Button
 	
 	text:			""
 	enabled:		false
+	visible:		source.visible && target.visible
+
 	iconSource:		leftToRight ? iconToRight : iconToLeft
 
 	control.buttonPadding:	2
@@ -50,39 +52,38 @@ Button
 	function setIconToLeft()	{ if (rightSource.activeFocus)	leftToRight = false;	}
 	function setState()
 	{
-		var result = source.enabled && target.enabled;
-		if (result && source.model && source.model.selectedItems().length > 0)
+		var isEnabled = source.enabled && target.enabled;
+		if (isEnabled && source.model && source.model.selectedItems().length > 0)
 		{
 			if (target.allowedColumns.length > 0)
 			{
-				result = false;
+				isEnabled = false;
 				var sourceSelectedItemsTypes = source.model.selectedItemsTypes()
 				for (var i = 0; i < sourceSelectedItemsTypes.length; i++)
 				{
 					var itemType = sourceSelectedItemsTypes[i];
 					if (target.allowedColumns.includes(itemType))
-						result = true;
+						isEnabled = true;
 				}
 			}
 		}
 
-		if (result && interactionControl)
+		if (isEnabled && interactionControl)
 		{
 			if (target.addInteractionOptions && source.model)
 			{
 				var nb = source.model.selectedItems().length
-				interactionControl.enabled = result
+				interactionControl.enabled = isEnabled
 				var enabledOptions = [ true, nb > 1, nb > 2, nb > 3, nb > 4, true ]
 				interactionControl.enabledOptions = enabledOptions
 				if (!enabledOptions[interactionControl.currentIndex])
-					result = false;
+					isEnabled = false;
 			}
 			else
 				interactionControl.enabled = false
 		}
 		
-		enabled = result
-		
+		enabled = isEnabled
 	}
 
 	onSourceChanged:	setState()
--- a/JASP-Desktop/components/JASP/Controls/ComponentsList.qml
+++ b/JASP-Desktop/components/JASP/Controls/ComponentsList.qml
@@ -104,6 +104,7 @@ JASPGridControl
 					id					: deleteMouseArea
 					anchors.fill		: parent
 					onClicked			: removeItem(index)
+					cursorShape			: Qt.PointingHandCursor
 				}
 			}
 		}
--- a/JASP-Desktop/components/JASP/Controls/FileSelector.qml
+++ b/JASP-Desktop/components/JASP/Controls/FileSelector.qml
@@ -17,6 +17,7 @@ TextField
 	{
 		id				: button
 		text			: qsTr("Browse")
+		z				: 5
 		anchors
 		{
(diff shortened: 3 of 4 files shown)


TimKDJ:
Those were definitely the two most important ones I had noticed too.
Although I think the following improvements would be great:

- [x] If the assigned variables list is hidden, then the assign button should also be hidden:
Schermafbeelding 2020-06-18 om 14 02 52

- [x] Can we get the little grabbing hand cursor on the `x` in the Effect table under Priors (also considering we use a fancy `+` should we also use the fancy red `x`? Not sure how that'd look):
Schermafbeelding 2020-06-18 om 14 38 11

- [x] This button also gets the wrong cursor when you hover over it:
Schermafbeelding 2020-06-18 om 14 41 29

- [ ] The components under `Plots` and `MCMC Diagnostics` have a lot of empty, wasted space. I also think there are groups without titles? This makes it unclear to the user if things really belong together or are just grouped together "by accident"; basically the entire right hand side:
Schermafbeelding 2020-06-18 om 14 50 27

- [x] In the `Advanced` section the interface would look much better if the various text inputs would align with each other. Like how it's done under "Estimation settings (MCMC)":
Schermafbeelding 2020-06-18 om 14 54 10

boutinb:
_Can we get the little grabbing hand cursor on the x in the Effect table under Priors (also considering we use a fancy + should we also use the fancy red x? Not sure how that'd look):_
I found the fancy red x too... fancy.

I am no so sure how to rearrange the Plots and MCMC Diagnostics: @FBartos explicitly chose for this layout. For example in Plots, _Type_ group has apparently to be at the level as the _Forest plot_ group, and the _Conditional models only_ group has apparently to be as the same level as the _Individual models_

TimKDJ:
@FBartos are the groupings under `Plots` ([this](https://github.com/jasp-stats/jasp-desktop/blob/bd4a3cbdfbf91bf4af40570a5b0ac979eeda40e1/Resources/Meta%20Analysis/qml/RobustBayesianMetaAnalysis.qml#L1577-L1603) and [this](
https://github.com/jasp-stats/jasp-desktop/blob/stable/Resources/Meta%20Analysis/qml/RobustBayesianMetaAnalysis.qml#L1656-L1709), mainly the checkboxes) done because they belong together and are logical groups, or to avoid weird placement of items? I kinda want to get rid of some of the unused space so it's good to know the thought process behind the current placement.

FBartos:
These are paired because they belong together in these groups:
image

Here, there are only two groups that needs to be together, the rest is free.
image
However, I didn't add "Repeatability" under "Estimation settings (MCMC)" since it covers much more and I want its headings to be on the same level as "Estimation settings (MCMC)" (ie. not to be a subcategory as is "Adaptation" ie.)

FBartos:
I think that this would look nicer, but I didn't know to merge the left column together and have the proper level of headings:

image

boutinb:
I spoke to @FBartos, and it is now what we think how the groups should be aligned.

TimKDJ:
> These are paired because they belong together in these groups:
> image

I would not have guessed those groupings from the interface.
Considering we usually (always?) make them vertical, what do you think about this @FBartos:
Schermafbeelding 2020-06-23 om 09 57 33

FBartos:
I find my version clearer and more esthetically pleasing (what a surprise given the fact I designed it), but we can go with your suggestion if that's the way how is it usually done in JASP @TimKDJ

FBartos:
A thing is that the right column was a control panel for all plots generated using the right column and this column spacing feels like as if it meshes this options together.

TimKDJ:
> I find my version clearer and more esthetically pleasing (what a surprise given the fact I designed it), but we can go with your suggestion if that's the way how is it usually done in JASP @TimKDJ

Hehe, yeah it makes sense that you like what you create :)
I do think consistency across the application is quite important.

> A thing is that the right column was a control panel for all plots generated using the right column and this column spacing feels like as if it meshes this options together.

I don't understand what you mean by this?

FBartos:
image

the pink parts of each group create the actual figures - either for effects, heterogeneity and etc..., the right part of the group then changes a style of those plots (and changes all of the plots created by the pink part). And this relationship is kinda broken if you move it bellow the plots

TimKDJ:
The problem is that to me it wasn't even clear that the right part belonged to the left part. Let alone that it served a special purpose of "modifying" the left.
This grouping doesn't happen anywhere else in JASP (as you could maybe tell because you had to use those separate `group {}` components to make it), so it needs to be made more transparant what is happening.

You made the analysis and if you feel that the analysis becomes worse by my suggested changes, then we don't do them. But if we don't, then we need to add group titles to the right hand side to clarify all the subtext that comes from their placement.

FBartos:
I could also break it down into two sections

Plots - Pooled Estimatest
and
Plots - Individual Models

and keep the left and right side columns as they are. What do you think about that?

FBartos:
Do the changes to the assigned variable list include a fix for this problem @boutinb ?
https://github.com/jasp-stats/jasp-test-release/issues/833

Basically, if a variable is assigned to the input value, a "t-statistics" and the input type is then switched from "Cohen's d" to "Correlation", the assigned value stays in the `AssignedVariablesList` even though it has been disabled. I know that the arrows are hidden now. But is the variable removed from the disabled `AssignedVariablesList` as well?

TimKDJ:
> I could also break it down into two sections
> 
> Plots - Pooled Estimatest
> and
> Plots - Individual Models
> 
> and keep the left and right side columns as they are. What do you think about that?

Yeah that sounds good. Then it's easier for the user to navigate

boutinb:
So In place of 2 apart sections for the plots, I have just added a Divider (a horizontal line).
Fixes jasp-stats/jasp-test-release#833

FBartos:
Awesome! I checked it out and it fixes the issues!

boutinb:
@TimKDJ I think we should merge this PR, since it solves already many problems and incomprehension with this analysis.

TimKDJ:
I haven't checked out your latest changes, but it's already a big improvement, I'd say go for it.

PR #4146, opened 2020-06-18
Change sign Bayesian Wilcoxon sign rank tests

JohnnyDoorn:
And add decorrelation step to decrease autocorrelation of mcmc chains

### DIFF
--- a/JASP-Engine/JASP/R/ttestbayesianindependentsamples.R
+++ b/JASP-Engine/JASP/R/ttestbayesianindependentsamples.R
@@ -288,9 +288,10 @@ TTestBayesianIndependentSamples <- function(jaspResults, dataset, options) {
 
       }
 
-      xVals <- currentVals[1:n1]
-      yVals <- currentVals[(n1+1):(n1+n2)]
-
+      decorStepResult <- .decorrelateStepTwoSample(currentVals[1:n1], currentVals[(n1+1):(n1+n2)], oldDeltaProp, sigmaProp = 0.5)
+      xVals <- decorStepResult[[1]]
+      yVals <- decorStepResult[[2]]
+      
       gibbsResult <- .sampleGibbsTwoSampleWilcoxon(x = xVals, y = yVals, nIter = nGibbsIterations,
                                                    rscale = cauchyPriorParameter)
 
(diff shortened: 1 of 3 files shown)


vandenman:
@JohnnyDoorn can you fix the failing unit test?

image

vandenman (changes requested):
So... why is this supposed to work? Could you leave some comments in the code explaining why/ how this works, and /or reference something that explains why/ how it works? 

If I try something like this:
```r
.decorrelateStepTwoSample <- function(x, y, muProp, sigmaProp = 0.5) {
  
  thisZ <- rnorm(1, 0, sigmaProp)
  
  newX <- x + thisZ
  newY <- y + thisZ
  
  denom <- sum(dnorm(x, (muProp-thisZ) * -0.5, log = TRUE)) + sum(dnorm(y, (muProp-thisZ) * 0.5, log = TRUE))
  num <- sum(dnorm(newX, muProp * -0.5, log = TRUE)) + sum(dnorm(newY, muProp * 0.5, log = TRUE))
  
  if(runif(1) < exp(num - denom) ) {
    return(list(x = newX, y = newY, accept = TRUE))
  } else {
    return(list(x = x, y = y, accept = FALSE))
  }
  
}

n1 <- 30
n2 <- 70
n  <- n1 + n2
obs <- numeric(n)
obs[1] <- rnorm(1)
for (i in 2:n) {
  obs[i] <- obs[i - 1] + rnorm(1)
}
acf(obs) # high autocorrelation

x <- obs[1:n1]
y <- obs[(n1+1):n2]

aa <- list(accept = FALSE)
while (!aa$accept)
  aa <- .decorrelateStepTwoSample(x, y, 0)
aa$accept

new_obs <- c(aa$x, aa$y)
acf(new_obs)
```
There is minimal improvement in the autocorrelation but nothing spectacular. At the same time, I wonder about the validity of the method because you're just shifting the samples by a (random) value.

JohnnyDoorn:
It's an improvement in the autorcorrelation of the chains of the delta parameter,. Though it does not completely fix it, it's a step, and more pronounced in cases of large effect sizes. 
It's outlined in our Wilcoxon paper, where we included an additive decorrelating step documented by Liu and Sabatti(2000) and Morey et al. (2008), Jeff Rouder and JP Fox (independend of each other) recommend doing this.
Morey,  R.  D.,  Rouder,  J.  N.,  and  Speckman,  P.  L.  (2008).   A  statistical  model  for  dis-criminating between subliminal and near-liminal performance.Journal of MathematicalPsychology, 52:21–36.
Liu, S. and Sabatti, C. (2000).  Generalised Gibbs sampler and multigrid Monte Carlo forBayesian computation.Biometrika, 87:353–369

I'll fix the unit test tomorrow.

vandenman:
@JohnnyDoorn Right, can you reference that paper in a comment in the function so that someone else who sees this could look it up later? For example,
```r
# uses the method described in van Doorn, ..., (xxxx) title / journal.
```

vandenman (approved):

PR #4148, opened 2020-06-18
add messages regarding random effects grouping factors

FBartos:
fixes 2/3 of: https://github.com/jasp-stats/jasp-test-release/issues/806

1) sets an error on the empty default table if no random effects grouping factor is selected to notify the user about it.
image

2) adds a note with the random effects grouping factors specification
image

### DIFF
--- a/JASP-Engine/JASP/R/MixedModelsCommon.R
+++ b/JASP-Engine/JASP/R/MixedModelsCommon.R
@@ -486,9 +486,6 @@
     }
     ANOVAsummary$dependOn(c(dependencies, seed_dependencies, "pvalVS"))
     
-    # add message about (lack of) random effect grouping factors 
-    ANOVAsummary$addFootnote(.mmMessageREgrouping(options$randomVariables))
-    
     # some error managment for GLMMS - and oh boy, they can fail really easily
     if (type %in% c("LMM", "GLMM") && !is.null(model)) {
       if (any(attr(model, "class") %in% c("std::runtime_error", "C++Error", "error"))) {
@@ -554,6 +551,17 @@
     
     
     if (is.null(model)) {
+      if (options$dependentVariable != "" &&
+          length(options$fixedVariables) > 0 &&
+          length(options$randomVariables) == 0) {
+        ANOVAsummary$addFootnote(.mmMessageMissingRE)
+      }
+      if (type == "GLMM") {
+        if (options$family == "binomial_agg" &
+            options$dependentVariableAggregation == "") {
+          ANOVAsummary$addFootnote(.mmMessageMissingAgg)
+        }
+      }
       return()
     }
     
@@ -592,6 +600,9 @@
       
       ANOVAsummary$addRows(temp_row)
     }
+    
+    # add message about (lack of) random effect grouping factors 
+    ANOVAsummary$addFootnote(.mmMessageREgrouping(options$randomVariables))
     
     # add warning messages
     # deal with type II multiple models stuff
@@ -2049,7 +2060,7 @@
         glmm_family <<- rstanarm::neg_binomial_2(link = glmm_link)
       } else if (options$family == "betar") {
         glmm_family <<- mgcv::betar(link = glmm_link)
-      } else{
+      } else if (options$family != "binomial_agg"){
         temp_family <<- options$family
         glmm_family <<- eval(call(temp_family, glmm_link))
       }
@@ -2058,7 +2069,7 @@
       if (options$family == "binomial_agg") {
         glmm_weight <<- dataset[, .v(options$dependentVariableAggregation)]
         
-        model <- stanova::stanova_lmer(
+        model <- stanova::stanova_glmer(
           formula           = as.formula(model_formula$model_formula),
           check_contrasts   = "contr.bayes",
           data              = dataset,
@@ -2394,9 +2405,6 @@
       temp_table <- createJaspTable(title = table_name)
       STANOVAsummary[[paste0("summary_", i)]] <- temp_table
       
-      # add message about (lack of) random effects grouping factors
-      temp_table$addFootnote(.mmMessageREgrouping(options$randomVariables))
-      
       if (var_name != "Intercept" && nrow(temp_summary) > 1) {
         temp_table$addColumnInfo(name = "level",
                                  title = gettext("Level"),
@@ -2431,6 +2439,17 @@
                                type = "number")
       
       if (table_name == "Model summary") {
+        if(options$dependentVariable != "" &&
+           length(options$fixedVariables) > 0 &&
+           length(options$randomVariables) == 0) {
+          temp_table$addFootnote(.mmMessageMissingRE)
+        }
+        if (type == "BGLMM") {
+          if (options$family == "binomial_agg" &
+              options$dependentVariableAggregation == "") {
+            temp_table$addFootnote(.mmMessageMissingAgg)
+          }
+        }
         return()
       }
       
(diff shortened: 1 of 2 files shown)


AlexanderLyNL:
Sorry, I should have made this clear before, but please don't put a big red box there, because that feels like a punishment. Instead, show an empty table with that message in the footnote, and the fixed effects labels, but with ... in the output, so it's clear to the user that the analysis is waiting for something.

FBartos:
changing 1) to a note instead
image

AlexanderLyNL:
- Yes that looks great. To make it even more obvious, I'd have contGamma on the left, or what ever variables entered in "contGamma" and then dots in the output. 
- The solution with the note is great, the only problem I now have is that it's not that obvious, whereas the previous red box was too obvious
- @TimKDJ what would be good user feedback here? A highlighted "Random effects grouping factors box" <- I noticed something happening with this type of user feedback via the upgrader? Or the note text in green? Or instead of dots in the table the message "Please also specify 'random effects' " but not as a big red box?

FBartos:
So, this would be the output with the dots
image

And I also fixed when "Number of trials" was not correctly enabled when "Binomial (aggregated)" was selected.

AlexanderLyNL:
- I think that a plug-in of "contGamma" under effects (and all other variables names specified under fixed effects variables) could be helpful. 
-  The reason is that it provides the user with feedback acknowledging the user "I understand that you added contGamma" and the dot output conveys that it's waiting for something. @TimKDJ is that still the rule? Or are we going for empty rows now?
- In principle the note already informs the user the reason why it's not producing any results yet
- It would have been easier if random effects were reported, then you can have a random effects table with a column "Random effects" and under it ... stating that the analysis needs a random effect. 
- The note takes over the role of the previous point.
- Again, we need something more prominent than the note, especially for those with old eyes. But that's a different issue. 

An alternative approach would be to change the UI:
- enter all the variables into "Covariates" (instead of fixed effects): Produce standard GLM result
- Section called "Random effects" takes in subset of variables which can be selected and specified as random effects
- The idea is loosely based on the UI of linear regression, where we specify nuisance parameters under the "Model" section

EJWagenmakers:
Without having given this specific suggestion much thought (Frantisek, Lexi, let me know if you want to discuss this in a zoom meeting or something!) I think we should at least give Frantisek/Henrik's GUI a go for 0.13 -- we are in the final, final stages of getting out a new version and unless there is a blatant error I think we should avoid making structural changes. We can always change the GUI for the next version.

TimKDJ:
I think for now a footnote should suffice. But maybe it should read `This analysis requires at least one random effects grouping factor to run`. At that point everyone should be able to work out why the analysis is in a waiting state.

The long term solution is to make the analysis work without the grouping factor as described in [this issue](https://github.com/jasp-stats/INTERNAL-jasp/issues/932); it's obviously better to make the analysis do what a user would reasonably expect it to :)

FBartos:
So, I changed the message as suggested (it's shown only if the Random-effects grouping factor is missing):
image

and added a message to address this issue (https://github.com/jasp-stats/jasp-test-release/issues/836)- it's not clear that one needs to specify the "Number of trials" variable in order to run the analysis with "Binomial (aggregated)" family 
image

I also fixed one issue that would prevent the BGLMM with "Binomial (aggregated)" family to run

JASP-Engine/JASP/R/MixedModelsCommon.R
```diff
@@ -552,7 +551,17 @@
...
     
     if (is.null(model)) {
-      ANOVAsummary$setExpectedSize(1)
+      if (options$dependentVariable != "" &&
+          length(options$fixedVariables) > 0 &&
+          length(options$randomVariables) == 0) {
+        ANOVAsummary$addFootnote(.mmMessageMissingRE)
+      }
+      if (type == "GLMM") {
+        if (options$family == "binomial_agg" &
```
TimKDJ: `&&`

JASP-Engine/JASP/R/MixedModelsCommon.R
```diff
@@ -2904,6 +2927,18 @@
...
 }
+.mmMessageREgrouping    <- function(RE_grouping_factors) {
+  gettextf(
+    "The following %s used %s random effects grouping %s: %s.",
+    ifelse(length(RE_grouping_factors) < 2, "variable is", "variables are"),
+    ifelse(length(RE_grouping_factors) < 2, "as a", "as"),
+    ifelse(length(RE_grouping_factors) < 2, "factor", "factors"),
+    ifelse(length(RE_grouping_factors) == 0, "...", paste0("'", RE_grouping_factors, "'", collapse = ", "))
+  )
+}
```
TimKDJ: Use `sprintf(ngettext())` for this. Aside from the strings not being in `gettext()` the structure of this sentence might need to be entirely different in another language.

TimKDJ (changes requested):

JASP-Engine/JASP/R/MixedModelsCommon.R
FBartos: Yeah, thanks. I keep forgetting those.

What do you advise to do about it then?

FBartos: I could maybe create two complete sentences for each case

TimKDJ: Something like:
```
sprintf(ngettext(length(RE_grouping_factors) < 2,
                 "The following variable is used as random effects grouping factor: %s.",
                 "The following variables are used as random effects grouping factors: %s."),
        ifelse(length(RE_grouping_factors) == 0, "...", paste0("'", RE_grouping_factors, "'", collapse = ", ")))
```

untested though

JASP-Engine/JASP/R/MixedModelsCommon.R
```diff
@@ -2986,32 +2968,25 @@
...
-  if (value < 2) {
-    return(gettextf("%i observation was removed due to missing values.",
-                    value))
-  } else{
-    return(gettextf("%i observations were removed due to missing values.",
-                    value))
-  }
+  sprintf(
+    ngettext(
+      value < 2,
```
TimKDJ: Unless you specifically want to group `0` and `1` together you should use `ngettext(length(value), "..", "..")` (so no boolean, but an integer as argument to `ngettext()`)

`0 observation was removed due to missing values.` is not correct.

This goes for quite a few of these changed lines

TimKDJ (changes requested):

JASP-Engine/JASP/R/MixedModelsCommon.R
FBartos: the 's' is actually correct at the end - there might be more missing values within one observation. We do not check those.

These messages aren't called when the value == 0

TimKDJ: We've solved this over Slack, but:

`0 observation was removed` is correct not though :)
These are your sentences:
```
 "%i observation was removed due to missing values.",
 "%i observations were removed due to missing values."
```

FBartos: 0 observation was removed cannot happen. The function is called only if `value > 0`

TimKDJ: I know, but the English is not correct. It was in response to this:

> the 's' is actually correct at the end - there might be more missing values within one observation. We do not check those.

FBartos: oh, my English is probably not good enough to understand the problem with my English :D

TimKDJ (approved):

PR #4151, opened 2020-06-22
T-tests: show footnote when the test value is 0

vandenman:
Fixes jasp-stats/jasp-test-release/issues/842

For consistency, I also changed the Bayesian one-sample so that it shows a footnote when the test value is 0.

@TimKDJ I considered changing the footnote to have more if statements so that `"greater"` and `"smaller"` wouldn't be substituted as you recommended in https://github.com/jasp-stats/jasp-desktop/pull/4119#discussion_r438700524. I'd also like to refactor the footnote into one function for both analyses. 

However, since we're close to a release I didn't do either of that so that the changes are as small as possible. Let me know if you do want that though.


TimKDJ (approved):
Definitely good to clean this table code up a bit, but I agree it's better to do post-release

PR #4152, opened 2020-06-22
Resizing or rewriting images now set pkg modified

JorisGoosen:
- Also updates width&height of plot in results. Otherwise they get squished or stretched after reloading
- Fixes https://github.com/jasp-stats/jasp-test-release/issues/824

### DIFF
--- a/JASP-Desktop/analysis/analysis.h
+++ b/JASP-Desktop/analysis/analysis.h
@@ -170,6 +170,7 @@ signals:
 	void				imageEditedSignal(		Analysis * analysis);
 	void				resultsChangedSignal(	Analysis * analysis);
 	void				userDataChangedSignal(	Analysis * analysis);
+	void				somethingModified(); //Possibly not emitted everywhere. But it should be ^^ (well, eventually, but take into account loading stuff etc, thats why it wasnt added everywhere it should be in the first place)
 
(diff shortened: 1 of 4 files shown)


JASP-Desktop/analysis/analysis.h
```diff
@@ -170,6 +170,7 @@ class Analysis : public QObject
 	void				imageEditedSignal(		Analysis * analysis);
 	void				resultsChangedSignal(	Analysis * analysis);
 	void				userDataChangedSignal(	Analysis * analysis);
+	void				somethingModified(); //Possibly not emitted everywhere. But it should be ^^ (well, eventually, but take into account loading stuff etc, thats why it wasnt added everywhere it should be in the first place)
```
boutinb: I don't think that a new signal should be added.
I am not for adding a signal for each purpose. The name of a signal should only described what has changed, and it should be connected to all needed slots, eg. analysis.titleChanged -> analyses.somethingChanged. It has the advantage that if titleChanged signal is emitted somewhere else, then you don't have to remember that you have to call also the analysis somethingChanged signal.

JorisGoosen: I guess that makes sense.

JorisGoosen: Ok, so thats fixed.

boutinb (approved):

PR #4154, opened 2020-06-22
Fix anova labels in 2x2x2 setup

### LINKED ISSUE #797: Incorrect Values in 2x2x2 ANCOVA Table
When running a 2x2x2 ANCOVA in JASP 0.12.2 on a mid-2017 MacBook Pro 13 inch, the ANCOVA table reports incorrect values for the CV and all interaction factors. Essentially, the values for these rows are swapped. The reported values for IV1*IV2 are replaced with those for IV2*IV3; the values for IV1*IV3 are replaced with those from IV1*IV2; the values for IV2*IV3 are replaced with those from the CV; the values for IV1*IV2*IV3 are replaced with those for IV1*IV3; and finally, the CV values are replaced with those from IV1*IV2*IV3. 

This issue is not seen in version 0.12.2 when running a 2x2 ANCOVA. It is also not seen when running a 2x2 or 2x2x2 ANCOVA in an earlier version of JASP (0.11.1). 

It is possible that 0.12.2 is actually reporting the correct values for the CV and interaction terms, and it is earlier versions that were reporting these values incorrectly. However, I did run a linear regression using all terms (including interaction terms) and found that on both versions of JASP that I have installed (0.12.2 and 0.11.1), the interaction and CV term p-values line up identically with those reported in the ANCOVA table from version 0.11.1, not 0.12.2. Hence, I believe that version 0.12.2 is reporting the 2x2x2 ANCOVA results incorrectly.

JohnnyDoorn:
Fix jasp-stats/jasp-issues#797
now the cases column just stays true to the output of car::Anova instead of attempting to reshuffle

### DIFF
--- a/JASP-Engine/JASP/R/ancova.R
+++ b/JASP-Engine/JASP/R/ancova.R
@@ -355,9 +355,7 @@ Ancova <- function(jaspResults, dataset = NULL, options) {
     
   }
 
-  # Make sure that the order of the result is same order as reordered modelterms
-  result <- result[.mapAnovaTermsToTerms(rownames(result), c(termsBase64, "Residuals")), ]
-  result[['cases']] <- c(termsNormal, "Residuals")
+  result[['cases']] <- gsub(x = rownames(result), pattern = ":", " \u273B ")
   result <- as.data.frame(result)
   result[['.isNewGroup']] <- c(TRUE, rep(FALSE, nrow(result)-2), TRUE)
   if (length(options$covariates) > 0)
--- a/JASP-Engine/JASP/R/ancova.R
+++ b/JASP-Engine/JASP/R/ancova.R
@@ -354,8 +354,10 @@ Ancova <- function(jaspResults, dataset = NULL, options) {
     result['SSt'] <- sum(result["Sum Sq"], na.rm = TRUE)
     
   }
-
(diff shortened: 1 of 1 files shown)


Kucharssim:
@JohnnyDoorn Now I get the variable names as `JaspColumn_.x._Encoded` (but the reshuffling seems fixed)
Screenshot 2020-06-26 at 10 35 55

JohnnyDoorn:
@Kucharssim  Ah yes I hadn't considered that - now it should be ok!

Kucharssim (approved):
Now it's working again 🙂 

@JohnnyDoorn just a question: why did not we catch this in a unit test? Did it only happen in specifically 2x2x2 design or more generally if you have more than two-way interaction? If the latter then I would perhaps suggest creating additional unit test...

JohnnyDoorn:
Yes, I think this only happens in cases with higher order interactions, and the unit tests only go up to two order interactions I believe.

Kucharssim:
Can we maybe create a unit test for it then? It's strange that we do not have a higher order interaction example even in the data library, but it would be nice... If you don't have time, let me know, I can make a unit test for this in a separate PR after merging this

JohnnyDoorn:
Yeah let;s do this in a separate PR, after 0.13

PR #4156, opened 2020-06-23
Network analysis: fix jasp-test-release/issues/853, 854, & 855

vandenman:
Fixes jasp-stats/jasp-test-release/issues/853
Fixes jasp-stats/jasp-test-release/issues/854
Fixes jasp-stats/jasp-test-release/issues/855

- 853 was a regression from changing the way the input type for mgm is specified. Some old code wasn't adjusted properly (lines 710-728), for example, `options[["mgmVariableType"]]` doesn't exist anymore but I guess I didn't change that.

- 854 is an error where `bootnet` catches an error internally and returns a faulty ggplot2 object (a ggplot object that throws an error when you plot it..). I now check whether the returned object is faulty or not and attribute that to bootnet. This functions are quite unstable, hopefully updating `bootnet` to 1.4.3 fixes this (we're using 1.2.4).

- 855 is an error due to bad data. This is hard to catch beforehand because it happends during cross validation. So there is some bootstrapped dataset that has no variance, even though the overall variable does have variance. I now add something to the error message for this particular case.

### DIFF
--- a/JASP-Engine/JASP/R/networkanalysis.R
+++ b/JASP-Engine/JASP/R/networkanalysis.R
@@ -50,6 +50,7 @@ NetworkAnalysis <- function(jaspResults, dataset, options) {
   if (options[["missingValues"]] == "listwise")
     exclude <- vars2read
   dataset <- .readDataSetToEnd(columns.as.numeric = vars2read, exclude.na.listwise = exclude)
+  # dataset <- .readDataSetToEnd(columns = vars2read, exclude.na.listwise = exclude) # jasptools need this, JASP the line above
 
   if (options[["groupingVariable"]] == "") { # one network
     dataset <- list(dataset) # for compatability with the split behaviour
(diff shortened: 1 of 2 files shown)


JASP-Engine/JASP/R/networkanalysis.R
```diff
@@ -1528,6 +1544,26 @@ NetworkAnalysis <- function(jaspResults, dataset, options) {
...
+  # NOTE: this fails if glmnet decides to translate their error messages one day.
+  # Unfortunately, these error appear for particular subsets of the data (from cross validation),
+  # so it's difficult to traceable particular errors to the complete data.
+
+  errmsg <- .extractErrorMessage(e)
+  # possibly add other checks here in the future
+  dataIssue <- startsWith(errmsg, "y is constant") || endsWith(errmsg, "standardization step")
+
+  ans <- if (dataIssue) {
+    gettextf("bootnet crashed with the following error message:\n%s\n\nPlease check if there are variables in the network with little variance or few distinct observations.", errmsg)
```
Kucharssim:
```suggestion
    gettextf("bootnet crashed with the following error message:\n%s. <ul><li>Please check if there are variables in the network with little variance or few distinct observations.</li></ul>", errmsg)
```

Kucharssim: Maybe this is more pleasing to the eye

JASP-Engine/JASP/R/networkanalysis.R
```diff
@@ -1528,6 +1544,26 @@ NetworkAnalysis <- function(jaspResults, dataset, options) {
 
 }
 
+.networkAnalysisCheckKnownErrors <- function(e) {
+
+  # NOTE: this fails if glmnet decides to translate their error messages one day.
```
Kucharssim, line 1549: Well it would be nice to handle this inside of bootnet itself... That does not mean you are supposed to do anything about it, I am just complaining that you have to do this matching of strings.

Kucharssim (changes requested):
Just a small change to the error message formatting, otherwise it's good to be merged.

JASP-Engine/JASP/R/networkanalysis.R
vandenman: I fully agree. R has built-in support for custom errors, so you could throw "InputError" or "DataError". It's just that almost no R packages use that... So to understand what goes wrong we have to resort to string parsing :(

Kucharssim (approved):

PR #4157, opened 2020-06-23
Bayesian T-test - Enable sequential and robustness analysis only for default priors

Kucharssim:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/845

Pls check if the solution using `property alias` is the way how this should be done.

### DIFF
--- a/JASP-Desktop/components/JASP/Widgets/SubjectivePriors.qml
+++ b/JASP-Desktop/components/JASP/Widgets/SubjectivePriors.qml
@@ -26,6 +26,7 @@ Section
 {
 	title: qsTr("Prior")
     property alias informedPriorsEnabled: informedPriors.enabled
+	property alias defaultPriorsChecked: defaultPriors.checked
 
 	RadioButtonGroup
 	{
@@ -46,6 +47,7 @@ Section
 			enabled: standardized.checked
 			RadioButton
 			{
+				id: defaultPriors
 				label: qsTr("Default"); name: "default"; checked: true
 				RadioButtonGroup
 				{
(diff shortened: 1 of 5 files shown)


vandenman:
I can still get this:

image

First, check the robustness plot and then change the prior to informed.

If you make the following 3 changes that should also be gone.

1. https://github.com/jasp-stats/jasp-desktop/blob/6752f2ed3e14ad3a55d7c726a70fa069c49d6973/JASP-Engine/JASP/R/commonbayesianttest.R#L844 
adjust the if statement to
```r
if (options[["plotBayesFactorRobustness"]] && options[["effectSizeStandardized"]] == "default") {
```
2. https://github.com/jasp-stats/jasp-desktop/blob/6752f2ed3e14ad3a55d7c726a70fa069c49d6973/JASP-Engine/JASP/R/commonbayesianttest.R#L865
same as 1.

3. https://github.com/jasp-stats/jasp-desktop/blob/6752f2ed3e14ad3a55d7c726a70fa069c49d6973/JASP-Engine/JASP/R/commonbayesianttest.R#L749

if you insert 

```r
if (options[["effectSizeStandardized"]] != "default")
    opts <- opts[1L]
```
then I *think* it should work (I only tested it with jasptools though).

I don't know if it's better to solve this in QML or to solve this in R. In QML you could uncheck the robustness and sequential plots upon checking informed but no analysis does that except maybe Audit.

Kucharssim:
Ah shoot... Did not think of this 😓  

I'll go for skipping this in R rather than unchecking the boxes in qml indeed. I feel it's not very nice to change user input under their hainds without notifying them.

Thanks @vandenman for basically writing the code that fixes it, it should be fine now.

vandenman (approved):

PR #4160, opened 2020-06-23
(B)RM ANOVA fix test issues

vandenman:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/837
Fixes https://github.com/jasp-stats/jasp-test-release/issues/838
Fixes https://github.com/jasp-stats/jasp-test-release/issues/839
Fixes https://github.com/jasp-stats/jasp-test-release/issues/856

also adds a unit test for Single Model Inference in the Bayesian Repeated Measures.

### DIFF
--- a/JASP-Engine/JASP/R/commonAnovaBayesian.R
+++ b/JASP-Engine/JASP/R/commonAnovaBayesian.R
@@ -79,26 +79,28 @@
     hasDV  <- !any(options$repeatedMeasuresCells == "")
     hasIV  <- any(lengths(options[c("betweenSubjectFactors", "covariates")]) != 0)
     fixed  <- options$betweenSubjectFactors
-    covariates <- c(options$covariates)
+    covariates <- unlist(options$covariates)
     noVariables <- !hasDV
+    target <- c(covariates, fixed)
 
-    if (length(covariates) > 0L || length(fixed) > 0L) {
+    if (length(target) > 0L) {
       .hasErrors(
         dataset = dataset,
         type    = c("infinity", "observations", "variance", "factorLevels", "duplicateColumns"),
-        infinity.target     = covariates,
+        infinity.target     = target,
         variance.target     = covariates,
-        observations.target = covariates,
+        observations.target = target,
         observations.amount = paste("<", length(options$modelTerms) + 1),
         factorLevels.target = fixed,
         factorLevels.amount = " < 2",
-        duplicateColumns.target = c(covariates, fixed),
+        duplicateColumns.target = target,
         exitAnalysisIfErrors = TRUE
       )
     }
 
     if (!noVariables) {
 
+      # messages from commonmessages.R
       customChecks <- list(
         infinity = function() {
           x <- dataset[, .BANOVAdependentName]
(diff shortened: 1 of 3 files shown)


Kucharssim (commented):
This looks fine, or at least the code looks ok and I cannot reproduce the issues that are linked. I just don't know whether https://github.com/jasp-stats/jasp-test-release/issues/839 is indeed fixed as I don't know how to obtain the error in the first place.

Regarding https://github.com/jasp-stats/jasp-test-release/issues/856, I don't understand how this issue is fixed with this PR, I don't see any changes to files that are used in (classical) RM ANOVA. However, I do not get the error Koen reported so I guess it's fixed somehow 😅 What I get now (zipped jasp file attached) is that the analysis does not throw an error (except for Sphericity test), and proceeds reporting uninformative results. In my opinion this *should* technically error out (with something informative), as you have a single observation per level of `Celebrity` between subjects factor. The sphericity test should also error out, but not because "there are only two levels of the RM factor", but because not enough observations.

[Bush2.zip](https://github.com/jasp-stats/jasp-desktop/files/4836411/Bush2.zip)

To summarise I am uncertain about the fix to the last issue, otherwise I think it's ok to be merged. @vandenman if you think I am just confused, disregard this point. I don't have a strong opinion about this, but I think in this case the output of the analysis is so uninformative that it is as good as just telling the user to think about the analysis design, instead of letting them figure it out by themselves after seeing bunch of `NAs` in the tables.

vandenman:
#### jasp-stats/jasp-test-release#839
I don't know how to obtain the error either. Nevertheless, that particular if statement can no longer throw an error:
```r
# isTRUE should handle a state issue, see https://github.com/jasp-stats/jasp-test-release/issues/839
if (isTRUE(model[["analysisType"]] != "RM-ANOVA" && options[["dependent"]] != "")) {
```
`isTRUE` should convert any missing value to `FALSE`.


#### https://github.com/jasp-stats/jasp-test-release/issues/856

> I don't understand how this issue is fixed with this PR, I don't see any changes to files that are used in (classical) RM ANOVA.

The frequentist repeated measures uses the same error handling function as the Bayesian repeated measures: https://github.com/jasp-stats/jasp-desktop/blob/d387e33521b71e61d51367771e485ba6c1084919/JASP-Engine/JASP/R/anovarepeatedmeasures.R#L36

As such, jasp-stats/jasp-test-release#837 and jasp-stats/jasp-test-release#856 have the same cause.

>  In my opinion this should technically error out (with something informative), as you have a single observation per level of Celebrity between subjects factor. The sphericity test should also error out, but not because "there are only two levels of the RM factor", but because not enough observations.

I agree, I'll see if I can fix this later today.

vandenman:
@Kucharssim okay so the issue was that [first](https://github.com/jasp-stats/jasp-desktop/pull/4160/commits/d8f48195b556e927cdf1b06a8b089aa90002985c) I split up the check for duplicate columns into one check with the covariates and between-subject factors and another check for just the subject variable. As a consequence this scenario:

> this should technically error out (with something informative), as you have a single observation per level of Celebrity between subjects factor. 

Was no longer caught. I've fixed that and now you get a somewhat informative error message (basically the same message as before). I've also added a unit test that checks whether this dataset leads to a validationError.

Kucharssim:
Hm, this is weird, on my JASP build the analysis still does not catch the error, but run through jasptools it does. Any idea what could cause this?

vandenman:
> Any idea what could cause this?

I forgot to call `.v` so within JASP, `target` would not exist in the column names of the dataset. jasptools doesn't use encoding anymore so there it doesn't matter if you use `.v` or not.

Kucharssim (approved):
Ok, looks good to go.

TimKDJ (approved):
Everything seems fixed. Nice!

PR #4163, opened 2020-06-23
Fixes multiple errors within the Mixed Models analysis

FBartos:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/862
Fixes https://github.com/jasp-stats/jasp-test-release/issues/863
Fixes https://github.com/jasp-stats/jasp-test-release/issues/864
Fixes https://github.com/jasp-stats/jasp-test-release/issues/865
Fixes https://github.com/jasp-stats/jasp-test-release/issues/866
Fixes https://github.com/jasp-stats/jasp-test-release/issues/892
Fixes https://github.com/jasp-stats/jasp-test-release/issues/893

Fixes https://github.com/jasp-stats/jasp-test-release/issues/875
Fixes https://github.com/jasp-stats/jasp-test-release/issues/878
Fixes https://github.com/jasp-stats/jasp-test-release/issues/879
Fixes https://github.com/jasp-stats/jasp-test-release/issues/881
Fixes https://github.com/jasp-stats/jasp-test-release/issues/895

### DIFF
--- a/JASP-Engine/JASP/R/MixedModelsCommon.R
+++ b/JASP-Engine/JASP/R/MixedModelsCommon.R
@@ -2063,7 +2063,7 @@
     
     JASP:::.setSeedJASP(options)
     if (type == "BLMM") {
-      model <- stanova::stanova_lmer(
+      model <- tryCatch(stanova::stanova_lmer(
         formula           = as.formula(model_formula$model_formula),
         check_contrasts   = "contr.bayes",
         data              = dataset,
@@ -2072,7 +2072,7 @@
         warmup            = options$warmup,
         adapt_delta       = options$adapt_delta,
         control           = list(max_treedepth = options$max_treedepth)
-      )
+      ), error = function(e)e)
       
     } else if (type == "BGLMM") {
       # needs to be evaluated in the global environment
@@ -2090,7 +2090,7 @@
       if (options$family == "binomial_agg") {
         glmm_weight <<- dataset[, .v(options$dependentVariableAggregation)]
         
-        model <- stanova::stanova_glmer(
+        model <- tryCatch(stanova::stanova_glmer(
           formula           = as.formula(model_formula$model_formula),
           check_contrasts   = "contr.bayes",
           data              = dataset,
@@ -2101,10 +2101,10 @@
           control           = list(max_treedepth = options$max_treedepth),
           weights           = glmm_weight,
           family            = eval(call("binomial", glmm_link))
-        )
+        ), error = function(e)e)
         
       } else{
-        model <- stanova::stanova_glmer(
+        model <- tryCatch(stanova::stanova_glmer(
           formula           = as.formula(model_formula$model_formula),
           check_contrasts   = "contr.bayes",
           data              = dataset,
@@ -2114,7 +2114,7 @@
           adapt_delta       = options$adapt_delta,
           control           = list(max_treedepth = options$max_treedepth),
           family            = glmm_family
-        )
+        ), error = function(e)e)
         
       }
       
@@ -2370,9 +2370,9 @@
   function(jaspResults, dataset, options, type = "BLMM") {
     if (!is.null(jaspResults[["STANOVAsummary"]]))
       return()
-    
+
     model <- jaspResults[["mmModel"]]$object$model
-    if (!is.null(model)) {
+    if (!is.null(model) && !class(jaspResults[["mmModel"]]$object$model) %in% c("simpleError", "error")) {
       model_summary <-
         summary(
           model,
@@ -2470,6 +2470,10 @@
               options$dependentVariableAggregation == "") {
             temp_table$addFootnote(.mmMessageMissingAgg)
           }
+        }
+        
+        if(class(jaspResults[["mmModel"]]$object$model) %in% c("simpleError", "error")) {
+          STANOVAsummary$setError("The model could not be estimated. Please, check the options and dataset for errors.")
         }
         return()
       }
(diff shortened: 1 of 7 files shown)


vandenman:
@FBartos if you start your message with "Fixes \<link to issue\>" then the linked issues will appear on the right under linked issues and can be automatically closed when merging this PR.

vandenman:
So all the errors linked in the issues are fixed, great! 

I did notice one more issue though:
image

The variable names (e.g., `JaspColumn_.15._Encoded`) are not decoded before they are shown in the error message. To reproduce:

1. `debString` & `contGamma` to fixed effects variables.
2. `conBinom` as random effects grouping factors.
3. Family is `Gaussian`.
4. dependent is `contNormal`.

FBartos:
thanks Don, I will be adding the links now.

also, fixing this issue
image

vandenman:
@FBartos let me know when this is ready to review!

FBartos:
I think that this should fix all that was fixable within https://github.com/jasp-stats/jasp-test-release/issues/768, https://github.com/jasp-stats/jasp-test-release/issues/769, https://github.com/jasp-stats/jasp-test-release/issues/770, and https://github.com/jasp-stats/jasp-test-release/issues/771 as of 24. 06 (20:45). Could you review it again please? @vandenman

JASP-Engine/JASP/R/MixedModelsCommon.R
```diff
@@ -491,18 +492,27 @@
...
+          ANOVAsummary$setError(.unv(model$message))
         }
         
-        jaspResults[["ANOVAsummary"]] <- ANOVAsummary
-        jaspResults[["mmModel"]]      <- NULL
+        
+        hasERROR <- createJaspState()
+        hasERROR$object <- TRUE
+        hasERROR$dependOn(c(dependencies, seed_dependencies))
+        jaspResults[["hasERROR"]] <- hasERROR
```
vandenman: Do we need this? Can't we just do `ANOVAsummary$dependOn(c(dependencies, seed_dependencies))` before the return and `ANOVAsummary$dependOn(c(dependencies, seed_dependencies, "pvalVS"))` after the if statement?

JASP-Engine/JASP/R/MixedModelsCommon.R
```diff
@@ -2877,6 +2906,18 @@
 }
 
 
+.mmCustomChecks <- list(
+ collinCheck = function(dataset){
+   cor_mat       <- cor(apply(dataset,2,as.numeric))
+   diag(cor_mat) <- 0
+   cor_mat[lower.tri(cor_mat)] <- 0
+   if(any(1 - abs(cor_mat) < 1e-5)){
+     var_ind   <- which(abs(cor_mat) == 1, arr.ind = TRUE)
```
vandenman:
```suggestion
   nearOne <- 1 - abs(cor_mat) < sqrt(.Machine$double.eps)
   if(any(nearOne)){
     var_ind   <- which(nearOne, arr.ind = TRUE)
```
The issue is that you check with a tolerance in the if statement but without a tolerance in `which`. 

Using `sqrt(.Machine$double.eps)` (~1.490116e-08) should make this less machine specific, feel free to use a different root though (e.g., `.Machine$double.eps^(1/3)`).

vandenman (changes requested):
Two errors remain, the others seem fixed.

Error message for equal variables doesn't use the variable names (see comment in the code):
image


There's still an error if warmup == iterations:

image

JASP-Engine/JASP/R/MixedModelsCommon.R
FBartos: It's a little akward but I cannot set the `jaspResults[["mmModel"]]` to `NULL` in order to skip the block with summary and plotting functions because the model would be reffited, and it would skip the main summary that was already produced (inside of which it's set to NULL, I cannot do that before since I need to keep the error message). This object keeps track on whether the model failed and allows to skip the summary and plotting functions. I could alternatively create a designated function that would set the model to null after creating the summary table and it wouldn't be dependent on it. It might be a better solution.

FBartos: however, it would resulted on refiting the failed model every time, even when requested

vandenman: > This object keeps track on whether the model failed

Okay, but if you set an error on any jaspResults object you can check for an error using `$getError()`.

FBartos: oh, great. That's much better solution. Thanks

FBartos:
Thanks for the review Don. I found that I mistakenly changed the default value instead of the min value after I rebased on the updated qml interface. That's fixed now. I incorporated the suggested machine tolerance and restructured the check for errors as suggested.

vandenman:
Changes to the code look fine!

Two more bugs:

#### 1
image

1. contBinom to fixed
2. facGender to Random effects
3. `Adapt delta` to 0.0
4. contNormal to dependent.

perhaps `Adapt delta` needs a larger minimum.


#### 2
You can't set `Iterations` below `Warmup + 2`, but it does work the other way around.

Set:
1. `Warmup` = 1
2. `Iterations` = 3
3. `Warmup` = 100

The interface adds a red box around `Iterations` but still the analysis crashes. Error message was the same as in the screenshot above.

FBartos:
I'm not sure whether we can prevent (2) from withing the qml. Are you OK if I add an error check next to the datachecks and exit the analysis?

FBartos:
Also should
fix https://github.com/jasp-stats/jasp-test-release/issues/884
but it's not really testable.

That should be all the changes

vandenman (approved):

PR #4172, opened 2020-06-25
Add High Order Interaction Check in Mixed Models

boutinb:
Fixes jasp-stats/jasp-test-release#894
Arrange the layout of the random slopes by using a Variables List.
Rename some functions to understand better what they do


FBartos (approved):
it behaves as expected

PR #4173, opened 2020-06-25
rewritten gather-r-package-info.R to take info from sources instead of from CRAN.

JorisGoosen:
- also made the script less monolithic and hopefully more understandable.
- Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/926

### DIFF
--- a/Tools/flatpak/gather-r-package-info.R
+++ b/Tools/flatpak/gather-r-package-info.R
@@ -1,16 +1,32 @@
-# very loosely based on https://github.com/dewittpe/R-install-dependencies
 options(warn=1) #print warnings as they occur
-options(nCpus=8)
+options(nCpus=parallel::detectCores())
 
-
-
-expEnv     <- new.env(hash = TRUE, parent = parent.frame())
 CRAN       <- "https://cran.rstudio.com/"
 	#"https://cran.r-project.org/" #
 
-giveOrderedDependencies <- function()
+
+#define vars
+expEnv          <- NULL
+downloaded      <- NULL
+pkgDeps         <- NULL
+available_pkgs  <- NULL
+destDir         <- NULL
+destTmpDir      <- NULL
+pkgs            <- NULL
+base_pkgs       <- NULL
+specials        <- NULL
+
+initialize <- function()
 {
-  available_pkgs  <- available.packages(repos = CRAN)
+  #Create necessary environments (only hash available)
+  expEnv     <<- new.env(hash = TRUE, parent = parent.frame())
+  downloaded <<- new.env(hash = TRUE, parent = parent.frame())
+  pkgDeps    <<- new.env(hash = TRUE, parent = parent.frame())
+
+  #what packages are available? Do we care?
+  available_pkgs  <<- available.packages(repos = CRAN)
+  destDir         <<- "pkg-source-files/"
+  destTmpDir      <<- "pkg-source-tmp-files/" 
 
   # Define the base packages.  These are packages which come with R upon install
   # of R.  These packages include: "base", "compiler", "datasets", "graphics",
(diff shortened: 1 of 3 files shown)


JorisGoosen:
@TimKDJ & @vandenman so this code does not necessarily need to be in jasp-desktop before releasing or anything. I used it to generate a new list of R-pkgs to build in order and thank god it works.

But as I am not a very experienced R programmer I would like to hear if you guys see anything stupid in my code, or you know, general improvements.

But only if you have the time of course.

Tools/flatpak/gather-r-package-info.R
```diff
@@ -1,16 +1,32 @@
-# very loosely based on https://github.com/dewittpe/R-install-dependencies
 options(warn=1) #print warnings as they occur
 options(nCpus=8)
```
vandenman:
```suggestion
options(nCpus=parallel::detectCores())
```

vandenman (changes requested):
Before I continue, the script contains a ton of global variables and functions with side-effects. Do we want that? It doesn't matter that much, but that choice does determine what I provide feedback on.

Big issues so far:

- `installRequiredPackages` doesn't call `initialize` so `pkgs` doesn't exist.
- relative paths like `source("../../JASP-Engine/JASP/R/packagecheck.R", local=TRUE)` can be difficult to figure out. I'd restructure those files to be in the same directory and I'd add a check to see if those files exist.

Also, I see this script as the only way for Linux users (and masochists) to install all required R packages manually. That does set some assumptions in terms of user-friendliness IMO.

JorisGoosen:
> Before I continue, the script contains a ton of global variables and functions with side-effects. Do we want that? It doesn't matter that much, but that choice does determine what I provide feedback on.


Yeah, I mostly intended the script as a one-off to generate the json files for flatpak or, and I kind of neglected to look at that part this time, also to install pkgs on linux. Because the only way to get a proper hashmap was to use environments and Im a bit hazy on how to clean them up I did it like this. Closing R entirely is a good way to clean up I figured.

And I didn't really know what would be a good datastructure to store all of that stuff in. What would you suggest?

> Big issues so far:
> 
>     * `installRequiredPackages` doesn't call `initialize` so `pkgs` doesn't exist.

Oops, but, I didnt look at it yet, because I just wanted to get flathub to compile. I will check that code later today and make it work.

>     * relative paths like `source("../../JASP-Engine/JASP/R/packagecheck.R", local=TRUE)` can be difficult to figure out. I'd restructure those files to be in the same directory and I'd add a check to see if those files exist.

packagecheck.R is part of the JASP R package and is used for the check on startup. I really want that to be the exact same file as the source for info. So Im afraid we're stuck with this being there until JASP is a proper pkg.

> Also, I see this script as the only way for Linux users (and masochists) to install all required R packages manually. That does set some assumptions in terms of user-friendliness IMO.

Ah yeah, I hadn't really looked at it that way. I guess the fact that running it through [installExpectedPackages.R](https://github.com/jasp-stats/jasp-desktop/blob/stable/Tools/installExpectedPackages.R) makes it slightly more userfriendly?
As in, what more would someone want? You only need those pkgs if you build on linux which means you have these files right?

JorisGoosen:
I suppose I could turn it into an R6 object with a constructor and destructor?

vandenman:
> I suppose I could turn it into an R6 object with a constructor and destructor?

That would work and would probably be the cleanest solution. At the same time, it might be overkill.

vandenman:
Ideally, the dependencies for the script are also as minimal as possible, I think we should be able to do this with just base R and `remotes`.

JorisGoosen:
Well, I also use `stringr` and nothing else. I think R6 would be overkill. Which is why I set it up like it is now. The globals and basically just running the script once.

Tools/flatpak/gather-r-package-info.R
```diff
@@ -20,63 +36,187 @@ giveOrderedDependencies <- function()
...
+
+  #make sure the source pkg is in the pkg-source folder:
+  pkgFile <- downloadPkg(pkgName)
+
+  deps    <- remotes::dev_package_deps(pkgFile, repos = NULL, type = "source")$package
+  deps    <- deps[!(deps %in% base_pkgs)] #remove base pkgs
+
+  pkgDeps[[pkgName]] <- deps
+
+  padding <- paste(rep("", 30 - stringr::str_length(pkgName)), sep=' ', collapse=' ')
```
vandenman:
```suggestion
  padding <- strrep(" ", 29 - nchar(pkgName))
```
- `strrep` rather than `paste(rep("", ...))`. It does need a non-empty string to rep, so 30 becomes 29.
- using `base::nchar` rather than `stringr::str_length` means one dependency less.

JorisGoosen:
So I will just merge this as I think @TimKDJ won't have time to look at it before vacation anyway. Thanks for the feedback Don, I think I processed it all.

PR #4175, opened 2020-06-25
Removed all the "f" added at the end of optionnames in ReliabilityFreq

JorisGoosen:
- Why were they there anyway?
- Fixes https://github.com/jasp-stats/jasp-test-release/issues/890 (upgrader for reliability)

### DIFF
--- a/JASP-Desktop/modules/upgrader/upgrader.cpp
+++ b/JASP-Desktop/modules/upgrader/upgrader.cpp
@@ -106,6 +106,11 @@ bool Upgrader::upgradeAnalysisData(Json::Value & analysis, UpgradeMsgs & msgs) c
 	try
 	{
 		_upgradeOptionsFromJaspFile(analysis, msgs, stepsTaken);
+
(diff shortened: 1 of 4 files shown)


TimKDJ:
@juliuspf any reason in particular the added `f`'s were there?

TimKDJ:
@JorisGoosen: Even if it is not necessary it might not be nice to make these changes, because all the references to these option names also need to be changed in the R files

JorisGoosen:
Ooh yeah right... Its just because in the bayesian QML they were all without the "f"'s.
I will just also change them in the R code then.

They just don't make sense.

Unless @juliuspf has a really good reason for them?

juliuspfadt:
Well, embarrassingly enough I had the f's in there to avoid name clashes with the Bayesian file, because in a previous version both analysis were part of one R-file. There is no more meaningful reason, you can easily remove them.

TimKDJ (approved):

PR #4176, opened 2020-06-25
Getting an engine reply for an analysis while closing JASP crashes it

JorisGoosen:
- This because analysisform tries to set something on a dangling pointer...
- Sovled by giving setting the relevant Analysis as qobject::parent to AnalysisForm, this way it gets destroyed at the same times and connections will be reset, etc!
- Fixes https://github.com/jasp-stats/jasp-test-release/issues/826 (Closing JASP while bootstrapping makes it crash and not kill the engines)

### DIFF
--- a/JASP-Desktop/analysis/analysisform.cpp
+++ b/JASP-Desktop/analysis/analysisform.cpp
@@ -732,6 +732,7 @@ void AnalysisForm::_formCompletedHandler()
 	if (!analysisVariant.isNull())
 	{
 		_analysis	= qobject_cast<Analysis *>(analysisVariant.value<QObject *>());
+		setParent(_analysis); //This object should be destroyed whenever it's Analysis is destroyed.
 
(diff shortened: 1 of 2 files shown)


JASP-Desktop/analysis/analysisform.cpp
```diff
@@ -732,6 +732,7 @@ void AnalysisForm::_formCompletedHandler()
 	if (!analysisVariant.isNull())
 	{
 		_analysis	= qobject_cast<Analysis *>(analysisVariant.value<QObject *>());
+		setParent(_analysis); //This object should be destroyed whenever it's Analysis is destroyed.
```
boutinb, line 735: I get a crash when I close an analysis.
I think this is not a good idea to set the parent object to something else than its QQuickItem parent.

JorisGoosen: Ok, but maybe you can fix the problem then. I get a crash on closing because the code tries to `_setErrorMessages` while the qml is already gone.

How would you fix that?

boutinb:
I make a new PR to solve this.
I can add the KILLTIME also in my PR.

PR #4155, opened 2020-06-23
Fixes reliability module issues

juliuspfadt:
fixes https://github.com/jasp-stats/jasp-test-release/issues/847
fixes https://github.com/jasp-stats/jasp-test-release/issues/848
fixes https://github.com/jasp-stats/jasp-test-release/issues/846
fixes https://github.com/jasp-stats/jasp-test-release/issues/887
fixes https://github.com/jasp-stats/jasp-test-release/issues/889
fixes https://github.com/jasp-stats/jasp-test-release/issues/891

PR #4178, opened 2020-06-26
Reliability updated 

juliuspfadt:
fixes https://github.com/jasp-stats/jasp-test-release/issues/847
fixes https://github.com/jasp-stats/jasp-test-release/issues/848
fixes https://github.com/jasp-stats/jasp-test-release/issues/846
fixes https://github.com/jasp-stats/jasp-test-release/issues/887
fixes https://github.com/jasp-stats/jasp-test-release/issues/889
fixes https://github.com/jasp-stats/jasp-test-release/issues/891
fixes https://github.com/jasp-stats/jasp-test-release/issues/886
fixes https://github.com/jasp-stats/jasp-test-release/issues/909

In addition, the PR fixes issues with
i) the footnotes when Bayesian - missings - listwise deletion- footnotes
ii) the footnote when variables are deleted from the window, and the footnote would be empty, which crashed then
iii) being able to select no reliability coefficient, but item statistics such as item-rest correlation

### DIFF
--- a/JASP-Engine/JASP/R/reliabilityAnalysisBayesian.R
+++ b/JASP-Engine/JASP/R/reliabilityAnalysisBayesian.R
@@ -68,21 +68,17 @@ reliabilityBayesian <- function(jaspResults, dataset, options) {
   .checkInverse <- function() {
     if (length(options[["variables"]]) > 2) {
       use.cases <- "everything"
-      if (any(is.na(dataset))) {
+      if (anyNA(dataset)) {
         if (options[["missingValues"]] == "excludeCasesPairwise") 
           use.cases <- "pairwise.complete.obs"
         else if (options[["missingValues"]] == "excludeCasesListwise")
           use.cases <- "complete.obs"
       }
-      if (!("matrix" %in% class(try(solve(cov(dataset, use = use.cases)),silent=TRUE)))) {
-        return(gettext("Data covariance matrix is not invertible"))
-      } else {
-        return(NULL)
-      }
-        
-    } else {
-      return(NULL)
-    }
+      if (isTryError(try(solve(cov(dataset, use = use.cases)),silent=TRUE))) {
+        return(gettext("The covariance matrix of the data is not invertible"))
+      } 
+    } 
+    return(NULL)
   }
 
   .hasErrors(dataset = dataset, options = options, perform = "run",
@@ -145,7 +141,7 @@ reliabilityBayesian <- function(jaspResults, dataset, options) {
                                       
       missing <- "none" 
       options[["missings"]] <- "everything"
-      if (any(is.na(dataset))) {
+      if (anyNA(dataset)) {
         if (options[["missingValues"]] == "excludeCasesPairwise") {
           missing <- "pairwise"
           options[["missings"]] <- "pairwise.complete.obs"
@@ -274,7 +270,7 @@ reliabilityBayesian <- function(jaspResults, dataset, options) {
   names(cri) <- names(samps)
   
   for (nm in names(samps)) {
-    if (any(is.na(samps[[nm]]))) {
+    if (anyNA(samps[[nm]])) {
       cri[[nm]] <- c(NA_real_, NA_real_)
     } else {
       cri[[nm]] <- coda::HPDinterval(coda::mcmc(samps[[nm]]), prob = criValue)
--- a/JASP-Engine/JASP/R/reliabilityAnalysisBayesian.R
+++ b/JASP-Engine/JASP/R/reliabilityAnalysisBayesian.R
@@ -1,5 +1,6 @@
 reliabilityBayesian <- function(jaspResults, dataset, options) {
-
+  
+  
 	dataset <- .reliabilityReadData(dataset, options)
 
 	.reliabilityCheckErrors(dataset, options)
@@ -88,7 +89,7 @@ reliabilityBayesian <- function(jaspResults, dataset, options) {
       footnote <- sprintf(ngettext(length(variables[idx]),
                                    "The following item correlated negatively with the scale: %s. ",
                                    "The following items correlated negatively with the scale: %s. "),
-                          paste(variables[idx], collapse = ", "))      
+                          paste(variables[idx], collapse = ", "))
       return(footnote)
     }
   } else {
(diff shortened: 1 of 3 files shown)


TimKDJ:
Continuation of https://github.com/jasp-stats/jasp-desktop/pull/4155

JASP-Engine/JASP/R/reliabilityAnalysisBayesian.R
```diff
@@ -436,9 +439,23 @@ reliabilityBayesian <- function(jaspResults, dataset, options) {
       tb <- cbind(tb, newtb)
     }
 		itemTable$setData(tb)
+		
+		if (!is.null(unlist(options[["reverseScaledItems"]]))) {
+		  itemTable$addFootnote(sprintf(ngettext(length(options[["reverseScaledItems"]]),
```
vandenman: you're mixing tabs and spaces here. Please only use spaces in R code.

Resources/Reliability/qml/ReliabilityBayesian.qml
```diff
@@ -208,34 +208,40 @@ Form
...
-			{ 
-				id:                 probTable
-				name:               "probTable"; 
-				label:              qsTr("Probability for: ")
-				childrenOnSameRow:  true
+            {
+                id:                 probTable
+                name:               "probTable";
+                label:              qsTr("Probability for:")
+                childrenOnSameRow:  true
```
vandenman: Please use tabs only in QML files. If you use Qt Creator for editing QML files, see https://doc.qt.io/qtcreator/creator-indenting-code.html#indenting-qml-files.

vandenman (changes requested):
All issues are fixed, great! Two minor comments about style, see below.

JASP-Engine/JASP/R/reliabilityAnalysisBayesian.R
juliuspf: okay. So I should restructure the R-files to only have spaces instead of tabs? I know it is a mess right now, because I didn't start out with that premise :( How can I see where I used tabs and where I used spaces?

Resources/Reliability/qml/ReliabilityBayesian.qml
juliuspfadt: So what do I do here? I thought I used tabs here.

vandenman: Well if you use RStudio to edit the files it will automatically change the tabs into spaces (unless you changed that in the settings). Otherwise, Qt Creator or some other text editor (Atom, VS code) should do the trick.

juliuspfadt: Sorry, I am confused. You are saying I should open the qml file in Qt and use tabs? But where exactly? Just at the spot that you quoted here?

vandenman: In QML, we indent using tabs, never spaces. See also the [style guide](https://github.com/jasp-stats/jasp-desktop/blob/stable/Docs/development/jasp-qml-style-example.qml). So everywhere, not just the spot I quoted. Looking at your last commit I think you already figured that out though.

vandenman (approved):

juliuspfadt:
There is another issue by Suzanne that I will work on now. Would you have me make another PR then, or would you wait with merging this, until I fixed it?

vandenman:
@juliuspf just add it to this PR

vandenman:
@juliuspf let me know when this is ready for review!

juliuspfadt:
oh. I think it is ready

JASP-Engine/JASP/R/reliabilityAnalysisBayesian.R
```diff
@@ -64,19 +64,35 @@ reliabilityBayesian <- function(jaspResults, dataset, options) {
 
 .reliabilityCheckErrors <- function(dataset, options) {
 
-  .hasErrors(dataset = dataset, perform = "run",
+  # check for existing inverse
+  .checkInverse <- function() {
+    if (length(options[["variables"]]) > 2) {
+      use.cases <- "everything"
+      if (any(is.na(dataset))) {
```
vandenman:
```suggestion
      if (anyNA(dataset)) {
```

JASP-Engine/JASP/R/reliabilityAnalysisBayesian.R
```diff
@@ -64,19 +64,35 @@ reliabilityBayesian <- function(jaspResults, dataset, options) {
...
+      if (!("matrix" %in% class(try(solve(cov(dataset, use = use.cases)),silent=TRUE)))) {
+        return(gettext("Data covariance matrix is not invertible"))
+      } else {
+        return(NULL)
+      }
+        
+    } else {
+      return(NULL)
+    }
+  }
```
vandenman:
```suggestion
      }
    }
    return(NULL)
  }
```
if you make the default return value NULL then you don't need the else statements.

JASP-Engine/JASP/R/reliabilityAnalysisBayesian.R
```diff
@@ -64,19 +64,35 @@ reliabilityBayesian <- function(jaspResults, dataset, options) {
...
+    if (length(options[["variables"]]) > 2) {
+      use.cases <- "everything"
+      if (any(is.na(dataset))) {
+        if (options[["missingValues"]] == "excludeCasesPairwise") 
+          use.cases <- "pairwise.complete.obs"
+        else if (options[["missingValues"]] == "excludeCasesListwise")
+          use.cases <- "complete.obs"
+      }
+      if (!("matrix" %in% class(try(solve(cov(dataset, use = use.cases)),silent=TRUE)))) {
+        return(gettext("Data covariance matrix is not invertible"))
```
vandenman: rather than checking for "not a matrix" it makes more sense to check for is a try-error imo:
```suggestion
      if (isTryError(try(solve(cov(dataset, use = use.cases)),silent=TRUE))) {
        return(gettext("The covariance matrix of the data is not invertible"))
```

vandenman (changes requested):
the bug is fixed, some minor comments below.

JASP-Engine/JASP/R/reliabilityAnalysisBayesian.R
juliuspf: Well it seems to me that there is not function "isTryError"

TimKDJ (approved):
Everything fixed, code changes seem fine

PR #4182, opened 2020-06-26
Reorder Data Library to follow analyses on ribbon and add Reliability

TimKDJ:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/888

The data library didn't really follow any order after `6. Factor`. I made it follow the module ordering:

Schermafbeelding 2020-06-26 om 17 51 10

Schermafbeelding 2020-06-26 om 17 50 36

Not sure if we want this, but considering it was just a bit of shuffling around it's not a big deal to undo


boutinb (approved):
I tested it, and it seems to work.

PR #4183, opened 2020-06-29
Mixed models additional changes

FBartos:
fixes: https://github.com/jasp-stats/jasp-test-release/issues/899
fixes: https://github.com/jasp-stats/jasp-test-release/issues/897
fixes: https://github.com/jasp-stats/jasp-test-release/issues/898
fixes: https://github.com/jasp-stats/jasp-test-release/issues/900
fixes the additional issues raised here: https://github.com/jasp-stats/jasp-test-release/issues/800
(I did not rename the df estimation to CI method or something else since it also refers to the test against 0 or other specified value)

### DIFF
--- a/JASP-Engine/JASP/R/MixedModelsCommon.R
+++ b/JASP-Engine/JASP/R/MixedModelsCommon.R
@@ -904,7 +904,7 @@
   else if (type == "GLMM") dependencies <- .mmDependenciesGLMM
 
   if(options$method == "PB"){
-    seed_dependencies <- "seed", "setSeed"
+    seed_dependencies <- c("seed", "setSeed")
   }else{
     seed_dependencies <- NULL
   }
--- a/JASP-Engine/JASP/R/MixedModelsCommon.R
+++ b/JASP-Engine/JASP/R/MixedModelsCommon.R
@@ -2118,7 +2118,7 @@
         warmup            = options$warmup,
         adapt_delta       = options$adapt_delta,
         control           = list(max_treedepth = options$max_treedepth),
-        seed              = if (isTRUE(options[["setSeed"]])) set.seed(options[["seed"]])
+        seed              = if (isTRUE(options[["setSeed"]])) options[["seed"]]
       )
       
     } else if (type == "BGLMM") {
@@ -2148,7 +2148,7 @@
           control           = list(max_treedepth = options$max_treedepth),
           weights           = glmm_weight,
           family            = eval(call("binomial", glmm_link)),
-          seed              = if (isTRUE(options[["setSeed"]])) set.seed(options[["seed"]])
+          seed              = if (isTRUE(options[["setSeed"]])) options[["seed"]]
         )
         
       } else{
@@ -2162,7 +2162,7 @@
           adapt_delta       = options$adapt_delta,
           control           = list(max_treedepth = options$max_treedepth),
           family            = glmm_family,
-          seed              = if (isTRUE(options[["setSeed"]])) set.seed(options[["seed"]])
+          seed              = if (isTRUE(options[["setSeed"]])) options[["seed"]]
         )
         
       }
--- a/JASP-Engine/JASP/R/MixedModelsCommon.R
+++ b/JASP-Engine/JASP/R/MixedModelsCommon.R
@@ -3263,8 +3263,8 @@
   sprintf(
     ngettext(
       iterations,
-      "There was %i transition exceeding maximum treedepth indication problems with the efficiency of Hamiltonian Monte Carlo. Consider carefully increasing 'Maximum treedepth'.",
-      "There were %i transitions exceeding maximum treedepth indication problems with the efficiency of Hamiltonian Monte Carlo. Consider carefully increasing 'Maximum treedepth'."
+      "There was %i transition exceeding maximum tree depth indication problems with the efficiency of Hamiltonian Monte Carlo. Consider carefully increasing 'Maximum tree depth'.",
+      "There were %i transitions exceeding maximum tree depth indication problems with the efficiency of Hamiltonian Monte Carlo. Consider carefully increasing 'Maximum tree depth'."
     ),
     iterations
   )
--- a/JASP-Engine/JASP/R/MixedModelsCommon.R
+++ b/JASP-Engine/JASP/R/MixedModelsCommon.R
@@ -2097,8 +2097,7 @@
     mmModel$dependOn(dependencies)
     
     model_formula <- .mmModelFormula(options, dataset)
-    
-    JASP:::.setSeedJASP(options)
+
     if (type == "BLMM") {
       model <- stanova::stanova_lmer(
         formula           = as.formula(model_formula$model_formula),
@@ -2108,7 +2107,8 @@
         iter              = options$iteration,
         warmup            = options$warmup,
         adapt_delta       = options$adapt_delta,
-        control           = list(max_treedepth = options$max_treedepth)
+        control           = list(max_treedepth = options$max_treedepth),
+        seed              = if (isTRUE(options[["setSeed"]])) set.seed(options[["seed"]])
       )
       
     } else if (type == "BGLMM") {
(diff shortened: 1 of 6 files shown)


TimKDJ:
Are you still working on this?

FBartos:
yes, still tackling some issues raised by Lexi

TimKDJ:
Whoever gets around to reviewing this first. If you guys have time before tomorrow morning then go for it, otherwise I will do it

Resources/MixedModels/qml/common/MixedModelsMCMCDiagnostics.qml
```diff
@@ -40,17 +40,16 @@ Section
 		{
 			singleVariable:	true
 			name:			"samplingVariable1"
-			title:			qsTr("Selected term")
+			title:			if(samplingPlot.currentValue == "stan_scat"){qsTr("Horizontal axis")}else{qsTr("Plotted term")}
```
vandenman: how about a ternary?
```suggestion
			title:			samplingPlot.currentValue == "stan_scat" ? qsTr("Horizontal axis") : qsTr("Plotted term")
```

Resources/MixedModels/qml/common/MixedModelsPlots.qml
```diff
@@ -171,7 +173,7 @@ Section
 			visible:		plotsGeom.currentValue == "geom_jitter" | plotsGeom.currentValue == "geom_boxjitter"
```
vandenman: A single pipe is [bitwise or](https://stackoverflow.com/questions/6194950/what-does-the-single-pipe-do-in-javascript)
```suggestion
			visible:		plotsGeom.currentValue == "geom_jitter" || plotsGeom.currentValue == "geom_boxjitter"
```

JASP-Engine/JASP/R/MixedModelsCommon.R
```diff
@@ -659,16 +636,17 @@
         ANOVAsummary$addFootnote(.mmMessageOmmitedTerms2(removed_te[[i]], names(removed_te)[i]), symbol = gettext("Warning:"))
       }
     }
-    if (length(added_re) > 0) {
-      for (i in 1:length(added_re)) {
+
+    if (length(added_re) > 0)
+      for (i in 1:length(added_re))
```
vandenman: This is a nice use case for `seq_along`. That returns `1:length(added_re)` unless `length(added_re)` is 0, in which case it returns `integer(0)` and the whole body of the loop is never run.
```suggestion
    for (i in seq_along(added_re))
```

JASP-Engine/JASP/R/MixedModelsCommon.R
```diff
@@ -2174,7 +2122,8 @@
         iter              = options$iteration,
         warmup            = options$warmup,
         adapt_delta       = options$adapt_delta,
-        control           = list(max_treedepth = options$max_treedepth)
+        control           = list(max_treedepth = options$max_treedepth),
+        seed              = if (isTRUE(options[["setSeed"]])) set.seed(options[["seed"]])
```
vandenman: these and others can just be:
```suggestion
        seed              = .setSeedJASP(options)
```
we've abstracted this into a function to avoid people from having to do this manually and so we can easily change the implementation of the seed relatively easily.

JASP-Engine/JASP/R/MixedModelsCommon.R
```diff
@@ -2892,10 +2828,33 @@
...
+      coefs_name <-
+        paste(.unv(unlist(
+          strsplit(dimnames(samples)$Parameter[cf], ",")
+        )), collapse = ":")
+      coefs_name <- gsub(" ", "", coefs_name, fixed = TRUE)
+      coefs_name <-
+        .mmVariableNames(coefs_name, options$fixedVariables)
+      
+      if(length(coefs_trend) > 0){
+        for(cft in coefs_trend){
```
vandenman:
```suggestion
        for(cft in coefs_trend){
```
the loop is not entered if `length(coefs_trend)` is 0.

JASP-Engine/JASP/R/MixedModelsCommon.R
```diff
@@ -2918,30 +2877,48 @@
...
+            }
+          }
+        }
+        if(length(coefs_trend) > 0){
+          for(cft in coefs_trend){
+            if(cft %in% strsplit(pars[[2]], ":")[[1]] && !grepl(.unv(cft), coefs2_name)){
+              coefs2_name <- paste0(coefs2_name, ":", .unv(cft))
+            }
+          }
+        }
```
vandenman: shouldn't this be a function? The exact same code is also on lines 2846-2852, and here twice.

vandenman (changes requested):
Most changes look fine to me although I did not have time to test it in JASP. I've left some comments below.

Resources/MixedModels/qml/common/MixedModelsMCMCDiagnostics.qml
FBartos: ah, never used that before. Thanks

JASP-Engine/JASP/R/MixedModelsCommon.R
FBartos: yes, that's a good point. there is more of these there, I will change them at once

JASP-Engine/JASP/R/MixedModelsCommon.R
FBartos: but the `.setSeedJASP` function set's the seed in R and doesn't return the seed if it's specified.:

```
> JASP:::.setSeedJASP
function (options) 
{
    if (is.list(options) && c("setSeed", "seed") %in% 
        names(options)) {
        if (isTRUE(options[["setSeed"]])) 
            set.seed(options[["seed"]])
    }
}
```

Here, I need to pass it into the rstan function which handles seeds differently.

JASP-Engine/JASP/R/MixedModelsCommon.R
FBartos: yes, that was the intention. Is there a problem with that?

JASP-Engine/JASP/R/MixedModelsCommon.R
FBartos: yes, true. It is supposed to be a temporary solution - it's fixing a problem in `stanova` - we should get rid of this code with the next release. Changed

JASP-Engine/JASP/R/MixedModelsCommon.R
vandenman: but you had
```r
seed              = if (isTRUE(options[["setSeed"]])) set.seed(options[["seed"]])
```
shouldn't it then be
```r
seed              = if (isTRUE(options[["setSeed"]])) options[["seed"]]
```
without calling `set.seed` (which returns NULL)?

JASP-Engine/JASP/R/MixedModelsCommon.R
vandenman: the loop `for(cft in coefs_trend){` is not entered if the length of `coefs_trend` is 0, so the if statement `if(length(coefs_trend) > 0){` is redundant. Compare:
```r
for (a in list(a = 1))
  print(a)

for (a in list())
  print(a)
```

JASP-Engine/JASP/R/MixedModelsCommon.R
FBartos: yes, it definitely should've. Thanks for noticing this.

JASP-Engine/JASP/R/MixedModelsCommon.R
FBartos: oh, I see now. Thanks

JASP-Engine/JASP/R/MixedModelsCommon.R
FBartos: interesting thing is that both of the ways worked for making the results reproducible

vandenman: perhaps it defaults to `.Random.seed` when the seed is NULL?

vandenman (approved):
I did not have time to test it in JASP but the code looks good to me.

TimKDJ (approved):
I had a little trouble going through the linked issues as they had a lot of going back and forth in the comments, but the main points seem implemented. Code looks fine, although the string manipulations to obtain the variable names seem a little scary. I appreciate packages do not always make this easy though.

PR #4185, opened 2020-06-29
Closing JASP while bootstrapping makes it crash

boutinb:
Fixes jasp-stats/jasp-test-release#826

. Make the Analysis object as child of the AnalysisForm
. Do not destroy the Analysis object from Analyses: as child of
AnalysisForm, it will be destroyed automatically.
. When the AnalysisForm starts to be destroyed, signal it to all
objects that have a pointer to an analysis (EngineRepresentation and
PlotEditorModel), so that this pointer should not be used anymore.

### DIFF
--- a/Desktop/analysis/analyses.cpp
+++ b/Desktop/analysis/analyses.cpp
@@ -170,7 +170,8 @@ void Analyses::bindAnalysisHandler(Analysis* analysis)
 	connect(analysis,	&Analysis::titleChanged,						this, &Analyses::somethingModified					);
 	connect(analysis,	&Analysis::imageChanged,						this, &Analyses::somethingModified					);
 	connect(analysis,	&Analysis::userDataChangedSignal,				this, &Analyses::analysisOverwriteUserdata			);
-
+	connect(analysis,	&Analysis::destructionSignal,					this, &Analyses::analysisRemoved					);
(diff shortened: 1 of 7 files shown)


JorisGoosen:
I'll just download your branch and play around with it, because this really needs to be checked well.

JorisGoosen:
But that'll be tomorrow.

boutinb:
I think this is OK for 0.13, but this is not a Must Have. So if you find it too dangerous, let's skip it.

AlexanderLyNL:
Let's skip this for now, I'll move this to 0.14.

JorisGoosen:
Yeah that is for the best I think.

boutinb:
I have checked again this PR. I think this is OK now

JorisGoosen:
What happens if we switch to loaders for AnalysisForm though?
Then the Analysis gets destroyed whenever the form goes out of view.

boutinb:
The problem is not reproduceable anymore, maybe due to https://github.com/jasp-stats/jasp-desktop/commit/f7993fc2ec9d9d0fd5560a67e43a882bd3ebc578
And this fix would prevent https://github.com/jasp-stats/INTERNAL-jasp/issues/737 being implemented

PR #4187, opened 2020-06-30
fix rounding of p-values to 0 in post-hoc

Kucharssim:
Fixes one of the issues listed in https://github.com/jasp-stats/jasp-test-release/issues/907: occurence of a p-value equal to 0.

The problem is inside of emmeans where some corrections can (presumably) underflow to zero. The fix is nasty (replace the zero with something very very tiny), I am not sure what else to do with this.

### DIFF
--- a/JASP-Engine/JASP/R/ancova.R
+++ b/JASP-Engine/JASP/R/ancova.R
@@ -860,10 +860,19 @@ Ancova <- function(jaspResults, dataset = NULL, options) {
     postHocCorrections <- c("tukey", "scheffe", "bonferroni", "holm", "sidak")
 
     ## Computation
-    resultPostHoc <- lapply(postHocCorrections, function(x)
-      summary(emmeans::contrast(postHocRef[[postHocVarIndex]], method = "pairwise"), 
-              adjust = x, infer = c(TRUE, TRUE), level = options$confidenceIntervalIntervalPostHoc))
-
+    resultPostHoc <- lapply(postHocCorrections, function(x) {
+      post <- summary(emmeans::contrast(postHocRef[[postHocVarIndex]], method = "pairwise"), 
+                      adjust = x, infer = c(TRUE, TRUE), level = options$confidenceIntervalIntervalPostHoc)
+      
+      # some p-value corrections can underflow: we replace them with the smalles representable number
+      if(any(post$p.value == 0)) 
+        postHocStandardContainer[[thisVarName]]$addFootnote(message = gettext("Some p-values are approximate."))
+      
+      post$p.value[post$p.value == 0] <- .Machine$double.xmin
+      
+      return(post)
+    })
+    
     allContrasts <- strsplit(as.character(resultPostHoc[[1]]$contrast), split = " - ")
     
     if (nrow(resultPostHoc[[1]]) > 1)


JohnnyDoorn:
Seems like an OK fix! What does .Machine$double.xmin do exactly?

TimKDJ:
It won't really be the pvalue though, just [some boundary value](https://stackoverflow.com/questions/38165221/r-largest-smallest-representable-numbers/38165593#38165593). It feels a little iffy to me.
What about adding a footnote that points to the cell and explains underflow?

It's pretty much this problem: https://github.com/jasp-stats/INTERNAL-jasp/issues/924

EJWagenmakers:
But we do know that the p-value is smaller than this boundary value. I assume the new code shows p < .Machine$double.xmin, not p = .Machine$double.xmin. If it shows the smaller sign, I'm cool with this solution.

JohnnyDoorn:
Yes I agree

JohnnyDoorn (approved):

TimKDJ (changes requested):
It doesn't currently show `< ..`.
I'm not entirely sure if our js will like that, but it can definitely be tried

Right now we went from



To

Schermafbeelding 2020-06-30 om 15 38 21

Kucharssim:
> It won't really be the pvalue though, just [some boundary value](https://stackoverflow.com/questions/38165221/r-largest-smallest-representable-numbers/38165593#38165593). It feels a little iffy to me.

Yeah, I don't like this as well.

> What about adding a footnote that points to the cell and explains underflow?

I already added a footnote "Some p-values are approximate", but I agree it should be clearer (will be ugly to specify the correct table row and column. But I'll look into it.

> It's pretty much this problem: [jasp-stats/INTERNAL-jasp#924](https://github.com/jasp-stats/INTERNAL-jasp/issues/924)

Ok, that is really the same issue, so let's be consistent with the solution there (i.e., properly clarify in a footnote).

Just as a side note - would not it be just better to handle this somewere inside the jaspTable/javascript? I.e., for column type "pvalue" add some check if it exactly equals 0, and if so, automatically add a standardized footnote under the table? It's not that I cannot fix it for this particular table, but fixing this in the javascript would mean we need to fix it once, instead of possibly in many analyses (should they have similar issues).

JohnnyDoorn:
My bad, I thought the "<" sign would be in there.  What I would suggest is to just show p < 0.01 (if user specified 2 decimals), p < 0.001 (if users specified 3 decimals) etc. with some limit, such as 6/7 to keep the tables from blowing up. But that might be the same as @Kucharssim is suggesting.

Kucharssim:
> It doesn't currently show < ...
> I'm not entirely sure if our js will like that, but it can definitely be tried

I just tried, does not work... I think it should technically work if I pass it to the jaspTable as a list, where each column is a list of itself, to be able to have a character and a number in a single column, right? That would need some brutal rewrite 😦 

> What I would suggest is to just show p < 0.01 (if user specified 2 decimals), p < 0.001 (if users specified 3 decimals) etc. with some limit, such as 6/7 to keep the tables from blowing up. 

Yeah, something like that + potentially a footnote explaning why it does not show "exact" p-value, would be ideal.

AlexanderLyNL:
> What I would suggest is to just show p < 0.01 (if user specified 2 decimals), p < 0.001 (if users specified 3 decimals) etc. with some limit, such as 6/7 to keep the tables from blowing up.

I adviced Bruno to do the following: https://github.com/jasp-stats/jasp-desktop/pull/4189 

Even if user specified 2 decimals, and "Exact p-values" is off, then if p is very small, we have p <0.001. I hope you agree with that. It caused some confusion here:

https://github.com/jasp-stats/jasp-test-release/issues/906

EJWagenmakers:
I agree with Lexi

Kucharssim:
> > What I would suggest is to just show p < 0.01 (if user specified 2 decimals), p < 0.001 (if users specified 3 decimals) etc. with some limit, such as 6/7 to keep the tables from blowing up.
> 
> I adviced Bruno to do the following: #4189
> 
> Even if user specified 2 decimals, and "Exact p-values" is off, then if p is very small, we have p <0.001. I hope you agree with that. It caused some confusion here:
> 
> [jasp-stats/jasp-test-release#906](https://github.com/jasp-stats/jasp-test-release/issues/906)

Yes, this is related but slightly different issue. I think here we need to figure out what to show when the "Exact p-values" is on. Because if it is off, then it will just correctly show `< ...` for p-value = 0, the problem is what do we show if the p-value = 0 and the user wants to see the "exact value".

TimKDJ:
There are two different conversations in this pull request.
(1) when to switch from < 0.001 to < 0.01 or whatever and (2) what to do when the pvalue is exactly 0 and we show exact pvalues.

This pr is not about (1) `p < 0.001` vs `p < 0.01` -> see [this pr](https://github.com/jasp-stats/jasp-desktop/pull/4189).
This pr is about (2) the pvalue sometimes underflowing and being 0.

I don't think we can come up with a satisfactory solution for this before the release.

Kucharssim:
> I don't think we can come up with a satisfactory solution for this before the release.

Thanks for the summary - that should make it clear. I will make a separate issue for this and close this PR.

PR #4189, opened 2020-06-30
When exact p-values is off, display always at least 3 decimals for p values

boutinb:
Fixes jasp-stats/jasp-test-release#906

### DIFF
--- a/JASP-Desktop/html/js/utils.js
+++ b/JASP-Desktop/html/js/utils.js
@@ -79,7 +79,7 @@ function formatColumn(column, type, format, alignNumbers, combine, modelFootnote
             if (window.globSet.pExact) {
                 sf = 4;
 			} else {
-				p = 1/(Math.pow(10,dp));
+				p = f.substring(2);
             }
         }
 
@@ -276,7 +276,7 @@ function formatColumn(column, type, format, alignNumbers, combine, modelFootnote
                 var decimalsExpon = fixDecimals ? dp : sf - 1;
                 let _sign = (html) ? "&minus;" : "-";
                 var exponentiated = content.toExponential(decimalsExpon).replace(/-/g, _sign)
-                var paddingNeeded = Math.max(maxFSDOE - fSDOE(content), 0)
+				var paddingNeeded = Math.max(maxFSDOE - fSDOE(content), 0)
 
                 var split = exponentiated.split("e")
                 var mantissa = split[0]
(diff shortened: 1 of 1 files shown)


JASP-Desktop/html/js/utils.js
```diff
@@ -79,6 +79,7 @@ function formatColumn(column, type, format, alignNumbers, combine, modelFootnote
             if (window.globSet.pExact) {
                 sf = 4;
 			} else {
+				if (dp < 3) dp = 3
```
TimKDJ: If you fix the number of decimals to values 3+ then this will keep getting smaller (e.g., `< 0.00001`). `< 0.001` should be the lower bound.

~~On a somewhat different note: do we want to go "exact" unless `< 0.001` or do we also want to show the `< 0.05` and `< 0.01` @AlexanderLyNL?~~ -> If I look at 0.11.1 we didn't used to do that either. Which is definitely a good thing.

TimKDJ (changes requested):

JASP-Desktop/html/js/utils.js
TimKDJ: So I guess the rules are:

If "Display exact p-values" is **off**:
if value < 0.001 ---> `< .001` (**does not** follow the "Fix the number of decimals" setting)
if value >= 0.001 ---> value (**does** follow the "Fix the number of decimals" setting)

If "Display exact p-values" is **on**:
if value < 0.001 ---> value (**does** follow the "Fix the number of decimals" setting)
if value >= 0.001 ---> value (**does** follow the "Fix the number of decimals" setting)

AlexanderLyNL: I'm stupid, ignore my comment

TimKDJ: Are we sure we want `< 0.01` and `< 0.05`? It is not something we ever did and they're also not very convincing values to report like that. We only ever used to do `< 0.001`.

boutinb: The rules are:
If "Display exact p-values" is off:
     if number of decimals is < 3, then number of decimals is 3 so that
          if value < 0.001 (eg. 0.000406) --> < .001 
          if value > 0.001 (eg. 0.00406)   --> .004
     if number of decimals is 4 (for example)
          if value < 0.0001 --> < .0001 
          if value > 0.0001 --> value with 4 decimals
If "Display exact p-values" is on:
it takes always the specified number of decimals. So it it 2
     if value < 0.01 (eg. 0.00406) -> 4.06e-3
     if value > 0.01 (eg. 0.0406) -> 0.04

TimKDJ: No we don't want `.0001`, `.001` is the lowest we go.
See also e.g., [this resource](https://support.jmir.org/hc/en-us/articles/360000002012-How-should-P-values-be-reported-)

boutinb: OK. I put the situation back to before https://github.com/jasp-stats/jasp-desktop/pull/4065, and I solve the issue (https://github.com/jasp-stats/jasp-issues/issues/706) by setting a scientific notation when the p-value is bigger than 0.001 but needs more than the 'number of decimals' (for example: number of decimals is 2, and p-value is 0.00412).

That means that when exact p-value is off:

- value < 0.001 is always written with '< .001' (does not depend on number of decimals)
- value >= 0.001
if 'number of decimals' >= 3: value is written always in decimal format
if 'number of decimals' = 2 (or 1):
if value < 0.01 then value is written in scientific format
if value >= 0.01 then value is written in decimal format

JASP-Desktop/html/js/utils.js
```diff
@@ -372,8 +372,29 @@ function formatColumn(column, type, format, alignNumbers, combine, modelFootnote
...
+						strContent = mantissa + "e&thinsp;" + exponentSign + exponentNum;
+					} else {
+						strContent = mantissa + "e" + exponentSign + exponentNum;
+					}
+
+				} else {
+					let _content = (html) ? "&minus;" : "-";
+					strContent = content.toFixed(dp).replace(/-/g, _content)
+				}
+				formatted = { content: strContent, "class": "number" }
```
TimKDJ, line 397: Can you fix the indentation?

TimKDJ (approved):
Code looks fine (apart from indentation..) and it seems to work as intended!

JASP-Desktop/html/js/utils.js
boutinb: Do you mean that I use tab and the rest of this files uses spaces for indentation? The rest of our code uses tab, so I can transform all the spaces in tabs. Or do you mean something else?

TimKDJ: Yeah I mean making the indentation consistent within the file.
If you make everything tabs can you do it in a separate commit? So we can see the actual changes you made in the future

JorisGoosen:
I will fix the indents

PR #4193, opened 2020-07-01
0 should not be displayed with scientific format.

boutinb:
The error was that the test to check that the number is too small for a decimal format (smaller than 10^(-numberDecimals) should also check that the number is not 0
Remove copy/paste by adding toExponent function.

### DIFF
--- a/JASP-Desktop/html/js/utils.js
+++ b/JASP-Desktop/html/js/utils.js
@@ -274,31 +274,9 @@ function formatColumn(column, type, format, alignNumbers, combine, modelFootnote
 			else if (Math.abs(content) >= upperLimit || Math.abs(content) <= Math.pow(10, -dp)) {
 
 				var decimalsExpon = fixDecimals ? dp : sf - 1;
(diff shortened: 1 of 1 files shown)


JASP-Desktop/html/js/utils.js
```diff
@@ -374,22 +352,8 @@ function formatColumn(column, type, format, alignNumbers, combine, modelFootnote
...
-					var exponentSign = exponent.substr(0, 1)
-					var exponentNum = exponent.substr(1)
-
-					if (html) {
-						strContent = mantissa + "e&thinsp;" + exponentSign + exponentNum;
-					} else {
-						strContent = mantissa + "e" + exponentSign + exponentNum;
-					}
-
+				if (content != 0 && Math.abs(content) < 1/(Math.pow(10,dp))) {
```
TimKDJ: This creates the edge case that `-0.001` will be displayed as `-1.000e-3`. Can we use `<=` here? It's not a big deal though

TimKDJ (approved):

PR #4200, opened 2020-07-13
Fix bug in benfords law

### LINKED ISSUE #815: Audit Module: check the frequency of last digits
* Enhancement: in addition to Benford's law, it would be useful to check the frequency of last digits.
* Purpose: Improve fraud detection.
* Use-case: Example: http://datacolada.org/77

Thanks!

koenderks:
Make sure last numbers are read in correctly.

Fixes https://github.com/jasp-stats/jasp-issues/issues/815

[testfile.zip](https://github.com/jasp-stats/jasp-desktop/files/4911304/testfile.zip)

Accompanied by https://github.com/jasp-stats/Audit/pull/10

### DIFF
--- a/JASP-Engine/JASP/R/auditClassicalBenfordsLaw.R
+++ b/JASP-Engine/JASP/R/auditClassicalBenfordsLaw.R
@@ -149,7 +149,7 @@ auditClassicalBenfordsLaw <- function(jaspResults, dataset, options, ...){
               type=c("infinity", "variance", "observations"),
               all.target = values, 
               message = "short", 
-              observations.amount= "< 10",
+              observations.amount= "< 2",
               exitAnalysisIfErrors = TRUE)
 }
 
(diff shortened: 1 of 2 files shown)


JorisGoosen:
Nitpicky but: if you choose first & second digit the table still says "Leading digit":
Schermafbeelding 2020-07-13 om 12 50 02

JorisGoosen:
And Im still a bit confused why in a number like "123.0123" the first digit wouldn't be 0?
Especially in uniform distribution?

Or do you always skip that one?
Because the percentages you give here wouldn't make a lot of sense:
Schermafbeelding 2020-07-13 om 12 52 34

JorisGoosen:
For the rest this all looks fine!

koenderks:
> And Im still a bit confused why in a number like "123.0123" the first digit wouldn't be 0?
> Especially in uniform distribution?
> 
> Or do you always skip that one?
> Because the percentages you give here wouldn't make a lot of sense:
> Schermafbeelding 2020-07-13 om 12 52 34

Actually, the first digit would be a 1 in this instance right?

Thanks for the nitpicky part, I'll adjust that in just a minute.

Furthermore, I always just ignore the 0 as a starting digit. Then it will just be the second number.

AlexanderLyNL:
I'm still concerned with Benford's law for the last digit, if it ends with a zero.

JorisGoosen:
Well, I'll leave the number interpretations and zeroes up to @koenderks and @AlexanderLyNL .
If you both think the current approach (or after fixes) is ok this PR can be merged.

koenderks:
Yes, it should probably count the last digit that is not zero.

AlexanderLyNL:
> Yes, it should probably count the last digit that is not zero.

Should, or is? What does the Benford's law for last digit procedure do for entries of the form shown below?

`4.60, 5.70, 12.10, 32.99`

koenderks:
image

AlexanderLyNL:
Seems to work well then. So for `4.60` it just studies `6`, which is great. I'm not sure whether Benford's law is reasonable for last digits, though, but that's a different discussion. If people want to study Benford's law for last digit, then so be it.

koenderks:
I think they mostly want to compare the distribution of last digits to a uniform distribution, hence the inclusion of that in this PR.

PR #4207, opened 2020-07-30
adding classical Selection models to Meta Analysis

FBartos:
This PR adds a new analysis - Selection Models - the analysis is pretty straight-forward and similar to RoBMA. However, the fitting process is lightning fast in comparison to it (no sampling needed).

We want to release it hand in hand with RoBMA which builds upon the selection models because we discuss both analyses in our upcoming tutorial paper

this PR is not an update for RoBMA (it will come later)
the same PR towards the Meta Analysis module: https://github.com/jasp-stats/MetaAnalysis/pull/6

I will add a help file and unit tests later


Resources/Meta Analysis/Description.qml
```diff
@@ -19,11 +19,24 @@ Description
...
+		title:	qsTr("Bayesian Meta-Analysis")
 		func:	"BayesianMetaAnalysis"
 	}
+
+	Analysis
+	{
+		title:		qsTr("Robust Bayesian Meta-Analysis")
+		func:		"RobustBayesianMetaAnalysis"
+		requiresData:	false
+	}
```
FBartos, line 45: why isn't the RoBMA menu accessible without data?

image

FBartos: fixed thanks to Joris

JorisGoosen:
@Kucharssim you have time to check out this PR?

Kucharssim:
> @Kucharssim you have time to check out this PR?

Uh, maybe... When do we want to have another release? I will try to get to it asap 😦

JorisGoosen:
Well, maybe someone else can do it. I don't think there is any need to hurry like  crazy though.

AlexanderLyNL:
I can't build at the moment to check properly.

JorisGoosen:
Well if you can think of a good reviewer then that would be fine too

AlexanderLyNL:
Simon?

Kucharssim:
> Well, maybe someone else can do it. I don't think there is any need to hurry like crazy though.

If we are not in a hurry, I'll do it...

JorisGoosen:
Haha, ok, great!

Kucharssim (commented):
This looks good, the code is clean, the analysis seems to run. I cannot judge whether it does what it's supposed to do, because I don't know what it is for.

I have one minor comment, which I don't know whether is a bug or not (as I don't know what do you need to run the analysis).

1) If I fill field `Effect size` and `Effect size standard error`, the analysis runs. Cool.
2) If I fill field `Effect size` and `P-value`, the analysis is idle, as if it waits for something. That gives me a hint that you actually need the `Effect size standard error` every time.
3) If I fill all three fields, the analysis ends with the following error:
Screenshot 2020-08-06 at 12 35 31


Attached `.jasp` file as `.zip` (rename to `.jasp`) 
[BCG Vaccine selection.zip](https://github.com/jasp-stats/jasp-desktop/files/5034275/BCG.Vaccine.selection.zip)

Kucharssim:
Also - A unit test would be great!

JorisGoosen:
In fact, it would be almost mandatory so that it can actually be tested for regression...

FBartos:
I fixed the bug and added a note saying that both ES and SE are required if user specifies p-values but not both of them.

I was trying to add unit tests but I had problems running them afterward. Tim told me that he will add the functionality for unit tests within the specific modules within a day, so I will add them there then.

JorisGoosen:
So after @TimKDJ adds that functionality this PR can be finished up with the unittests? Great!

FBartos:
yes, but I will add the tests only in the Meta Analysis module if I understood correctly. (The tests are already done, I just need to test that they can be run)-

JorisGoosen:
Can't the tests be added to both?

FBartos:
They can, but I wasn't able to run them after adding them and I have been told that it's not worth fixing it when the tests will be working in the modules separately.

JorisGoosen:
Well, I don't know how much effort that fixing would entail so we could wait for @TimKDJ to start working tomorrow? He was the one who told you right?
Because this branch will still be used for the next release and it would make me happy to know that the inevitable bugfixes that will be coming will actually not break anything here.
This could also be done in the module, but Im guessing that under stress the changes will be made here first...

TimKDJ:
I figured it was not worth the trouble to get them to work in stable as we're switching to modules soon anyway (and all code ends up in both places). But I can take a quick look to see if it's an easy fix. What branch do the tests live in?

JorisGoosen:
Well, I suggested this only because I though it would be fairly little work, I thought it was just a matter of moving the unittests to some test-folder. And adding them maybe to some list of tests to perform.

TimKDJ:
It turned out to be an easy (renaming) fix :)

FBartos:
thanks, it's working now. so, this should be all

PR #4208, opened 2020-07-31
Check when items overflow the form space

boutinb:
Fixes jasp-stats/INTERNAL-jasp#754

### DIFF
--- a/JASP-Desktop/analysis/analysisform.cpp
+++ b/JASP-Desktop/analysis/analysisform.cpp
@@ -749,6 +749,9 @@ void AnalysisForm::_formCompletedHandler()
 	{
 		_analysis	= qobject_cast<Analysis *>(analysisVariant.value<QObject *>());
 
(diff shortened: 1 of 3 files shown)


JASP-Desktop/analysis/analysisform.cpp
```diff
@@ -899,6 +902,9 @@ QString AnalysisForm::metaHelpMD() const
 
 QString AnalysisForm::helpMD() const
 {
+	if (!PreferencesModel::prefs()->generateMarkdown())
+		return ""; // It should not come here, but when debugging QML, it crashes here because apparently _analysis is not set yet. Weird...
```
JorisGoosen, line 906: Hmm, so it would still crash if the pref isn't set?

And now the property is broken forever whenever you disable the option...
I guess we can leave it for now though.

JorisGoosen (approved):

PR #4211, opened 2020-08-04
Contingency tables properly reverse column labels

### LINKED ISSUE #914: Column order option does not work
* JASP version: 0.13.1

* OS name and version: MacOS 10.15.6
* Analysis: Contingency tables
* Bug description: Ascending and descending column order option does not work, the data of the table does not change the order
* Expected behaviour: the order of the column should be modified

#### Steps to reproduce:
1. Open the attached file
2. Scroll down to Options section
3. Click on Ascending or Descending order option
4. Observe that the data of the Contingency table is not updated

[Chi cuadrado asociación ejemplo 1.jasp.zip](https://github.com/jasp-stats/jasp-issues/files/4999799/Chi.cuadrado.asociacion.ejemplo.1.jasp.zip)

vandenman:
Fixes https://github.com/jasp-stats/jasp-issues/issues/914

I'm not entirely sure if this a good solution but at least the radio buttons `Row Order` and `Column Order` are now consistent (they both flip the table). I don't know what this has to with "Ascending" or "Descending" though. I also have no idea why the old implementation sorted the column labels.

Furthermore, in my opinion, we should delete these two radio buttons. Reordering a table should be done by clicking on a variable and reordering its levels. This also works already.

### DIFF
--- a/JASP-Engine/JASP/R/contingencytables.R
+++ b/JASP-Engine/JASP/R/contingencytables.R
@@ -362,7 +362,7 @@ ContingencyTables <- function(jaspResults, dataset, options, ...) {
     else  lvls <- unique(dataset[[ .v(analysis$columns) ]])
       
     if (options$columnOrder == "descending")
-        lvls <- sort(lvls, decreasing = TRUE)
+        lvls <- rev(lvls)
 
     overTitle            <- unlist(analysis$columns)
     useColumnNameAsTitle <- TRUE


Kucharssim (commented):
Looks good, fixes the linked issue, and can be merged as it is.

> Furthermore, in my opinion, we should delete these two radio buttons. Reordering a table should be done by clicking on a variable and reordering its levels. This also works already.

I agree that seems like a reasonable thing to do. Or at least I cannot think of a use case which cannot be solved by sorting the variables themselves.

If we don't delete them, however, we need to add dependency of the ordinal coefficients on the Ascending/Descending options. For example, it seems that Ordinal Gamma does not update if I switch one of the radio buttons (when it fact it should switch the sign).

PR #4215, opened 2020-08-05
Enable save as svg

### LINKED ISSUE #873: Save image as SVG
As suggested/requested in https://github.com/jasp-stats/jasp-issues/issues/863 

It would indeed be nice to support this. Especially given the fact that we already support rendering all plots to svg anyway.
Those few remaining issues with that will have to be solved.

vandenman:
Fixes https://github.com/jasp-stats/jasp-issues/issues/873
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/620

Plots shown in JASP are still png files. 

This no longer has the error with multiple pages (just refresh [abtestPlotPagingError.jasp.zip](https://github.com/jasp-stats/jasp-desktop/files/5027749/abtestPlotPagingError.jasp.zip)) if you use the development version of svglite (https://github.com/r-lib/svglite).

### DIFF
--- a/JASP-Desktop/mainwindow.cpp
+++ b/JASP-Desktop/mainwindow.cpp
@@ -783,18 +783,19 @@ void MainWindow::_analysisSaveImageHandler(Analysis* analysis, QString options)
 	parser.parse(utf8, root);
 
 	QString selectedFilter;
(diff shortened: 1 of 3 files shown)


JorisGoosen:
Hah it is quite nice to be able to open a plot in inkscape! :)

PR #4216, opened 2020-08-05
Updates - Distributions module

Kucharssim:
Details here https://github.com/jasp-stats/DiscoverDistributions/pull/1 + new unit tests


Kucharssim:
Travis failed to build...

Kucharssim:
@AlexanderLyNL would you be able to check it out? If not let's ask someone else...

AlexanderLyNL:
Perhaps Don or Johnny can have a look at this.

boutinb:
@JohnnyDoorn If you can't review it, maybe ask @vandenman too do it.

JASP-Engine/JASP/R/ldLogistic.R
```diff
@@ -0,0 +1,200 @@
...
+}
+
+.recodeOptionsLDLogistic <- function(options){
+  options[['parValNames']] <- c("mu", "sigma")
+  
+  options[['pars']]   <- list(location = options[['mu']], scale = options[['sigma']])
+  options[['pdfFun']] <- dlogis
+  options[['cdfFun']] <- plogis
+  options[['qFun']]   <- qlogis
+  options[['rFun']]   <- rlogis
```
vandenman:
```suggestion
  options[['pdfFun']] <- stats::dlogis
  options[['cdfFun']] <- stats::plogis
  options[['qFun']]   <- stats::qlogis
  options[['rFun']]   <- stats::rlogis
```
a little bit pedantic, but when we switch to modules we'll either need to prepend all of these with `stats::` or otherwise we need to import in the namespace (using roxygen2, ideally). Doing it now already might alleviate that step. If you prefer doing this with roxygen2 you can ignore this comment :)

JASP-Engine/JASP/R/ldLogistic.R
```diff
@@ -0,0 +1,200 @@
...
+}
+
+.ldFormulaLogisticCDF <- function(options){
+  if(options$parametrization == "scale"){
+    text <- "<MATH>
+    F(x; <span style='color:red'>&beta;</span>) = 
+    </MATH>"
+  } else{
+    
+  }
```
vandenman: is this else intentionally empty?

JASP-Engine/JASP/R/ldLogistic.R
```diff
@@ -0,0 +1,200 @@
...
+}
+
+.ldFormulaLogisticQF <- function(options){
+  if(options$parametrization == "rate"){
+    text <- "<MATH>
+    Q(p; <span style='color:red'>&beta;</span>) = 
+    </MATH>"
+  } else{
+    
+  }
```
vandenman: also intentionally empty?

JASP-Engine/JASP/R/ldLognormal.R
```diff
@@ -0,0 +1,200 @@
...
+}
+
+.recodeOptionsLDLognormal <- function(options){
+  options[['parValNames']] <- c("mu", "sigma")
+  
+  options[['pars']]   <- list(meanlog = options[['mu']], sdlog = options[['sigma']])
+  options[['pdfFun']] <- dlnorm
+  options[['cdfFun']] <- plnorm
+  options[['qFun']]   <- qlnorm
+  options[['rFun']]   <- rlnorm
```
vandenman: `stats::` or ignore if you want to use roxygen2

JASP-Engine/JASP/R/ldLognormal.R
```diff
@@ -0,0 +1,200 @@
...
+    options[['highlightmax']] <- options[['max']]
+  } else if(options[['highlightType']] == "lower"){
+    options[['highlightmin']] <- options[['range_x']][1]
+    options[['highlightmax']] <- options[['lower_max']]
+  } else if(options[['highlightType']] == "upper"){
+    options[['highlightmin']] <- options[['upper_min']]
+    options[['highlightmax']] <- options[['range_x']][2]
+  } else{
+    options[['highlightmin']] <- options[['highlightmax']] <- NULL
+  }
```
vandenman: shouldn't this be a function? I'm sure it works fine but there is some code duplication going on.

JASP-Engine/JASP/R/ldLognormal.R
```diff
@@ -0,0 +1,200 @@
...
+}
+
+.ldFormulaLognormalQF <- function(options){
+  if(options$parametrization == "rate"){
+    text <- "<MATH>
+    Q(p; <span style='color:red'>&beta;</span>) = 
+    </MATH>"
+  } else{
+    
+  }
```
vandenman: empty else

JASP-Engine/JASP/R/ldLognormal.R
```diff
@@ -0,0 +1,200 @@
...
+}
+
+.ldFormulaLognormalCDF <- function(options){
+  if(options$parametrization == "scale"){
+    text <- "<MATH>
+    F(x; <span style='color:red'>&beta;</span>) = 
+    </MATH>"
+  } else{
+    
+  }
```
vandenman: empty else

vandenman (changes requested):
I haven't tested it in JASP yet, but most of the code looks ok to me.

I've probably said this before but there is quite some code duplication. It might be worthwhile to think of some ways to minimize that. For example, the bodies of `LDlogistic` and `LDlognormal` look nearly identical to me. There are plenty of ways to circumvent duplicating the code like this (e.g., using a function, using S3 classes, or using R6 classes).

JohnnyDoorn:
I will test later today, or else tomorrow!

Kucharssim:
Thanks @vandenman for the feedback! I reduced some code duplication and added `stats::` wherever needed. There is still some code duplication left, but the most obvious sources are reduced, I think. I will need to think about the leftovers next time.

vandenman (changes requested):
So I'm not sure if this is intended:

image

Open the Chi-square distribution, sample 100 new observations to a new variable with k = 5 and lambda = 0. Then try to estimate the parameters of this distribution. The result is an error. It could be that lambda = 0 is a corner case as other values do work (e.g., lambda = 2). If so, then should we maybe change the default value for lambda to something else?

The rest of the changes look good to me!

Kucharssim:
> So I'm not sure if this is intended:

Yeah, this is annoying. You are correct that lambda = 0 is a boundary value and this causes problems when fitting. One problem was that I set the bounds of lambda incorrectly, and that is fixed now. It does not really fix the issue, though.

> then should we maybe change the default value for lambda

The problem is that lambda is the noncentrality parameter, and so defaulting it to 0 makes sense, as that leads to the standard chi-squared distribution. I am more inclined to rather include an option where you can specify which parameters to estimate and which parameters to keep fixed at their specified values. In case of chi-squared, we could have default lambda = 0 and fixed (but possibility to let it free). I am currently writing this feature in a different branch, but it's not ready yet.

Also, I need to improve the robustness of MLE in general. I thought using the specialised `fitdistplus` package would be fine as it usually works well, but in reality it just makes the implementation in JASP more awkward and harder to debug than just to call `optim()` - in hindsight, I would do it differently. But I will need to do it later.

vandenman:
> Also, I need to improve the robustness of MLE in general. I thought using the specialised fitdistplus package would be fine as it usually works well, but in reality it just makes the implementation in JASP more awkward and harder to debug than just to call optim() - in hindsight, I would do it differently. But I will need to do it later.

Yeah a package like fitdistrplus is nice when first writing it up but it may turn out to be not flexible enough later. Either way, the code looks fine to me, if Travis is happy then I'm happy.

EJWagenmakers:
"Optim" is very buggy though -- it sometimes forgets previous better solutions and reports an inferior one
E.J.

vandenman (approved):

AlexanderLyNL:
@Kucharssim Could you please make a corresponding PR to the module?

Kucharssim:
> @Kucharssim Could you please make a corresponding PR to the module?

Ah yeah, it's here https://github.com/jasp-stats/jaspDistributions/pull/1, but I need to add the changes based on Don's feedback. I forgot about it - i'll do it asap.

PR #4221, opened 2020-08-06
Add possibility a control of a list as source for another list

boutinb:
This is needed for the Learning Bayes module

### DIFF
--- a/JASP-Desktop/widgets/listmodel.cpp
+++ b/JASP-Desktop/widgets/listmodel.cpp
@@ -520,10 +520,8 @@ const Terms &ListModel::terms(const QString &what) const
 
 	for (const QString& condition : allConditions)
 	{
-		if (condition.startsWith(typeIs))
-			useTheseVariableTypes = condition.right(condition.length() - typeIs.length());
-		if (condition.startsWith(controlIs))
-			useThisControl = condition.right(condition.length() - controlIs.length());
+		if (condition.startsWith(typeIs))		useTheseVariableTypes	= condition.right(condition.length() - typeIs.length());
+		if (condition.startsWith(controlIs))	useThisControl			= condition.right(condition.length() - controlIs.length());
 	}
 
 	if (!useTheseVariableTypes.isEmpty())
@@ -559,10 +557,9 @@ const Terms &ListModel::terms(const QString &what) const
 			if (rowControls)
 			{
 				JASPControlWrapper* controlWrapper = rowControls->getJASPControl(useThisControl);
-				if (controlWrapper)
-					controlTerms.add(controlWrapper->getItemProperty("value").toString());
-				else
-					Log::log() << "Could not find control " << useThisControl << " in list view " << name() << std::endl;
+
+				if (controlWrapper)	controlTerms.add(controlWrapper->getItemProperty("value").toString());
+				else				Log::log() << "Could not find control " << useThisControl << " in list view " << name() << std::endl;
 			}
 		}
 		specialTerms = controlTerms;
--- a/JASP-Desktop/widgets/listmodel.cpp
+++ b/JASP-Desktop/widgets/listmodel.cpp
@@ -77,35 +77,32 @@ void ListModel::_initTerms(const Terms &terms, const RowControlsOptions& allOpti
 	_rowControlsOptions = allOptionsMap;
 	endResetModel();
 
-	if (setupControlConnections && listView()->sourceModels().size() > 0)
-	{
-		QMap<ListModel*, Terms> map = getSourceTermsPerModel();
-		QMapIterator<ListModel*, Terms> it(map);
-
-		while (it.hasNext())
-		{
-			it.next();
-			ListModel* sourceModel = it.key();
-			const Terms& terms = it.value();
-			QMLListView::SourceType* sourceType = listView()->getSourceTypeFromModel(sourceModel);
-			if (sourceType)
-			{
-				for (const QMLListView::SourceType::ConditionVariable& conditionVariable : sourceType->conditionVariables)
-					for (const Term& term : terms)
-						_connectSourceControl(sourceModel->getRowControl(term.asQString(), conditionVariable.controlName));
-			}
-		}
-	}
+	if (setupControlConnections)
+		for (QMLListView::SourceType* sourceType : listView()->sourceModels())
+			_connectSourceControls(sourceType->model, sourceType->usedControls);
 }
 
-void ListModel::_connectSourceControl(JASPControlWrapper* control)
+
+
+void ListModel::_connectSourceControls(ListModel* sourceModel, const QSet<QString>& controls)
 {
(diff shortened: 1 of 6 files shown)


boutinb:
@FBartos can check it

FBartos:
I have seen it and it works. I love it!

JASP-Desktop/widgets/listmodel.cpp
```diff
@@ -77,38 +77,35 @@ void ListModel::_initTerms(const Terms &terms, const RowControlsOptions& allOpti
...
-			QMLListView::SourceType* sourceType = listView()->getSourceTypeFromModel(sourceModel);
-			if (sourceType)
+			JASPControlWrapper* control = sourceModel->getRowControl(term.asQString(), controlName);
+			BoundQMLItem* boundControl = dynamic_cast<BoundQMLItem*>(control);
+			if (boundControl && !_rowControlsConnected.contains(boundControl))
 			{
-				for (const QMLListView::SourceType::ConditionVariable& conditionVariable : sourceType->conditionVariables)
-					for (const Term& term : terms)
-						_connectSourceControl(sourceModel->getRowControl(term.asQString(), conditionVariable.controlName));
+				boundControl->boundTo()->changed.connect(boost::bind(&ListModel::_sourceTermsChangedHandler, this, _1));
```
JorisGoosen, line 102: Why are you using boost connections instead of qt connections?

JASP-Desktop/widgets/listmodel.cpp
```diff
@@ -506,12 +504,33 @@ const QString &ListModel::name() const
 const Terms &ListModel::terms(const QString &what) const
 {
 	const QString typeIs = "type=";
+	const QString controlIs = "control=";
 
-	if (what.startsWith(typeIs))
+	if (!what.startsWith(controlIs) && !what.startsWith(typeIs))
+		return _terms; // in most cases, it comes here.
```
JorisGoosen, line 510: "it ends here" ?

JASP-Desktop/widgets/listmodel.cpp
```diff
@@ -506,12 +504,33 @@ const QString &ListModel::name() const
...
+
+	// This method was defined as const (it is used in other const methods), and returns a reference to a const Terms
+	// It returns 99% of the cases _terms
+	// But in case that a condition is used (what is not empty), then it cannot return a local variable nor a member of this class
+	// So a static variable is used.
```
JorisGoosen, line 515: Thanks for the comment :)

JASP-Desktop/widgets/listmodel.cpp
```diff
@@ -506,12 +504,33 @@ const QString &ListModel::name() const
...
+	static Terms specialTerms;
+
+	QStringList allConditions = what.split(",");
+	QString useTheseVariableTypes, useThisControl;
+
+	for (const QString& condition : allConditions)
 	{
-		static Terms terms;
+		if (condition.startsWith(typeIs))
```
JorisGoosen: Isn't this more readable:
```C++
if (condition.startsWith(typeIs))		useTheseVariableTypes	= condition.right(condition.length() - typeIs.length());
if (condition.startsWith(controlIs))	useThisControl			= condition.right(condition.length() - controlIs.length());
```

JorisGoosen: Like, *much more* readable

JASP-Desktop/widgets/listmodel.cpp
```diff
@@ -521,18 +540,35 @@ const Terms &ListModel::terms(const QString &what) const
...
+	if (!useThisControl.isEmpty())
+	{
+		Terms controlTerms;
+		for (const Term& term : specialTerms)
+		{
+			RowControls* rowControls = _rowControlsMap[term.asQString()];
+			if (rowControls)
+			{
+				JASPControlWrapper* controlWrapper = rowControls->getJASPControl(useThisControl);
+				if (controlWrapper)
```
JorisGoosen:
```
//It nicer to have these simple ifelse's as two line I think
if ...
else ...
```

JASP-Desktop/widgets/qmllistview.cpp
```diff
@@ -63,6 +64,51 @@ QList<QVariant> QMLListView::_getListVariant(QVariant var)
...
+}
+
+QMap<QString, QVariant> QMLListView::_readSource(const QVariant& source, QString& sourceName, QString& sourceControl, QString& sourceUse)
+{
+	QMap<QString, QVariant> map;
+
+	JASPControlBase* sourceItem = source.value<JASPControlBase*>();
+	if (sourceItem)
+		sourceName = sourceItem->name();
+	else if (source.canConvert<QString>())
```
JorisGoosen: So and int in a QVariant *can* be converted to a QString, wouldn't it be safer to check for ["type()"](https://doc.qt.io/qt-5/qvariant.html#type) ?

JorisGoosen: Or you always want it to just be converted to a string unless it is a map? Because in that case I would put the map first, it seems like it would be a more restrictive "canConvert"

JASP-Desktop/widgets/qmllistview.cpp
```diff
@@ -75,86 +121,51 @@ void QMLListView::setSources()
...
+		{
+			QList<QVariant> conditionVariablesList = _getListVariant(map["conditionVariables"]);
 
-				for (const QVariant& conditionVariablesVar : conditionVariablesList)
-				{
-					if (conditionVariablesVar.canConvert<QMap<QString, QVariant> >())
-						conditionVariables.push_back(conditionVariablesVar.toMap());
-				}
+			for (const QVariant& conditionVariablesVar : conditionVariablesList)
+			{
```
JorisGoosen: The braces are a waste of space here...

JorisGoosen:
Well it all looks fine, some minor points, but I'll let you decide on them.
Mostly style issues.

Only thing I wonder about is the way "canConvert" is used as it seems like it could lead to weird behaviour if more types are used in that variable.

JASP-Desktop/widgets/qmllistview.cpp
boutinb: Maybe Qt may decide one day that a property value like '{"name": "test" }' can be converted to a string... you never know... OK, I change the order of the if else.

JASP-Desktop/widgets/listmodel.cpp
boutinb: Option class is not a Qt class.

JorisGoosen: Ooh ja...

JASP-Desktop/widgets/qmllistview.cpp
JorisGoosen: Wouldn't it be better to just check the type instead?

PR #4223, opened 2020-08-07
Add Formula Array Input type

boutinb:
Does not have to be implemented for 0.14
Fixes jasp-stats/INTERNAL-jasp#918

### DIFF
--- a/JASP-Desktop/components/JASP/Controls/FormulaField.qml
+++ b/JASP-Desktop/components/JASP/Controls/FormulaField.qml
@@ -25,11 +25,13 @@ TextField
 					id:					formulaField
 					value:				"0"
 	property double realValue:			0
+	property var	realValues:			[]
 	property double	min:				-Infinity
 	property double	max:				Infinity
 	property int	inclusive:			JASP.MinMax
 	property alias	defaultValue:		formulaField.value
+	property bool	multiple:			false
 	property bool	parseDefaultValue:	true
-					inputType:			"formula"
+					inputType:			multiple ? "formulaArray" : "formula"
 					fieldWidth:			jaspTheme.textFieldWidth / 2
 }
--- a/JASP-Desktop/components/JASP/Controls/TableView.qml
+++ b/JASP-Desktop/components/JASP/Controls/TableView.qml
@@ -201,6 +201,7 @@ JASPControl
 						onPressed:				tableView.colSelected = columnIndex
 						onEditingFinished:		tableView.itemChanged(columnIndex, rowIndex, value, inputType)
 						editable:				itemEditable
+						multiple:				itemInputType === "formulaArray"
 					}
 				}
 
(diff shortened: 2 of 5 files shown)


boutinb:
If you set multiple property to true in a FormulaField, then it accepts expression that returns array values.

vandenman:
So I tried this out with the JAGS module and it didn't work (perhaps I did something wrong though).

I added `multiple: true` in `~/jasp-desktop/JASP-Desktop/components/JASP/Controls/TableView.qml` (which drives `BasicThreeButtonTableView` which drives `JagsTableView`) but I still got the old error message. As input I used `rnorm(10)`. When debugging, `_inputType == TextInputType::FormulaType` was `true` but `result` was `""`. I looked a little deeper into the code and noticed the following:

```c++
	// lines 446-456 of jasprcpp.cpp
	SEXP result = jaspRCPP_parseEval(rCodeTryCatch);

	static std::string staticResult;
	try
	{
		staticResult = Rf_isString(result) ? Rcpp::as<std::string>(result) : NullString;
	}
	catch(...)
	{
		staticResult = NullString;
	}
```
If I understand this correctly then `jasprcpp.cpp` tries to convert the result to a `std::string`, but this fails because the SEXP contains a vector of strings. This looks a bit odd to me because [`Rf_isString`](https://lukasstadler.github.io/RAPI/html/_rinternals_8h.html#aa46b54d700f2c05b4ccafc9b386c0e6a) does return true for a vector of strings but I don't think `Rcpp::as<std::string>` can handle that.

boutinb:
The TableView is a bit tricky... it sets the type of the input field in c++. I add a commit so that JAGS table gets it. Can you try again?

vandenman:
The code looks good to me, JAGS works as intended! Something that could perhaps be a little more informative is the part:
```
item()->addControlError(tr("The expression did not return a number."));
```
If you use a non-whitelisted illegal function (but correct code) then the code won't be run but all a user sees is "The expression did not return a number.". But as far as I understand, we're not actually passing the error message back but only a bool with success/ fail.

Unfortunately, Travis exceeded the maximum time for running the unit tests and failed. I've restarted Travis, let's hope it runs a bit faster now.

vandenman:
@boutinb, @AlexanderLyNL would it be possible to put this in 0.14? 

A colleague in the states wants to use the JAGS module (partially for teaching) but they can't fit a particular model right now because the functionality in this PR is missing.

AlexanderLyNL:
What else do we need to do for this?

vandenman:
I think only testing? It worked fine for me on Linux and the code also looks ok to me. We could make the error messages more specific but that's probably a lot of work.

boutinb:
I think this PR is quite safe. The feature is used only for JAGS, and @vandenman tested it. It would be nicer to have a better error message, but that would indeed change the way the R results are handled, and would need much more testing.
As last check: @vandenman, is this code `paste(as.array(" + _value + "), collapse=\"|\")` the right R code to check that a value can be converted to an array (the "|" token is used afterwards to split the results)?

vandenman:
@boutinb `paste(as.array(" + _value + "), collapse=\"|\")` is fine. We should probably check it once more when we're going to R4 but on R3.6.x though. Maybe `paste(" + _value + ", collapse=\"|\")` also works.

PR #4224, opened 2020-08-09
trying to fix a few typos

JoKeyser:
Hi, this is my attempt to point out (and fix) what I think are typos.
(Though I'm not 100% sure about `Tu(r)key`).

I hope this is the correct file to fix this - if not, please point me to the correct one.

### DIFF
--- a/JASP-Engine/JASP/po/R-JASP.pot
+++ b/JASP-Engine/JASP/po/R-JASP.pot
@@ -19,22 +19,22 @@ msgstr ""
 msgid "The number of iterations must be at least 2 iterations higher than the burnin"
 msgstr ""
 
-msgid "%s requres that the dependent variable is positive."
+msgid "%s requires that the dependent variable is positive."
 msgstr ""
 
-msgid "%s requres that the dependent variable is an integer."
+msgid "%s requires that the dependent variable is an integer."
 msgstr ""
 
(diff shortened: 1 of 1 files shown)


AlexanderLyNL:
Hi Johannes, 

Thanks for pointing out these typos. To have these changed, we have to adapt the strings in R, so R-JASP.pot is generated accordingly. @FBartos could you please change the strings in the main code? 

Cheers,
Alexander

JoKeyser:
@AlexanderLyNL , thanks for the info, sorry that this PR isn't directly usable.
If I knew the correct files, I could also change this PR.

For example, searching for 'Turkey' (https://github.com/search?q=org%3Ajasp-stats+turkey&type=Code) yields 2 candidates: 
- https://github.com/jasp-stats/MixedModels/blob/master/R/MixedModelsCommon.R#L3226
- https://github.com/jasp-stats/jasp-desktop/blob/stable/JASP-Engine/JASP/R/MixedModelsCommon.R#L3226
But then the next question is which branch is correct?

FBartos:
Hi Johannes,

first of all, thank you for reporting those typos. No need to worry about updating the files, I will take care of it.

Regarding the duplicity of the entries - this is a result of switching to a more modular version of JASP, as of now, all code is at two places at ones.

Cheers,
František

PR #4225, opened 2020-08-10
fix unit tests for audit

vandenman:
updated unit tests with jfa 0.2.0.


koenderks:
Hmm I was under the impression that the files `test-auditBayesianWorkflow.csv` and `test-auditClassicalWorkflow.csv` needed a change for the unit tests as well, but it seems to work this way.

koenderks (approved):

PR #4226, opened 2020-08-10
Bayesian Linear Regression/ ANOVAs: updates for posterior summary with matched models

### LINKED ISSUE #852: Analysis of Effects: Across Matched Models
Maybe not a bug, but I agree with this user (https://forum.cogsci.nl/index.php?p=/discussion/6290/how-to-compute-the-bayes-factor-inclusion-in-jasp#latest) that it is strange and potentially confusing that the table of effects shows prior and posterior inclusion probabilities that do not compute to the inclusion BF. It seems more straightforward to report the prior and posterior inclusion probabilities that form the basis of the inclusion BF. @vandenman @AlexanderLyNL

vandenman:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/795
Fixes https://github.com/jasp-stats/jasp-issues/issues/852
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/914

The screenshot below shows the new posterior effects table. The table is ordered so that the first columns relate to variable selection and the last columns relate to the parameter estimates.

In addition, the posterior exclusion probabilities are now shown when `Matched models` is ticked. This allows people to compute the inclusion Bayes factor from this table, without having to compute the exclusion probabilities themselves from the other table (e.g., for `contcor1`, `(0.207 / 0.690) / (.222 / .444) = 0.600`.

image

The changes in the error handling function of the ANOVAs silence the unit tests of the repeated measures ANOVAs (https://travis-ci.org/github/vandenman/jasp-desktop/builds/716464613, line 1715 `Error in .subset(x, j) : invalid subscript type 'list'` compared to line 1715 of https://travis-ci.org/github/vandenman/jasp-desktop/builds/716516437 (this PR)).


AlexanderLyNL (approved):
Looks good to me. The reason why `P(exclude | data)` is not in there for across all models is because that's just `1-P(include | data)` isn't it? The fact that the column `P(exclude | data)` "all of a sudden" appears might scare the user.

AlexanderLyNL:
Including `P(exclude | data)` to the table for both cases might be illustrative I think, even though it's a bit redundant for across all models. Either way is fine for me, but having `P(exclude | data)` in the table by default makes it easier to show what across `Matched models` does.

vandenman:
> The reason why P(exclude | data) is not in there for across all models is because that's just 1-P(include | data) isn't it?

Yes.

> Either way is fine for me, but having P(exclude | data) in the table by default makes it easier to show what across Matched models does.

I don't have very strong feelings about that either, just let me know what you prefer and I'll change it to that.

AlexanderLyNL:
Let's have `P(exclude | data)` in by default. Thanks for improving this analysis by the way, it looks really great.

vandenman:
Exclusion probabilities are now always shown.

boutinb:
@TimKDJ: can we merge tis PR?

JASP-Tests/R/tests/testthat/test-anovarepeatedmeasuresbayesian.R
```diff
@@ -90,7 +90,7 @@ test_that("Effects table results match", {
   options$effects <- TRUE
 
   refTables <- list(
-    allModels = list("<unicode>", "RM_FACTOR_1", 0.4, 0, 0.6, 1, 2.08122183080571,
+    allModels = list(1000799917193443, "RM_FACTOR_1", 0.4, 0, 0.6, 1, 2.08122183080571,
```
TimKDJ: Maybe you could use less extreme data so this will pass on different operating systems?

vandenman: I didn't change the input data for this test, but I will now. If you don't mind I'll merge the test for the effects table and those for the main table since the two are independent (and the unit tests timed out a couple of times already).

TimKDJ (approved):

TimKDJ:
Can you make a PR to the module too?

vandenman:
here are the PRs to the modules:

https://github.com/jasp-stats/jaspRegression/pull/7
https://github.com/jasp-stats/jaspAnova/pull/8

these do not contain the changes to the unit tests, because those repos don't have any tests yet ¯\\_(ツ)_/¯

PR #4229, opened 2020-08-10
Label saver

### LINKED ISSUE #935: Minor display glitch in spreadsheet
* JASP version: 0.13.1
* OS name and version: macOS 11.0 Big Sur (public beta)
* Analysis: N/A (spreadsheet)
* Bug description: JASP displays "Column Name/Title here" instead of variable name
* Expected behaviour: the variable name should be displayed

1. Open JASP;
2. open a data set;
3. click on the name of a nominal variable;
4. see the "Column Name/Title here".

This seems to happen only in these conditions:
- only with nominal variables;
- only in the first data set opened after starting JASP: if you close the data set without quitting JASP, and then open any (same or other) data set, the bug does not appear again;
- only for the first time a nominal variable name is clicked: if you click on a subsequent nominal variable name, the bug does not appear again.

variable_name_bug

JorisGoosen:
Fixes https://github.com/jasp-stats/jasp-issues/issues/935 and https://github.com/jasp-stats/jasp-issues/issues/933

### DIFF
--- a/JASP-Desktop/data/datasetpackage.cpp
+++ b/JASP-Desktop/data/datasetpackage.cpp
@@ -170,7 +170,8 @@ QModelIndex DataSetPackage::parent(const QModelIndex & index) const
 
 QModelIndex DataSetPackage::parentModelForType(parIdxType type, int column) const
 {
-	if(type == parIdxType::root) return QModelIndex();
+	if(type == parIdxType::root || column < 0)
+		return QModelIndex();
 
 	return index(int(type), column, QModelIndex());
 }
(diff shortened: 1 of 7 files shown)


JASP-Desktop/data/datasetpackage.cpp
```diff
@@ -409,10 +410,10 @@ void DataSetPackage::resetFilterAllows(size_t columnIndex)
 	_dataSet->column(columnIndex).resetFilter();
 
 	QModelIndex parentModel = parentModelForType(parIdxType::data);
-	emit dataChanged(index(0, columnIndex,	parentModel),	index(rowCount(), columnIndex, parentModel), {int(specialRoles::filter)} );
+	emit dataChanged(DataSetPackage::index(0, columnIndex,	parentModel),	DataSetPackage::index(rowCount(), columnIndex, parentModel), {int(specialRoles::filter)} );
```
boutinb, line 414: Why do you add DataSetPackage prefix here?

JASP-Desktop/data/labelmodel.cpp
```diff
@@ -117,3 +118,11 @@ void LabelModel::columnAboutToBeRemoved(int column)
 	if(proxyParentColumn() == column)
 		setVisible(false);
 }
+
+void LabelModel::columnDataTypeChanged(std::string colName)
+{
+	int colIndex = DataSetPackage::pkg()->getColumnIndex(colName);
+
+	if(colIndex == proxyParentColumn())
+		invalidate();
```
boutinb, line 127: I'm no sure I understand this invalidate(). It invalidates the current filter? But if I try it (I open the VaraiblesWindow, uncheck some labels, and change the type of the column), the filter still exists.

JASP-Desktop/data/datasetpackage.cpp
JorisGoosen: I did it in other places in the file as well and I wanted to be sure I got the custom `index()` function and not some other one. So better safe than sorry.

JASP-Desktop/data/labelmodel.cpp
JorisGoosen: the `invalidate()` is for the QSortFilterProxy, not our filter. It is to make sure that the new items are loaded if the columntype changes.
Normally it would reset after you change the `proxyParentColumn()` (which defines which parent column in the root of DataSetPackageis used). But if you have a column selected and the labelModel is set to it, you then change the type to "scale" and back to "ordinal" then `proxyParentColumn` doesn't change so nothing gets reloaded.

It would be better to emit some kind of `dataChanged` from DataSetPackage, but I couldn't get that to work so I did this.

PR #4236, opened 2020-08-12
Missing Unicode characters on Windows

FransMeerhoff:
- Rewrite of the Reliablity Analysis placed some gettext() around variables.
  This is fixed.
- Removed obsolete reliabilityanalysis.R


JASP-Engine/JASP/R/MixedModelsCommon.R
```diff
@@ -3114,7 +3114,7 @@
 .mmMessageSingularFit <-
   gettext("Model fit is singular. Specified random effects parameters (random intercepts and random slopes) cannot be estimated from the available data. Carefully reduce the random effects structure, but this practice might inflate the reported p-value, and invalidates the analysis.")
 .mmMessageVovkSellke <-
-  gettext("Vovk-Sellke Maximum <em>p</em>-Ratio: Based on a two-sided <em>p</em>-value, the maximum possible odds in favor of H\u2081 over H\u2080 equals 1/(-e <em>p</em> log(<em>p</em>)) for <em>p</em> \u2264 .37 (Sellke, Bayarri, & Berger, 2001).")
+  gettextf("Vovk-Sellke Maximum <em>p</em>-Ratio: Based on a two-sided <em>p</em>-value, the maximum possible odds in favor of H%1$s over H%2$s equals 1/(-e <em>p</em> log(<em>p</em>)) for <em>p</em> %3$s .37 (Sellke, Bayarri, & Berger, 2001).","\u2081","\u2080","\u2264")
```
TimKDJ, line 3117: But this would mean undoing all the unicode code points inside gettext again in all R files... which makes life more difficult for both programmers and translators.

It works on macOS; R and gettext are capable of having unicode code points inside the strings, granted that UTF-8 encoding is maintained throughout.
Somewhere on Windows the encoding is lost. Did you try to see where in the process this could be?
I'll grant you it's not an easy problem to solve and it'll probably require quite a bit of fiddling with encoding / locales, but this is not really optimal as far as solutions go...

JASP-Engine/JASP/R/descriptives.R
```diff
@@ -205,7 +205,7 @@ Descriptives <- function(jaspResults, dataset, options) {
   variables               <- unlist(options$variables)
   equalGroupsNo           <- options$percentileValuesEqualGroupsNo
   percentilesPercentiles  <- unique(options$percentileValuesPercentilesPercentiles)
-  stats                   <- createJaspTable(gettext("Descriptive Statistics"))
+  stats                   <- createJaspTable( gettextf("BF%s%s", "\u2081", "\u2080"))
```
TimKDJ: Don't think you meant to push this

TimKDJ (changes requested):

PR #4237, opened 2020-08-12
Add values attribute to source property

boutinb:
Fixes jasp-stats/INTERNAL-jasp#991
Fixes jasp-stats/INTERNAL-jasp#968
'values' and 'source' properties were handled differently. 'values' is a list of strings (or labels/values), and source is a reference (or a list of references) to other controls. These properties are used in DropDown, VariablesList, ComponentList, InputList, TabView controls.
If we want to be able to mix 'values' with 'sources', 'values' has to be made as a kind of 'source'. This needs a bit of rewrite, so that all these controls use the same code to handle source & values.
One advantage also by handling 'values' as a kind of source is that all the work done to handle changes in the sources, can be used by the values as well: values can also dynamically change, and the model of the control has to be updated as well.

### DIFF
--- a/Desktop/widgets/listmodel.cpp
+++ b/Desktop/widgets/listmodel.cpp
@@ -108,64 +108,18 @@ void ListModel::_connectSourceControls(ListModel* sourceModel, const QSet<QStrin
 
 Terms ListModel::getSourceTerms()
 {
-	const QList<QMLListView::SourceType*>& sourceItems = listView()->sourceModels();
-
 	Terms termsAvailable;
-	if (sourceItems.size() == 0)
-		return termsAvailable;
-
 	Terms termsToCombine;
 	Terms termsToBeCombinedWith;
 
-	for (QMLListView::SourceType* sourceItem : sourceItems)
+	for (const std::pair<QMLListView::SourceType *, Terms>& source : listView()->getTermsPerSource())
 	{
+		QMLListView::SourceType* sourceItem = source.first;
+		const Terms& sourceTerms = source.second;
 		ListModel* sourceModel = sourceItem->model;
 		if (sourceModel)
 		{
-			Terms sourceTerms = sourceModel->terms(sourceItem->modelUse);
-
-			for (const QMLListView::SourceType& discardModel : sourceItem->getDiscardModels())
-				sourceTerms.discardWhatDoesContainTheseComponents(discardModel.model->terms(discardModel.modelUse));
-
 			_connectSourceControls(sourceModel, sourceItem->usedControls);
-
-			if (!sourceItem->conditionExpression.isEmpty())
-			{
-				Terms filteredTerms;
-				QJSEngine jsEngine;
-
-				for (const Term& term : sourceTerms)
-				{
-					for (const QMLListView::SourceType::ConditionVariable& conditionVariable : sourceItem->conditionVariables)
-					{
-						JASPControlWrapper* control = sourceModel->getRowControl(term.asQString(), conditionVariable.controlName);
-						if (control)
-						{
-							QJSValue value;
-							QVariant valueVar = control->getItemProperty(conditionVariable.propertyName);
-
-							switch (valueVar.type())
-							{
-							case QVariant::Type::Int:
-							case QVariant::Type::UInt:		value = valueVar.toInt();		break;
-							case QVariant::Type::Double:	value = valueVar.toDouble();	break;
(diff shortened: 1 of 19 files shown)


JorisGoosen:
Not related to an issue?

JorisGoosen:
Big changes for something without an issue! :o

I like that you generalized the combobox into a listview btw.
I think this can be merged (I have not tested it out, but it is a bit of a too large change to be able to grok where I need to look exactly) but I would like it if you amended the commit message and described what the changes do exactly.

JorisGoosen (changes requested):
Please add a description in the commit of what this big commit does.

boutinb:
Maybe this PR is not completed needed for 0.14. We have to see with @FBartos: In jasp-stats/INTERNAL-jasp#991, you need to add 'Intercept' in the list, but I showed that there was a workaround with a hidden DropDown. Would it be enough for 0.14?

FBartos:
Yes, I think so. Also, I think that we can keep this change to Mix Models till the next release. I will probably add more features to them (allowing LMs and GLMs to be fitted as well), so we can bundle it together.

boutinb:
OK, nice. This PR was a bit too dangerous just before the release.

Desktop/widgets/listmodel.cpp
```diff
@@ -108,65 +108,19 @@ void ListModel::_connectSourceControls(ListModel* sourceModel, const QSet<QStrin
...
-
 	Terms termsAvailable;
-	if (sourceItems.size() == 0)
-		return termsAvailable;
-
 	Terms termsToCombine;
 	Terms termsToBeCombinedWith;
 
-	for (QMLListView::SourceType* sourceItem : sourceItems)
+	for (const std::pair<QMLListView::SourceType *, Terms>& source : listView()->getTermsPerSource())
```
JorisGoosen, line 115: `const auto & sourceTerms` works quite well too ;)

Desktop/widgets/listmodellabelvalueterms.cpp
```diff
@@ -164,3 +89,20 @@ int ListModelLabelValueTerms::getIndexOfValue(const QString &value)
...
 }
 
+void ListModelLabelValueTerms::setLabelValues(const QMLListView::LabelValueMap &labelvalues)
+{
+	_valueToLabelMap.clear();
+	_labelToValueMap = labelvalues;
+	_terms.clear();
+
+	QMapIterator<QString, QString> it(labelvalues);
+	while (it.hasNext())
```
JorisGoosen, line 99: A nice for-loop here would have been cool though not necessary

Desktop/widgets/qmllistview.cpp
```diff
@@ -179,7 +247,17 @@ void QMLListView::setSources()
...
 		for (SourceType* sourceItem : _sourceModels)
 		{
-			ListModel* sourceModel = form()->getModel(sourceItem->name);
+			ListModel* sourceModel = nullptr;
+			if (!sourceItem->isValuesSource)
+			{
+				if (form())
+					sourceModel = form()->getModel(sourceItem->name);
+				else
+					Log::log() << "Cannot use a source property outside of an Anlysis Form" << std::endl;
```
JorisGoosen, line 256: Anlysis -> Analysis

Desktop/widgets/qmllistview.cpp
```diff
@@ -292,6 +386,65 @@ void QMLListView::setTermsAreInteractions()
...
+		{
+			Terms sourceTerms = sourceModel->terms(sourceItem->modelUse);
+
+			for (const SourceType& discardModel : sourceItem->getDiscardModels())
+				sourceTerms.discardWhatDoesContainTheseComponents(discardModel.model->terms(discardModel.modelUse));
+
+			if (!sourceItem->conditionExpression.isEmpty())
+			{
+				Terms filteredTerms;
+				QJSEngine jsEngine;
```
JorisGoosen, line 406: Would it be nice to have this static?
Or is the initialization of QJSEngine light enough that we don't need to care?

JorisGoosen: Another option would be to just use the QQmlEngine we already have running?
https://doc.qt.io/qt-5/qqmlengine.html shows me it inherits QJSEngine so it should be able to do everything we want and because QQMLListView is already a QML element it should already have access to it I think.

JorisGoosen (approved):

PR #4238, opened 2020-08-13
RoBMA updates

FBartos:
adding a possibility to analyze OR
fixing "too tight" secondary y-axis labels
(most of the code changes are just reformating)

relevant things included in RoBMA 1.04
- more and better error messages

things we might want from RoBMA 1.05 (not released yet, won't get to CRAN in time)
- https://github.com/FBartos/RoBMA/tree/development
- fix for inability to plot a bivariate mu vs tau estimates plot
- range of x-values always cover 0 (or 1 when ORs are analyzed) for individual models parameters plots

**read before testing**
- the development version of the package needs to be installed (https://github.com/FBartos/RoBMA/tree/development)
- fitting models with default settings take long (especially with an increasing number of studies) = using debug dataset is a very bad idea (including the fact that the variables there were not generated by a meta-analytic model, which further impacts estimation time)
- you can use this file instead ([MA_test.txt](https://github.com/jasp-stats/jasp-desktop/files/5070115/MA_test.txt)), it contains all of the data input types, just note that the values in the rows don't correspond, so different columns will lead do different results
- decreasing the MCMC settings number also significantly decreases fitting time 
image
- or load a prefitted [model](https://drive.google.com/file/d/1tc8pu8qp5kOBLjlKhEWQdk8RPLHNFtum/view?usp=sharing) using the "Fitted model" option
- clicking multiple options in short succession can crash JASP engine and the whole model might as a results need refitting

### DIFF
--- a/JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
+++ b/JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
@@ -19,68 +19,68 @@
 RobustBayesianMetaAnalysis <- function(jaspResults, dataset, options, state = NULL) {
 
   # clean fitted model if it was changed
-  if(!.RoBMAready(options))
-    .RoBMAcleanModel(jaspResults)
+  if(!.robmaReady(options))
+    .robmaCleanModel(jaspResults)
  
   # load data
-  if (.RoBMAready(options))
-    dataset <- .RoBMAdataGet(options, dataset)
+  if (.robmaReady(options))
+    dataset <- .robmaGetData(options, dataset)
   
   # get the priors
-  .RoBMApriorsGet(jaspResults, options)
+  .robmaGetPriors(jaspResults, options)
   
   # show the model preview
   if (is.null(jaspResults[["model"]]))
-    .RoBMAmodelPreview(jaspResults, options)
+    .robmaModelPreviewTable(jaspResults, options)
 
   # fit model model
-  if (is.null(jaspResults[["model_notifier"]]) && .RoBMAready(options))
-    .RoBMAfitModel(jaspResults, dataset, options)
+  if (is.null(jaspResults[["model_notifier"]]) && .robmaReady(options))
+    .robmaFitModel(jaspResults, dataset, options)
   
   ### Priors plot
   if (options[["priors_plot"]])
-    .RoBMApriorsPlots(jaspResults, options)
+    .robmaPriorsPlots(jaspResults, options)
   
   ### Inference, Plots, and Diagnostics are accessible only if a model is fitted
   if (!is.null(jaspResults[["model"]])) {
     ### Inference
     # defaul summary
-    .RoBMAsummary(jaspResults, options)
+    .robmaSummaryTable(jaspResults, options)
     # models overview
     if (options[["results_models"]])
-      .RoBMAmodelsOverview(jaspResults, options)
+      .robmaModelsOvervievTable(jaspResults, options)
     # models summary
     if (options[["results_individual"]])
-      .RoBMAmodelsSummary(jaspResults, options)
+      .robmaModelsSummaryTable(jaspResults, options)
 
     ### Plots
     # pooled estimates plots
     if (options[["plots_theta"]])
-      .RoBMAplots(jaspResults, options, "theta")
+      .robmaPlots(jaspResults, options, "theta")
     if (options[["plots_mu"]])
-      .RoBMAplots(jaspResults, options, "mu")
+      .robmaPlots(jaspResults, options, "mu")
     if (options[["plots_tau"]])
-      .RoBMAplots(jaspResults, options, "tau")
+      .robmaPlots(jaspResults, options, "tau")
     if (options[["plots_tau"]]                &&
         options[["plots_mu"]]                 &&
         options[["plots_type"]] == "averaged" &&
         !options[["plots_priors"]])
-      .RoBMAplots(jaspResults, options, c("mu", "tau"))
+      .robmaPlots(jaspResults, options, c("mu", "tau"))
     if (options[["plots_omega"]])
-      .RoBMAplots(jaspResults, options, "omega")
+      .robmaPlots(jaspResults, options, "omega")
     
     # individual models
     if (options[["plots_individual_mu"]])
-      .RoBMAindividualPlots(jaspResults, options, "mu")
+      .robmaModelsPlots(jaspResults, options, "mu")
     if (options[["plots_individual_tau"]])
-      .RoBMAindividualPlots(jaspResults, options, "tau")
+      .robmaModelsPlots(jaspResults, options, "tau")
     if (options[["plots_individual_omega"]])
-      .RoBMAindividualPlots(jaspResults, options, "omega")
+      .robmaModelsPlots(jaspResults, options, "omega")
     
     ### Diagnostics
     # overview
     if (options[["diagnostics_overview"]])
-      .RoBMAdiagnosticsOverview(jaspResults, options)
+      .robmaDiagnosticsOverviewTable(jaspResults, options)
     # plots
     if ((
       options[["diagnostics_mu"]]    ||
@@ -93,18 +93,18 @@ RobustBayesianMetaAnalysis <- function(jaspResults, dataset, options, state = NU
       options[["diagnostics_autocorrelation"]] ||
       options[["diagnostics_samples"]]
     ))
-      .RoBMAdiagnosticsPlots(jaspResults, options)
+      .robmaDiagnosticsPlots(jaspResults, options)
     
     ### Save the model
     if (options[["save_path"]] != "" &&
         is.null(jaspResults[["model_saved"]]))
-      .RoBMAsaveModel(jaspResults, options)
+      .robmaSaveModel(jaspResults, options)
   }
   
   return()
 }
 
-.RoBMAdependencies <- c(
+.robmaDependencies <- c(
   "measures",
   "fitted_path",
   "cohensd_testType",
@@ -147,8 +147,8 @@ RobustBayesianMetaAnalysis <- function(jaspResults, dataset, options, state = NU
   "advanced_mu_transform"
 )
 # priors related functions
-.RoBMAoptions2priors      <- function(options_prior) {
-  options_prior <- .RoBMAoptions2priorsEval(options_prior)
+.robmaOptions2Priors      <- function(options_prior) {
+  options_prior <- .robmaOptions2PriorsEval(options_prior)
   
   if (options_prior[["type"]] == "normal") {
     return(
@@ -298,7 +298,7 @@ RobustBayesianMetaAnalysis <- function(jaspResults, dataset, options, state = NU
     )
   }
 }
-.RoBMAoptions2priorsClean <- function(x) {
+.robmaOptions2PriorsClean <- function(x) {
   
   x <- trimws(x, which = "both")
   x <- trimws(x, which = "both", whitespace = "c")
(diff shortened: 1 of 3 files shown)


Resources/Meta Analysis/qml/RobustBayesianMetaAnalysis.qml
```diff
@@ -28,6 +28,16 @@ Form
...
+		onValueChanged:	if(measures_correlation.checked) {
+			advanced_mu_transform.value				= "cohens_d"
+			// TODO: make this work >>
+			advanced_mu_transform_log_OR.enabled	= true
+			advanced_mu_transform_fishers_z.enabled	= false
+		} else if (measures_OR.checked){
+			advanced_mu_transform.value				= "log_OR"
+			advanced_mu_transform_log_OR.enabled	= false
+			advanced_mu_transform_fishers_z.enabled	= true
+		}
```
FBartos, line 40: Maybe @boutinb  could help?
I manage to change the default value of the at line 1852 using the line 32 and 37, but I can't figure out how to disable the "log_OR" option when correlations are selected and "fishers_z" option when "OR" are selected. 

I also tried setting the values and value directly here - the correct value and options are passed, but the dropdown box is missing labels:
```
onValueChanged:	if(measures_correlation.checked) {
			advanced_mu_transform.values = 	
			[
				{ label: qsTr("Cohen's d"),		value: "cohens_d"},
				{ label: qsTr("Fisher's z"),	value: "fishers_z"}
			]
			advanced_mu_transform.value  = "cohens_d"
		} else if (measures_OR.checked){
			advanced_mu_transform.values	 = 	
			[
				{ label: qsTr("log(OR)"),		value: "log_OR"},
				{ label: qsTr("Cohen's d"),		value: "cohens_d"}
			]
			advanced_mu_transform.value  = "log_OR"
		}
```

FBartos:
also, a known problem is that not all symbols are properly visualized in JASP generated plots (Rstudio versions are OK), as reported here: https://github.com/jasp-stats/INTERNAL-jasp/issues/1003n and with the same issue in Learn Bayes and Distributions modules. The missing symbols are infinities for a change.

image

JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
```diff
@@ -16,96 +16,95 @@
...
-          is.null(jaspResults[["model_saved"]]))
-        .RoBMA_save_model(jaspResults, options)
-    }
-    
-    return()
+RobustBayesianMetaAnalysis <- function(jaspResults, dataset, options, state = NULL) {
+
+  # clean fitted model if it was changed
+  if(!.RoBMAready(options))
+    .RoBMAcleanModel(jaspResults)
```
AlexanderLyNL: So your function is more an action

.cleanModelRobma

JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
```diff
@@ -16,96 +16,95 @@
...
+ 
+  # load data
+  if (.RoBMAready(options))
+    dataset <- .RoBMAdataGet(options, dataset)
```
AlexanderLyNL: So your function is more an action

.getDataRobma

JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
```diff
@@ -16,96 +16,95 @@
...
+  
+  # get the priors
+  .RoBMApriorsGet(jaspResults, options)
```
AlexanderLyNL: So your function is more an action

.getPriorsRobma

JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
```diff
@@ -16,96 +16,95 @@
...
+  
+  # show the model preview
+  if (is.null(jaspResults[["model"]]))
+    .RoBMAmodelPreview(jaspResults, options)
```
AlexanderLyNL: So your function is more an action

.previewModelRobma

JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
```diff
@@ -16,96 +16,95 @@
...
+
+  # fit model model
+  if (is.null(jaspResults[["model_notifier"]]) && .RoBMAready(options))
+    .RoBMAfitModel(jaspResults, dataset, options)
```
AlexanderLyNL: .fitModelRobma

JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
```diff
@@ -16,96 +16,95 @@
...
+  
+  ### Priors plot
+  if (options[["priors_plot"]])
+    .RoBMApriorsPlots(jaspResults, options)
```
AlexanderLyNL: plotPriorsRobma

JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
```diff
@@ -16,96 +16,95 @@
...
+  # show the model preview
+  if (is.null(jaspResults[["model"]]))
+    .RoBMAmodelPreview(jaspResults, options)
+
+  # fit model model
+  if (is.null(jaspResults[["model_notifier"]]) && .RoBMAready(options))
+    .RoBMAfitModel(jaspResults, dataset, options)
+  
+  ### Priors plot
+  if (options[["priors_plot"]])
```
AlexanderLyNL, line 41: plotPriors

JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
```diff
@@ -16,96 +16,95 @@
...
+    .RoBMApriorsPlots(jaspResults, options)
+  
+  ### Inference, Plots, and Diagnostics are accessible only if a model is fitted
+  if (!is.null(jaspResults[["model"]])) {
+    ### Inference
+    # defaul summary
+    .RoBMAsummary(jaspResults, options)
```
AlexanderLyNL: What does the summary function do? 

.printSummaryRobma
.makeSummaryRobma
.makeSummaryTableRobma

?

JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
```diff
@@ -16,96 +16,95 @@
...
+    # models overview
+    if (options[["results_models"]])
```
AlexanderLyNL, line 50: I'm not sure what resultsModels is and what it corresponds to from the qml input

JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
```diff
@@ -16,96 +16,95 @@
...
+      .RoBMAmodelsOverview(jaspResults, options)
```
AlexanderLyNL: What does this function do? Does it make a table?

JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
```diff
@@ -16,96 +16,95 @@
...
+    if (options[["save_path"]] != "" &&
+        is.null(jaspResults[["model_saved"]]))
+      .RoBMAsaveModel(jaspResults, options)
   }
+  
+  return()
+}
 
-.RoBMA_dependencies <- c(
+.RoBMAdependencies <- c(
```
AlexanderLyNL: I'd recommend using camelCased variable names

JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
```diff
@@ -1238,98 +1062,73 @@ RobustBayesianMetaAnalysis <-
...
+  
   # estimated studies table
   if (options[["results_theta"]]) {
-    studies_summary <-
-      createJaspTable(title = gettextf("Model Averaged Estimated Studies' Effects (%s)", "\u03B8"))
+    studies_summary <- createJaspTable(title = gettextf("Model Averaged Estimated Studies' Effects (%s)", "\u03B8"))
     studies_summary$position <- 4
-    studies_summary <-
-      .RoBMA_table_fill_studies(studies_summary, s.fit[["averaged"]], s.fit[["add_info"]], options)
+    studies_summary <- .RoBMAtableFillStudies(studies_summary, s.fit[["averaged"]], s.fit[["add_info"]], options)
```
AlexanderLyNL: Is there a reason why you now use a dot in this variable name? s.fit?

AlexanderLyNL (commented):
I went through this very quickly and noticed some style things

JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
Kucharssim: If I may weight in, I disagree with this suggestion. I perceive the `.RoBMA` as a prefix for the RoBMA functions, similarly as we have for many analyses (e.g., `.cfaReadData()`, `.cfaComputeResults()`, `.bainCleanModelInput()`). Moving the prefix to the end would break consistency of the naming patterns.

AlexanderLyNL: The reason for this suggestion is that I have a particular hard time reading the code. This might be due to 
- Github formatting
- RoBMA in capitals shouting at me
- Functions names such as `.RoBMAdataGet`, `.RoBMApriorGet` emphasising RoBMA and the Get part, rather than "it getting data", which is the information I need. Furthermore, it's not clear from the function name where to break off the statement. Capitals in camelCasing denote a new word, but with `RoBMAdataGet` I get confused. It's hard for me to figure out where to chop things off, because the capitals confuse me and the fact that `data` does not start with a capital also confuse me. 
- Function names `.robmaGetData` and `.getDataRobma` provide me with the right information directly. The emphasis of the action is immediate with the last two options, as it is easier for me to chop off `robma` resulting in `GetData`. The version `.getDataRobma` is easiest for me to read, as I get the information immediately, and I can just stop reading when I hit Rob... The information that it's a RoBMA function is not relevant for me because I know I'm reviewing the RoBMA analysis. I agree that the latter is not consistent with some other analyses, but it definitely makes things clearer.
- Some other functions names being a bit generic, such as `.RoBMAsummary` rather than `makeSummaryTableRobma` etc etc. I don't know what `.RoBMAmodelsOverview` does Does it produce a table? A plot? A computation?

Kucharssim: Oh yeah, that I agree with - `.robmaGetData()` is better than `.RoBMAgetData()`, and I do agree to stress the actions more. As for whether or not to keep the prefixes (which is the point of disagreement), we can have a discussion whether the name of the analysis can be read better as prefix or a suffix, but I would still vote for a prefix `.robmaNameOfFunction()` instead of suffix `.nameOfFunctionRobma()`, as that's what we were doing consistently (or at least trying to) for a while.

FBartos: addressed by another renaming

JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
FBartos: changed

JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
FBartos: changed

JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
FBartos: changed

JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
FBartos: it makes overview tables, changed

JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
FBartos: no, changed

JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
FBartos: changed

JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
FBartos: changed

AlexanderLyNL (approved):
Some function names can still be more descriptive, but I think it'll be good to do that later.

AlexanderLyNL:
Please also make this PR to https://github.com/jasp-stats/jaspMetaAnalysis

PR #4239, opened 2020-08-14
Selection Models - adding correlation input

FBartos:
small changes allowing for using correlations and sample sizes as an input into the selection models analysis

### DIFF
--- a/JASP-Engine/JASP/R/selectionmodels.R
+++ b/JASP-Engine/JASP/R/selectionmodels.R
@@ -17,46 +17,46 @@
 
 SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
   
-  if (.SMready(options)) {
-    dataset <- .SMdataGet(dataset, options)
-    dataset <- .SMdataCheck(jaspResults, dataset, options)
+  if (.smReady(options)) {
+    dataset <- .smGetData(dataset, options)
+    dataset <- .smCheckData(jaspResults, dataset, options)
   }
   
   # fit the models
-  if (.SMready(options))
-    .SMfit(jaspResults, dataset, options)
+  if (.smReady(options))
+    .smFit(jaspResults, dataset, options)
   
   # main summary tables
-  .SMfitTests(jaspResults, dataset, options)
-  .SMfitEstimates(jaspResults, dataset, options)
+  .smTestsTables(jaspResults, dataset, options)
+  .smEstimatesTables(jaspResults, dataset, options)
   
   # the p-value frequency tables
   if (options[["p_table"]])
-    .SMpFrequency(jaspResults, dataset, options)
+    .smPFrequencyTable(jaspResults, dataset, options)
   
   # figures
   if (options[["FE_weightfunction"]])
-    .SMweightsPlot(jaspResults, dataset, options, "FE")
+    .smWeightsPlot(jaspResults, dataset, options, "FE")
   if (options[["RE_weightfunction"]])
-    .SMweightsPlot(jaspResults, dataset, options, "RE")
+    .smWeightsPlot(jaspResults, dataset, options, "RE")
   if (options[["plot_models"]])
-    .SMestimatesPlot(jaspResults, dataset, options)
+    .smEstimatesPlot(jaspResults, dataset, options)
   
   return()
 }
 
-.SMdependencies <- c(
+.smDependencies <- c(
   "auto_reduce", "cutoffs_p", "selection_twosided", "effect_direction", "measures", "mu_transform",
   "input_ES", "input_SE", "input_N", "input_p"
 )
-.SMready              <- function(options) {
+.smReady                   <- function(options) {
   if (options[["measures"]] == "general") {
     return(options[["input_ES"]] != "" && options[["input_SE"]] != "")
   } else if (options[["measures"]] == "correlation") {
     return(options[["input_ES"]] != "" && options[["input_N"]] != "")
   }
 }
-.SMdataGet            <- function(dataset, options) {
+.smGetData                 <- function(dataset, options) {
   if (!is.null(dataset)) {
     return(dataset)
   } else {
@@ -68,7 +68,7 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
     )))
   }
 }
-.SMdataCheck          <- function(jaspResults, dataset, options) {
+.smCheckData               <- function(jaspResults, dataset, options) {
   
   dataset_old <- dataset
   dataset     <- na.omit(dataset)
@@ -79,7 +79,7 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
       ommited_observations <- jaspResults[["ommited_observations"]]
     } else {
       ommited_observations <- createJaspState()
-      ommited_observations$dependOn(c(.SMdependencies))
+      ommited_observations$dependOn(c(.smDependencies))
       jaspResults[["ommited_observations"]] <- ommited_observations
     }
     ommited_observations$object <- nrow(dataset_old) - nrow(dataset)
@@ -118,7 +118,7 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
   
   return(dataset)
 }
-.SMpcutoffsGet        <- function(options) {
+.smGetCutoffs              <- function(options) {
   
   x <- options[["cutoffs_p"]]
   x <- trimws(x, which = "both")
@@ -151,7 +151,7 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
   
   return(x)
 }
-.SMautoreduce         <- function(steps, pval) {
+.smAutoreduceCutoffs       <- function(steps, pval) {
   
   ### this is probably the most complex part of the analysis
   # the fit function protests when there is any p-value interval with lower than 4 p-values
@@ -187,7 +187,7 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
   
   return(steps)
 }
-.SMfillEstimates      <- function(jaspResults, table, fit, options) {
+.smFillEstimates           <- function(jaspResults, table, fit, options) {
   
   CI_overtitle <- gettext("95% Confidence Interval")
   
@@ -228,8 +228,8 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
         uCI  = fit[["ci.ub_adj"]][mean_pos,1]
       ))
       
-      note_messages    <- .SMnotes(jaspResults, fit, options)      
-      warning_messages <- .SMwarningMessages(fit)
+      note_messages    <- .smSetNoteMessages(jaspResults, fit, options)      
+      warning_messages <- .smSetWarningMessages(fit)
       for(i in seq_along(note_messages)) {
         table$addFootnote(symbol = gettext("Note:"),    note_messages[i])
       }
@@ -238,7 +238,7 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
       }
       
     } else {
-      table$setError(.SMerrorMessage(fit))
+      table$setError(.smSetErrorMessage(fit))
     }    
   }
   
@@ -247,7 +247,7 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
   
   return(table)
 }
-.SMfillHeterogeneity  <- function(jaspResults, table, fit, options) {
+.smFillHeterogeneity       <- function(jaspResults, table, fit, options) {
   
   CI_overtitle <- gettext("95% Confidence Interval")
   
@@ -286,8 +286,8 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
           if (options[["mu_transform"]] == "cohens_d") gettext("Cohen's <em>d</em>") else gettext("Fisher's <em>z</em>")
         ))
       
-      note_messages    <- .SMnotes(jaspResults, fit, options)      
-      warning_messages <- .SMwarningMessages(fit)
+      note_messages    <- .smSetNoteMessages(jaspResults, fit, options)      
+      warning_messages <- .smSetWarningMessages(fit)
       for(i in seq_along(note_messages)) {
         table$addFootnote(symbol = gettext("Note:"),    note_messages[i])
       }
@@ -296,7 +296,7 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
       }
       
     } else {
-      table$setError(.SMerrorMessage(fit)) 
+      table$setError(.smSetErrorMessage(fit)) 
     }
   }
   
@@ -305,7 +305,7 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
   
   return(table)
 }
-.SMfillWeights        <- function(jaspResults, table, fit, options) {
+.smFillWeights             <- function(jaspResults, table, fit, options) {
   
   CI_overtitle <- gettext("95% Confidence Interval")
   p_overtitle  <- gettext("<em>p</em>-values interval(one-sided)")
@@ -353,8 +353,8 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
         table$addRows(temp_row)
       }
       
-      note_messages    <- .SMnotes(jaspResults, fit, options)
-      warning_messages <- .SMwarningMessages(fit)
+      note_messages    <- .smSetNoteMessages(jaspResults, fit, options)
+      warning_messages <- .smSetWarningMessages(fit)
       for(i in seq_along(note_messages)) {
         table$addFootnote(symbol = gettext("Note:"),    note_messages[i])
       }
@@ -363,31 +363,31 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
       }
       
     } else {
-      table$setError(.SMerrorMessage(fit)) 
+      table$setError(.smSetErrorMessage(fit)) 
     }
   }
   
   return(table)
 }
-.SMfit                <- function(jaspResults, dataset, options) {
+.smFit                     <- function(jaspResults, dataset, options) {
   
   if (!is.null(jaspResults[["models"]])) {
     return()
   } else {
     models <- createJaspState()
-    models$dependOn(c(.SMdependencies, "p_table"))
+    models$dependOn(c(.smDependencies, "p_table"))
     jaspResults[["models"]] <- models
   }
   
   # get the p-value steps
-  steps <- .SMpcutoffsGet(options)
+  steps <- .smGetCutoffs(options)
   
   # get the p-values
-  pval  <- .SMmodelInputPval(dataset, options)
+  pval  <- .smGetInputPval(dataset, options)
   
   # remove intervals that do not contain enought(3) p-values
   if (options[["auto_reduce"]]) {
-    steps <- tryCatch(.SMautoreduce(steps, pval), error = function(e)e)
+    steps <- tryCatch(.smAutoreduceCutoffs(steps, pval), error = function(e)e)
     
     if (class(steps) %in% c("simpleError","error")) {
       models[["object"]] <- list(
@@ -400,16 +400,16 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
   
   # fit the models
   fit_FE <- tryCatch(weightr::weightfunct(
-    effect = .SMmodelInputES(dataset, options),
-    v      = .SMmodelInputVAR(dataset, options),
+    effect = .smGetInputES(dataset, options),
+    v      = .smGetInputVAR(dataset, options),
     pval   = pval,
     steps  = steps,
     fe     = TRUE
   ),error = function(e)e)
   
   fit_RE <- tryCatch(weightr::weightfunct(
-    effect = .SMmodelInputES(dataset, options),
-    v      = .SMmodelInputVAR(dataset, options),
+    effect = .smGetInputES(dataset, options),
+    v      = .smGetInputVAR(dataset, options),
     pval   = pval,
     steps  = steps,
     fe     = FALSE
@@ -417,14 +417,14 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
   
   # take care of the possibly turned estimates
   if (options[["effect_direction"]] == "negative") {
-    fit_FE <- .SMturnEstimates(fit_FE)
-    fit_RE <- .SMturnEstimates(fit_RE)  
+    fit_FE <- .smTurnEstimatesDirection(fit_FE)
+    fit_RE <- .smTurnEstimatesDirection(fit_RE)  
   }
   
   # take care of the transformed estimates
   if (options[["measures"]] == "correlation") {
-    fit_FE <- .SMtransformEstimates(fit_FE, options)
-    fit_RE <- .SMtransformEstimates(fit_RE, options)
+    fit_FE <- .smTransformEstimates(fit_FE, options)
+    fit_RE <- .smTransformEstimates(fit_RE, options)
   }
   
   models[["object"]] <- list(
@@ -434,7 +434,7 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
   
   return()
 }
-.SMfitTests           <- function(jaspResults, dataset, options) {
+.smTestsTables             <- function(jaspResults, dataset, options) {
   
   if (!is.null(jaspResults[["fit_tests"]])) {
     return()
@@ -442,7 +442,7 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
     # create container
     fit_tests <- createJaspContainer(title = gettext("Model Tests"))
     fit_tests$position <- 1
-    fit_tests$dependOn(.SMdependencies)
+    fit_tests$dependOn(.smDependencies)
     jaspResults[["fit_tests"]] <- fit_tests
   }
   
@@ -481,13 +481,13 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
     heterogeneity_test$addRows(row_heterogeneity)
     
     note_messages <- unique(c(
-      .SMnotes(jaspResults, models[["FE"]], options), .SMnotes(jaspResults, models[["RE"]], options)
+      .smSetNoteMessages(jaspResults, models[["FE"]], options), .smSetNoteMessages(jaspResults, models[["RE"]], options)
     ))
     warning_messages <- unique(c(
-      .SMwarningMessages(models[["FE"]]), .SMwarningMessages(models[["RE"]])
+      .smSetWarningMessages(models[["FE"]]), .smSetWarningMessages(models[["RE"]])
     ))
     error_messages   <- unique(c(
-      .SMerrorMessage(models[["FE"]], "FE"), .SMerrorMessage(models[["RE"]], "RE")
+      .smSetErrorMessage(models[["FE"]], "FE"), .smSetErrorMessage(models[["RE"]], "RE")
     ))
     for(i in seq_along(note_messages)) {
       heterogeneity_test$addFootnote(symbol = gettext("Note:"), note_messages[i])
@@ -499,8 +499,8 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
       heterogeneity_test$addFootnote(symbol = gettext("Error:"),   error_messages[i])
     }
   } else {
-    if (!.SMready(options) && options[["input_p"]] != "")
-      heterogeneity_test$addFootnote(symbol = gettext("Note:"), .SMnotes(NULL, NULL, options))
+    if (!.smReady(options) && options[["input_p"]] != "")
+      heterogeneity_test$addFootnote(symbol = gettext("Note:"), .smSetNoteMessages(NULL, NULL, options))
   }
   
   
@@ -546,13 +546,13 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
     bias_test$addRows(row_bias_heterogeneity)
     
     note_messages <- unique(c(
-      .SMnotes(jaspResults, models[["FE"]], options), .SMnotes(jaspResults, models[["RE"]], options)
+      .smSetNoteMessages(jaspResults, models[["FE"]], options), .smSetNoteMessages(jaspResults, models[["RE"]], options)
     ))
     warning_messages <- unique(c(
-      .SMwarningMessages(models[["FE"]]), .SMwarningMessages(models[["RE"]])
+      .smSetWarningMessages(models[["FE"]]), .smSetWarningMessages(models[["RE"]])
     ))
     error_messages   <- unique(c(
-      .SMerrorMessage(models[["FE"]], "FE"), .SMerrorMessage(models[["RE"]], "RE")
+      .smSetErrorMessage(models[["FE"]], "FE"), .smSetErrorMessage(models[["RE"]], "RE")
     ))
     for(i in seq_along(note_messages)) {
       bias_test$addFootnote(symbol = gettext("Note:"),    note_messages[i])
@@ -565,13 +565,13 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
     }
     
   } else {
-    if (!.SMready(options) && options[["input_p"]] != "")
-      bias_test$addFootnote(symbol = gettext("Note:"), .SMnotes(NULL, NULL, options))
+    if (!.smReady(options) && options[["input_p"]] != "")
+      bias_test$addFootnote(symbol = gettext("Note:"), .smSetNoteMessages(NULL, NULL, options))
   }
   
   return()
 }
-.SMfitEstimates       <- function(jaspResults, dataset, options) {
+.smEstimatesTables         <- function(jaspResults, dataset, options) {
   
   models   <- jaspResults[["models"]]$object
   
@@ -580,7 +580,7 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
     # create container
     FE_estimates <- createJaspContainer(title = gettext("Fixed Effects Estimates"))
     FE_estimates$position <- 2
-    FE_estimates$dependOn(c(.SMdependencies, "FE_estimates"))
+    FE_estimates$dependOn(c(.smDependencies, "FE_estimates"))
     jaspResults[["FE_estimates"]] <- FE_estimates
   } else {
     FE_estimates <- jaspResults[["FE_estimates"]]    
@@ -594,7 +594,7 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
     ))
 FE_estimates_mean$position  <- 1
 FE_estimates[["FE_mean"]] <- FE_estimates_mean
-FE_mean <- .SMfillEstimates(jaspResults, FE_estimates_mean, models[["FE"]], options)    
+FE_mean <- .smFillEstimates(jaspResults, FE_estimates_mean, models[["FE"]], options)    
   }
   
   # weights estimates
@@ -603,7 +603,7 @@ FE_mean <- .SMfillEstimates(jaspResults, FE_estimates_mean, models[["FE"]], opti
     FE_weights$position  <- 2
     FE_weights$dependOn(c("FE_weights"))
     FE_estimates[["FE_weights"]] <- FE_weights
-    FE_weights <- .SMfillWeights(jaspResults, FE_weights, models[["FE"]], options)
+    FE_weights <- .smFillWeights(jaspResults, FE_weights, models[["FE"]], options)
   }
   
   
@@ -612,7 +612,7 @@ FE_mean <- .SMfillEstimates(jaspResults, FE_estimates_mean, models[["FE"]], opti
     # create container
     RE_estimates <- createJaspContainer(title = gettext("Random Effects Estimates"))
     RE_estimates$position <- 3
-    RE_estimates$dependOn(c(.SMdependencies, "RE_estimates"))
+    RE_estimates$dependOn(c(.smDependencies, "RE_estimates"))
     jaspResults[["RE_estimates"]] <- RE_estimates
   } else {
     RE_estimates <- jaspResults[["RE_estimates"]]    
@@ -626,7 +626,7 @@ FE_mean <- .SMfillEstimates(jaspResults, FE_estimates_mean, models[["FE"]], opti
     ))
     RE_estimates_mean$position <- 1
     RE_estimates[["RE_mean"]] <- RE_estimates_mean
-    RE_estimates_mean <- .SMfillEstimates(jaspResults, RE_estimates_mean, models[["RE"]], options)    
+    RE_estimates_mean <- .smFillEstimates(jaspResults, RE_estimates_mean, models[["RE"]], options)    
   }
   
   # tau estimates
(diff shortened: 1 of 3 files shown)


Resources/Meta Analysis/qml/SelectionModels.qml
```diff
@@ -23,6 +23,30 @@ import JASP				1.0
...
+		radioButtonsOnSameRow:	true
+		columns:				2
+
+		RadioButton
+		{
+			label: qsTr("Effect sizes & SE")
+			value: "general"
+			id: 	measures_general
+			checked:true
+		}
```
Kucharssim, line 40: The indenting/alignment goes to hell here, can you fix it?

JASP-Engine/JASP/R/selectionmodels.R
```diff
@@ -77,22 +81,36 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
...
+             type                  = c("infinity", "observations"),
              observations.amount   = "< 2",
              exitAnalysisIfErrors  = TRUE)
   
-  .hasErrors(dataset               = dataset,
-             type                  = c("negativeValues"),
-             negativeValues.target = options[["input_SE"]],
-             exitAnalysisIfErrors  = TRUE)
-  
+  if(options[["input_SE"]] != ""){
```
Kucharssim:
```suggestion
  if (options[["input_SE"]] != "") {
```

Kucharssim: And further where appropriate (see the style guide: https://github.com/jasp-stats/jasp-desktop/blob/stable/Docs/development/r-style-guide.md)

JASP-Engine/JASP/R/selectionmodels.R
```diff
@@ -593,12 +618,15 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
   
   # mean estimates
   if(is.null(RE_estimates[["RE_mean"]]) && options[["RE_estimates"]]){
-    RE_estimates_mean <- createJaspTable(title = gettextf("Mean Estimates (%s)", "\u03BC"))
+    RE_estimates_mean <- createJaspTable(title = gettextf(
+      "Mean Estimates (%s)",
+      ifelse(options[["measures"]] == "correlation", "\u03C1" , "\u03BC")
```
Kucharssim: at some places like this you use `ifelse()` on another you do plain simple `if() ... else ...`. I think you should prefer the latter here and in use cases like that, right?

Kucharssim (approved):
Ok, the code in general is good. Only couple of styling issues. (I am just going to mention the snake_casing against the JASP guildelines, but I understand that it's too much work to change that now)

Kucharssim (changes requested):
I accidentally approved the PR, but I think it would be nice to address some of the style comments first.

JASP-Engine/JASP/R/selectionmodels.R
```diff
@@ -77,22 +81,36 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
...
 
+  if(options[["input_N"]] != ""){
+    .hasErrors(dataset               = dataset,
+               type                  = c("negativeValues"),
+               negativeValues.target = options[["input_N"]],
+               exitAnalysisIfErrors  = TRUE)
+  }
+  
   if(options[["input_p"]] != ""){
     if(any(dataset[,.v(options[["input_p"]])] <= 0 | dataset[,.v(options[["input_p"]])] >= 1)){
```
AlexanderLyNL: In accordance to the jasp (as well as google) R style guide: 
- Space after the comma in `dataset[, .v(options[["inputP"]])]`
- Instead of a single `|`, you might want to use the double one `||`
- Please change the snake_cases to camelCases
- Space after if
- Space before {

see https://github.com/jasp-stats/jasp-desktop/blob/stable/Docs/development/r-style-guide.md

JASP-Engine/JASP/R/selectionmodels.R
```diff
@@ -77,22 +81,36 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
...
       JASP:::.quitAnalysis(gettextf("Error in %s: All p-value needs to be between 0 and 1.", options[["input_p"]]))
     }
   }
-
+  
+  if(options[["input_ES"]] != "" && options[["measures"]] == "correlation"){
+    if(any(dataset[,.v(options[["input_ES"]])] <= -1 | dataset[,.v(options[["input_ES"]])] >= 1)){
+      JASP:::.quitAnalysis(gettextf("Error in %s: All correlation coefficients needs to be between -1 and 1.", options[["input_ES"]]))
```
AlexanderLyNL: All correlation coefficients need to be

JASP-Engine/JASP/R/selectionmodels.R
```diff
@@ -215,7 +222,7 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
         lCI  = fit[["ci.lb_adj"]][mean_pos,1],
```
AlexanderLyNL: Is the dot there for column folding? If not please just use camelCase

JASP-Engine/JASP/R/selectionmodels.R
AlexanderLyNL: ifelse is vectorised and can behave oddly. Especially in this case you'd like to use the standard if else statement.

JASP-Engine/JASP/R/selectionmodels.R
```diff
@@ -698,7 +726,7 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
   if(!.SM_ready(options)){
```
AlexanderLyNL: Regarding single line if statements, see https://github.com/jasp-stats/jasp-desktop/blob/stable/Docs/development/r-style-guide.md

JASP-Engine/JASP/R/selectionmodels.R
```diff
@@ -940,12 +971,12 @@ jaspResults[["plot_estimates"]] <- plot_estimates
...
+  
   # in the case that the expected direction was negative, the estimated effect sizes will be in the opposite direction
   # this function turns them around
   
   if(!class(fit) %in% c("simpleError","error")){
-  
+    
     if(fit[["fe"]]){
       mean_pos <- 1    
     }else{
```
AlexanderLyNL: also space after }

JASP-Engine/JASP/R/selectionmodels.R
```diff
@@ -21,7 +21,7 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
     dataset <- .SM_data_get(dataset, options)
     dataset <- .SM_data_check(jaspResults, dataset, options)
   }
-
+  
   # fit the models
   if(.SM_ready(options)).SM_fit(jaspResults, dataset, options)
```
AlexanderLyNL: Please indent this properly

AlexanderLyNL (changes requested):
I went over this very quickly and only provided style comments. It would be good to follow our style guide: https://github.com/jasp-stats/jasp-desktop/blob/stable/Docs/development/r-style-guide.md

Resources/Meta Analysis/qml/SelectionModels.qml
FBartos: that's weird, Maybe a Github issue? It looks normal in both Visual studio code and Qt
image
image

JASP-Engine/JASP/R/selectionmodels.R
FBartos: no, the dot is there because that's how the `weightr` package names the object

JASP-Engine/JASP/R/selectionmodels.R
FBartos: yes, changed

JASP-Engine/JASP/R/selectionmodels.R
FBartos: done

JASP-Engine/JASP/R/selectionmodels.R
FBartos: thanks, changed

JASP-Engine/JASP/R/selectionmodels.R
FBartos: > Instead of a single |, you might want to use the double one ||

no, I'm using vectors there

FBartos:
I added the formating changes, - spaces, brackets, indentation etx... I camel-cased all functions that are called but did not alter the variable names and qml interface since that would create way too much changes

JASP-Engine/JASP/R/selectionmodels.R
AlexanderLyNL: I don't think you'd like to vectorise this, because an inputP cannot be both smaller than zero and larger than one. 

Instead, I think you'd like to take action if any of the provide inputP is smaller or equal to zero or if any of the provided inputP is larger or equal to one. The code implementing this thought is given by 

`if (any(dataset[, .v(options[["inputP"]])] <= 0) || any(dataset[, .v(options[["inputP"]])] >= 1)) { `

FBartos: I changed it, but both versions do exactly the same thing. You can read mine as: take an action if any of the values is either lower or equal to zero OR higher or equal to one.

AlexanderLyNL: No you can't. Your code says to specifically test for `inputP[i] <= 0 or inputP[i] >= 1` for each i, because it's vectorised. This implies much stricter semantics, which is totally unnecessary. It's better to remove complexity as it makes code easier for your future self, reviewers and debuggers.

AlexanderLyNL: What I'd like to say is that the bounds are uniform, thus allows you to write that does not depend on having the comparison for lower and upper bound at each index. The result is somewhat simpler code that does not depend on the difference between `|` and `||` making it a bit more robust.

AlexanderLyNL: Thinking back on it, the single `|` is quite elegant, but the fact that it caused some confusion makes it a bit troublesome. I can go either way now, but I was told myself that `||` is recommended in if statements, see https://stat.ethz.ch/R-manual/R-devel/library/base/html/Logic.html

FBartos: I already changed that in the last commit, so we are going with "||"

boutinb:
@AlexanderLyNL Can we merge this PR?

AlexanderLyNL:
The naming of functions and variables can be further improved. The same remarks given here: 

https://github.com/jasp-stats/jasp-desktop/pull/4238#discussion_r474737124 

might help.

FBartos:
thanks @AlexanderLyNL , the names are updated accordingly now

JASP-Engine/JASP/R/selectionmodels.R
```diff
@@ -16,86 +16,109 @@
...
+  
+  if (.smReady(options)) {
+    dataset <- .smGetData(dataset, options)
+    dataset <- .smCheckData(jaspResults, dataset, options)
   }
-
+  
   # fit the models
-  if(.SM_ready(options)).SM_fit(jaspResults, dataset, options)
+  if (.smReady(options))
```
AlexanderLyNL: any reason why you go through `if (.smReady(options))` twice?

JASP-Engine/JASP/R/selectionmodels.R
```diff
@@ -16,86 +16,109 @@
...
+    .smFit(jaspResults, dataset, options)
   
   # main summary tables
```
AlexanderLyNL, line 29: Testing the tables?

JASP-Engine/JASP/R/selectionmodels.R
```diff
@@ -16,86 +16,109 @@
...
-  .SM_fit_tests(jaspResults, dataset, options)
-  .SM_fit_estimates(jaspResults, dataset, options)
+  .smTestsTables(jaspResults, dataset, options)
+  .smEstimatesTables(jaspResults, dataset, options)
   
   # the p-value frequency tables
-  if(options[["p_table"]]).SM_p_frequency(jaspResults, dataset, options)
+  if (options[["p_table"]])
```
AlexanderLyNL: What is a p-value frequency table? Is this a separate table for p-values? I'm not familiar with this analysis, but other frequentist analyses show the p-values in the main table.

JASP-Engine/JASP/R/selectionmodels.R
```diff
@@ -16,86 +16,109 @@
...
+    .smPFrequencyTable(jaspResults, dataset, options)
   
   # figures
-  if(options[["FE_weightfunction"]]).SM_weights_plot(jaspResults, dataset, options, "FE")
-  if(options[["RE_weightfunction"]]).SM_weights_plot(jaspResults, dataset, options, "RE")
-  if(options[["plot_models"]]).SM_estimates_plot(jaspResults, dataset, options)
+  if (options[["FE_weightfunction"]])
+    .smWeightsPlot(jaspResults, dataset, options, "FE")
```
AlexanderLyNL, line 39: plot weights? I'm not familiar with this analysis, so I'm not sure what weights plots are.

JASP-Engine/JASP/R/selectionmodels.R
```diff
@@ -16,86 +16,109 @@
...
   return()
 }
 
-.SM_dependencies <- c(
-  "auto_reduce", "cutoffs_p", "selection_twosided", "effect_direction",
-  "input_ES", "input_SE", "input_p"
+.smDependencies <- c(
+  "auto_reduce", "cutoffs_p", "selection_twosided", "effect_direction", "measures", "mu_transform",
+  "input_ES", "input_SE", "input_N", "input_p"
 )
```
AlexanderLyNL, line 51: Could you please add line breaks between functions? I think they should be separate blocks.

JASP-Engine/JASP/R/selectionmodels.R
```diff
@@ -16,86 +16,109 @@
...
-.SM_ready              <- function(options){
-  return(options[["input_ES"]] != "" && options[["input_SE"]] != "")
+.smReady                   <- function(options) {
```
AlexanderLyNL: Shouldn't this be `.smCheckReady`? I think that such a name is more informative than `.smReady`.

JASP-Engine/JASP/R/selectionmodels.R
```diff
@@ -16,86 +16,109 @@
...
+  if (options[["measures"]] == "general") {
+    return(options[["input_ES"]] != "" && options[["input_SE"]] != "")
+  } else if (options[["measures"]] == "correlation") {
+    return(options[["input_ES"]] != "" && options[["input_N"]] != "")
+  }
```
AlexanderLyNL, line 58: Don't you want to have an else clause that returns FALSE? Otherwise, it'll produce NULL, which might lead to some odd errors depending on how R deals with a NULL in a conditional clause.

JASP-Engine/JASP/R/selectionmodels.R
```diff
@@ -16,86 +16,109 @@
...
-
   dataset_old <- dataset
   dataset     <- na.omit(dataset)
   
   # store the number of missing values
-  if(nrow(dataset_old) > nrow(dataset)){
-    if(!is.null(jaspResults[["ommited_observations"]])){
+  if (nrow(dataset_old) > nrow(dataset)) {
+    if (!is.null(jaspResults[["ommited_observations"]])) {
       ommited_observations <- jaspResults[["ommited_observations"]]
```
AlexanderLyNL: Note the use of snake_case instead of camelCase. The naming ommitted_observations suggests that you keep track of the whole row/column. I see that the object added is nrow(dataset_old) - nrow(dataset), which suggests that you're saving `nOmitted` or `nMissing` or `numberOfMissingValues` and not the actual observations. What's stored here? If it's the latter, could you please change the variable name accordingly.

JASP-Engine/JASP/R/selectionmodels.R
```diff
@@ -16,86 +16,109 @@
...
+  
+  
+  if (options[["input_p"]] != "") 
+    if (any(dataset[, .v(options[["input_p"]])] < 0) || any(dataset[, .v(options[["input_p"]])] > 1))
+      JASP:::.quitAnalysis(gettextf("Error in %s: All p-values need to be between 0 and 1.", options[["input_p"]]))
+  
+  
+  
+  if (options[["input_ES"]] != "" && options[["measures"]] == "correlation")
+    if (any(dataset[, .v(options[["input_ES"]])] <= -1) || any(dataset[, .v(options[["input_ES"]])] >= 1))
```
AlexanderLyNL: Shouldn't this need to be strictly smaller than -1 and strictly larger than 1?

JASP-Engine/JASP/R/selectionmodels.R
```diff
@@ -16,86 +16,109 @@
...
-  if(options[["input_p"]] != ""){
-    if(any(dataset[,.v(options[["input_p"]])] <= 0 | dataset[,.v(options[["input_p"]])] >= 1)){
-      JASP:::.quitAnalysis(gettextf("Error in %s: All p-value needs to be between 0 and 1.", options[["input_p"]]))
-    }
-  }
-
   return(dataset)
 }
-.SM_pcutoffs_get       <- function(options){
+.smGetCutoffs              <- function(options) {
```
AlexanderLyNL, line 125: It's not clear to me what this function does. The body of the function tells me that the cut off relate to p-values, but I'm a bit confused about the text manipulation here. 

Perhaps a QML element that takes a list of numbers might be of use here. There are certain validators in QML that can quickly check the input, but that requires some additional work. I remember a similar problem related to multiple quantiles in descriptives, and Joris mentioned the QML validator.

JASP-Engine/JASP/R/selectionmodels.R
```diff
@@ -109,47 +132,36 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
   x <- trimws(x, which = "both")
   x <- x[x != ""]
   
-  if(anyNA(as.numeric(x)))
+  if (anyNA(as.numeric(x)))
     JASP:::.quitAnalysis(gettext("The p-value cutoffs were set incorectly."))
-  if(length(x) == 0){
+  if (length(x) == 0) {
```
AlexanderLyNL: I think that it would be more clear if you added a line break between these two if clauses. The brackets of the second if statement seem to be redundant.

JASP-Engine/JASP/R/selectionmodels.R
```diff
@@ -109,47 +132,36 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
...
   # the autoreduce should combine all p-value intervals that have too few p-values
-
+  
   # create p-value table
   cutoffs_table <- table(cut(pval, breaks = c(0, steps)))
-
+  
   # start removing from the end, but never 1 at the end
-  while(cutoffs_table[length(cutoffs_table)] < 4){
+  while(cutoffs_table[length(cutoffs_table)] < 4) {
```
AlexanderLyNL: spacing

JASP-Engine/JASP/R/selectionmodels.R
```diff
@@ -158,24 +170,24 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
     cutoffs_table <- table(cut(pval, breaks = c(0, steps)))
   }
   
-  # and then go from the start (go one by one - two neiboring intervals with 2 p-values will become one with 4 instead of removing all of them)
-  while(any(cutoffs_table < 4)){
+  # and then go from the start(go one by one - two neiboring intervals with 2 p-values will become one with 4 instead of removing all of them)
+  while(any(cutoffs_table < 4)) {
```
AlexanderLyNL: Same, but you should check the spacing yourself. So I won't comment on them from here on anymore, but they should still be fixed.

JASP-Engine/JASP/R/selectionmodels.R
```diff
@@ -158,24 +170,24 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
...
     stop("No steps")
-  }
+  
   
   return(steps)
 }
-.SM_fill_estimates     <- function(jaspResults, table, fit, options){
+.smFillEstimates           <- function(jaspResults, table, fit, options) {
   
   CI_overtitle <- gettext("95% Confidence Interval")
```
AlexanderLyNL: snake_case to camelCase

AlexanderLyNL: There's probably more to say about the code from here onwards.

JASP-Engine/JASP/R/selectionmodels.R
```diff
@@ -109,47 +132,36 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
...
-      dataset[,.v(options[["input_ES"]])] /dataset[,.v(options[["input_SE"]])],
-      lower.tail = options[["effect_direction"]] == "negative") 
-  }else{
-    pval <- dataset[,.v(options[["input_p"]])]
-  }
-  return(pval)
-}
-.SM_autoreduce         <- function(steps, pval){
-
+.smAutoreduceCutoffs       <- function(steps, pval) {
```
AlexanderLyNL: Why the auto reduce? Why isn't it just a reduce?

AlexanderLyNL (changes requested):

JASP-Engine/JASP/R/selectionmodels.R
FBartos: I thought it looked better structured, changed

JASP-Engine/JASP/R/selectionmodels.R
FBartos: sorry? I don't understand this comment

JASP-Engine/JASP/R/selectionmodels.R
FBartos: yes, it's a table that shows frequency of p-values in pre-determined p-value intervals. P-values are one of the inputs for this analysis and this has nothing to do with inferential tests for coefficient or etc. I added a comment explaining the analysis.

FBartos:
The Selection Models analysis is based on `weightr` package and implements selection models into JASP. Selection models are an extension on meta-analytic models that corrects for publication bias operating on p-values. They assume, that p-values of individual studies determine probability of he studies being published. They model this assumption by changing the normal likelihood function (as in classical meta-analytic models) into weighted-normal likelihood function. The weighted-normal likelihood (f^w) is basically a normal likelihood (f) that is multiplied by a weight function (w) that is being simultaneously estimated from the data (the models are usually fitted using test-statistics likelihood, but that's nor relevant for the general understanding). 
image
Usually, a step function is used for convenience and the cut-points correspond to some critical (specified by user) p-values.
image

JASP-Engine/JASP/R/selectionmodels.R
FBartos: yes, a function that produces plot of weights, added a comment that explains the analysis

JASP-Engine/JASP/R/selectionmodels.R
FBartos: changed

JASP-Engine/JASP/R/selectionmodels.R
FBartos: added the lines

JASP-Engine/JASP/R/selectionmodels.R
FBartos: I would rather not. These two are exhaustive options as given by the interface and I would rather get an odd error somewhere if the analysis is changed that it does not follow the intended design.

JASP-Engine/JASP/R/selectionmodels.R
FBartos: It stores the number of omitted observations. Changed to `nOmitted` accordingly.

JASP-Engine/JASP/R/selectionmodels.R
FBartos: no, because correlations are transformed to either Cohen's d or Fisher's z for the fitting purposes and both of them would be undefined. Also, you can't practically get a correlation of |1|.

JASP-Engine/JASP/R/selectionmodels.R
FBartos: The functions extracts the cutoffs on p-values as specified by a user and checks that it's a valid vector of p-value cut-offs. It's done in the exactly same way as in RoBMA. I haven't heard about this input type before. Do you mean `inputType:	"doubleArray"` for a `TextField` component? As in line 175 for `Descriptives.qml`?

JASP-Engine/JASP/R/selectionmodels.R
FBartos: added the linebreak and remove the brackets

JASP-Engine/JASP/R/selectionmodels.R
FBartos: I automatically fixed spacing in all `if` `else` statements and forgot that there were two `while` statement. It's fixed now

JASP-Engine/JASP/R/selectionmodels.R
FBartos: fixed

JASP-Engine/JASP/R/selectionmodels.R
FBartos: changed to `overtitleCI` since `CIOvertitle` would look terrible, even though conforming to the camelCase...

FBartos: did some additional changes, but they clearly shows the problems of camelCase:
i.e.: when I have want to use a shortcut in the name of variable (RE / FE) the first letter of the following words get smashed together it, making it much more difficult to understand. Or when I need to use a shortcut starting with small letter (uCI) -> UCI doesn't make sense or the lower case "u" gets connected to the previous world...

TimKDJ: `ciOvertitle` would work just fine

JASP-Engine/JASP/R/selectionmodels.R
FBartos: changed

FBartos: the option in UI is called "Automatically reduce p-value intervals" though it was closer to the original naming. It was renamed by then, so I changed it to `smJoinCutoffs`

JASP-Engine/JASP/R/selectionmodels.R
AlexanderLyNL: p-values are continuous, so is there a specific binning involved here? Does this function make a table? .smPFrequencyTable

JASP-Engine/JASP/R/selectionmodels.R
```diff
@@ -271,18 +271,18 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
...
       rowREUnadjustedTau  <- c(rowREUnadjustedTau, list(
-        est  = sqrt(fit[["unadj_est"]][1,1]),
-        stat = fit[["z_unadj"]][1,1],
-        pval = fit[["p_unadj"]][1,1],
-        lCI  = sqrt(ifelse(fit[["ci.lb_unadj"]][1,1] < 0, 0, fit[["ci.lb_unadj"]][1,1])),
-        uCI  = sqrt(fit[["ci.ub_unadj"]][1,1])
+        est  = sqrt(fit[["unadj_est"]][1, 1]),
+        stat = fit[["z_unadj"]][1, 1],
+        pval = fit[["p_unadj"]][1, 1],
+        lCI  = sqrt(ifelse(fit[["ci.lb_unadj"]][1, 1] < 0, 0, fit[["ci.lb_unadj"]][1, 1])),
```
AlexanderLyNL: lowerCi and upperCi might be a more descriptive name

JASP-Engine/JASP/R/selectionmodels.R
```diff
@@ -271,18 +271,18 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
...
     if (!class(fit) %in% c("simpleError","error")) {
       rowREUnadjustedTau  <- c(rowREUnadjustedTau, list(
-        est  = sqrt(fit[["unadj_est"]][1,1]),
-        stat = fit[["z_unadj"]][1,1],
-        pval = fit[["p_unadj"]][1,1],
-        lCI  = sqrt(ifelse(fit[["ci.lb_unadj"]][1,1] < 0, 0, fit[["ci.lb_unadj"]][1,1])),
-        uCI  = sqrt(fit[["ci.ub_unadj"]][1,1])
+        est  = sqrt(fit[["unadj_est"]][1, 1]),
+        stat = fit[["z_unadj"]][1, 1],
+        pval = fit[["p_unadj"]][1, 1],
```
AlexanderLyNL: perhaps pValue or pVal might be more descriptive

JASP-Engine/JASP/R/selectionmodels.R
```diff
@@ -398,7 +398,7 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
   if (options[["auto_reduce"]]) {
     steps <- tryCatch(.smJoinCutoffs(steps, pval), error = function(e)e)
     
-    if (class(steps) %in% c("simpleError","error")) {
+    if (class(steps) %in% c("simpleError", "error")) {
```
AlexanderLyNL, line 401: The camelCase versions of fit_FE and fit_RE would be fitFE and fitRE

JASP-Engine/JASP/R/selectionmodels.R
```diff
@@ -582,7 +582,7 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
   return()
 }
 
-.smEstimatesTables         <- function(jaspResults, dataset, options) {
+.smMakeEstimatesTables     <- function(jaspResults, dataset, options) {
   
   models   <- jaspResults[["models"]]$object
```
AlexanderLyNL, line 588: I can't comment at the right line 676. I think that lowerRange and upperRange are too generic. You might want to change that to lowerPRange and upperPRange

pFrequency$addColumnInfo(name = "lowerRange", title = gettext("Lower"),     type = "number", overtitle = overtitle)

JASP-Engine/JASP/R/selectionmodels.R
```diff
@@ -843,10 +843,10 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
...
-    model = c(gettext("Fixed effects"),   gettext("Fixed effects (adjusted)"),   gettext("Random effects"),  gettext("Random effects (adjusted)")),
-    mean  = c(FE[["unadj_est"]][1,1],     FE[["adj_est"]][1,1],                  RE[["unadj_est"]][2,1],     RE[["adj_est"]][2,1]),
-    lCI   = c(FE[["ci.lb_unadj"]][1,1],   FE[["ci.lb_adj"]][1,1],                RE[["ci.lb_unadj"]][2,1],   RE[["ci.lb_adj"]][2,1]),
-    uCI   = c(FE[["ci.ub_unadj"]][1,1],   FE[["ci.ub_adj"]][1,1],                RE[["ci.ub_unadj"]][2,1],   RE[["ci.ub_adj"]][2,1])
+    model = c(gettext("Fixed effects"),    gettext("Fixed effects (adjusted)"),    gettext("Random effects"),   gettext("Random effects (adjusted)")),
+    mean  = c(FE[["unadj_est"]][1, 1],     FE[["adj_est"]][1, 1],                  RE[["unadj_est"]][2, 1],     RE[["adj_est"]][2, 1]),
+    lCI   = c(FE[["ci.lb_unadj"]][1, 1],   FE[["ci.lb_adj"]][1, 1],                RE[["ci.lb_unadj"]][2, 1],   RE[["ci.lb_adj"]][2, 1]),
+    uCI   = c(FE[["ci.ub_unadj"]][1, 1],   FE[["ci.ub_adj"]][1, 1],                RE[["ci.ub_unadj"]][2, 1],   RE[["ci.ub_adj"]][2, 1])
   )
   estimates <- estimates[4:1,]
```
AlexanderLyNL, line 852: L 855. "since one of the estimate is not a number."

JASP-Engine/JASP/R/selectionmodels.R
```diff
@@ -1043,13 +1043,13 @@ SelectionModels <- function(jaspResults, dataset, options, state = NULL) {
...
+    fit[["unadj_est"]][posMean, 1] <- .smInvTransform(fit_old[["unadj_est"]][posMean, 1], options[["mu_transform"]])
+    fit[["ci.lb_adj"]][posMean, 1] <- .smInvTransform(fit_old[["ci.lb_adj"]][posMean, 1], options[["mu_transform"]])
+    fit[["ci.ub_adj"]][posMean, 1] <- .smInvTransform(fit_old[["ci.ub_adj"]][posMean, 1], options[["mu_transform"]])
     
-    fit[["adj_est"]][posMean,1]     <- .smInvTransform(fit_old[["adj_est"]][posMean,1],     options[["mu_transform"]])
-    fit[["ci.lb_unadj"]][posMean,1] <- .smInvTransform(fit_old[["ci.lb_unadj"]][posMean,1], options[["mu_transform"]])
-    fit[["ci.ub_unadj"]][posMean,1] <- .smInvTransform(fit_old[["ci.ub_unadj"]][posMean,1], options[["mu_transform"]])
+    fit[["adj_est"]][posMean, 1]     <- .smInvTransform(fit_old[["adj_est"]][posMean, 1],     options[["mu_transform"]])
+    fit[["ci.lb_unadj"]][posMean, 1] <- .smInvTransform(fit_old[["ci.lb_unadj"]][posMean, 1], options[["mu_transform"]])
+    fit[["ci.ub_unadj"]][posMean, 1] <- .smInvTransform(fit_old[["ci.ub_unadj"]][posMean, 1], options[["mu_transform"]])
```
AlexanderLyNL: I can't comment properly after this line.

AlexanderLyNL (changes requested):
This looks much better. I do think that it's worthwhile to remove snake_case names from the QML and change that to camelCase. Otherwise, we have to correct that later using the QML upgrader which is quite a chore. 

You might also want to change the function names that relate to plotting. .smWeightsPlot suggest that it weights the plot, whereas I think it plots the weights. Perhaps change that to .smPlotWeights. Similar for the other plot functions.

JASP-Engine/JASP/R/selectionmodels.R
```diff
@@ -939,36 +989,137 @@ jaspResults[["plot_estimates"]] <- plot_estimates
...
+  
+  # do a scale transform if neccessary
+  if (options[["measures"]] == "general") {
+    return(ES)
+  } else if (options[["measures"]] == "correlation") {
+    return(.smTransform(ES, transformation = options[["mu_transform"]], what = "ES"))
+  }
+}
+
+.smGetInputVAR             <- function(dataset, options) {
```
AlexanderLyNL: What's VAR?

JASP-Engine/JASP/R/selectionmodels.R
```diff
@@ -939,36 +989,137 @@ jaspResults[["plot_estimates"]] <- plot_estimates
...
+  
+  # do a scale transform if neccessary
+  if (options[["measures"]] == "general") {
+    return(dataset[, .v(options[["input_SE"]])]^2)
+  } else if (options[["measures"]] == "correlation") {
+    return(.smTransform(ES, dataset[, .v(options[["input_N"]])], transformation = options[["mu_transform"]], what = "VAR"))
+  }
+}
+
+.smGetInputPval            <- function(dataset, options) {
```
AlexanderLyNL: PVal

JASP-Engine/JASP/R/selectionmodels.R
```diff
@@ -939,36 +989,137 @@ jaspResults[["plot_estimates"]] <- plot_estimates
...
+  # weightfunc uses one-sided p-values as input!
+  if (options[["input_p"]] == "") {
+    pval <- pnorm(
```
AlexanderLyNL: pVal

JASP-Engine/JASP/R/selectionmodels.R
```diff
@@ -939,36 +989,137 @@ jaspResults[["plot_estimates"]] <- plot_estimates
...
+    pval <- dataset[, .v(options[["input_p"]])]
+  }
+  return(pval)
+}
+
+.smTransform               <- function(ES, N = NULL, transformation, what) {
+  if (what == "ES") {
+    return(switch(
+      transformation,
+      "cohens_d"  = psych::r2d(ES),
```
AlexanderLyNL: cohensD etc etc

JASP-Engine/JASP/R/selectionmodels.R
FBartos: changed, thanks, I missed those previously

JASP-Engine/JASP/R/selectionmodels.R
FBartos: changed to pVal

JASP-Engine/JASP/R/selectionmodels.R
FBartos: changed

JASP-Engine/JASP/R/selectionmodels.R
FBartos: fixed

JASP-Engine/JASP/R/selectionmodels.R
FBartos: changed

JASP-Engine/JASP/R/selectionmodels.R
FBartos: shortcut for variance

JASP-Engine/JASP/R/selectionmodels.R
FBartos: changed

JASP-Engine/JASP/R/selectionmodels.R
FBartos: changed

JASP-Engine/JASP/R/selectionmodels.R
FBartos: changed to `lowerCI`, as in `binomial.R` for example

JASP-Engine/JASP/R/selectionmodels.R
AlexanderLyNL: Then `.smGetInputVar` would do I reckon.

FBartos: in the same way as ES stands for Effect Size

AlexanderLyNL: so it's VARiance then?

FBartos:
ok, all of the requests are incorporated now

JASP-Engine/JASP/R/selectionmodels.R
```diff
@@ -939,36 +989,137 @@ jaspResults[["plot_estimates"]] <- plot_estimates
...
+
+.smTransform               <- function(ES, N = NULL, transformation, what) {
+  if (what == "ES") {
+    return(switch(
+      transformation,
+      "cohensD"  = psych::r2d(ES),
+      "fishersZ" = psych::fisherz(ES)
+    ))
+  } else if (what == "VAR") {
+    if (is.null(N))stop("The effect size needs to be specified for the VAR transformation.")
```
AlexanderLyNL: I'm a bit confused about this. What is the VAR transformation?

JASP-Engine/JASP/R/selectionmodels.R
FBartos: I renamed it to `.smGetInputVar`

JASP-Engine/JASP/R/selectionmodels.R
FBartos: it computes variances from the effect sizes. I wanted to bundle both of them in one function

AlexanderLyNL: I think that "No effect sizes specified to compute the studies' sample variances" might convey that message. I'm guessing variance refers to the sample variance of an individual study, but it might be something else.

FBartos: it does refer to the sampling variances. Changed

AlexanderLyNL:
This is now ready for testing. @FBartos could you please add this PR to the module as well? Note the possible adaptation regarding the namespaces such as jaspGraphs, etc.

PR #4240, opened 2020-08-17
Errors and warnings for a form are now a property

JorisGoosen:
- No more need for directly manipulating QQuickItem*...
- Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/757
- Refactored some code a bit
- Added a property for Analysis * as well

### DIFF
--- a/JASP-Desktop/analysis/analysisform.cpp
+++ b/JASP-Desktop/analysis/analysisform.cpp
@@ -54,9 +54,9 @@ using namespace std;
 AnalysisForm::AnalysisForm(QQuickItem *parent) : QQuickItem(parent)
 {
 	setObjectName("AnalysisForm");
-
-	connect(this, &AnalysisForm::formCompleted, this, &AnalysisForm::formCompletedHandler	);
 	connect(this, &AnalysisForm::infoChanged,	this, &AnalysisForm::helpMDChanged			);
+	connect(this, &AnalysisForm::formCompleted, this, &AnalysisForm::formCompletedHandler   );
+
 }
 
 QVariant AnalysisForm::requestInfo(const Term &term, VariableInfo::InfoType info) const
@@ -218,11 +218,11 @@ void AnalysisForm::addControl(JASPControlBase *control)
 			if (label.isEmpty())
 				label = control->property("title").toString();
 
-			if (label.isEmpty())	_formErrorMessages.append(tr("Control with label '%1' has no name").arg(label));
-			else					_formErrorMessages.append(tr("A control has no name"));
+			if (label.isEmpty())	_formErrors.append(tr("Control with label '%1' has no name").arg(label));
+			else					_formErrors.append(tr("A control has no name"));
 		}
 		else if (_controls.keys().contains(name))
-			_formErrorMessages.append(tr("2 controls have the same name: %1").arg(name));
+			_formErrors.append(tr("2 controls have the same name: %1").arg(name));
 		else
 		{
 			JASPControlWrapper* wrapper = control->getWrapper();
@@ -242,11 +242,6 @@ void AnalysisForm::_setUpControls()
 	_orderExpanders();
 	_setUpRelatedModels();
 	_setUpItems();
-
-	if (!_formErrorMessagesItem)
-		Log::log()  << "No errorMessages Item found!" << std::endl;
-
-	_setErrorMessages();
 }
 
 void AnalysisForm::_setUpRelatedModels()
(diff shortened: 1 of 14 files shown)


JASP-Desktop/analysis/analysisform.cpp
```diff
@@ -721,49 +703,59 @@ void AnalysisForm::clearControlError(JASPControlBase* control)
...
 
+
+void AnalysisForm::formCompletedHandler()  { QTimer::singleShot(0, this, &AnalysisForm::_formCompletedHandler); }
 void AnalysisForm::_formCompletedHandler()
 {
-	QVariant analysisVariant = QQmlProperty(this, "analysis").read();
-	if (!analysisVariant.isNull())
-	{
-		_analysis	= qobject_cast<Analysis *>(analysisVariant.value<QObject *>());
+	qmlContext(this)->setContextProperty("form", this);
```
JorisGoosen: Is this really necessary?

JorisGoosen:
I'm also adding https://github.com/jasp-stats/INTERNAL-jasp/issues/1018 to this branch

JorisGoosen:
That is, if I get it to work. But it can be separated from this PR without problems if @boutinb is a quick man ;)

JASP-Desktop/analysis/analysisform.cpp
boutinb: Yes: the form id is known is all forms, but not inside a JASP Widget. Like that it is known everywhere.

JorisGoosen: But I keep seeing warnings in the output `QQmlContext: Cannot set property on internal context.`
I don't know if it is related to this though.

And also, I do not understand "the form id is known is all forms", what it would mean and how it is different for a JASP Widget?

boutinb: I added this in this commit https://github.com/jasp-stats/jasp-desktop/commit/ffb7478a78660c76bf0c141a9c6b5cd60ab13953 to fix overflow issues in the form. I saw that sometimes the GridLayout did not have the form object, and I thought it was because GridLayout is not a JASPControl. But I have just tested it without using the context setting, and it still works. So you can remove it.

JASP-Desktop/analysis/analysisform.cpp
```diff
@@ -218,11 +218,11 @@ void AnalysisForm::addControl(JASPControlBase *control)
 			if (label.isEmpty())
 				label = control->property("title").toString();
 
-			if (label.isEmpty())	_formErrorMessages.append(tr("Control with label '%1' has no name").arg(label));
-			else					_formErrorMessages.append(tr("A control has no name"));
+			if (label.isEmpty())	_formErrors.append(tr("Control with label '%1' has no name").arg(label));
+			else					_formErrors.append(tr("A control has no name"));
```
boutinb: use addFormError in place of _formErrors.append: if not, no signal would be emitted.

boutinb (changes requested):
I have tested it, but it does not work well with malformed QML form.
For example if a control has no name, it generates an error, but the analysis is run, and clears the errors: you have no time to see the error.
Maybe a better solution would be to have a close icon (a cross) in the error message, so that the user can himself close the error message.

JASP-Desktop/analysis/analysisform.cpp
JorisGoosen: Oops!

JASP-Desktop/analysis/analysisform.cpp
JorisGoosen: Ok good!

JorisGoosen:
Yeah it is a problem it clears the errors on refresh...
It would probably be better to do it the way you describe.

How would we clear the warning on an option then? 
As in, the errors on options are timed so solve themselves...

JorisGoosen:
Ok so I've processed the requested changes and fixed a small bug.

Now form errors and warnings remain unless the user closes them.
Warnings/errors on options are not cleared but I think this will probably work out fine anyway.

The only warnings not set through known-issues are set because the "name" is missing or doubled or something. So they should stay. And if there are errors on options you can't run the analysis anyway.

JASP-Desktop/analysis/analysisform.cpp
```diff
@@ -721,49 +692,55 @@ void AnalysisForm::clearControlError(JASPControlBase* control)
...
+	Analysis * analysisPointer = qobject_cast<Analysis *>(analysis.value<QObject *>());
+
+	if(_analysis == analysisPointer) return;
+
+	if(_analysis && analysisPointer)
+		throw std::runtime_error("An analysis of an analysisform was replaced by another analysis, this is decidedly NOT supported!");
+
+	_analysis = analysisPointer;
+
+	setAnalysisUp();
```
boutinb, line 717: I don't see the need to call that here.
Why just to set _analysis here and let formCompleted to call the setup? This makes more complicated than it is.... Now you need this extra boolean _formCompleted.

boutinb (changes requested):
Just a small change. The rest seems to be perfect. It clears up the code!

JASP-Desktop/analysis/analysisform.cpp
JorisGoosen: Because maybe `setAnalysis` is only set/called *after* `formCompleted`.
I don't see how right now, but I prefer to safeguard the code for it like this.

And I would also like to get rid of all that 'formCompleted' stuff at some point because I have the feeling it makes things more rigid than necessary. Because everything has to be set then and won't be changed later if users change properties or whatever.

boutinb (approved):

PR #4241, opened 2020-08-18
Set fonts via Preferences

boutinb:
Fixes jasp-stats/INTERNAL-jasp#1019


JASP-Desktop/components/JASP/Theme/Theme.qml
```diff
@@ -33,22 +33,22 @@ JaspTheme
 
 	fontLabel.bold:										true
 	fontLabel.pixelSize:								Math.round(18 * uiScale)
-	fontLabel.family:									jaspFont + (MACOS ? " Bold" : "")
+	fontLabel.family:									jaspFont
```
JorisGoosen, line 36: Did you make sure the font s are still bold/demibold etc when they ought to be?

JASP-Desktop/components/JASP/Widgets/FileMenu/PrefsResults.qml
```diff
@@ -130,12 +130,41 @@ ScrollView
...
+		{
+			id:		fontGroup
+			title:	qsTr("Font")
+
+			TextArea
+			{
+				id:					resultFont
+				width:				300 * jaspTheme.uiScale
+				height:				150 * jaspTheme.uiScale
+				title:				qsTr("List of Font")
```
JorisGoosen: "List of fonts"

JASP-Desktop/components/JASP/Widgets/FileMenu/PrefsResults.qml
```diff
@@ -130,12 +130,41 @@ ScrollView
...
+				toolTip:			qsTr("Font family used by the results")
```
JorisGoosen, line 150: If it is just for the results then why is it changing QML?

JASP-Desktop/components/JASP/Widgets/FileMenu/PrefsResults.qml
JorisGoosen: Also, can we just do this through a userselectable list of systemfonts instead of them having to type?

JASP-Desktop/components/JASP/Widgets/FileMenu/PrefsUI.qml
```diff
@@ -69,12 +69,42 @@ ScrollView
...
+			{
+				id:					uiFont
+				width:				300 * jaspTheme.uiScale
+				height:				150 * jaspTheme.uiScale
+				title:				qsTr("List of Font")
+				toolTip:			qsTr("Font family used by the interface")
+				text:				preferencesModel.interfaceFont.replace(/,/g, "\n")
+				onApplyRequest:
+				{
+					preferencesModel.interfaceFont = text.split("\n")
```
JorisGoosen: Code duplication...
Could you not do that? 

Maybe move the function to C++ and use it there?
Although if you change to a combobox this will probably not be necessary anymore.

JASP-Desktop/gui/preferencesmodel.h
```diff
@@ -38,7 +38,8 @@ class PreferencesModel : public QObject
 	Q_PROPERTY(bool			safeGraphics			READ safeGraphics				WRITE setSafeGraphics				NOTIFY safeGraphicsChanged				)
 	Q_PROPERTY(QString		cranRepoURL				READ cranRepoURL				WRITE setCranRepoURL				NOTIFY cranRepoURLChanged				)
 	Q_PROPERTY(int			plotPPI					READ plotPPI														NOTIFY plotPPIPropChanged				)
-	Q_PROPERTY(QFont		defaultFont				READ defaultFont				WRITE setDefaultFont				NOTIFY defaultFontChanged				)
+	Q_PROPERTY(QString		interfaceFont			READ interfaceFont				WRITE setInterfaceFont				NOTIFY interfaceFontChanged				)
```
JorisGoosen, line 44: Also a problem, by changing QFont to QString we will no longer be able to use non-system fonts.
Like the open source sansserif font included with JASP.
Which it would be great to use!

Because it contains all mathematical symbols for sure.

JASP-Desktop/qquick/jasptheme.h
```diff
@@ -563,6 +563,7 @@ public slots:
 	void setIconPath(QString iconPath);
 	void setThemeName(QString themeName);
 	void setFontConsole(QFont fontConsole);
+	void setDefaultFont(QString font);
```
JorisGoosen: Why the QString and not the QFont?

Resources/Help/preferences/prefsui.md
```diff
@@ -32,3 +32,6 @@ If this is enabled JASP will run in software rendering mode, which means your in
 Some users on certain systems (Windows) can run into trouble with the default native (or system) file dialogs.
 When they try to open or save a file in JASP by clicking "Browse" in one of the file menus JASP crashes.
 If you disable this option we use Qt file dialogs and they might not crash. Most users will not need this option disabled though.
+
+### Font
+You can specify here which font will be used by the interface. Default is `SansSerif`
```
JorisGoosen: "Here you can specify which font will be used by the interface. Default is `SansSerif`"

JorisGoosen: Should this also be added to the results thingy?

JorisGoosen (changes requested):

JorisGoosen:
I think having a combobox would greatly improve this functionality, in what application do users have to guess the name of the font they want to use?
And I think using a QFont makes more sense then QString, especially since this is one of the few datatypes incuded with QML.

JorisGoosen:
And I didn't even test it yet ;)

boutinb:
I see that Qt can retrieve the system fonts https://doc.qt.io/qt-5/qfontdatabase.html. I can try to use that.
But what I understand from fonts is that sometimes, the browser can decide to use different fonts for the same string if part of the string cannot be displayed with the first specified font, but with the second one. That's why there is a list of fonts. So one ComboBox is not enough. I could make a list of ComboBox...

JorisGoosen:
> I see that Qt can retrieve the system fonts https://doc.qt.io/qt-5/qfontdatabase.html. I can try to use that.

Good!

> But what I understand from fonts is that sometimes, the browser can decide to use different fonts for the same string if part of the string cannot be displayed with the first specified font, but with the second one. That's why there is a list of fonts. So one ComboBox is not enough. I could make a list of ComboBox...

I have never heard of a browser using different fonts for different parts of a string. Those lists are there for when a system doesn't have the first font, then it tries the second one. But if the user selects from systemfonts how can they not have that font?

Of course, if Im wrong about this I would like to hear it but I don't think so.

boutinb:
https://stackoverflow.com/questions/884177/how-can-i-determine-what-font-a-browser-is-actually-using-to-render-some-text
According to this discussion, the browser may choose another font if the current one does not support some characters. It's still a bit strange: if it was true, why do we get sometimes empty squares for some non-latin characters...
I have another issue: for the WebEngine we have now a list of fonts (Lucida Grande,Helvetica,Arial,sans-serif,Helvetica Neue,freesans,Segoe UI"). I could set it to Lucinda Grande per default, but if the computer does not have this font, it will get the default font, which is maybe different than the Helvetica (what he gets now if the PC has this font). Maybe small issue...
In QML it is now per default SansSerif, which does not exist on my Mac, and it gives apparently the default font ".SF NS Text" (which is per chance a Sans Serif font). I think I have to add a 'default system font' option.

JASP-Desktop/components/JASP/Theme/Theme.qml
boutinb: I see no difference on my Mac if I remove this. It seems anyway quite redundant with the bold property.
And it must be removed because most of the fonts do not have a 'Bold' version: then it falls back to the default font.

JASP-Desktop/gui/preferencesmodel.h
boutinb: But we do not load FreeSans yet. And I think if we add it in the application, we can call it by its name (we will have to anyway, by selecting it in the Dropdown).

JorisGoosen: Ooh yeah, ok. 

But I still think QFont is the way to go.

JorisGoosen:
> https://stackoverflow.com/questions/884177/how-can-i-determine-what-font-a-browser-is-actually-using-to-render-some-text
> According to this discussion, the browser may choose another font if the current one does not support some characters. It's still a bit strange: if it was true, why do we get sometimes empty squares for some non-latin characters...

Because Arial on the user's system might be different because who knows what Microsoft does to support other locales?

> I have another issue: for the WebEngine we have now a list of fonts (Lucida Grande,Helvetica,Arial,sans-serif,Helvetica Neue,freesans,Segoe UI"). I could set it to Lucinda Grande per default, but if the computer does not have this font, it will get the default font, which is maybe different than the Helvetica (what he gets now if the PC has this font). Maybe small issue...
> In QML it is now per default SansSerif, which does not exist on my Mac, and it gives apparently the default font ".SF NS Text" (which is per chance a Sans Serif font). I think I have to add a 'default system font' option.

SansSerif is a generic font family which allows someone to not have to specify exactly what font to use.

I think having an option called "default" for the results is fine, it would just not use anything special. And to get back to it after the user changed something you could just reload the css.

JorisGoosen:
Nice!
Works pretty awesome :)

And it is hilarious to see JASP change as I move through the different fonts for the interface.

I did find a small bug still with the "default" option for the results. It says next to it that the "default" is "Deja vu sans" which Im guessing is the system sansserif I have. But if I turn the checkbox on the results turn to a serif font...

A la:
Screenshot_2020-08-19_173819

Then to the actual selected font:
Screenshot_2020-08-19_173823

Then I tried restarting with "default" on but it shows it with the selected font instead:
Screenshot_2020-08-19_173906

So there is definitely something still wrong there.

boutinb:
I think the problem is that the DejaVu Sans is not recognized by The WebEngine. Could you start the Chrome inspector and ask what the value of the actual font (in computed tab, at the bottom), and whether the body gets the DejaVu Sans font?

JorisGoosen:
I think there are two problems:

1. When I start JASP and load a file while "default font" is checked for results it still uses the dropdown font. So that is bad
2. When I turn default off and on again it switches to a different font than selected in dropdown. But it is: "font-family: SansSerif;"

What I would suggest for turning on default font is: simply reload the "theme css" and then *not set any font*. Is that an option?

JASP-Desktop/gui/preferencesmodel.cpp
```diff
@@ -33,6 +34,13 @@ PreferencesModel::PreferencesModel(QObject *parent) :
 	connect(this,					&PreferencesModel::disableAnimationsChanged,	this, &PreferencesModel::animationsOnChanged			);
 
 	connect(LanguageModel::lang(),	&LanguageModel::currentIndexChanged,			this, &PreferencesModel::languageCodeChanged);
+
+	connect(this,					&PreferencesModel::useDefaultInterfaceFontChanged,	[&](){	emit realInterfaceFontChanged(realInterfaceFont()); } );
```
JorisGoosen: Are you aware you can connect signals to signals? Would be prettier than this.
Also, I don't think it is necessary to send the font with it?

JASP-Desktop/gui/preferencesmodel.cpp
```diff
@@ -107,7 +115,41 @@ GET_PREF_FUNC_STR(	currentThemeName,			Settings::THEME_NAME								)
...
+QString PreferencesModel::resultFont() const
+{
+	QString result = Settings::value(Settings::RESULT_FONT).toString();
 
+	if (result.contains(","))
+	{
+		// The default is a list of fonts.
+		// Select the first one which is available.
+		for (const QString& font : result.split(","))
+		{
```
JorisGoosen, line 135: Why the braces?

JASP-Desktop/results/resultsjsinterface.cpp
```diff
@@ -73,6 +73,7 @@ void ResultsJsInterface::setResultsLoaded(bool resultsLoaded)
 		emit runJavaScript("window.setAppVersion('" + version + "')");
 
 		setGlobalJsValues();
+		setFontFamily(Settings::value(Settings::RESULT_FONT).toString());
```
JorisGoosen: Just check here if "default is on" and if so don't set the result font especially.

JorisGoosen:
I think it would work better to set the results font in the way Ive described and then  you can also remove the "Deja Vu Sans" or whatever font next to it. Because it might not be the right font anyway. 
Just show it as "default" instead.

JorisGoosen:
And a little bit of an extension: maybe we could also make the font for the console/code-windows changeable?

JorisGoosen:
(After the code actually works I will do an in-depth code review btw ;) )

JASP-Desktop/components/JASP/Theme/Theme.qml
JorisGoosen: I think you would see the difference if you would specify the fonts as QFont instead of as QString, but if it works it works so whatever.

JorisGoosen:
I think once you have the FiraCode commited this can be merged!

PR #4242, opened 2020-08-18
Known issues

JorisGoosen:
Should be merged AFTER https://github.com/jasp-stats/jasp-desktop/pull/4240 !

Screenshot_2020-08-18_130523
Screenshot_2020-08-18_130534

### DIFF
--- a/JASP-Desktop/gui/jaspversionchecker.cpp
+++ b/JASP-Desktop/gui/jaspversionchecker.cpp
@@ -21,6 +21,8 @@
 #include "utilities/qutils.h"
 #include "gui/aboutmodel.h"
 #include "appinfo.h"
+#include "analysis/knownissues.h"
+#include "log.h"
 
 JASPVersionChecker::JASPVersionChecker(QObject *parent) : QObject(parent)
 {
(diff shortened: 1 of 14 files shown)


JorisGoosen:
Maybe I need to rebase the branch first after https://github.com/jasp-stats/jasp-desktop/pull/4240 is merged because it seems those commits are also in here...

EJWagenmakers:
Can we replace the warning text with this:
"WARNING. A bug has been reported for this analysis. Click on this text for details."
Hmm I am not too happy with that either...How about:
"WARNING. This analysis contains an error, and we are in the process of fixing it. Click on this text for details. Proceed at your own risk."
?!

JorisGoosen:
The warning text is part of the json file on the server so it can be tweaked to your heart's desire at any time. The one in the screenshot is just to show the mechanism

JorisGoosen:
So Ive updated this now that the errormessages PR was merged, so this can be checked and merged if given the OK.

JASP-Desktop/gui/jaspversionchecker.cpp
```diff
@@ -47,6 +56,14 @@ void JASPVersionChecker::downloadFinished()
 	if (latest > cur)
 		emit showDownloadButton(downloadfile);
 
+	if(KnownIssues::issues()->downloadNeededOrLoad())	downloadKnownIssues();
```
boutinb, line 59: This is double: downloadKnownIssues is already called in downloadNeededOrLoad

JASP-Desktop/gui/jaspversionchecker.cpp
```diff
@@ -47,6 +56,14 @@ void JASPVersionChecker::downloadFinished()
...
+	else												deleteLater(); //Remove yourself!
+}
```
boutinb, line 61: I would set a JASPVersionChecker as member mainWindow and a add a method to start the check. This deleteLater while waiting for a network response is a bit scary... and is not really needed.

boutinb (changes requested):

boutinb:
The cross to remove the warning does not work.

JASP-Desktop/gui/jaspversionchecker.cpp
JorisGoosen: No it isn't.

Show me where then?

JASP-Desktop/gui/jaspversionchecker.cpp
JorisGoosen: The deleteLater isn't called while waiting for a network response...

JorisGoosen:
Aah yeah, well, I think it is good to have the warning permanently there. Because they are known issues after all and I want people to be aware of them.

JorisGoosen:
Hmm, but it is more difficult to do now that they aren't cleared all the time. I'll make sure the cross works. But then the warnings on the options should probably also be removed.

boutinb (approved):

PR #4243, opened 2020-08-18
Fix JAGS crash

JorisGoosen:
- Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/1016
- Also some small styletweaking

### DIFF
--- a/JASP-Desktop/widgets/listmodeljagsdatainput.cpp
+++ b/JASP-Desktop/widgets/listmodeljagsdatainput.cpp
@@ -47,28 +47,34 @@ void ListModelJAGSDataInput::sourceTermsChanged(const Terms *, const Terms *)
 	if (_values.length() > 0)
 	{
 		QMap<QString, QVariant> mapping;
-		const QVector<QVariant>& firstCol = _values[0];
-		const QVector<QVariant>& secondCol = _values[1];
+		const QVector<QVariant> & firstCol  = _values[0],
+								& secondCol = _values[1];
 		int row = 0;
+
 		for (const QVariant& key : firstCol)
 		{
 			mapping[key.toString()] = secondCol[row];
 			row++;
 		}
+
 		_values.clear();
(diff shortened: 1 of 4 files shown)


JASP-Desktop/widgets/listmodeltableviewbase.cpp
```diff
@@ -45,7 +45,7 @@ QVariant ListModelTableViewBase::data(const QModelIndex &index, int role) const
 	int		column	= index.column(),
 			row		= index.row();
 
-	if (column < 0 || column >= columnCount() || row < 0 || row >= _rowNames.length())
+	if (column < 0 || column >= columnCount() || row < 0 || row >= rowCount())
```
boutinb, line 48: It is surely better, but I still not understand why it solves the problem

JorisGoosen: This isnt what solves the problem actually. Ill comment at the line where I did.

JASP-Desktop/widgets/listmodeljagsdatainput.cpp
```diff
@@ -164,10 +171,10 @@ void ListModelJAGSDataInput::initValues(OptionsTable * bindHere)
 	if (_values.size() > 0 && int(_values[0].size()) != _rowNames.size())
 		Log::log() << "Number of rows specifed in Options for ListModelJAGSDataInput does not match number of rows in values!" << std::endl;
 
-
 	beginResetModel();
 
-	_columnCount = _colNames.size();
+	_rowCount		= _rowNames.size();
```
JorisGoosen, line 176: This line solves it, because `_rowCount` wasn't the right size `_values` wasn't the right size I think, which caused it to crash.

I think we should clean up these model a bit more by the way, and make sure that the size is only set in 1 location. (Not in _rowNames.length and _rowCount for instance)

boutinb (approved):

PR #4244, opened 2020-08-18
Speed up An(c)ova and .v

### LINKED ISSUE #918: RM-ANOVA in 1.13.1 *extremely slow* on Linux.
* JASP version: 1.13.1

* OS name and version: Ubuntu 18.04
* Analysis: Bayesian RM-ANOVA
* Bug description: analysis is many times slower than in previous versions (hours in 1.13.1 vs seconds or minutes in older version (0.9ish).

vandenman:
Should fix https://github.com/jasp-stats/jasp-issues/issues/918 (but let's not close that issue when merging this, let's wait with that until the contributors in that thread can verify the timings).

Here a zip file with the profiling of this PR vs Stable.

[profileAnova.zip](https://github.com/jasp-stats/jasp-desktop/files/5090352/profileAnova.zip)

The slowdown/ speedup in a nutshell:

Take the flame graph for stable:

image

The total time is 20930 ms of which 15340 ms are spend inside `.anovaCheckErrors`. If we zoom in on all the tiny bars we see:

image

For some reason, `.v` is executed a million times (or more) and it's rather slow. The culprit here seems `.findFun` / `getAnywhere`. I've changed that into:

```r
if (jaspResultsCalledFromJasp())
  return(get(name, .GlobalEnv))
else
  return(getFromNamespace(name, asNamespace("jasptools")))
```

Furthermore, `.v` is called so often because `.applyOnGroups` calls it inside its lambda. I've now ensured that it's only called once outside of the lambda.

A final optimization I did was inside `.anovaCheckErrors`. The key there is that multiple calls to `.hasErrors` could be combined to one and that we don't need to loop over all model terms, but instead over the model terms with highest interactions. For example, if the variance is nonzero after grouping on A & B & C, it's certain that the variance is also nonzero when grouping on A, B, C, A & B, A & C, and B & C (which is what the old code did twice).

Here's the flame graph for this PR:

image

The total run time is now 5390 ms of which 360 ms are spend inside `.anovaCheckErrors`.


EJWagenmakers:
nice!

JASP-Engine/JASP/R/commonerrorcheck.R
```diff
@@ -335,13 +335,16 @@
     result <- func(dataset[[.v(target)]])
     
   } else {
-    
-    result <- plyr::ddply(dataset, .v(grouping), 
-      function(data, target) {
-        if (any(is.na(data[.v(grouping)])) == FALSE) {
-          func(data[[.v(target)]])
+
```
AlexanderLyNL, line 338: Not really a part of this PR, but can we use the same structure using plyr in the above clause.

AlexanderLyNL (commented):
It looks fine to me, but I'm not sure if I'm the best person to look into this.

AlexanderLyNL:
@TimKDJ Could you please have a look this one as well? Then if it's good we can merge it.

JASP-R-Interface/jaspResults/DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Package: jaspResults
 Type: Package
 Title: Easy results for your JASP analysis
-Version: 1.14
+Version: 1.15
```
TimKDJ, line 4: Can you also adjust the `JASP_R_INTERFACE_MINOR_VERSION` in JASP.pri?

JASP-R-Interface/jaspResults/R/writeImage.R
```diff
@@ -263,10 +263,10 @@ decodeColNames <- function(x, strict = FALSE, fun = NULL, ...) {
...
   
   if (!exists(name) || !is.function(obj)) {
-    location <- getAnywhere(name)
-    for (i in seq_along(location[["objs"]]))
-      if (is.function(location[["objs"]][[i]]))
-        return(location[["objs"]][[i]])
+    if (jaspResultsCalledFromJasp())
+      return(get(name, .GlobalEnv))
+    else
+      return(getFromNamespace(name, asNamespace("jasptools")))
```
TimKDJ, line 268: This work for `jasptools`, but it will break `jaspTools` ([I dump this stuff into the globalEnv now](https://github.com/jasp-stats/jaspTools/blob/f2f3d1bd7a4cc1577748e35868860837cf36aca2/R/rbridge.R#L4-L16)) . To make sure everything keeps on working, maybe you could edit `jasptools` to put [the stuff](https://github.com/jasp-stats/jasp-desktop/blob/bf833e55c58f8e186dd2760ffeb46fb8f456f2e8/Tools/jasptools/R/rbridge.R#L3-L11) in globalEnv too?
Then this `.findFun` function can actually just go altogether.

JASP-Engine/JASP/R/ancova.R
```diff
@@ -143,33 +143,65 @@ Ancova <- function(jaspResults, dataset = NULL, options) {
...
-  for(i in length(factorModelTerms):1) {
+      observations.amount  = paste("<", length(options$dependent)+1),
+      exitAnalysisIfErrors = TRUE
+    )
+
     .hasErrors(
-      dataset = dataset, 
-      type = c('infinity', 'factorLevels'),
-      all.target = factorModelTerms[[i]][['components']], 
+      dataset              = dataset,
```
TimKDJ: Do we also require this `.hasErrors()` to be inside the loop?
Aren't we just checking to see if all factors have enough factor levels?

JASP-Engine/JASP/R/ancova.R
```diff
@@ -143,33 +143,65 @@ Ancova <- function(jaspResults, dataset = NULL, options) {
...
+
+  nWayInteractions <- lengths(factorModelTerms)
+  if (any(nWayInteractions > 1L)) {
+
+    # ensure that the largest n-way interaction effects come last
+    factorModelTerms <- factorModelTerms[order(nWayInteractions)]
+
+    # For each model term, check if it is a strict subset of another term.
+    # For example, if grouping on a three-way interaction doesn't violate any error checks
+    # then all the two-way interactions composed of variables from the three-way interaction will pass.
```
TimKDJ, line 165: Nice :)

TimKDJ (changes requested):
Good job on the investigative work.
Theoretically it sounds ok, but I could not verify that this solves the linked issue.
I tried the dataset linked [here](https://github.com/jasp-stats/jasp-issues/issues/918#issuecomment-667514078), but this already runs in about 2 seconds for me in the latest nightly.

JASP-R-Interface/jaspResults/R/writeImage.R
vandenman: we could do that, or something like:
```r
    if ("jasptools" %in% loadedNamespaces())
      return(getFromNamespace(name, asNamespace("jasptools")))
    return(get(name, .GlobalEnv))
```
I'd rather not adjust jasptools :)

TimKDJ: Fine by me too. We'll just remove the line once `jasptools` is put to rest

JASP-Engine/JASP/R/ancova.R
vandenman: I think you're absolutely right, I'll add this to the first call to `.hasErrors` that also checks for infinity. Furthermore, the check `factorLevels` seems to check the number of unique observations when you pass continuous variables as the target. But we're already checking the number of observations and the variance so we never needed to do that... I'll also change that.

TimKDJ (approved):

PR #4250, opened 2020-08-20
Export Results to PDF

JorisGoosen:
- Implements https://github.com/jasp-stats/INTERNAL-jasp/issues/173
- For some reason runs ***very slow*** in dark mode, and leads to ridiculous filesizes.
-- Example 6 page export in light theme: ~280kb rendered in seconds
-- The same in dark theme: 6MB and rendered in minutes...
-- But it *does* work in dark theme, so whatever I guess? We have very little control over how QWebEngine/Chromium prints to pdf
- Only small problem remaining is that very wide tables or images do no fit on a page, but Im not sure what would be the best way around that
- All (pdf)printing related css settings are in "printing.css"
- Also refactored/restyled darkTheme.css a little bit

### DIFF
--- a/JASP-Desktop/widgets/filemenu/actionbuttons.cpp
+++ b/JASP-Desktop/widgets/filemenu/actionbuttons.cpp
@@ -21,12 +21,12 @@ void ActionButtons::loadButtonData(std::vector<ActionButtons::DataRow> &data)
 {
 	_data =
 	{
-		{FileOperation::Open,			tr("Open"),				true,	{ResourceButtons::RecentFiles,	ResourceButtons::Computer,	ResourceButtons::DataLibrary, ResourceButtons::OSF }		},
+		{FileOperation::Open,			tr("Open"),				true,	{ResourceButtons::Computer, ResourceButtons::OSF, ResourceButtons::RecentFiles,	ResourceButtons::DataLibrary }		},
 		{FileOperation::Save,			tr("Save"),				false,	{}																														},
 		{FileOperation::SaveAs,			tr("Save As"),			false,	{ResourceButtons::Computer, ResourceButtons::OSF }																		},
 		{FileOperation::ExportResults,	tr("Export Results"),	false,	{ResourceButtons::Computer, ResourceButtons::OSF }																		},
 		{FileOperation::ExportData,		tr("Export Data"),		false,	{ResourceButtons::Computer, ResourceButtons::OSF }																		},
-		{FileOperation::SyncData,		tr("Sync Data"),		false,	{ResourceButtons::CurrentFile, ResourceButtons::Computer, ResourceButtons::OSF }										},
+		{FileOperation::SyncData,		tr("Sync Data"),		false,	{ResourceButtons::Computer, ResourceButtons::OSF, ResourceButtons::CurrentFile}										},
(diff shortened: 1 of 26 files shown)


JASP-Desktop/widgets/filemenu/actionbuttons.cpp
```diff
@@ -21,12 +21,12 @@ void ActionButtons::loadButtonData(std::vector<ActionButtons::DataRow> &data)
 {
 	_data =
 	{
-		{FileOperation::Open,			tr("Open"),				true,	{ResourceButtons::RecentFiles,	ResourceButtons::Computer,	ResourceButtons::DataLibrary, ResourceButtons::OSF }		},
+		{FileOperation::Open,			tr("Open"),				true,	{ResourceButtons::Computer, ResourceButtons::OSF, ResourceButtons::RecentFiles,	ResourceButtons::DataLibrary }		},
```
boutinb, line 24: Why did you change the order?

boutinb (commented):
Seems to work really nicely! Well done!
Only with darktheme it does not work really well... but a pdf in dark mode is quite strange I find. Maybe we could force it to be in light theme when exporting in pdf.
Such a nice feature is maybe not easy to find (who uses the Export menu?). Maybe we could also add this feature in the dropdown menu of the results: a 'save as' menu item.

JorisGoosen:
> 
> 
> Seems to work really nicely! Well done!

:)

> Only with darktheme it does not work really well... but a pdf in dark mode is quite strange I find. Maybe we could force it to be in light theme when exporting in pdf.

Yeah I thought about it, but maybe someone really wants it in dark theme? 
It does work fine, just very very slow...

> Such a nice feature is maybe not easy to find (who uses the Export menu?). Maybe we could also add this feature in the dropdown menu of the results: a 'save as' menu item.

Oh come on! 
That is where I put it originally, which was way less work and then I moved it to the filemenu because that is where @EJWagenmakers wanted it...

JorisGoosen:
I'll have to rebase it though...

JASP-Desktop/widgets/filemenu/actionbuttons.cpp
JorisGoosen: To make it look better and the order doesnt matter anyway I think

boutinb:
I don't see the problem to put it on both places.

JorisGoosen:
But isn't it a bit weird to have it in two different places?
And should the normal export results not also be there then?

And why not add a "Export data" to the dataviewer while we're at it anyway?

TimKDJ:
I didn't test it, but I'd definitely +1 forcing light mode export.
I actually think it might be a bit unexpected for users if their pdf came out all dark.
Combining that with it apparently running more smoothly for light mode...

EJWagenmakers:
I agree with Tim.

JorisGoosen:
> I actually think it might be a bit unexpected for users if their pdf came out all dark.

But their results onscreen are also dark...

EJWagenmakers:
Yeah, but that is to reduce the glare from the screen. 
If people request that the pdf has a black background then we can add that option later. My guess is that nobody will request it.

JorisGoosen:
Ok, the pdf is now always printed with the lightTheme on and can be exported directly from resultsmenu as well.

boutinb (approved):

PR #4258, opened 2020-08-26
Learn Bayes analysis crashes when 2 models is added

boutinb:
. Hopefully this fixes other crashes fount by Franticek
. In Learn Bayes, Binomial Estimation, the model is renamed to its old
value when you add a new model because the value in QML uses the key.
This is a but that will be solved later, but for the time being if you
do not use the key, it will not be updated.


FBartos:
it seems to fix the problem and also the other problems in RoBMA

FBartos (approved):

PR #4259, opened 2020-08-26
RoBMA - fixing model refitting

FBartos:
I forget to delete a `debug()` function, lol

### DIFF
--- a/JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
+++ b/JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
@@ -942,8 +942,8 @@ RobustBayesianMetaAnalysis <- function(jaspResults, dataset, options, state = NU
       ),error = function(e)e)
       
     } else{
-      debug()
-      fit <- tryCatch(RoBMA::update.robma(
+
+      fit <- tryCatch(update(
         object  = fit,
         study_names  = if (options[["input_labels"]] != "") dataset[, .v(options[["input_labels"]])],
         chains  = options[["advanced_chains"]],


JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
```diff
@@ -942,8 +942,8 @@ RobustBayesianMetaAnalysis <- function(jaspResults, dataset, options, state = NU
       ),error = function(e)e)
       
     } else{
-      debug()
-      fit <- tryCatch(RoBMA::update.robma(
+
+      fit <- tryCatch(update(
```
vandenman, line 946: So one comment about your usage of `tryCatch`. This is mainly a style thing to keep in mind for the future (I wouldn't change it for this PR).

Consider:
```r
i_crash <- function() {1 + "a"}

a <- tryCatch(i_crash(),
              error = function(e) e
)

b <- try(i_crash())

str(a)
List of 2
 $ message: chr "non-numeric argument to binary operator"
 $ call   : language 1 + "a"
 - attr(*, "class")= chr [1:3] "simpleError" "error" "condition"
str(b)
 'try-error' chr "Error in 1 + \"a\" : non-numeric argument to binary operator\n"
 - attr(*, "condition")=List of 2
  ..$ message: chr "non-numeric argument to binary operator"
  ..$ call   : language 1 + "a"
  ..- attr(*, "class")= chr [1:3] "simpleError" "error" "condition"
```
the results are basically the same. `tryCatch` stores the call and message as a list whereas `try` stores that same list as an attribute. The idea behind `tryCatch` is that you can put all the error handling in a separate function that is called whenever there is an error (or warning). The idea behind `try` is that you get an error object back and then can do whatever you like with it. In the code, you're kinda mixing the two approaches because a few lines further there is:

```r
if(any(class(fit) %in% c("simpleError", "error"))){
  JASP:::.quitAnalysis(fit[["message"]])
}
```
but since you're using `tryCatch` you don't need this if statement (in fact that `if` was already executed under the hood of the `tryCatch` function.

I'd recommend either this approach with `tryCatch`:
```r
a <- tryCatch(i_crash(),
              error = function(e) JASP:::.quitAnalysis(e[["message"]]))
```
or this approach with `try`:
```r
a <- try(i_crash(), silent = TRUE) # if you want it to be silent (I'd do that in this case since you're showing it anyway)
if (isTryError(a)) { # convenience function to test for class "try-error"
  JASP:::.quitAnalysis(as.character(a)) # same message as with tryCatch
}
```
the functionality should be identical and the choice is up to you.

vandenman (approved):

vandenman:
also, out of curiosity, what is inside your `debug` function? Logging to a file? `browser()`?

JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
FBartos: thanks! I will use this in the future

FBartos:
> also, out of curiosity, what is inside your `debug` function? Logging to a file? `browser()`?

actually, this was just a bare R debug that I must've typed into the code instead of the console. But I have a function that automatically saves options as RDS so I can conveniently load the directly to R for browser and a function that logs into messages into a file.

PR #4261, opened 2020-08-30
RoBMA - unit tests (+small bug fixes)

FBartos:
adds unit tests for Robust Bayesian Meta-Analysis

fixes the following bug:
- crashing while trying to plot a scatter-plot of model-averaged mu and tau estimates
- model preview being incorrectly deleted after showing prior distribution plots

improvements:
- grouping the prior distribution plots by effect/heterogeneity/publication bias & null/alternative so all figures do not need to be refreshed if only one of the groups is changed


FBartos:
Travis failed due to time out (all of the RoBMA unit tests proceeded successfully), the unit tests took 300s which is not that much more than Bayesian T-test (200s) which is much less complex. Maybe this won't be an issue when the unit tests are in separate modules?

TimKDJ:
Yeah we can't really add any more lengthy tests to jasp-desktop, as we've reached the time limit of Travis.
I'd just add these directly to the module instead, see also https://github.com/jasp-stats/INTERNAL-jasp/issues/1027

FBartos:
ok, I added a skip to the tests so they don't get triggered on travis. It's probably good to keep them in so I can test bug fixes I do before the release.

FBartos:
+ fixing the value of effect size transformation being changed but not properly passed to R from qml

JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
```diff
@@ -1488,13 +1513,13 @@ RobustBayesianMetaAnalysis <- function(jaspResults, dataset, options, state = NU
...
-    if(! (parameters == "theta" || (parameters == "omega" && options[["plots_omega_function"]]) ) ){
-      if(p$plot_env$any_density && p$plot_env$any_spikes){
-        p <- p + ggplot2::theme(
-          axis.title.y.right = ggplot2::element_text(vjust = 3),
-          plot.margin = ggplot2::margin(t = 3, r = 12, b = 0, l = 1))
-      }
-    }
+    if (!all(parameters %in% c("mu", "tau")))
+      if (! (parameters == "theta" || (parameters == "omega" && options[["plots_omega_function"]]) ) )
+        if(p$plot_env$any_density && p$plot_env$any_spikes)
```
vandenman, line 1518: This line looks a bit like black magic to me. Perhaps this should be part of the fit object?

vandenman (approved):

vandenman:
Something to sort out / double-check is what package version to use for RobMA. 

@JorisGoosen I used the flatpack script for Linux to install the required R pkgs but I did not get the correct version of `RoBMA`.

@FBartos If I look at the [DESCRIPTION file on the master branch of your R pkg](https://github.com/FBartos/RoBMA/blob/master/DESCRIPTION) then it has an older version (1.0.4) than [jasp-required-files](https://github.com/jasp-stats/jasp-required-files/blob/MacOS-Original/Frameworks/R.framework/Versions/3.6/Resources/library/RoBMA/DESCRIPTION) (1.0.5). At the same time, http://static.jasp-stats.org/RPkgs/ uses version 0.0.0.9000. We should probably make sure that FBartos/RoBMA is always ahead or equal to JASP and that the two JASP repos contain identical packages.

JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
FBartos: Yeah, I agree that this is not a very nice solution, however, the fit object does not need to know whether the model-averaged posterior distribution is a mixture of spikes or densities or not, especially since it varies for each parameter. Therefore, the computations are done internally in the plotting function. I think that the problem is on side of `themeJasp()` because it doesn't space the secondary y-axis from the figure in the same way as it does with the primary y-axis.

FBartos:
Yes, the master branch does not matches the JASP version because it's using the developmental version `https://github.com/FBartos/RoBMA/tree/development`. I will merge the developmental into the master and then update the CRAN once the JASP testing is finished and I do some last updates if they are needed.

vandenman:
it's fine to use a separate branch, but then flatpak should also use that branch :) Right now I had to figure out what to install from where in order to make the analysis not crash immediately after it opens.

FBartos:
sorry, should've mentioned that in the PR. I didn't realize that you need to download them on Linux. However, I have no idea what to do with Flatpack

vandenman:
no worries, this is more something for @JorisGoosen to fix.

AlexanderLyNL:
@FBartos could you please make a corresponding PR to the module?

PR #4263, opened 2020-09-01
Special Unicode characters are still not well displayed on Windows

### LINKED ISSUE #851: Mangled name in 0.13
In the table note for the Analysis of Effects -- Matched models (Bayesian ANOVA's) we say:
"Analysis suggested by Sebastiaan Matht" His last name is "Mathot", with an "accent circumflex" on the o, so "o-dakje". So this entire letter dropped out. In addition to correcting this, maybe we can include it in a unit test, because this is not the first time we experience this issue.

image

boutinb:
Fixes jasp-stats/jasp-issues#851

On Windows the gettext R function is apparently depending on the locale of the machine, and sometimes transforms for example a \u03C9 to a <U+03C9> instead of a ω character. This is afterwards considered as an
HTML tag by the WebEngine (and stays invisible). After many attempts to understand why the gettext does not work correctly on Windows (sometimes), a workaround has been decided to be added: transform in the Engine any strings containing such a string ‘<U+FFFF>’ to its UTF8 character.

### DIFF
--- a/JASP-Common/utils.cpp
+++ b/JASP-Common/utils.cpp
@@ -34,6 +34,8 @@
 #endif
 
 #include "utilenums.h"
+#include <codecvt>
+#include <regex>
 
 using namespace std;
 using namespace boost::posix_time;
@@ -398,4 +400,37 @@ std::string Utils::doubleToString(double dbl)
 	std::stringstream conv; //Use this instead of std::to_string to make sure there are no trailing zeroes (and to get full precision)
 	conv << dbl;
 	return conv.str();
+}
+
+// hex should be 4 hexadecimals characters
+std::string Utils::_convertEscapedUnicodeToUTF8(const std::string& hex)
+{
+	std::string result = hex;
+	std::istringstream iss(hex);
+
+	uint32_t bytes;
+	std::wstring_convert<std::codecvt_utf8<char32_t>, char32_t> conv;
+
+	// Read the 4 hexadecimals as bytes, and convert these bytes into UTF8.
+	if (iss >> std::hex >> bytes) result = conv.to_bytes(char32_t(bytes));
+
+	return result;
+}
+
+// Replace all <U+FFFF> in str by their UT8 characters.
+void Utils::convertEscapedUnicodeToUTF8(std::string& inputStr)
+{
+	static const std::regex unicodeExpression ("<U\\+([0-9a-fA-F]{4})>");
+
+	std::smatch match;
+	auto begin	= inputStr.cbegin();
+
+	while (std::regex_search(begin, inputStr.cend(), match, unicodeExpression))
+	{
+		std::string utf8 = _convertEscapedUnicodeToUTF8(match[1].str()); // match 1 is the first group of the regexp: that is the 4 hexadecimals.
+		auto pos = match.position(0); // position of the whole sequence in str.
+		inputStr.replace(begin + pos, begin + pos + 8, utf8);
+		// str iterators cannot be trusted after replace. They must be recalculated from str.
+		begin = inputStr.begin() + pos;
+	}
 }
--- a/JASP-Common/utils.cpp
+++ b/JASP-Common/utils.cpp
@@ -429,7 +429,7 @@ void Utils::convertEscapedUnicodeToUTF8(std::string& inputStr)
 	{
 		std::string utf8 = _convertEscapedUnicodeToUTF8(match[1].str()); // match 1 is the first group of the regexp: that is the 4 hexadecimals.
 		auto pos = match.position(0); // position of the whole sequence in str.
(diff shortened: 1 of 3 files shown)


TimKDJ:
~~Nice, it works like a charm on my Windows machine!~~

This build <-> A recent nightly (notice the ω in the footnote):
working-unicode

Actually it still doesn't work quite right. While the ω is properly shown (yay!), the string returned by gettext is still English, although I can see a Dutch translation for it in the .po file

[footnote_reliability.zip](https://github.com/jasp-stats/jasp-desktop/files/5155917/footnote_reliability.zip)

JASP-Common/utils.cpp
```diff
@@ -399,3 +401,36 @@ std::string Utils::doubleToString(double dbl)
...
+	static const std::regex unicodeExpression ("<U\\+([0-9a-fA-F]{4})>");
+
+	std::smatch match;
+	auto begin	= inputStr.cbegin();
+
+	while (std::regex_search(begin, inputStr.cend(), match, unicodeExpression))
+	{
+		std::string utf8 = _convertEscapedUnicodeToUTF8(match[1].str()); // match 1 is the first group of the regexp: that is the 4 hexadecimals.
+		auto pos = match.position(0); // position of the whole sequence in str.
+		inputStr.replace(begin + pos, begin + pos + 8, utf8);
```
JorisGoosen: Where does 8 come from?

JorisGoosen: If it is supposed to be the length of utf8 I would prefer `utf8.size()` for instance.

JorisGoosen:
But shouldn't we just set the locale gettext is supposed to be using properly?

Like, right now we have just LC_LANGUAGE or something like that, if we find out what locale setting gettext uses we could just also change that and then everything would work fine?

TimKDJ:
> 
> 
> But shouldn't we just set the locale gettext is supposed to be using properly?
> 
> Like, right now we have just LC_LANGUAGE or something like that, if we find out what locale setting gettext uses we could just also change that and then everything would work fine?

Yup that's the theory. @FransMeerhoff has been exploring this possibility the past month, but he could not figure out how. Frans maybe you can post here what you've tried?

FransMeerhoff:
@TimKDJ 
Mostly tested (and everyting of course on my - perhaps special - Windows machine)
1 trying to set the locale properly 
2 playing with the set Unicode flag on windows
3 switching to other languages
The strange thing in this issue that the locale does not seem to play a role at all.
See the difference on my Windows machine from an R-prompt and R-gui
Screen Shot 2020-09-02 at 09 59 48

 I'm aware of course that what I tested was only on my Virtual windows and really not knowing what the gettext is doeiing under the hood.
Did you find any time to look at this issue too, as you promised?

JorisGoosen:
@FransMeerhoff you could make VMs with other installs of windows you know. Ive been testing stuff on a german and a korean one. Quite helpful.

boutinb:
I think this fix can be considered as a fallback feature: when the gettext generated such <U+FFFF> strings, the least we can do is to change these strings into the right Unicode character. This does not solve the translation problem, but at least the english version is then correct.
The right fix can come only if jasp-stats/INTERNAL-jasp#1034 is fixed.

JASP-Common/utils.cpp
```diff
@@ -399,3 +401,36 @@ std::string Utils::doubleToString(double dbl)
...
 }
+
+// hex should be 4 hexadecimals characters
+std::string Utils::_convertEscapedUnicodeToUTF8(const std::string& hex)
+{
+	std::string result = hex;
+	std::istringstream iss(hex);
+
+	uint32_t bytes;
+	std::wstring_convert<std::codecvt_utf8<char32_t>, char32_t> conv;
```
JorisGoosen: Could be `static const` maybe? you are reusing it quite a lot.

JASP-Common/utils.cpp
```diff
@@ -399,3 +401,36 @@ std::string Utils::doubleToString(double dbl)
 	conv << dbl;
 	return conv.str();
 }
+
+// hex should be 4 hexadecimals characters
+std::string Utils::_convertEscapedUnicodeToUTF8(const std::string& hex)
```
JorisGoosen: If you change the argument to `std::string hex` you do not need to define `std::string results = hex;` anymore.

JorisGoosen (changes requested):
I think those two small changes will make the _convertEscapedUnicodeToUTF8 function a bit nicer.
After that it can just be merged.

JASP-Common/utils.cpp
boutinb: This function is used only exceptionally, only when a gettext returns a <U+FFFF> string. But a static is not wrong.

PR #4264, opened 2020-09-01
Adding a lot of labels might require the dataset to be enlarged

### LINKED ISSUE #938: Large dataset: Crash when changing variable type
* JASP version: 0.14 (nightly build of August 14th, 075f70b4e3b7cd994b0efe9bf78da79cc6468b83)
* OS name and version: macOS 11.0 Big Sur (Public beta)
* Analysis: N/A (spreadsheet)
* Bug description: JASP crashes when changing variable type in a large dataset
* Expected behaviour: JASP should not crash

**Steps to reproduce**
1. Open the attached dataset, which contains 85,938 cases.
2. In the spreadsheet, change the variable type of Order ID from Scale to Nominal.
3. JASP crashes immediately.

The dataset contains the first 85,938 cases of the (even larger) "100000 Sales Records" dataset available from [E for Excel](http://eforexcel.com/wp/downloads-18-sample-csv-files-data-sets-for-testing-sales/), with which the same crash occurs. I removed cases to make the dataset smaller. I found that the crash does not occur when the dataset contains the first 85,157 cases.

JASP 0.10.2 shows the same behaviour as JASP 0.14:
85,157 cases => no crash upon changing variable type
85,938 cases => crash upon changing variable type
&nbsp;

[85938 Sales Records.csv.zip](https://github.com/jasp-stats/jasp-issues/files/5078300/85938.Sales.Records.csv.zip)

JorisGoosen:
- fixes https://github.com/jasp-stats/jasp-issues/issues/938

### DIFF
--- a/JASP-Desktop/data/datasetpackage.cpp
+++ b/JASP-Desktop/data/datasetpackage.cpp
@@ -714,7 +714,9 @@ bool DataSetPackage::setColumnType(int columnIndex, columnType newColumnType)
 	if (_dataSet == nullptr)
 		return true;
 
-	bool changed = _dataSet->column(columnIndex).changeColumnType(newColumnType);
+	bool changed;
+
+	enlargeDataSetIfNecessary([&](){ changed = _dataSet->column(columnIndex).changeColumnType(newColumnType); }, "setColumnType");
 
 	if (changed)
 	{


JASP-Desktop/data/datasetpackage.cpp
```diff
@@ -714,7 +714,9 @@ bool DataSetPackage::setColumnType(int columnIndex, columnType newColumnType)
 	if (_dataSet == nullptr)
 		return true;
 
-	bool changed = _dataSet->column(columnIndex).changeColumnType(newColumnType);
+	bool changed;
+
+	enlargeDataSetIfNecessary([&](){ changed = _dataSet->column(columnIndex).changeColumnType(newColumnType); }, "setColumnType");
```
boutinb, line 719: I'm afraid that this enlargeDataSetIfNecessary should be called from other places (Column::setName, Columns:setColumnAsScale...).
Also this enlargeDataSetIfNecessary should throw forward exceptions that are not of type bad_alloc.
But maybe you wanted to get rid of this shared memory, and this fix is temporary?

JorisGoosen: I think if anything breaks inside `enlargeDataSetIfNecessary` we are fucked anyway I think.

Well, this fix is necessary to have JASP not crash in this particular way. I'm sure there are more ways we could make it crash... If someone changes there columnname to something large enough to be too big for the dataset... Or who knows?
But the problem is that I can't really use it inside `Column` or `Label` because those are part of JASP-Common.
And `enlargeDataSetIfNecessary` needs to (possibly) update the pointer to the `DataSet` in `DataSetPackage` which isn't available in JASP-Common.

Definitely part of the reason why I want to get rid of this sharedmemory thing, it is so very fragile.

PR #4269, opened 2020-09-03
Some problems in Learn Bayes analyses

boutinb:
Fixes jasp-stats/INTERNAL-jasp#1035
Fixes jasp-stats/INTERNAL-jasp#1036

. When a TextField has a binding for its value (e.g TextField { value:
“TEST ” + myCheckBox.checked ? “YES” : “NO” } ), then if you change the
value manually, and then change the value of the binding (in the
example, you click the myCheckBox), then the value of the TextField is
changed according to the binding.
The startValue property has been added in order to avoid this
behaviour: the startValue is used only when the TextField is created,
and the binding will not affect anymore the value of the TextField.

. When a ComponentsList (or a VariablesList) uses a rowComponent having
the same name as the optionKey, no error was thrown (this generates the
problem in issue jasp-stats/INTERNAL-jasp#1036). An error is now
displayed.

. the JASP constants was not imported in the QML analyses (import JASP
1.0)

. A reference to the id plotsPosteriorMarginalLower did not exist in
LStestinginterface.qml

### DIFF
--- a/JASP-Desktop/components/JASP/Controls/TextField.qml
+++ b/JASP-Desktop/components/JASP/Controls/TextField.qml
@@ -37,6 +37,7 @@ JASPControl
 	property alias	label:				beforeLabel.text
 	property alias	text:				beforeLabel.text
 	property alias	value:				control.text
+	property string startValue:			""
 	property string lastValidValue:		""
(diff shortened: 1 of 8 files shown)


FBartos (approved):
Thanks Bruno! I tested it and it fixes both raised issues.

PR #4272, opened 2020-09-04
Add simple table view

boutinb:
. A simple TableView needed for Learn Bayes module. The TableViews must be refactored after 0.14.
. The AnovaCustomContrast is removed as it is not used anymore
. The Language files should not be generated automatically, as it sets another datum in the po files, and then modify all po and mo files.


FBartos:
It's looking good. I have just shown it to EJ and we had three small feature requests and one bug report:
```
	SimpleTableView
	{
		name:				"test"
		cornerText:			"Player"
		buttonAddText:		"Add player"
		buttonDeleteText:	"Delete player"
		values:				["Pr(win a point)", "Points gained"]
		columnName:			""
	}
```
image

bug:
- the `columnName` variable does not seem to work

feature requests:
- could the automatically-generated column names be "A", "B", "C", ... , "Z" instead of 1, 2, 3, .. N ?
- can we set the default number of columns to 2?
- could the "Reset" button be aligned with the lower line of the box (according to the red line -- don't add the red line please :)
image

JoKeyser:
I've got a tiny (unsolicited) remark from a translation perspective: Maybe you label the field `Pr(win a point)` with `P(win 1 point)`, or `P(win)`? As it is, a literal translation to German would be quite lengthy: "P(Einen Punkt gewinnen)", and there's no good way to translate `Pr` directly.

EJWagenmakers:
It needs to be clear whether the win refers to the point or to the game. How about Pr(Punktgewin) and Pr(Sieg)? Sieg would be generally understood to refer to the entire game, not a single point. 
And yes we debated whether to use Pr, P, or p. For consistency, we should choose "p". But for this educational module, I felt that maybe "Pr" would be clearer....but if everybody tells me "p", we'll do "p"

JoKeyser:
@EJWagenmakers , indeed, "Sieg" would entail winning the entire game :).
To me, either "p(1 Punktgewinn)" or "p(1 Punkt gewinnen)" seem quite understandable; using a number makes it shorter.
This would correspond to "p(win 1 point)", or in Dutch "p(win 1 punt)".

Using "Pr" may be a nice reminder in English, but it doesn't translate.
And consistently using p() has the benefit of transfer to other modules etc.

Maybe it would be useful to add clarifying full sentences when you hover over the fields.

JASP-Engine/JASP-Engine.pro
```diff
@@ -50,37 +50,6 @@ win32:LIBS += -lole32 -loleaut32
 
 mkpath($$OUT_PWD/../R/library)
 
-exists(/app/lib/*) {
-	#Actually org.jaspstats.JASP.json already does this...
```
JorisGoosen: Why did you throw all of this away?

JorisGoosen (changes requested):

JASP-Engine/JASP-Engine.pro
boutinb: oeps, I have committed too quickly!

boutinb:
> It's looking good. I have just shown it to EJ and we had three small feature requests and one bug report:
> 
> ```
> 	SimpleTableView
> 	{
> 		name:				"test"
> 		cornerText:			"Player"
> 		buttonAddText:		"Add player"
> 		buttonDeleteText:	"Delete player"
> 		values:				["Pr(win a point)", "Points gained"]
> 		columnName:			""
> 	}
> ```
> 
> image
> 
> bug:
> 
> * the `columnName` variable does not seem to work
> 
> feature requests:
> 
> * could the automatically-generated column names be "A", "B", "C", ... , "Z" instead of 1, 2, 3, .. N ?
> * can we set the default number of columns to 2?
> * could the "Reset" button be aligned with the lower line of the box (according to the red line -- don't add the red line please :)
>   image

The initialColumnCount property can be set to 2.
You  can now set the column header text in this way:
```
SimpleTableView
	{
		name: "test"
		values: ["one", "two"]
		buttonAddText: "Add Player"
		buttonDeleteText: "Delete Player"
		cornerText: ""
		initialColumnCount: 2
		//showResetButton: false
		buttonsInRow: true

		function getColHeaderText(defaultName, colIndex) { return String.fromCharCode(65 + colIndex); }
	}
```
This will set 'A' , 'B', etc as column names.

For the reset button, there are 2 solutions: either you remove it (showResetButton to false), or you can also set the buttons in a row above the table (buttonsInRow to true).

boutinb:
@FBartos Did you try my last changes?

FBartos:
sorry, I missed the update. Just tried it and it works wonderfully. Exactly what I needed. Thanks!

FBartos (approved):

PR #4280, opened 2020-09-09
Setting an interface font makes data disappears

boutinb:
The main problem was that when the interface font is changed, it should call (like for uiScale change) the viewPort change slot of the DataSetView. 

The DropDown settings in PrefsUI.qml were confusing: the value was set to be the preferencesModel.interfaceFont, but onValueChanged signal sets the preferencesModel.interfaceFont to value.
The value should not be directly set in a DropDown, only the currentIndex. If no currentIndex is known, then a new property is made ‘startValue’ that sets internally the right currentIndex.

Moreover, when an item is being initialised, the properties are changed from empty to the specified value. So onValueChanged is called twice (empty and specified value), without real interaction with the user. These calls should be avoided, so the property ‘initialized’ has been added, and is set to true only when the control is fully initialised, and can be used to avoid these extra onValueChanged calls.

Remove obsolete isDirectModel property in ComboBox

### DIFF
--- a/JASP-Desktop/analysis/jaspcontrolbase.cpp
+++ b/JASP-Desktop/analysis/jaspcontrolbase.cpp
@@ -87,7 +87,11 @@ void JASPControlBase::componentComplete()
 	if (!hasContextForm)
 	{
 		if (_form)	_form->addControl(this);
-		else		_wrapper->setUp();
+		else
+		{
+			_wrapper->setUp();
+			setInitialized();
+		}
 	}
 	else
 	{
@@ -95,6 +99,7 @@ void JASPControlBase::componentComplete()
 		if (!noDirectSetup)
 			_wrapper->setUp();
 
+		setInitialized();
 		QMLListView* listView = nullptr;
 
 		QVariant listViewVar = context->contextProperty("listView");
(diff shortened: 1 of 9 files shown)


JASP-Desktop/analysis/jaspcontrolbase.cpp
```diff
@@ -87,14 +87,19 @@ void JASPControlBase::componentComplete()
 	if (!hasContextForm)
 	{
 		if (_form)	_form->addControl(this);
-		else		_wrapper->setUp();
+		else
+		{
+			_wrapper->setUp();
+			setInitialized();
```
JorisGoosen, line 93: Why isn't this called from the constructor of JASPControlBase/Wrapper?

JASP-Desktop/analysis/jaspcontrolbase.h
```diff
@@ -175,6 +179,7 @@ public slots:
 							_title,
 							_parentListViewKey;
 	bool					_isBound				= true,
+							_initialized			= false,
```
JorisGoosen, line 182: Ok, the wrapper then? ;')
Hmm.. Nvm

JorisGoosen (approved):

PR #4282, opened 2020-09-11
Learn Bayes - unit tests

FBartos:
adds unit tests to the Learn Bayes module


JorisGoosen:
@vandenman @TimKDJ does it makes sense that a million svg's are changed?

FBartos:
these are completely new unit tests - they are changed since they are freshly generated

vandenman:
So I'm afraid that these tests will make Travis time out. @JorisGoosen it makes sense that there a million svg files, they didn't change but are new files. 

However, I do think that both the test files and the stuff tested could be shorter. For example, all stuff with `options$.meta <-` is just unnecessary stuff generated by jasptools.

FBartos:
> For example, all stuff with options$.meta <- is just unnecessary stuff generated by jasptools.

Oh, didn't know this. Thanks. I will delete it.

> However, I do think that both the test files and the stuff tested could be shorter.

Regarding the test cases, I would rather not delete them since there is way too many different visualizations in the module and I'm pretty certain that I will break some during updates. I tried to cover just the ones that are most likely to be used/broken by additional changes.
Also, the module splitting should take care of the Travis timing out for the future releases, right? But it would be helpful for me to be able to test the analysis for this release if I have to add some changes during testing. I can add skip to the tests for now.

vandenman:
Well, let's just make sure we don't test the same pathway twice. Also, the unit tests failed because `there is no package called ‘HDInterval’`. Do we really need that package though? What's wrong with `coda::HPDinterval`?

FBartos:
`coda::HPDinterval` uses samples from posterior distribution - my analysis does not use any sampling at all. However, `HDInterval::hdi` allows to use quantile functions which I need for the analyses.

Joris just added the package, so it shouldn't fail because of it and I removed the.meta stuff

kornbrot:
Really appreciate all you have done so far.
Is there any hope of my being able to use Bayes Generalized linear on the file I sent you.
Have now succeeded in R but have many files and would really prefer JASP if possible - am so close!
Rcode
bayeslgt <- brm(freq|trials(Nmax) ~ wiithin + (wiithin|pno), pars="all", data=kex12,family=binomial(link="logit”))
Message also on GitHub
Best
Diana
____________
University of Hertfordshire
College Lane, Hatfield, Hertfordshire AL10 9AB, UK
+44 (0) 208 444 2081
+44 (0) 7403 18 16 12
d.e.kornbrot@herts.ac.uk<mailto:d.e.kornbrot@herts.ac.uk>
http://dianakornbrot.wordpress.com/
skype:  kornbrotme
Save our in-boxes! http://emailcharter.org
 __________________

vandenman:
@FBartos sure it won't fail now but at the same time, you're setting an extremely low bar for adding new packages. I'd prefer to keep the dependencies to a bare minimum even if that means reimplementing small functions from other packages. `HDInterval` is not exactly doing that much (see `HDInterval:::hdi.function`). You could remove the dependency by adding:

```r
my_hdi <- function(object, credMass = 0.95, ...) {
  intervalWidth <- function(lowTailPr, ICDF, credMass, ...) {
    ICDF(credMass + lowTailPr, ...) - ICDF(lowTailPr, ...)
  }
  optInfo <- stats::optimize(intervalWidth, c(0, 1 - credMass), ICDF = object, credMass = credMass, tol = tol, ...)
  HDIlowTailPr <- optInfo$minimum
  return(c(lower = object(HDIlowTailPr, ...), upper = object(credMass + HDIlowTailPr, ...)))
}

# example
my_hdi(qnorm)
    lower     upper 
-1.959964  1.959964 
```

vandenman:
`The job exceeded the maximum time limit for jobs, and has been terminated.`

I'll keep restarting until it doesn't. 
Times restarted = 2.

FBartos:
ok, I stole the functions that I needed from the HDI package and I added `skip_on_travis()` to the tests (I ran them locally and verified that they work)

Kucharssim:
> ok, I stole the functions that I needed 

Good. I agree with @vandenman as it's really not ideal to install a million packages, each for one simple function. I would not call it "stealing" though, although we derive the code from the original package. I quickly checked that HDInterval is under GNU GPL-3 License. I think this is legal (I am not sure though whether we need to attribute the authors of the original code), but maybe we should pay attention to these things? @EJWagenmakers

EJWagenmakers:
I am hardly the authority on these issues, but it does seem nice to acknowledge the package (even though we don't use it any more) and indicate this clearly in the code as well.
E.J.

vandenman:
I'd put the source of the code in a comment as the first line inside the function, for example, `# adapted from HDIinterval:::hdi.function`. Since the vast majority of R packages are under extremely permissive licenses (like GNU GPL-3) this is not stealing and crediting the package is by not a requirement (I'm not a laywer though). Nevertheless, it's a good idea to put the location of the source in there so that we can look it up in the future. A quick search for `adapted from` shows that we already do so for a number of functions (either to reduce dependencies or because the function needed to be changed in some spots).

FBartos:
ok, I added the line about the source of the function. I think it should be fine now

FBartos:
and I added some small additional changes to the interface/plotting

vandenman:
(Since 383db7b)

`The job exceeded the maximum time limit for jobs, and has been terminated.`

I'll keep restarting until it doesn't. 
Times restarted = 2.

FBartos:
hmm, that's weird since I set `skip_on_travis()` to both of the tests

FBartos:
if you check the travis log, you will see that it didn't even comes to the tests. It fails somewhere earlier
```

The command "eval brew update >/dev/null " failed. Retrying, 2 of 3.
fatal: unable to access 'https://github.com/Homebrew/homebrew-core/': transfer closed with outstanding read data remaining
Error: Fetching /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core failed!
The command "eval brew update >/dev/null " failed. Retrying, 3 of 3.
$ brew install checkbashisms || true
Updating Homebrew...
```

vandenman:
Well, we have so many tests that the tests were already failing sporadically before this PR. `skip_on_travis` still takes a very tiny amount of time...

vandenman:
Here you can see the overview of the builds:

https://travis-ci.org/github/jasp-stats/jasp-desktop/pull_requests

and here your previous build:

https://travis-ci.org/github/jasp-stats/jasp-desktop/builds/727933819

the current build is looking kinda weird though...

FBartos:
I'm sorry but I have no idea what that's supposed to tell me

vandenman:
@FBartos it tells you why the previous one failed (it timed out) and how to see previous logs in general :) As for the current build, it doesn't provide any information...

FBartos:
so, we need someone who understand the build and travis to fix those?

FBartos:
@vandenman now I see that it passed? so, is it ok to merge it now?

Kucharssim:
> so, is it ok to merge it now?

I'll still have a quick check

Kucharssim (approved):
Ok, it looks to me as good to go. The new tests also pass on my computer.

PR #4283, opened 2020-09-14
The description for "Install Module" should come from QML

JorisGoosen:
- Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/1012

### DIFF
--- a/JASP-Desktop/components/JASP/Widgets/ModuleInstaller.qml
+++ b/JASP-Desktop/components/JASP/Widgets/ModuleInstaller.qml
@@ -127,8 +127,8 @@ Popup
 						onLinkActivated:	Qt.openUrlExternally(link)
 						clip:				true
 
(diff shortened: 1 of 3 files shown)


boutinb:
If the Description.qml is missing, you get a confusing error message: 
Screen Shot 2020-09-14 at 14 08 59
A user-friendler error message would be welcome!

JorisGoosen:
This issue is for "Install Module" not for "Install Developer Module" and so this error is entirely unrelated.

JorisGoosen:
Well, one shouldn't be allowed to install it if the description is missing...

boutinb:
> Well, one shouldn't be allowed to install it if the description is missing...

For sure, but with a message telling what is missing would be nicer.

JorisGoosen:
Exactly, that is why i fixed it, the message isnt very clear but people should just put Description.qml in there, it's like nr 1 in to-do list of making a module :p

PR #4286, opened 2020-09-15
update augmented values Wilcoxon test

JohnnyDoorn:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/1046#event-3769843777


TimKDJ:
Any idea why this was not caught by a unit test?

JohnnyDoorn:
I think the case in the unit test still led to a very similar Bayes factor with/without this small addition (because the unit test uses a normally distributed variable), so I just chalked it up to the random variation of MCMC sampling. For Lexi's data set (debEqual1) this was not the case, so maybe it's better to replace the unit test with a non-normal variable.

AlexanderLyNL:
The code looks sensible. Let's double check in the next nightly.

PR #4293, opened 2020-09-21
JAGS: make data in model available for initial values and userdata

vandenman:
So since https://github.com/jasp-stats/jasp-desktop/pull/4223 JAGS supports multiple values. I noticed yesterday that when QML verifies the R code, the R code is evaluated in an environment with access to the column names in the dataset. However, when parsing the code in R you'd still get an error because there the data was not available. 

This PR ensures that `eval(parse(...))` also has access to the data used in the model.

I'd like this to be in 0.14 since (1) the changes are kinda small and (2) evaluating the R code in a different environment in QML as opposed to the analysis could be seen as a bug. At the same time, I understand that it's kinda late to merge this though.

### DIFF
--- a/JASP-Engine/JASP/R/jagsModule.R
+++ b/JASP-Engine/JASP/R/jagsModule.R
@@ -84,10 +84,10 @@ JAGS <- function(jaspResults, dataset, options, state = NULL) {
   }
   
   # Evaluate user R code, terminate early if the code doesn't work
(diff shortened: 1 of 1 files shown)


JorisGoosen (approved):

JorisGoosen:
This change looks benign enough to merge and definitely an improvement!

PR #4295, opened 2020-09-22
RM ANOVA Dont make referenceGrid when error

JohnnyDoorn:
Fix https://github.com/jasp-stats/jasp-test-release/issues/983


JohnnyDoorn:
@vandenman I added a commit here to deal with an issue in descriptive plots in BANOVA (usePooledSE was not defined in the right place). Didn't think it necessary to make an issue/separate PR because it's only one line, but just keeping you in the loop about altering your precious BANOVA code ;)

vandenman (approved):

PR #4297, opened 2020-09-23
Fix scaling of cells in DataSetViewer

JorisGoosen:
- Fixes https://github.com/jasp-stats/jasp-test-release/issues/914
- Removed "font" from datasetview because it wasn't being used consistently anyway, and is unnecessary with JaspTheme being available
- getTextSize was mistakenly multiplying a size by uiScale that already had it's pixelSize multiplied by that, leading to ridiculously large cells when zoomed in far

### DIFF
--- a/JASP-Desktop/qquick/datasetview.cpp
+++ b/JASP-Desktop/qquick/datasetview.cpp
@@ -7,22 +7,23 @@
 #include "timers.h"
 #include "log.h"
 #include "gui/preferencesmodel.h"
+#include "qquick/jasptheme.h"
 
 DataSetView * DataSetView::_lastInstancedDataSetView = nullptr;
 
-DataSetView::DataSetView(QQuickItem *parent) : QQuickItem (parent), _metricsFont(_font)
+DataSetView::DataSetView(QQuickItem *parent) : QQuickItem (parent)
 {
 	setFlag(QQuickItem::ItemHasContents, true);
 	setFlag(ItemIsFocusScope);
 
 	material.setColor(Qt::gray);
 
-	connect(this, &DataSetView::parentChanged, this, &DataSetView::myParentChanged);
+	connect(this, &DataSetView::parentChanged,					this, &DataSetView::myParentChanged);
 
-	connect(this, &DataSetView::viewportXChanged, this, &DataSetView::viewportChanged);
-	connect(this, &DataSetView::viewportYChanged, this, &DataSetView::viewportChanged);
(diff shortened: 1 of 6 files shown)


JASP-Desktop/qquick/datasetview.cpp
```diff
@@ -351,6 +332,13 @@ void DataSetView::addLine(float x0, float y0, float x1, float y1)
 	_lines[_linesActualSize++] = y1;
 }
 
+QSizeF DataSetView::getTextSize(const QString & text) const
+{
+	QFontMetricsF metricsFont(JaspTheme::currentTheme()->font());
```
boutinb: The QFontMetricsF should be a property of the currentTheme: it should not be created each time it is needed, but only when the font is changed.

JorisGoosen: Yeah, but the problem is that I'm not entirely sure if it responds properly when the `pixelSize` gets set in `Theme.qml`.
And I also don't really know how to check it.

If you look at: https://github.com/jasp-stats/jasp-desktop/blob/8a6fdb171a816bbba653b1943e8e016399504db6/JASP-Desktop/components/JASP/Theme/Theme.qml#L29
Does a change to uiScale get propagated? 

Although I guess I could just reset the stored QFontMetricsF whenever uiScale changes?
Because that wouldn't be very different from how I do it there right?
Yeah I can change it to something like that.

boutinb: There is maybe a concurrency problem here (I love concurrency problem): if uiScale is changed, the data view change can be first called before the fontmetrics is changed...

JorisGoosen: good point, check the latest commit.

JASP-Desktop/qquick/jasptheme.cpp
```diff
@@ -133,7 +138,12 @@ void JaspTheme::setCurrentTheme(JaspTheme * theme)
 	_currentTheme = theme;
 
 	if(_currentTheme)
+	{
 		emit _currentTheme->jaspThemeChanged(theme);
+		_currentTheme->updateFontMetrics();
```
boutinb, line 142: Last comment: I would inverse these 2 lines

JorisGoosen: Yeah ok, that is reasonable

boutinb (approved):

PR #4300, opened 2020-09-23
Try the renaming of *.pdf.tmp to *.pdf several times

JorisGoosen:
- Only on windows and sleep in between...
- Fixes https://github.com/jasp-stats/jasp-test-release/issues/957


TimKDJ:
This fixes the .tmp problem for me

boutinb:
Gool old school coding, perfect!

PR #4301, opened 2020-09-24
Update the translation files for Weblate

FransMeerhoff:


JASP-Desktop/po/jasp_de.po
```diff
@@ -19873,9 +19855,10 @@ msgctxt "LDgamma|"
 msgid "Scale:"
 msgstr ""
 
+#, fuzzy
 msgctxt "LDgamma|"
 msgid "Rate:"
-msgstr ""
+msgstr "Rate:"
```
TimKDJ, line 19861: @boutinb, @FransMeerhoff : So will filled in fuzzies be marked as translated on Weblate? Because then we'll want to empty them before pushing

boutinb: fuzzy strings get a 'Needs Action' warning in WebLate. You can also select all Needing action strings to check them in WebLate.

TimKDJ: Ah, neat! Good to hear it isn't too barebones

PR #4303, opened 2020-09-24
Replacing some unicode in R by either different ascii chars etc

JorisGoosen:
- implements https://github.com/jasp-stats/jasp-test-release/issues/984
- these changes should not be copied to the modules' repos because we will just fix `do_gettext` later


JorisGoosen:
After this is merged I'll build a nightly and check it out with some crazy usernames or in some "strange" locale

JASP-Engine/JASP/R/correlation.R
```diff
@@ -401,7 +401,7 @@ Correlation <- function(jaspResults, dataset, options){
       }
       
       result$vsmpr <- JASP:::.VovkSellkeMPR(result$p.value)
-      result$vsmpr <- ifelse(result$vsmpr == "∞", Inf, result$vsmpr)
+      result$vsmpr <- ifelse(result$vsmpr == "\u221E", Inf, result$vsmpr)
```
boutinb, line 404: Did you test this??
I have tried, but I cannot get  a situation where this vsmpr  becomes infinite.

JASP-Engine/JASP/po/R-JASP.pot
```diff
@@ -622,7 +622,7 @@ msgstr ""
 msgid "Model fit is singular. Specified random effects parameters (random intercepts and random slopes) cannot be estimated from the available data. Carefully reduce the random effects structure, but this practice might inflate the reported p-value, and invalidates the analysis."
 msgstr ""
 
-msgid "Vovk-Sellke Maximum <em>p</em>-Ratio: Based on a two-sided <em>p</em>-value, the maximum possible odds in favor of H%1$s over H%2$s equals 1/(-e <em>p</em> log(<em>p</em>)) for <em>p</em> %3$s .37 (Sellke, Bayarri, & Berger, 2001)."
+msgid "Vovk-Sellke Maximum <em>p</em>-Ratio: Based on a two-sided <em>p</em>-value, the maximum possible odds in favor of H1 over H0 equals 1/(-e <em>p</em> log(<em>p</em>)) for <em>p</em> = .37 (Sellke, Bayarri, & Berger, 2001)."
```
boutinb: Did you forget to add this change in  MixedModelsCommon.R in your commit? How such a change can come in the pot file, if the original string is not changed?

JASP-Engine/JASP/R/correlation.R
JorisGoosen: No I didn't, but I looked up the codepoint for infinity and it is that: https://codepoints.net/U+221E

JASP-Engine/JASP/po/R-JASP.pot
JorisGoosen: Someone else changed it in `stable` I think, it appeared when I ran the "language generation" thing of jasp.
But maybe I made a mistake?

JorisGoosen: Hmm, I can't see where it came from to be honest. I'll revert that.

JASP-Engine/JASP/R/correlation.R
TimKDJ: This line can just be removed. In https://github.com/TimKDJ/jasp-desktop/commit/9a821494b6deda4800ea8f4538701d88c279007f# I changed the `.VovkSellkeMPR()` function so it returns a numeric value rather than a character.

JASP-Engine/JASP/R/correlation.R
```diff
@@ -431,7 +431,7 @@ Correlation <- function(jaspResults, dataset, options){
         }
       }
       result$vsmpr <- JASP:::.VovkSellkeMPR(result$p.value)
-      result$vsmpr <- ifelse(result$vsmpr == "∞", Inf, result$vsmpr)
+      result$vsmpr <- ifelse(result$vsmpr == "\u221E", Inf, result$vsmpr)
```
TimKDJ, line 434: This line can be removed too

TimKDJ (changes requested):
So I build your branch, removed all fuzzies and problematic translations, then compiled the translations and tested the two `LC_CTYPE` settings:

```
env.insert("LC_CTYPE",			"C");
```
withCSpecTranslation

-----
```
//env.insert("LC_CTYPE",			"C");
```
withoutCSpecTranslation

I guess this isn't going to work properly without our UTF-8 `do_gettext()`.

JorisGoosen:
Ah man, not even umlauts works :(
Perhaps the LC_CTYPE being set to C should be an option in the Windows version of JASP?
So that, when it doesn't work they can turn it off?
Bleh

TimKDJ:
> Perhaps the LC_CTYPE being set to C should be an option in the Windows version of JASP?
> So that, when it doesn't work they can turn it off?

I guess we could have it as a setting, but I'm doubtful anyone will figure out when they need it unless they post on our GitHub page...

Edit: Although I guess if we make a setting and we do not set it to `C` by default, then for most users this will be an improvement. Only when people use JASP in a different language could it pose a problem, and then they can specify they want it set to `C`. This would only be for one release, right?

JorisGoosen:
> Edit: Although I guess if we make a setting and we do not set it to `C` by default, then for most users this will be an improvement. Only when people use JASP in a different language could it pose a problem, and then they can specify they want it set to `C`. This would only be for one release, right?

Ah yes, that could work!
And then when a user switches to german (I'm guessing we didn't put anything strange in dutch) we could:
- Check where JASP is installed, if the path does not contain any non-ascii chars.
- Check whether the username does not contain any non-ascii chars.
If both checks are good we could just turn on LC_CTYPE anyway. 

Maybe we could just do this check anyway and turn it on/off based on that?
It shouldn't be too hard and then we don't need to bother the user with this setting. Although we can add it for manual override under "Advanced - Developer Mode" or something.

And yeah, that would be just for this release.

TimKDJ:
Dutch does have stuff like é and ë. There were too many errors in the Dutch translations file so I didn't test that one.
Otherwise I think what you propose sounds like a good solution (with the automatic checks), granted that it actually solves the path problem :P

JorisGoosen:
> 
> 
> Dutch does have stuff like é and ë. There were too many errors in the Dutch translations file so I didn't test that one.
> Otherwise I think what you propose sounds like a good solution (with the automatic checks), granted that it actually solves the path problem :P

Hahaha, yeah if it doesn't do that then this whole PR is quite pointless :p

JorisGoosen:
So in light of the new setting I'll be adding this PR will be pointless (given that german still looks bad).

JorisGoosen:
@TimKDJ is right, maybe we want this PR as well.

JorisGoosen:
It will have to be rebased after #4311 is merged

FBartos:
Actually, it seems like that the previous version worked better:
image

than this new one
image

However, compiling on my EN PC, the previous version with "š" is also broken

```
    stats$addColumnInfo(name="Variable",   title=gettext("Variable"), type="string")
    stats$addColumnInfo(name="Variable1",  title=gettext("\u03C4"), type="string")
    stats$addColumnInfo(name="Variable2",  title=gettextf("%s a", "\u03C4"), type="string")
    stats$addColumnInfo(name="Variable3",  title=gettextf("%s \u03C4", "a"), type="string")
    stats$addColumnInfo(name="Variable4",  title=gettext("š"), type="string")
    stats$addColumnInfo(name="Variable5",  title=gettextf("%s a", "š"), type="string")
    stats$addColumnInfo(name="Variable6",  title=gettextf("%s š", "a"), type="string")
```

JorisGoosen:
Yeah, this one doesn't actually fix anything like, @FBartos this one ought to: https://github.com/jasp-stats/jasp-desktop/pull/4311
But it seems to be behaving weirdly on my system now...

JorisGoosen:
Merged it with the other PR

PR #4305, opened 2020-09-24
Font should not change when a new theme is set

boutinb:
Fixes jasp-stats/jasp-test-release#1008

### DIFF
--- a/JASP-Desktop/qquick/jasptheme.cpp
+++ b/JASP-Desktop/qquick/jasptheme.cpp
@@ -4,38 +4,23 @@
 #include <QFontDatabase>
 
 JaspTheme			*	JaspTheme::_currentTheme	= nullptr;
-QFont					JaspTheme::_jaspFont		= QFont("SansSerif");
-QFont					JaspTheme::_jaspCodeFont	= QFont("FiraCode-Retina");
-QFontMetricsF			JaspTheme::_fontMetrics		= QFontMetrics(_jaspFont);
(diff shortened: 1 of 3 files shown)


JASP-Desktop/qquick/jasptheme.cpp
```diff
@@ -135,6 +135,9 @@ void JaspTheme::setCurrentTheme(JaspTheme * theme)
 	if(theme == _currentTheme)
 		return;
 
+	if (_currentTheme && theme)
+		theme->setFont(_currentTheme->font());
```
JorisGoosen: Shouldn't the same happen for `_jaspCodeFont`?

JorisGoosen: As in, the init does:
```
	setFont(						_jaspFont);
	setFontLabel(					_jaspFont);
	setFontRibbon(					_jaspFont);
	setFontGroupTitle(				_jaspFont);
	setFontPrefOptionsGroupTitle(	_jaspFont);
	setFontCode(					_jaspCodeFont);
	setFontRCode(					_jaspCodeFont);
```

So why not all of those here?

JorisGoosen: Maybe it is better to make all the fonts static?
There not really supposed to be set from the theme now that we have font-settings right?

boutinb: Yep, and my fix does not work well anyway...

boutinb:
@JorisGoosen I have changed my fix

PR #4307, opened 2020-09-25
Column should not be remove when invisible

boutinb:
Fixes jasp-stats/jasp-test-release#1011


boutinb:
@FBartos Can you check if it is OK now? The problem was that the action to remove a column was done when the Variables List was unvisible. But if the whole analysis form is unvisible, then the column was also removed.

FBartos (approved):
It works and it fixes the problem.

PR #4308, opened 2020-09-28
Font from Preferences should be used in Dynamic Module console, Welcome Page and About Box

boutinb:
Fixes jasp-stats/jasp-test-release#1009

### DIFF
--- a/JASP-Desktop/components/JASP/Widgets/AboutWindow.qml
+++ b/JASP-Desktop/components/JASP/Widgets/AboutWindow.qml
@@ -4,6 +4,7 @@ import QtQuick.Controls 2.5
 import QtQuick.Controls.Styles 1.4
 import QtWebEngine 1.7
 import JASP.Widgets 1.0
+import JASP.Controls 1.0 as JC
 
 
 Window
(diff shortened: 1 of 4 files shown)


JorisGoosen:
Welcome page looks great!

Aboutbox wasn't made for scaling though:
image

JorisGoosen:
Could you also fix the weird "Modul installieren" button:
image

Like the size of the buttons changes all the time depending on zooming in/out...
Zooming out, then in again I get: (after switch to english):
image

Zoom out
image

Zoom in 
image

JorisGoosen:
Aah and the module installer now looks great, except for the fact that when you zoom in a lot the text doesn't fit anymore in the popup and there is no scrollbar.

I guess it would be nicer if it didn't always just take to the middle segment of the screen but has a minimum size and adds a scrollbar but that might very well be out of scope for this PR?

Normal zoom looks awesome:
image

Bit hard to read all:
image

Not that bad actually, just leave that I guess we can do it later?

boutinb:
The Module Install button problem was due to a binding loop, which is now solved.
The Module Installer console can be in the next release reshaped (when it will be really used).

PR #4311, opened 2020-09-29
Work around problems with windows+codepages+R... Sigh

JorisGoosen:
Try to convert certain paths on windows to short version first (if available)
- Using this for R_HOME was hypothesized to mitigate some problems with spaces or weird characters in username
-- https://github.com/jasp-stats/INTERNAL-jasp/issues/1033
-- https://github.com/jasp-stats/INTERNAL-jasp/issues/778
-- But R seems to get the fullpath anyway later, while it does allow for the library to actually load things like "base"-pkg the engine still crashes afterwards. So marginal improvement at best
-- Im guessing it does not hurt though so Im leaving it in. The shortpath function doesn't do anything if the short path does not exist (and it doesnt need to)

Adding settings for LC_CTYPE="C" on windows
- you can either leave it at default, where it will check if the directory is the same in local codepage as in utf-8, then it sets LC_CTYPE to C.
-- In that case gettext will give nice output
-- Otherwise it will choose local codepage, which makes sure people using a korean system and having korean characters in their username can actually install JASP in their userfolder, gettext stuff will be messed up and certain code might even break. But better than not being able to
start JASP I suppose.
- Or you can specify whether it should be *on* manually
- Added command line arguments for JASP to start in LC_CTYPE C or system mode
-- JASP --setLC_CTYPE_C forces "C" while --setLC_CTYPE_system forces system-default

This is a replacement for the attempt at https://github.com/jasp-stats/jasp-desktop/pull/4303

I have tested this on a korean system and it solves the problem.
It does not however fix the problem where non-local-chars are in your username, aka my local user "ẞoogaloo" cannot install JASP to appdata because "ẞ" is not dutch.


JorisGoosen:
Aah and this PR also makes sure you can now build JASP into a folder with spaces in it!
For all those times you want to do that...

JASP-Desktop/engine/enginesync.h
```diff
@@ -66,6 +66,7 @@ public slots:
 	void		cleanUpAfterClose();
 	void		filterDone(int requestID);
 	std::string	currentState() const;
+	void		haveYouTriedTurningItOffAndOnAgain() { stopEngines(); restartEngines(); } // https://www.youtube.com/watch?v=DPqdyoTpyEs
```
boutinb, line 69: I did not know this serie. Zo herkenbaar!

JASP-Desktop/gui/preferencesmodel.cpp
```diff
@@ -458,3 +461,78 @@ void PreferencesModel::resetRememberedModules(bool setToRemember)
 {
 	setModulesRemembered(!setToRemember ? QStringList({}) : RibbonModel::singleton()->getModulesEnabled());
 }
+
+winLcCtypeSetting _getWinLcCtypeSetting()
```
boutinb: This should a method this class or of the Settings class

JASP-Desktop/gui/preferencesmodel.cpp
```diff
@@ -458,3 +461,78 @@ void PreferencesModel::resetRememberedModules(bool setToRemember)
...
+	winLcCtypeSetting	current = _getWinLcCtypeSetting(),
+						newSet	= newVal ? winLcCtypeSetting::alwaysC : winLcCtypeSetting::neverC; 
+	
+	
+	if(current != newSet)
+	{
+		Settings::setValue(Settings::LC_CTYPE_C_WIN, winLcCtypeSettingToQString(newSet));
+		
+		emit lcCtypeToCChanged();
+		emit lcCtypeCheckChanged();
```
boutinb: LcCtypeCheck is not depending on LcCtypeToC

JASP-Desktop/components/JASP/Widgets/FileMenu/PrefsAdvanced.qml
```diff
@@ -206,10 +206,44 @@ ScrollView
...
+				KeyNavigation.tab:	alwaysSetLC_CTYPE_C
+				KeyNavigation.down:	alwaysSetLC_CTYPE_C
+				
+			}
+			
+			CheckBox
+			{
+				id:					alwaysSetLC_CTYPE_C
+				label:				qsTr("Always set LC_CTYPE to \"C\".")
+				toolTip:			qsTr("If this is enabled and you have non-ascii characters in your install path JASP won't work anymore.  If you only have non-ascii characters in your username then installing modules will probably break. On the other hand, disabling this will make certain characters in the results look weird, sorry for the inconvenience, we are working on it and hopefully have this fixed next release.")
```
boutinb: It is unreadable in a tooltip due to its timeout. Make it shorter, and add it to the help file.

boutinb (commented):
It is quite a change that must be carefully tested.
I have tested it on my PC, but I cannot use another local account (with strange characters) remotly. I think @FBartos  can maybe test it better.
What I could see, is in Advanced Preferences, if I click too quickly the 2 LC_CTYPE checkboxes, then the JASP Engines crashes.

JASP-Desktop/gui/preferencesmodel.cpp
JorisGoosen: Yeah I didnt want to include "enginedefinitions.h" in the header, I could do a forward declare if you really want that?

JASP-Desktop/gui/preferencesmodel.cpp
JorisGoosen: Yeah it is if it is `true` it gets set to false because the underlying enum gets a different value.
And a little extra update is ok I think. Smaller chance of things going wrong because an old value gets stuck somewhere.

JASP-Desktop/components/JASP/Widgets/FileMenu/PrefsAdvanced.qml
JorisGoosen: Good idea

JorisGoosen:
> It is quite a change that must be carefully tested.
> I have tested it on my PC, but I cannot use another local account (with strange characters) remotly. I think @FBartos can maybe test it better.

I made a simple korean and a german Windows VM, it is useful to have different locale installs of windows anyway.

> What I could see, is in Advanced Preferences, if I click too quickly the 2 LC_CTYPE checkboxes, then the JASP Engines crashes.

Yeah, but these options should not really be used anyway. And they will go out in the next version once we fix gettext. So, you know, let's ignore that?

JorisGoosen:
Can't you add a french locale to your own system?

TimKDJ:
Don't we still want the R changes from https://github.com/jasp-stats/jasp-desktop/pull/4303? Otherwise the non-ASCII strings won't come out okay in English if people have non-ASCII characters in their path.

JorisGoosen:
good point

JASP-Desktop/gui/preferencesmodel.cpp
boutinb: Yes I think it's better

boutinb:
> > It is quite a change that must be carefully tested.
> > I have tested it on my PC, but I cannot use another local account (with strange characters) remotly. I think @FBartos can maybe test it better.
> 
> I made a simple korean and a german Windows VM, it is useful to have different locale installs of windows anyway.
> 
> > What I could see, is in Advanced Preferences, if I click too quickly the 2 LC_CTYPE checkboxes, then the JASP Engines crashes.
> 
> Yeah, but these options should not really be used anyway. And they will go out in the next version once we fix gettext. So, you know, let's ignore that?

That would be OK for me, but it can be really useful to understand why restarting the engines may lead to a crash? Here we have a nice crash test to understand why it happens. Or do you know already the reason?

JorisGoosen:
> That would be OK for me, but it can be really useful to understand why restarting the engines may lead to a crash? Here we have a nice crash test to understand why it happens. Or do you know already the reason?

No I had a quick look, but didn't delve too deeply. But there could be many reasons why it crashes, but I really don't think this PR is the place to go check out all of those...

JASP-Desktop/engine/enginesync.h
JorisGoosen: Haha, yeah, it is pretty funny :p
I think it is on Netflix nowadays!

JorisGoosen:
Ok, I rebased and implemented the two requested changes.

boutinb:
I can nearly not change the LCTYPE settings without making JASP crash...
I saw that the restartAllEngines is called 2 times when you change the 'Always set LC_CTYPE to "C"' (each check changes calls the change slot of the other check). I tried to remove that. But it still crashes. What helps is to add a 1 second sleep between the stopEngines and restartEngines. It still does not help completely: if I click the checkboxes too quickly. it still crashes because (I suppose) a second restartAllEngines is called too quickly after the first restartAllEngines. If I add also  a 1 second sleep after the restartEngines, then it does not crash anymore.
I know this checkboxes should normally not be used, but if they have to be used, it would be a pity if they are actually not useable (on a normal slow PC at least).
So I think a sleep after the stopEngines and a sleep after the restartEngines are needed. And also a mechanism so that clicking on one checkbox does not result in 2 calls to restartAllEngines.

JorisGoosen:
> on a normal slow PC at least

Haha, ok I will try to fix it a bit.

JorisGoosen:
Ok, I did not go for a sleep (I think you like them too much ;) ) but this should also fix it.

I am a bit confused right now though, if you load the jasp file inside of [TestGettextMathot.zip](https://github.com/jasp-stats/jasp-desktop/files/5327601/TestGettextMathot.zip) and refresh I get:

image

Then I turn LC_CTYPE back to C and refresh to get:
image

So it seems the functioning of this has flipped?
Maybe I messed up the pot-file or something?

JASP-Engine/JASP/po/R-JASP.pot
```diff
@@ -6,7 +6,7 @@ msgstr ""
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
 "MIME-Version: 1.0\n"
-"Content-Type: text/plain; charset=CHARSET\n"
+"Content-Type: text/plain; charset=UTF-8\n"
```
JorisGoosen: Could this have caused that?
@TimKDJ ?

JorisGoosen:
Ok, I don't have a clue why now suddenly Mathôt is working when LC_CTYPE != C...
If that is the case this entire PR is entirely pointless...

I don't understand.

Did something change in the language settings? Is it because I generated the languagefiles on a windows system? 
Is it encoding it somehow in my local locale?

JorisGoosen:
Could it be I need to first generate new languagefiles on Mac? That is the difference now I guess? @FransMeerhoff  what do you think?

FransMeerhoff:
@JorisGoosen  It could be that you have to generate the language files again. But I do not understand how this is related to the path problem.

JorisGoosen:
> 
> 
> @JorisGoosen It could be that you have to generate the language files again. But I do not understand how this is related to the path problem.

I'll go generate them on my mac.

JASP-Engine/JASP/po/R-JASP.pot
TimKDJ: No, the .pot file declarations don't do much

TimKDJ:
> > @JorisGoosen It could be that you have to generate the language files again. But I do not understand how this is related to the path problem.
> 
> I'll go generate them on my mac.

Be sure to check the compilation output to ensure the translations were actually installed. When I built your branch from the other PR I had to fix quite a few errors that prevented compilation of the .mo files (e.g., `%s` being replaced with `% s` screws it up).

JorisGoosen:
> Be sure to check the compilation output to ensure the translations were actually installed. When I built your branch from the other PR I had to fix quite a few errors that prevented compilation of the .mo files (e.g., `%s` being replaced with `% s` screws it up).

Aah, those errors really block it?
Damn... I'll be sure to fix them then.

boutinb:
At least, the crash issue is solved.

JorisGoosen:
God damn it.
Previously `gettext` would only show up when LC_CTYPE="C" and now it is the opposite. 
I'll go check if I missed some errors during the language generation stuff but I am going crazy, I'll tell you that.

JorisGoosen:
Ok thank the divine but after generating the language files on a mac the LC_CTYPE works again as "expected".

JorisGoosen:
> 
> 
> @JorisGoosen It could be that you have to generate the language files again. But I do not understand how this is related to the path problem.

@FransMeerhoff if you read the text I wrote at the top of this PR, maybe that explains it a bit?

JorisGoosen:
So I've solved the merge problems for this one, but to see if I can work around the "null" problem in citations Im now in the process of rebasing the branch for https://github.com/jasp-stats/jasp-desktop/pull/4303 on LC_CTYPE_Fixes so that that one can be merged (possibly).

JorisGoosen:
The nightly called `LC_CTYPE_Unicode_Fixes` can be used to test this PR + #4303

JorisGoosen:
This PR now includes https://github.com/jasp-stats/jasp-desktop/pull/4303

JorisGoosen:
@FBartos tested it for me and of course Mathôt looks good because of the unicode fixes, but that is not a bad thing.

Pretty sure this PR is good now, so maybe it can be squashmerged?

TimKDJ:
I'll test it tomorrow and have one more look through the changes

JASP-Desktop/JASP-Desktop.pro
```diff
@@ -8,7 +8,7 @@ message("AM_I_BUILDBOT: '$$[AM_I_BUILDBOT]'")
 COPY_BUILDBOTNESS = $$[AM_I_BUILDBOT] # We need to copy it to make sure the equals function below actually works...
 !equals(COPY_BUILDBOTNESS, "") {
 !equals(COPY_BUILDBOTNESS, "\"\"") { #this should be done less stupidly but I do not want to waste my time on that now
-     GENERATE_LANGUAGE_FILES = true
+	GENERATE_LANGUAGE_FILES = true
```
TimKDJ, line 11: Do we want this on by default?

JASP-Desktop/components/JASP/Widgets/FileMenu/PrefsAdvanced.qml
```diff
@@ -206,10 +206,45 @@ ScrollView
...
+
+			}
+		}
+		
+		PrefsGroupRect
+		{
+			id:			windowsSpecific
+			visible:	WINDOWS
+			enabled:	WINDOWS
+			title:		qsTr("Windows Workarounds")
```
TimKDJ: `qsTr("Windows workarounds")` -- we seem to not make the second one a capital

JASP-Desktop/components/JASP/Widgets/FileMenu/PrefsAdvanced.qml
```diff
@@ -206,10 +206,45 @@ ScrollView
...
+			
+			CheckBox
```
TimKDJ: I thought the checkbox behaviour was a bit strange.
One checkbox appears when you click the other, but they cannot both be on at the same time.
This sounds like a radiobutton job to me:
```
LC_CTYPE:
(x) Auto
( ) Set to "C"
( ) Disabled
```

TimKDJ: It also crashes JASP amazingly easy. I tried to interact with it in a regular fashion, but it is super sensitive to crashing. Should there be a little delay before JASP automatically does something with this value?

JASP-Engine/JASP/R/LSbinomialestimation.R
```diff
@@ -956,8 +956,8 @@ LSbinomialestimation   <- function(jaspResults, dataset, options, state = NULL){
                                                 options[["plotsIterativeBF"]])
           }
           
-          if (all(is.na(tempCIPP[1:2]))){
-            tempInt <- "∅"
+          if (all(is.na(temp_CIPP[1:2]))){
+            temp_int <- "\u2205"
```
TimKDJ: The `else` clause mentions `tempInt` -- I don't think we want this variable rename

JASP-Engine/JASP/R/correlation.R
```diff
@@ -401,7 +401,7 @@ Correlation <- function(jaspResults, dataset, options){
       }
       
       result$vsmpr <- JASP:::.VovkSellkeMPR(result$p.value)
-      result$vsmpr <- ifelse(result$vsmpr == "∞", Inf, result$vsmpr)
+      result$vsmpr <- ifelse(result$vsmpr == "\u221E", Inf, result$vsmpr)
```
TimKDJ:
```
> "∞" == "\u221E"
[1] FALSE
```
This would not be desired behaviour, but the code is actually absolete so it can be deleted:  https://github.com/jasp-stats/jasp-desktop/pull/4303#discussion_r494856062

JASP-Engine/JASP/R/correlation.R
```diff
@@ -431,7 +431,7 @@ Correlation <- function(jaspResults, dataset, options){
         }
       }
       result$vsmpr <- JASP:::.VovkSellkeMPR(result$p.value)
-      result$vsmpr <- ifelse(result$vsmpr == "∞", Inf, result$vsmpr)
+      result$vsmpr <- ifelse(result$vsmpr == "\u221E", Inf, result$vsmpr)
```
TimKDJ: This line can be deleted too

TimKDJ (changes requested):
So I tested it with a local user called `Héòäëè ïìî` and the fix seems to work well. If I start a normal nightly all engines crash on startup, but with this fix it actually runs. It is clear JASP chose correctly and did not set the `LC_CTYPE`, because all German non-ASCII fails.

I couldn't place a comment next to the binaries, but this looks a little iffy to me:
`BIN -2.42 KB (99%) JASP-Engine/JASP/inst/po/de/LC_MESSAGES/R-JASP.mo`
`BIN -4.14 KB (98%) JASP-Engine/JASP/inst/po/nl/LC_MESSAGES/R-JASP.mo`

Also: is the option hidden on MacOS (and if it isn't, we don't override our value of `UTF-8`, right)?

JASP-Desktop/JASP-Desktop.pro
boutinb: The Build Bot have to compile the po files

JorisGoosen: This is only on when it is buildbot, the normal define is 4 lines up.

JorisGoosen:
> So I tested it with a local user called `Héòäëè ïìî` and the fix seems to work well. If I start a normal nightly all engines crash on startup, but with this fix it actually runs. It is clear JASP chose correctly and did not set the `LC_CTYPE`, because all German non-ASCII fails.

Good!

> I couldn't place a comment next to the binaries, but this looks a little iffy to me:
> `BIN -2.42 KB (99%) JASP-Engine/JASP/inst/po/de/LC_MESSAGES/R-JASP.mo`
> `BIN -4.14 KB (98%) JASP-Engine/JASP/inst/po/nl/LC_MESSAGES/R-JASP.mo`

I think someone who knows better how all of this translation stuff works should regenerate that stuff again after this is merged, I just removed a bunch of commented lines because they were blocking the build and maybe that is why this got smaller?

> Also: is the option hidden on MacOS (and if it isn't, we don't override our value of `UTF-8`, right)

The option is hidden everywhere but on Windows and even if it wasn't it isn't being used anywhere outside windows specific define blocks.

JASP-Engine/JASP/R/LSbinomialestimation.R
JorisGoosen: Huh, maybe someone changed the name in between my first attempt at unicode fix and now, I'll fix it.

JASP-Desktop/components/JASP/Widgets/FileMenu/PrefsAdvanced.qml
JorisGoosen: Yeah maybe it is better to make it a radiobutton actually, I just didn't want people to get confused.
Also, I really don't care if it crashes JASP (also, it didn't do that for me, but I guess I have a supercomputer or something?)

Like, people really shouldn't be changing this anyway.

boutinb (approved):

JASP-Desktop/components/JASP/Widgets/FileMenu/PrefsAdvanced.qml
JorisGoosen: Okidoki, I just have a tendency to Capitalize a lot ;P

JASP-Desktop/components/JASP/Widgets/FileMenu/PrefsAdvanced.qml
JorisGoosen: And no more timers...

JorisGoosen:
So this is better right?

JASP-Engine/JASP/R/LSbinomialestimation.R
```diff
@@ -956,8 +956,8 @@ LSbinomialestimation   <- function(jaspResults, dataset, options, state = NULL){
                                                 options[["plotsIterativeBF"]])
           }
           
-          if (all(is.na(tempCIPP[1:2]))){
-            tempInt <- "∅"
+          if (all(is.na(temp_CIPP[1:2]))){
```
TimKDJ: Ah, here's another one, should be `tempCIPP`

JorisGoosen: Maybe I should just copy the LSbinomialestimation from stable?

JorisGoosen:
So yeah, a lot of stuff was changed in LSbinomialtesting. I do remember it being annoying during rebasing, so I guess I messed that up originally.

I think it should now be good. I basically copied it from stable and then replaced ∅ by \u2205 and left the rest as is.

TimKDJ (approved):

JorisGoosen:
Finally!

PR #4312, opened 2020-09-30
RoBMA fixes

FBartos:
explicitly call S3 methods from RoBMA - it seems like that summary / plot etc... functions exported from R packages can't be called before any other function from the package was already used during the JASP session - this was causing some very odd crashes

also,
fixes: https://github.com/jasp-stats/jasp-test-release/issues/972
fixes: https://github.com/jasp-stats/jasp-test-release/issues/1022
fixes: https://github.com/jasp-stats/jasp-test-release/issues/966


JorisGoosen:
I can test this out later (today) when my mac is updated.

vandenman:
> explicitly call S3 methods from RoBMA - it seems like that summary / plot etc... functions exported from R packages can't be called before any other function from the package was already used during the JASP session

You could also try `loadNamespace("RoBMA")`. I don't dislike calling the functions with `RoBMA::` though, it avoids ambiguity. Aside from that, the code looks fine to me and all errors seem fixed.

vandenman (approved):

JorisGoosen:
Well, Im not going to check all of this code, but the problem in https://github.com/jasp-stats/jasp-test-release/issues/966 does not occur anymore!

PR #4316, opened 2020-09-30
Component List should not remember items that were deleted

boutinb:
Fixes jasp-stats/jasp-test-release#993

### DIFF
--- a/JASP-Desktop/widgets/listmodeltermsassigned.cpp
+++ b/JASP-Desktop/widgets/listmodeltermsassigned.cpp
@@ -127,6 +127,24 @@ const Terms &ListModelTermsAssigned::terms(const QString &what) const
 		return ListModelAssignedInterface::terms(what);
 }
 
+void ListModelTermsAssigned::removeTerm(int index)
+{
+	if (index < 0 || index >= int(_terms.size())) return
+	_tempTermsToRemove.clear();
+
+	beginResetModel();
+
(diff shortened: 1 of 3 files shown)


JASP-Desktop/widgets/listmodeltermsassigned.cpp
```diff
@@ -127,6 +127,24 @@ const Terms &ListModelTermsAssigned::terms(const QString &what) const
...
+
+	beginResetModel();
+
+	const Term& term = _terms.at(size_t(index));
+
+	_rowControlsMap.remove(term.asQString());
+	_terms.remove(term);
+	_tempTermsToRemove.add(term);
+
+	endResetModel();
```
JorisGoosen, line 143: If I test your branch on linux JASP crashes right here when I remove the hypothesis.

JorisGoosen: Somewhere in qt code with setParentItem...

JorisGoosen: There are some other problems Im having with JASP on linux lately, so Ill just go try this on windows.

JorisGoosen: ok on windows this works, I assume I  just broke my linux install for now.

PR #4318, opened 2020-09-30
Remove successes and failures when changing selected variable in BinomialTesting

boutinb:
Fixes jasp-stats/jasp-test-release#988
Remove also the indentation in the levels, Susccesses & Failures Variables box


FBartos (approved):
I tested it and it fixes the issue

PR #4320, opened 2020-10-01
Exploratory factor analysis: improved error message when there are fewer observation than variables

### LINKED ISSUE #961: Exploratory Factor crashes
* JASP version: 0.13.1

* OS name and version: OSX Catalina 10.15.5
* Analysis: 
Exploratory Factor Analysis
* Bug description:
I read a csv file with likert scale items (1-7). It reads it as ordinal data (I had to change it manually to scale variables).
Running Factor Analysis results with a certain number of variables to the message below:

_**The following problem(s) occurred while running the analysis:
Data not valid: too many missing values in variable(s) (1.t)houtoven-bezit, (1.t)houtoven-gemaakt, (1.t)houtoven-consumentenrapporten, (1.t)houtoven-vergelijken, (1.t)houtoven-verschillende, (1.t)houtoven-voorkeur, (1.e)houtoven-vrienden en familie, (1.e)houtoven-noodzakelijk, (1.e)houtoven-uitziet, (2.e)merkkleding-uitziet, (2.e)merkkleding-samenstelling, (3.t)sieraad-bezit, (3.t)sieraad-gemaakt, (1.e)houtoven-samenstelling, (2.t)merkkleding-bezit, (2.t)merkkleding-gemaakt, (2.t)merkkleding-consumentenrapporten, (2.t)merkkleding-vergelijken, (2.t)merkkleding-verschillende .**_

However, there are NO missing values in any of these. This error occurs when adding more than a certain number of variables for the factor analysis. It doesn't matter which variables are added, when adding random one by one, at a certain moment this error occurs. 

I will add the dataset (only the first set of variables are set to scale type.

[Archive.zip](https://github.com/jasp-stats/jasp-issues/files/5174710/Archive.zip)

vandenman:
Fixes https://github.com/jasp-stats/jasp-issues/issues/961

On the left is the old error message and on the right is the new one:
image

I added the new error check inside the other one because otherwise, a user would see both error messages. The second error message about missing values is also very confusing and inaccurate when n < p. In the linked issue, for example, there aren't any missing variables yet the analysis complains about them!

Also, the analysis used to start with:
```r
ExploratoryFactorAnalysis <- function(jaspResults, dataset, options, ...) {
  jaspResults[["optionslist"]] <- createJaspHtml(paste(capture.output(str(options)), collapse = "\n"),
                                                 class = "jasp-code", position = 7, title = "Options")
}

ExploratoryFactorAnalysis <- function(jaspResults, dataset, options, ...) {
	...
}
```
defining the same function twice?!

I deleted the first one since the second one always overwrites the first one (I couldn't figure out if the first one had any purpose).

### DIFF
--- a/JASP-Engine/JASP/R/exploratoryfactoranalysis.R
+++ b/JASP-Engine/JASP/R/exploratoryfactoranalysis.R
@@ -16,11 +16,7 @@
 #
 
 ExploratoryFactorAnalysis <- function(jaspResults, dataset, options, ...) {
-  jaspResults[["optionslist"]] <- createJaspHtml(paste(capture.output(str(options)), collapse = "\n"),
-                                                 class = "jasp-code", position = 7, title = "Options")
-}
 
-ExploratoryFactorAnalysis <- function(jaspResults, dataset, options, ...) {
   jaspResults$addCitation("Revelle, W. (2018) psych: Procedures for Personality and Psychological Research, Northwestern University, Evanston, Illinois, USA, https://CRAN.R-project.org/package=psych Version = 1.8.12.")
 
   # Read dataset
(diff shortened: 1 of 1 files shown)


Kucharssim (approved):
Looks good! 

@vankesteren is probably to blame for the double definition of the function, I guess he used it to debug something and forgot to remove it 😄  see https://github.com/jasp-stats/jasp-desktop/commit/eec75e8a1480d6d2a9badab9de1897f3361fc916

PR #4326, opened 2020-10-02
RoBMA fixes

FBartos:
fixes: https://github.com/jasp-stats/jasp-test-release/issues/1021
fixes: https://github.com/jasp-stats/jasp-test-release/issues/1024 (in conjecture with the RoBMA package update, requires the development [branch](https://github.com/FBartos/RoBMA/tree/development))

- adds a note to the model preview tables informing that it takes a while to fit the models - feel  free to suggest better wording:
`The analysis will estimate multiple meta-analytic models using MCMC and might require a prolonged time to complete.`
- the new package version also improves the warning messages - groups the same messages into one, instead of repeating the same one multiple times

### DIFF
--- a/JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
+++ b/JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
@@ -135,6 +135,7 @@ RobustBayesianMetaAnalysis <- function(jaspResults, dataset, options, state = NU
   "advanced_omit_rhat",
   "advanced_omit",
   "advanced_autofit_error",
+  "advanced_autofit_rhat",
   "advanced_autofit_time_unit",
   "advanced_autofit_time",
   "advanced_autofit",
@@ -952,6 +953,7 @@ RobustBayesianMetaAnalysis <- function(jaspResults, dataset, options, state = NU
         control = list(
           autofit         = options[["advanced_autofit"]],
           max_error       = if (options[["advanced_autofit"]]) options[["advanced_autofit_error"]],
+          max_rhat        = if (options[["advanced_autofit"]]) options[["advanced_autofit_rhat"]],
           max_time        = if (options[["advanced_autofit"]]) paste0(options[["advanced_autofit_time"]], options[["advanced_autofit_time_unit"]]),
           adapt           = options[["advanced_adapt"]],
           bridge_max_iter = options[["advanced_bridge_iter"]],
--- a/JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
+++ b/JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
@@ -833,22 +833,23 @@ RobustBayesianMetaAnalysis <- function(jaspResults, dataset, options, state = NU
   
   
   ### create overview table
-  overall_summary <-
-    createJaspTable(title = gettext("Model Summary"))
+  overall_summary <- createJaspTable(title = gettext("Model Summary"))
   overall_summary$position <- 1
   
   overall_summary$addColumnInfo(name = "terms",     title = "",                type = "string")
   overall_summary$addColumnInfo(name = "models",    title = gettext("Models"), type = "string")
   overall_summary$addColumnInfo(name = "priorProb", title = gettext("P(M)"),   type = "number")
-  
-  for (i in 1:nrow(fitSummary[["overview"]])) {
-    temp_row <- list(
-      terms     = if (i == 1) gettext("Effect") else if (i == 2) gettext("Heterogeneity") else if (i == 3) gettext("Publication bias"),
-      models    = paste0(fitSummary[["overview"]][["Models"]][i], "/", fitSummary[["add_info"]][["n_models"]]),
-      priorProb = fitSummary[["overview"]][["Prior prob."]][i]
-    )
-    
-    overall_summary$addRows(temp_row)
+
+  if (!options[["measures"]] == "fitted") {
+    for (i in 1:nrow(fitSummary[["overview"]])) {
+      temp_row <- list(
+        terms     = if (i == 1) gettext("Effect") else if (i == 2) gettext("Heterogeneity") else if (i == 3) gettext("Publication bias"),
+        models    = paste0(fitSummary[["overview"]][["Models"]][i], "/", fitSummary[["add_info"]][["n_models"]]),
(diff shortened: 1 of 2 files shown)


JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
```diff
@@ -833,22 +834,24 @@ RobustBayesianMetaAnalysis <- function(jaspResults, dataset, options, state = NU
...
-  for (i in 1:nrow(fitSummary[["overview"]])) {
-    temp_row <- list(
-      terms     = if (i == 1) gettext("Effect") else if (i == 2) gettext("Heterogeneity") else if (i == 3) gettext("Publication bias"),
-      models    = paste0(fitSummary[["overview"]][["Models"]][i], "/", fitSummary[["add_info"]][["n_models"]]),
-      priorProb = fitSummary[["overview"]][["Prior prob."]][i]
-    )
-    
-    overall_summary$addRows(temp_row)
+
+  if (!options[["measures"]] == "fitted") {
```
vandenman:
```suggestion
  if (options[["measures"]] != "fitted") {
```

JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
```diff
@@ -865,17 +868,20 @@ RobustBayesianMetaAnalysis <- function(jaspResults, dataset, options, state = NU
...
-      number       = as.numeric(rownames(fitSummary[["models"]]))[i],
-      prior_mu     = fitSummary[["models"]][i, "Prior mu"],
-      prior_tau    = fitSummary[["models"]][i, "Prior tau"],
-      prior_omega  = fitSummary[["models"]][i, "Prior omega"],
-      priorProb    = fitSummary[["models"]][i, "Prior prob."]
-    )
-    
-    models_summary$addRows(temp_row)
+
+  if (!options[["measures"]] == "fitted") {
```
vandenman:
```suggestion
  if (options[["measures"]] != "fitted") {
```

Resources/Meta Analysis/qml/RobustBayesianMetaAnalysis.qml
```diff
@@ -1979,9 +1984,9 @@ Form
 
 			IntegerField
 			{
-				label:  		qsTr("Bridge sampling iterations:")
-				name:     		"advanced_bridge_iter"
-				defaultValue: 	10000
+				label:			qsTr("Bridge sampling iterations:")
```
vandenman, line 1987: this is the maximum number of bridge sampling iterations right? the algorithm terminates early if the r-value converges.

Resources/Meta Analysis/qml/RobustBayesianMetaAnalysis.qml
```diff
@@ -64,38 +64,43 @@ Form
...
 		{
-			label: qsTr("Cohen's d / t-statistics & N / SE")
-			value: "cohensd"
-			id: 	measures_cohensd
-			checked:true
+			label:		qsTr("Cohen's d / t-statistics & N / SE")
+			value:		"cohensd"
+			id:			measures_cohensd
+			enabled:	mainWindow.dataAvailable
+			checked:	mainWindow.dataAvailable
```
vandenman, line 71: this doesn't work for me, but the same line (`mainWindow.dataAvailable`) does work in the Distributions module to disable sections. @boutinb any ideas?

JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
```diff
@@ -833,22 +834,24 @@ RobustBayesianMetaAnalysis <- function(jaspResults, dataset, options, state = NU
...
-    temp_row <- list(
-      terms     = if (i == 1) gettext("Effect") else if (i == 2) gettext("Heterogeneity") else if (i == 3) gettext("Publication bias"),
-      models    = paste0(fitSummary[["overview"]][["Models"]][i], "/", fitSummary[["add_info"]][["n_models"]]),
-      priorProb = fitSummary[["overview"]][["Prior prob."]][i]
-    )
-    
-    overall_summary$addRows(temp_row)
+
+  if (!options[["measures"]] == "fitted") {
+    for (i in 1:nrow(fitSummary[["overview"]])) {
```
vandenman, line 845: If this is always `1:3`, you could also write that. If not please ignore this comment.

vandenman (changes requested):
So I'm still waiting for https://github.com/jasp-stats/jasp-test-release/issues/1024 to finish, but overall the changes look fine. Some small comments below. It's weird that `mainWindow.dataAvailable` doesn't work because the code looks correct to me.

vandenman:
another issue is that `full precision may not have been achieved
in 'pnt{final}'` is printed all over the place when running jasp-stats/jasp-test-release#1024 
image

it's spamming the entire log in qt. Can we suppress this somehow? If it's generated by e.g., `dt` then `suppressWarnings` should do the trick.

JASP-Engine/JASP/R/robustbayesianmetaanalysis.R
FBartos: for now, but it not might be in the future - I'm planning on adding another possible factor crossed with these, so I would rather keep it without plugging in numbers

Resources/Meta Analysis/qml/RobustBayesianMetaAnalysis.qml
FBartos: interesting, it works for me on windows. Which part is problematic? The enabling or changing the default selection?

Resources/Meta Analysis/qml/RobustBayesianMetaAnalysis.qml
FBartos: yes, that's true, the label is getting a bit too long already and then it starts compressing the second column, not really sure how to address that

FBartos:
We have been already over these messages (https://github.com/jasp-stats/jasp-desktop/pull/4139#issuecomment-646159183). The  `callr::r` is even implemented in the package as a silent option, however, for whatever reason it crashes on MacOS and we had to disable it.

vandenman (approved):
Ah, I forgot about that. But then it's not fixed, right? So we better make an issue for it so we don't forget. Either way, this PR can be merged IMO.

PR #4327, opened 2020-10-02
Bayesian Correlation: change DoubleField to FormulaField for the prior

vandenman:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/1085

The first commit contains all actual changes.
The second commit changes the QML file so it consistently uses tabs instead of mixing tabs and spaces.

adding `options$kappa <- 1.0` to the test file is necessary because jasptools doesn't parse the value for a FormulaField.


AlexanderLyNL:
Overall this looks good, the only thing I wonder is whether formula input can be used to bypass bounds set in the ui, but that's something we can test later. I'll wait until it passes travis.

PR #4328, opened 2020-10-03
[Do not Merge in 0.14] Interval plots

milenasophie:
I was not quite sure whether jaspResults[["IntervalPlots"]]$position <- 11 is right here. Also there were a variety of options to plot the means with the confidence intervals so I don't know whether plotmeans(len ~ x, data = d, frame = FALSE, mean.labels = TRUE, connect = FALSE) is the best option.

### DIFF
--- a/JASP-Engine/JASP/R/descriptives.R
+++ b/JASP-Engine/JASP/R/descriptives.R
@@ -117,6 +117,24 @@ Descriptives <- function(jaspResults, dataset, options) {
       }
     }
   }
+  # Interval plots
+
+  if (options$descriptivesIntervalPlot) {
+    if(is.null(jaspResults[["IntervalPlots"]])) {
+      jaspResults[["IntervalPlots"]] <- createJaspContainer(gettext("Intervalplots"))
+      jaspResults[["IntervalPlots"]]$dependOn(c("IntervalPlots", "splitby"))
+      jaspResults[["IntervalPlots"]]$position <- 11
+    }
+
+    IntervalPlots <- jaspResults[["IntervalPlots"]]
+
+    for (var in variables) {
+        if(is.null(IntervalPlots[[var]]) && .descriptivesIsNumericColumn(dataset.factors, var)) {
+          IntervalPlots[[var]] <- .descriptivesIntervalPlot(dataset = dataset, options = options, variable = var)
+          IntervalPlots[[var]]$dependOn(optionContainsValue=list(variables=var))
+        }
+    }
+  }
 
   # QQ plots
   if (options$descriptivesQQPlot) {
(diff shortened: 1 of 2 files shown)


JASP-Engine/JASP/R/descriptives.R
```diff
@@ -975,6 +993,68 @@ Descriptives <- function(jaspResults, dataset, options) {
...
+    list(ggplot2::geom_segment(data=d, ggplot2::aes(x=x, y=y, xend=xend, yend=yend), size = 0.75, inherit.aes=FALSE), ggplot2::scale_y_continuous(breaks=b))
+  }
+
+  thePlot <- createJaspPlot(title=variable, width=options$plotWidth, height=options$plotHeight)
+
+  errorMessage <- .descriptivesCheckPlotErrors(dataset, variable, obsAmount = "< 1")
+  if (!is.null(errorMessage)) {
+    thePlot$setError(gettextf("Plotting not possible: %s", errorMessage))
+  }
+  else {
```
vandenman, line 1011:
```suggestion
  } else {
```

JASP-Engine/JASP/R/descriptives.R
```diff
@@ -975,6 +993,68 @@ Descriptives <- function(jaspResults, dataset, options) {
...
+    } else {
+      group     <- as.factor(dataset[[.v(options$splitby)]])[!is.na(dataset[[.v(variable)]])]
+      xlab      <- options$splitby
+
+    }
+
+    plotDat <- data.frame(group = group, y = y)
+    row.names(plotDat) <- yIndexToActual
+
+    p <- plotmeans(len ~ x, data = d, frame = FALSE,
```
vandenman, line 1042: So there's 3 issues here:

1. this function comes from `gplots` right? If so it should be explicitly imported like this:
```suggestion
    p <- gplots::plotmeans(len ~ x, data = d, frame = FALSE,
```
2. the data `d` is not defined here and neither are `len` and `x`. Should this be:
```
plotmeans(y ~ group, data = plotDat, ...)
```
If yes, that still gave me an error :)

3. `gplots::plotmeans` uses base R plot functions and afterward you mix the result with ggplot2 functions which won't work. 

I suggest replicating the output from `plotmeans` using ggplot2 yourself.

JASP-Engine/JASP/R/descriptives.R
```diff
@@ -975,6 +993,68 @@ Descriptives <- function(jaspResults, dataset, options) {
   return(thePlot)
 }
 
+.descriptivesIntervalPlot <- function(dataset, options,  variable) {
```
vandenman, line 996:
```suggestion
.descriptivesIntervalPlot <- function(dataset, options, variable) {
```

JASP-Engine/JASP/R/descriptives.R
```diff
@@ -975,6 +993,68 @@ Descriptives <- function(jaspResults, dataset, options) {
...
+      xlab      <- "Total"
+
+    } else {
+      group     <- as.factor(dataset[[.v(options$splitby)]])[!is.na(dataset[[.v(variable)]])]
+      xlab      <- options$splitby
+
+    }
+
+    plotDat <- data.frame(group = group, y = y)
+    row.names(plotDat) <- yIndexToActual
```
vandenman, line 1040:
```suggestion
    rownames(plotDat) <- yIndexToActual
```
I know the result is the same but we use `rownames` almost everywhere in the codebase.

vandenman (changes requested):
Most of the code looks great and the style nicely fits with JASP.

A couple of general comments:

- It's considered an antipattern to make a pull request from `stable`. See step 4 here: https://www.attosol.com/create-and-merge-branches-using-github-desktop-client/ for how to create a branch. The main reason is that if you'd want to work on multiple issues simultaneously, it becomes unclear what branch is the main branch. For example, if you now create a new branch from the branch `stable`, it also includes the changes you made in this PR (which is probably not what you want).
- If you use a function from a library you always need to call it with `::`. You did this in some places but not everywhere. 
- If you edit your first comment and add "Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/1057" as the first line, then GitHub will automatically close the issue in the SKF project.
- Did you build JASP and run the code? Because if you do so you'll probably encounter this error:
image

Let me know if anything is unclear!

PR #4333, opened 2020-10-06
Return error message instead of crash in multinomial analysis

### LINKED ISSUE #996: Multinomial test issue
* JASP version: 0.13.1

* OS name and version: Windows 10 2004
* Analysis: 
* Bug description: Multinomial test issue
* Expected behaviour: table but something happen

.

Kucharssim:
Fixes https://github.com/jasp-stats/jasp-issues/issues/996 (turns out that the bug was caused by something much less sinister than reading the factor levels 🙂 )


vandenman (approved):
Looks good to me!

PR #4330, opened 2020-10-05
Tweak ribbonBar fadeout

JorisGoosen:
- This way you can see the upcoming or past ribbonbuttons under the menu-arrow-buttons
- Also show a "drag hand cursor" when above the items
- Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/1086

PR #4334, opened 2020-10-06
Firefox style ribbon

JorisGoosen:
So the code didn't get simplified but the UI seems clearest now
- I still like the glass look more than this, but this way at least it is glaringly obvious there are more buttons left or right
- Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/1086 differently from https://github.com/jasp-stats/jasp-desktop/pull/4330

image

image

image

image

I'll make some nightlies as well, they will be called `firefoxRibbon`

### DIFF
--- a/JASP-Desktop/components/JASP/Widgets/Ribbon/MenuArrowButton.qml
+++ b/JASP-Desktop/components/JASP/Widgets/Ribbon/MenuArrowButton.qml
@@ -28,11 +28,25 @@ Rectangle
 	implicitHeight				: jaspTheme.ribbonButtonHeight * 0.6
 	implicitWidth				: implicitHeight
 	// radius					: 5
-	color						: mice.pressed ? jaspTheme.grayLighter : jaspTheme.uiBackground
-
-	property bool	hamburger:	true
-	property bool	showArrow:	false
+	color						: pressed ? jaspTheme.grayLighter : jaspTheme.uiBackground
+	
+	enum ButtonType
+	{
+		Hamburger,
+		Plus,
+		LeftArrow,
+		RightArrow
+	}
+	
+	property int buttonType:	MenuArrowButton.ButtonType.Hamburger 
 	property string	toolTip:	""
+	
+			 property double iconScale:	0.7
+			 property bool showPressed: false
+	readonly property bool pressed:		mice.pressed || showPressed
+	readonly property bool hamburger:	buttonType == MenuArrowButton.ButtonType.Hamburger || buttonType == MenuArrowButton.ButtonType.LeftArrow
+	readonly property bool showArrow:	buttonType == MenuArrowButton.ButtonType.LeftArrow || buttonType == MenuArrowButton.ButtonType.RightArrow 
+	
 
 	ToolTip.text:				toolTip
 	ToolTip.timeout:			jaspTheme.toolTipTimeout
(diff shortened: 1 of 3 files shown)


JorisGoosen:
Mind you, I will change the color of the tooltip separately.

TimKDJ:
I think this is a great improvement in terms of behaviour and very clear for the user.
Three comments which I think would make it awesome:

1. I liked the way the old menu stayed the same color and had the opacity fade on the side. I think that in combination with the arrows would be the golden ticket for me.

So the love child of these two:
Screenshot 2020-10-06 at 16 28 52
Screenshot 2020-10-06 at 16 29 13

2. When you press the arrow once you move 7 items over (which is quite a lot) and you cannot long press it. What about a single press is move one item over and long press keeps moving it until the user lets go.

3. When you press on the hamburger or the plus, they should stay in the pressed state. So they should remain a little smaller when activated (i.e., when you press the hamburger and then hover over it, there shouldn't be any size animation -- the same as it is for the analysis buttons).

EJWagenmakers:
Great improvements, and I agree with Tim's suggestions

JorisGoosen:
>     1. I liked the way the old menu stayed the same color and had the opacity fade on the side. I think that in combination with the arrows would be the golden ticket for me.

Yeah I liked that too, but I changed the color because it looks like the button go "under" the arrow (like in firefox I might add)

> So the love child of these two:
> Screenshot 2020-10-06 at 16 28 52
> Screenshot 2020-10-06 at 16 29 13
> 
>     1. When you press the arrow once you move 7 items over (which is quite a lot) and you cannot long press it. What about a single press is move one item over and long press keeps moving it until the user lets go.

I'll see what I can do. 


>     2. When you press on the hamburger or the plus, they should stay in the pressed state. So they should remain a little smaller when activated (i.e., when you press the hamburger and then hover over it, there shouldn't be any size animation -- the same as it is for the analysis buttons).

Aah yeah that is a good one, missed something there.

JorisGoosen:
Ok, I mostly processed your comments, I will not change the shadow color to the ui-background color though, because it really doesn't look that good imho. But making it much more narrow should help mitigate the merry-go-round effect while making sure it still looks like there is more hidden underneath.

JorisGoosen:
Aah, and I turned on some builds for y'all.

TimKDJ:
For anyone wondering, this is what it looks like now:
Screenshot 2020-10-07 at 11 04 32

Again a great improvement over the previous iteration. The scrolling also works exactly like how I would expect it to work.

When you select an analysis you get vertical bars on either side of the button. Might be good to do this also on the arrows, plus and hamburger to be consistent? (In the screenshot: new ribbon with the plus selected (top), old ribbon with frequencies selected (bottom))
Screenshot 2020-10-07 at 11 02 27

I do have to get used a little to the hamburger and plus being so far in the corner. But at this point I've looked at the JASP ribbon thousands of times so a large part of it is familiarity. Considering the ribbon is a very prominent part of the JASP application, what do others think?

EJWagenmakers:
"Might be good to do this also on the arrows, plus and hamburger to be consistent?" -> yes
Looks good!
E.J.

JorisGoosen:
Hehe, ok I'll make those buttons also show an outline on press.

JorisGoosen:
Ok, looks like this now:
image

In this one the modulemenu is opened and Im pressing the rightarrow:
image

image

image

boutinb:
The hand cursor is not there anymore. I found it quite a good idea.

TimKDJ:
> In this one the modulemenu is opened and Im pressing the rightarrow:

Yes, nice, makes it a bit clearer.
I like it better in dark mode actually. Maybe because the shadow isn't so visible?

> The hand cursor is not there anymore. I found it quite a good idea.

But not the grabby hand one please, the pointy one; grabbing the bar will not be the main purpose when people interact with it (clicking analyses is).

JorisGoosen:
I'll just turn the grabbing off entirely, people can then either scroll vertical/horizontal (both work) or use the fancy new arrows.

And then the pointer can just be pointy on top of the buttons etc

JorisGoosen:
> Yes, nice, makes it a bit clearer.
> I like it better in dark mode actually. Maybe because the shadow isn't so visible?

Well, the dark mode just looks better alround ^^

EJWagenmakers:
What about the bright orange bar on top of the ribbon? Is this new? Grey is easier on the eyes, and that's what I get in my version (see screenshot).
image

JorisGoosen:
Yeah I just have my Windows set up in a bright orange theme, combined with dark backgrounds, it isn't part of JASP ;)

boutinb (approved):

PR #4336, opened 2020-10-07
Fix typo, "postive" -> "positive".

JoKeyser:
Noticed a typo during translation; hope this fixes the correct files (assuming the `.pot`, and `.po` files are derived, so the change will propagate to those).


TimKDJ:
Can you add this to the module @boutinb?

PR #4338, opened 2020-10-07
[Do not Merge in 0.15] Remove the special case for Windows

JorisGoosen:
- Disables native textrendering there
- Fixes https://github.com/jasp-stats/jasp-test-release/issues/1034
- Fixes https://github.com/jasp-stats/jasp-test-release/issues/1028

### DIFF
--- a/JASP-Desktop/main.cpp
+++ b/JASP-Desktop/main.cpp
@@ -295,10 +295,6 @@ int main(int argc, char *argv[])
 
 			QLocale::setDefault(QLocale(QLocale::English)); // make decimal points == .
 
-#ifdef _WIN32
-			QQuickWindow::setTextRenderType(QQuickWindow::NativeTextRendering); //Doesn't improve it on anything 'cept windows
-#endif
-
 			//Now we convert all these strings in args back to an int and a char * array.
 			//But to keep things easy, we are going to copy the old argv to avoid duplication (or messing up the executable name)
 			char** argvs = new char*[args.size()];


boutinb:
I have tried on my work PC, and the difference in minimal, cannot say which version is beter than the other one.
So if that solves also the Cutoff problem, I would say let's do it.

boutinb (approved):

TimKDJ:
The results page seems ok this time around, but the interface font has the same degradation as https://github.com/jasp-stats/jasp-test-release/issues/535. And this screen isn't even thàt bad (27" 1440p). I don't think we should do this (also based on https://github.com/jasp-stats/jasp-test-release/issues/1034#issuecomment-704842599), I think Windows + poor screen will constitute a sizeable portion of our user base.

JorisGoosen:
`27" 1440p` is really quite low DPI (108dpi).

What I could do is turn native rendering on if the dpi is below 120 and off when above. That should cover our bases.
Because DPI's will only be rising and it really does look much better on hi-dpi screens.

As in, on a pretty common laptop screen size of 1920x1080 at 17 inch we already get 130dpi and it looks great.

JorisGoosen:
No need to do this in 0.14 btw.

JorisGoosen:
And my comment that you linked on, that is on the old Dell screen from TOP that is connected to the labPC and it really is a horrible screen. As in, everything looks bad on it.

TimKDJ:
> `27" 1440p` is really quite low DPI (108dpi).

Not really for external monitors. With a quick search on Tweakers' pricewatch I find that 86% of all the external monitors they list are <= 109 dpi ([this](https://tweakers.net/monitoren/vergelijken/#filter:RY3LCoMwEEX_ZdZBEjWa5gOELrpyWboIcVoGfJFkUSr5944V7Go4957LbLCEAUNHOA5gIa7o6UneJVpmEEfZLyFx56I_E9au7KtKKQGre2FPHwSrJVMgjzeawTJEHNEnHI5BBHtXutFC16Z6iN-3jsaEgZsNVF3K_U77mKmQ_G9yb4ZKM2QB-_CvaE7bsmnN6SlpiosxjdaQc_4C) divided by [this](https://tweakers.net/monitoren/vergelijken/#filter:q1bKL0pJLQouSE32TFGy0jXUUSpITE8NzqxKVbIyNQDyijKTU30z85SsgJzi1JzU5JLUFIjyYiWraENTM1MdUxML41igLFDULTOnJLUIKFOtZGhiZACic0GagTw9AyUdpdzECiDH2BTIqa2tBQA)).

> What I could do is turn native rendering on if the dpi is below 120 and off when above. That should cover our bases.
> Because DPI's will only be rising and it really does look much better on hi-dpi screens.

This sounds like something to look into! But we'll also have to factor in scaling (I'm not sure if this will make any difference?), because I doubt our older generation will look at higher dpi without any scaling.

> No need to do this in 0.14 btw.

Sounds good, let's put a pin in this one and see if we can find the sweet spot.

JorisGoosen:
> > `27" 1440p` is really quite low DPI (108dpi).
> 
> Not really for external monitors. With a quick search on Tweakers' pricewatch I find that 86% of all the external monitors they list are <= 109 dpi ([this](https://tweakers.net/monitoren/vergelijken/#filter:RY3LCoMwEEX_ZdZBEjWa5gOELrpyWboIcVoGfJFkUSr5944V7Go4957LbLCEAUNHOA5gIa7o6UneJVpmEEfZLyFx56I_E9au7KtKKQGre2FPHwSrJVMgjzeawTJEHNEnHI5BBHtXutFC16Z6iN-3jsaEgZsNVF3K_U77mKmQ_G9yb4ZKM2QB-_CvaE7bsmnN6SlpiosxjdaQc_4C) divided by [this](https://tweakers.net/monitoren/vergelijken/#filter:q1bKL0pJLQouSE32TFGy0jXUUSpITE8NzqxKVbIyNQDyijKTU30z85SsgJzi1JzU5JLUFIjyYiWraENTM1MdUxML41igLFDULTOnJLUIKFOtZGhiZACic0GagTw9AyUdpdzECiDH2BTIqa2tBQA)).

Well, it is still low, but it doesn't matter anyway if we have JASP be intelligent about it.

> > What I could do is turn native rendering on if the dpi is below 120 and off when above. That should cover our bases.
> > Because DPI's will only be rising and it really does look much better on hi-dpi screens.
> 
> This sounds like something to look into! But we'll also have to factor in scaling (I'm not sure if this will make any difference?), because I doubt our older generation will look at higher dpi without any scaling.

No, the degradation you see is caused by the fact that the qml rendering is specifically made to be animated, scaled etc, and this causes the "fuzziness". But scaling the interface won't make the fuzziness any worse, in fact, the fuzziness is antialiasing and is there to make it look great. Which it does! On a good screen ^^
Mac for instance does this for all their computers also the small ones, but the tend to have hidpi so you don't notice any fuzziness.
The Windows rendering on the other hand uses "hinting" to make it look good on lowdpi screens.

See [this informative wiki article](https://en.wikipedia.org/wiki/Font_hinting)

> > No need to do this in 0.14 btw.
> 
> Sounds good, let's put a pin in this one and see if we can find the sweet spot.

I'll make an issue for it.

JorisGoosen:
https://github.com/jasp-stats/INTERNAL-jasp/issues/1100

Pretty sure this will be a small tweak, but I'll look at after I fix everything else for 0.14 and then we'll see if it could still go in or not.

JorisGoosen:
Oops

JorisGoosen:
I still need to work on this though

vandenman:
@JorisGoosen can this be closed? :stuck_out_tongue:

JorisGoosen:
Haha yes

PR #4339, opened 2020-10-07
[Do not Merge in 0.14] X and R charts 

tomishay:


JASP-Engine/JASP/R/descriptives.R
```diff
@@ -1427,4 +1455,89 @@ Descriptives <- function(jaspResults, dataset, options) {
...
+  XbarPlot <- createJaspPlot(title = "X bar chart", width=500, aspectRatio=1)
+  jaspResults[["Xbarchart"]] <- XbarPlot
+
+  if (!ready)
+    return()
+
+  .XbarFillPlotDescriptives(XbarPlot, dataset, options)
+
+  return()
+}
```
vandenman, line 1481: you can merge these functions if they're this small:
```suggestion
.Xbarchart <- function(jaspResults, dataset, options, variables){

  ready <- (length(options$variables) > 0)

  if (!ready || !is.null(jaspResults[["Xbarchart"]]))
    return()
    
  XbarPlot <- createJaspPlot(title = "X bar chart", width=500, aspectRatio=1)
  jaspResults[["Xbarchart"]] <- XbarPlot

  .XbarFillPlotDescriptives(XbarPlot, dataset, options)

  return()
}
```
that would also make it clear that you're having twice `if (ready)` in there.

JASP-Engine/JASP/R/descriptives.R
```diff
@@ -1427,4 +1455,89 @@ Descriptives <- function(jaspResults, dataset, options) {
...
+
+.XbarFillPlotDescriptives <- function(XbarPlot, dataset, options){
+
+  plot <- qcc::qcc(dataset, type="xbar", plot= F)
```
vandenman, line 1485: This is a nice package, but it doesn't return ggplot2 objects so once we have plot editing these plots cannot be modified. Can you remake the plots with `ggplot2`?

JASP-Engine/JASP/R/descriptives.R
```diff
@@ -1427,4 +1455,89 @@ Descriptives <- function(jaspResults, dataset, options) {
...
+  jaspResults[["Rchart"]] <- RchartPlot
+
+  if (!ready)
+    return()
+
+
+  .RchartFillPlotDescriptives(RchartPlot, dataset, options)
+
+  return()
+}
```
vandenman, line 1524: These functions can also be merged like the ones above.  Also, the empty whitespace and comments here are redundant. If you have a function title "xxxFillPlot" then a comment like "Filling function" doesn't add anything.

JASP-Engine/JASP/R/descriptives.R
```diff
@@ -184,6 +184,34 @@ Descriptives <- function(jaspResults, dataset, options) {
...
+
+  Xbarchart <- jaspResults[["Xbarchart"]]
+
+  for (var in variables){
+    if(is.null(Xbarchart[[var]]) && .descriptivesIsNumericColumn(dataset.factors, var)){
+        Xbarchart[[var]] <- .Xbarchart(jaspResults = jaspResults, dataset= dataset, options= options, variables= var)
+        jaspResults[["Xbarchart"]]$dependOn(c("Xbarchart", "variables", "Rchart"))
+    }
+  }
+}
```
vandenman, line 199: So there's a couple of issues here. 

1. `if(is.null(Xbarchart[[var]])` -- you also do this check within the function `.Xbarchart`, no need to do it twice.
2. you never defined `jaspResults[["Xbarchart"]]` to be a collection. As a consequence, JASP only shows a single plot even though (run the analysis with multiple variables).
3. `Xbarchart` depends on an unrelated option, "Rchart".

I'd do something like:
```r
if (is.null(jaspResults[["Xbarchart"]])) {
  jaspResults[["Xbarchart"]] <- createJaspCollection(title = "...")
  jaspResults[["Xbarchart"]]$dependOn("Xbarchart")
}

Xbarchart <- jaspResults[["Xbarchart"]]
for (var in variables){
    if(.descriptivesIsNumericColumn(dataset.factors, var)) {
        Xbarchart[[var]] <- .Xbarchart(jaspResults = jaspResults, dataset= dataset, options= options, variables= var)
        Xbarchart[[var]]$dependOn(optionContainsValue=list(variables=var))    
  }
}
```
the same goes for the Rcharts.

vandenman (changes requested):
Already looks functional, great! There are a couple of issues though, most of them I discuss below. The more important ones are:

1. Did you actually build JASP and try out the analysis? Because there are a few bugs that pop out once you run the analysis. For example, this one:
image

If there's only one variable, you get an error. I'm also not entirely sure whether the goal is to create a plot per variable (when using split) or whether the goals is to merge all variables into one figure. It looks like the output is doing the former but the code is written to do the latter.

2. The R package `qcc` does not return ggplot2 objects. Please rewrite those plot functions with `ggplot2` (otherwise they won't be compatible with plot editing).
3. When you fix an issue, please start the PR with Fixes \<link to the issue\>. That way there's a reference to the PR in the issue and it will automatically be closed when this PR is merged.

PR #4340, opened 2020-10-07
About JASP - Full Screen

FransMeerhoff:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/974

When in Full Screen and going to About in JASP , it opens quite ugly in a new full screen "page". After closing the About section, the created new full screen page still exists - but all black

### DIFF
--- a/JASP-Desktop/components/JASP/Widgets/AboutWindow.qml
+++ b/JASP-Desktop/components/JASP/Widgets/AboutWindow.qml
@@ -11,14 +11,12 @@ Window
 {
 	id:				aboutWindow
 
-	width:			850
-	height:			500
-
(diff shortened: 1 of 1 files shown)


JASP-Desktop/components/JASP/Widgets/AboutWindow.qml
```diff
@@ -33,7 +31,7 @@ Window
 	Connections
 	{
 		target:			mainWindow
-		onCloseWindows: aboutModel.visible = false;
+		onCloseWindows: {aboutModel.visible = false; MainWindow.show;}
```
boutinb: MainWindow: is it not mainWindow?

PR #4341, opened 2020-10-07
Do not show a column when no name is given in FilteredDataEntryModel

boutinb:
Fixes jasp-stats/jasp-test-release#1027

### DIFF
--- a/JASP-Desktop/widgets/listmodelfiltereddataentry.cpp
+++ b/JASP-Desktop/widgets/listmodelfiltereddataentry.cpp
@@ -28,7 +28,7 @@ ListModelFilteredDataEntry::ListModelFilteredDataEntry(BoundQMLTableView * paren
 	connect(this,				&ListModelFilteredDataEntry::filterChanged,		[&](){ _tableView->setItemProperty("filter",	_filter);	}						);
 	connect(this,				&ListModelFilteredDataEntry::colNameChanged,	[&](){ _tableView->setItemProperty("colName",	_colName);	}						);
 
-	if(_colNames.size() == 0)
+	if(_colNames.size() == 0 && !_colName.isEmpty())
 		_colNames.push_back(_colName);
 
 
@@ -154,16 +154,21 @@ void ListModelFilteredDataEntry::sourceTermsChanged(const Terms *, const Terms *
 	//std::cout << "ListModelFilteredDataEntry::sourceTermsChanged(Terms *, Terms *)" << std::endl;
 
 	Terms sourceTerms	= getSourceTerms();
-	QString colName		= _colNames[_editableColumn];
+	QString colName		= (_editableColumn >= 0 && _editableColumn < _colNames.size()) ? _colNames[_editableColumn] : "";
 	_dataColumns		= sourceTerms.asVector();
 	_colNames			= tq(_dataColumns);
 
 	if(_extraCol != "")
 		_colNames.push_back(_extraCol);
 
-	_editableColumn		= _colNames.size();	
+	if (!colName.isEmpty())
+	{
+		_editableColumn		= _colNames.size();
+		_colNames.push_back(colName);
+	}
+	else
+		_editableColumn = -1;
 
-	_colNames.push_back(colName);
 	_columnCount		= _colNames.size();
 
 	fillTable();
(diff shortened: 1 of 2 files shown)


JASP-Desktop/widgets/listmodelfiltereddataentry.cpp
```diff
@@ -370,23 +377,26 @@ void ListModelFilteredDataEntry::setColName(QString colName)
...
+	{
+		_colNames.push_back(colName);
+		_editableColumn = _colNames.size() - 1;
+	}
+	else if (colName.isEmpty())
+	{
+		if (_colNames.size() > 0) _colNames.pop_back();
+		_editableColumn = -1;
+	}
+	else if (_colNames.size() > _editableColumn && _editableColumn >= 0)
```
JorisGoosen: What happens if `_editableColumn == -1` ?
Is this even a possibility?

If it isn't this check is unnecessary and if it is, well then there needs to be an else right?

JorisGoosen: So looking at the file I don't think _editableColumn can be <0 if a colName was set before and will be set now right?

boutinb: Yes, if _colName is empty, then _editableColumn is -1. I find always a good idea to always check the index before using it in a array: in this way, even if there is a bug, it does not crash at least.
But if we want to remove the redundancy of information, I could beter remove the _editableColumn property: if _colName is empty it is always -1, and if not, it is always the last index of the _colNames array. This would remove the possibility to have an editable column which is not the last one... but ja, this TableView has to be refactored anyway.

boutinb: See https://github.com/jasp-stats/INTERNAL-jasp/issues/1098

JorisGoosen: But that means that if `_colName == ""` at the start of the function then `_editableColumn` is -1.
So if `colName != ""` then it gets set to `_colNames.size() - 1`.
If `colName == ""` then we don't even get to this if-else-if-else-if. Because then `_colName == colName` and we would've `return`ed already.

My point is, including `_editableColumn >= 0` is pointless, and when I read that code I wonder "Why are we checking for this? Is it possible?" But it isn't.

So then I would prefer either a crash, or at the very least and  `else Log::log() << "Shit is crazy WTF? _editableColumn == -1! How is it possible?" << std::endl;`

And refactoring would be glorious

boutinb:
@koenderks Can you use my branch to check it is what you want?

JASP-Engine/JASP-Engine.pro
```diff
@@ -50,37 +50,6 @@ win32:LIBS += -lole32 -loleaut32
 
 mkpath($$OUT_PWD/../R/library)
 
-exists(/app/lib/*) {
```
JorisGoosen: @boutinb ???

boutinb: Oeps...

boutinb: Removed.

JorisGoosen (approved):
After @koenderks tests it and affirms this is what he needed this PR can be merged.

koenderks:
Well, it seems like the workflows completely crash upon opening on my computer. Both of those use this widget.

JorisGoosen:
Hah wow, and how are you testing?
Rebasing your `uvb` `jasp-desktop` on Bruno's branch or loading your Audit module in Bruno's branch?

koenderks:
I have just tried to open the original audit module in bruno's branch.

koenderks:
By loading my uvb module in there, it also crashes

boutinb:
> By loading my uvb module in there, it also crashes

Did you test it with my branch?

koenderks:
Yes, I tested using `boutinb:stable`.

boutinb:
Hi @koenderks Can you test it again, just to be sure?

koenderks:
@boutinb This seems to be working perfectly now!

koenderks (approved):

PR #4343, opened 2020-10-07
Learn Bayes - fixes

FBartos:
fixes: jasp-stats/jasp-test-release#986
fixes: jasp-stats/jasp-test-release#997

+ additional fixes suggested by EJ
- typos in the explanatory text
- changing labels for prediction from "Number of successes" to "Predicted number of successes"
- making sequential analysis start with priors only for testing (starting at 0th iteration)
- adding prediction tables to Posterior predictions in Estimation and to Prior predictions in Testing


JASP-Engine/JASP/R/LSbinomialestimation.R
```diff
@@ -84,6 +84,7 @@ LSbinomialestimation   <- function(jaspResults, dataset, options, state = NULL){
   if (options[["plotsPredictions"]]){
     if (options[["predictionPlotType"]] != "individual").plotsPredictionsBinomialLS(jaspResults, data, ready, options)
     if (options[["predictionPlotType"]] == "individual").plotsPredictionsIndividualBinomialLS(jaspResults, data, ready, options)
+    if (options[["predictionPlotTable"]]).tablePosteriorPredictions(jaspResults, data, ready, options)
```
vandenman: This looks like one statement but actually, it's
```r
if (options[["predictionPlotTable"]])
  .tablePosteriorPredictions(jaspResults, data, ready, options)
```
right? could you add a space between the function call and the closing parenthesis of the if statement to make this obvious?

JASP-Engine/JASP/R/LSbinomialestimation.R
```diff
@@ -233,12 +234,14 @@ LSbinomialestimation   <- function(jaspResults, dataset, options, state = NULL){
   
   if (is.null(containerPlots[[paste0("plots",type)]])){
     
-    plotsSimple <- createJaspPlot(width = 530, height = 400, aspectRatio = 0.7)
+    plotsSimple <- createJaspPlot(
+      width  = if (options[[ifelse(type == "Prior", "plotsPriorType", "plotsPosteriorType")]] == "overlying") 700 else 530,
```
vandenman, line 256: This if else appears twice in a row and is kinda convoluted, can you factor it out?

JASP-Engine/JASP/R/LScommon.R
```diff
@@ -160,6 +160,9 @@ hdi.density    <- function(object, credMass=0.95, allowSplit=FALSE, ...) {
       tempArrow       <- allArrows[i,]
       tempArrow$yEnd <- tempArrow$yEnd * .scalingSpikes(allLines, allArrows)
       
+      if (tempArrow$yEnd == 0)
+        tempArrow$yEnd <- 1e-5
```
vandenman: if plotting goes wrong for exactly 0, leave as it is. If plotting would also go wrong for 1e-300, I'd change this to:
```suggestion
      if (tempArrow$yEnd < .Machine$double.eps)
        tempArrow$yEnd <- 1e-5
```
The same goes for the other checks for 0

vandenman (changes requested):
The code looks fine and https://github.com/jasp-stats/jasp-test-release/issues/997 is fixed. The other issue, https://github.com/jasp-stats/jasp-test-release/issues/986, is not fixed though:

- [ ] For `Binomial Testing`, I noticed overlapping text in the options:

image

- [ ] For Binomial estimation, the spike still can't be 0 or 1:

image

- [ ] The boxes for theta_0 and alpha are not nicely aligned. (EJ's comment)

image

JASP-Engine/JASP/R/LSbinomialestimation.R
FBartos: I'm not sure what do you mean by that. Can you elaborate please?

vandenman: Sure, you have:

```r
   plotsSimple <- createJaspPlot(
      width  = if (options[[ifelse(type == "Prior", "plotsPriorType", "plotsPosteriorType")]] == "overlying") 700 else 530,
      height = 400)

    plotsSimple$position <- 2
    plotsSimple$dependOn(c(
       .dataDependenciesBinomialLS,
       ifelse(options[[ifelse(type == "Prior", "plotsPriorType", "plotsPosteriorType")]] == "overlying", "colorPalette", ""))) 
```
the same `if` statement with a nested `ifelse` appears twice. You could write:
```r
if (options[[ifelse(type == "Prior", "plotsPriorType", "plotsPosteriorType")]] == "overlying") {
  width <- 700
  dependency <- "colorPalette"
} else {
  width <- 530
  dependency <- NULL
}
```
which is more lines of code but easier to read.

FBartos: hmm, the thing is that I'm using this structure thru the whole analysis to make the dispatching somewhat concise. If I were to rewrite this one occurrence, I would have to change another 700 lines and double them. Not sure whether that improves the readability overall

JASP-Engine/JASP/R/LScommon.R
FBartos: it's probably only zero, but I agree that this is safer. Changed

JASP-Engine/JASP/R/LSbinomialestimation.R
FBartos: changed

JASP-Engine/JASP/R/LSbinomialestimation.R
vandenman: Well, to be honest, I don't find nested `ifelse` (or `ifelse` within `if`) very readable. I had to read the same line of code multiple times before I properly understood it. I particularly pointed it out here because you're repeating the same code twice in a row.

If it's always the same `ifelse(type == "Prior", "plotsPriorType", "plotsPosteriorType")`, you could also write a function for it, e.g., 
```r
getPlotsType <- function(type) if (type == "Prior") "plotsPriorType" else "plotsPosteriorType"
```
that would also reduce the chances of any typos.

FBartos: the problem is that there is around 20-40 of these variations:

`ifelse(type == "Prior", "x", "z")`

some of them used only once, some more often and it doesn't make sense to create a dedicated function. I have been playing around with an idea to use a paste function that adds either Prior or Posterior to the name of the setting, but that would require more qml rewriting and possibly breaking more things

vandenman: If the structure is always like that, you could use a static table that you index by name:

```r
table <- matrix(c(
  "plotsPriorType", "plotsPosteriorType"
), ncol = 2, 
  dimnames <- list(
    c("plotsType"), 
    c("prior", "posterior")
))
table["plotsType", type] # to get information
```
This could also be a list, if the elements have different types. Then you'd get `table[["plotsType"]][type]`.

Sticking to a convention with either "prior" or "posterior" after a setting is also fine.

I agree that if there are around 20-40 of these variations, it's better to solve this for all of them. Can you create an issue for that on internal so that we don't forget about it? You can just link to the discussion here. Then I'd just leave this code as-is for now.

FBartos: ok, done

FBartos:
I pushed an update that resolves the previous issues. 

>  For Binomial Testing, I noticed overlapping text in the options:

This should be already solved by Bruno's fix in stable. I think that I didn't rebase on that versions since I'm scared of breaking everything by Git (again) but I don't have this problem. Maybe it's a Linux specific thing?

> For Binomial estimation, the spike still can't be 0 or 1: & The boxes for theta_0 and alpha are not nicely aligned. (EJ's comment)

Thanks, I must have missed that. Also fixed in the additional commit

vandenman:
>  I think that I didn't rebase on that versions since I'm scared of breaking everything by Git (again) but I don't have this problem.

I tried to rebase and immediately got merge conflicts... :P

The code looks fine and the issues seem fixed. The overlapping text has something to do with my choice of font which is a bit unfortunate.

vandenman (approved):

TimKDJ:
Could you solve the merge conflicts?

vandenman:
@TimKDJ GitHub doesn't show any merge conflicts, I only got them locally (I don't know why though).

TimKDJ:
> 
> 
> @TimKDJ GitHub doesn't show any merge conflicts, I only got them locally (I don't know why though).

They do for me. What happens when you force refresh?

mergeconflict

vandenman:
I should've tried that :p

FBartos:
interesting, I still don't see that, even if I force refresh

FBartos:
I created a new PR that gets rid off those problems: https://github.com/jasp-stats/jasp-desktop/pull/4346

PR #4353, opened 2020-10-12
Fix single column csv wrongfully choosing space as delimiter

JorisGoosen:
- Fixes https://github.com/jasp-stats/jasp-test-release/issues/1040

The original algorithm only looked at the first line and checked how many spaces, commas, tabs or semicolons there were.
This works reasonably well, only has two problems:
Often users have spaces in columnnames and they are not aware of the need to add "Quotes" to avoid ambiguities.
Normally this is ok, but when there is only 1 column there is no comma, and in that case the algorithm picked spaces
Ive changed it to take into account all lines and it decides that any "delimiter" that does not have the same amount of delimiters on each line cannot be a proper delimiter and ignores it.

### DIFF
--- a/JASP-Desktop/data/importers/csv.cpp
+++ b/JASP-Desktop/data/importers/csv.cpp
@@ -263,17 +263,17 @@ bool CSV::readUtf8()
 }
 
 
-void CSV::determineDelimiters()
+void CSV::determineDelimiters(size_t fromHere)
 {
-	bool inQuote = false;
-	bool eol = false;
+	bool	inQuote		= false,
+			eol			= false;
+	int		semicolons	= 0,
+			commas		= 0,
+			spaces		= 0,
+			tabs		= 0,
+			stopped		= 0;
 
-	int commas = 0;
-	int semicolons = 0;
-	int spaces = 0;
-	int tabs = 0;
-
-	for (int i = 0; i < _utf8BufferEndPos && eol == false; i++)
+	for (int i = fromHere; i < _utf8BufferEndPos && eol == false; i++)
 	{
 		char ch = _utf8Buffer[i];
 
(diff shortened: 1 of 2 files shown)


boutinb:
This does not handle wel missing values. If I have a file like this:
a;b
1
2;0

Then I get:
Screen Shot 2020-10-12 at 18 17 35

In place of:

Screen Shot 2020-10-12 at 18 17 43

JorisGoosen:
Well, maybe you shouldn't be giving such malformed csv?

How should the algorithm know you aren't giving it a single column file?

EDIT: I mean, if the CSV is malformed how could it be read properly anyhow?

boutinb:
Even a file like this: 
a;c
2;2
3;4
5

does not work. That was working. So I don't see this is acceptable.

JorisGoosen:
Ok, but where are these broken CSVs coming from?

They are exported right?
So why would they be missing characters?

The whole point is that it now just look at the topline and that also has problems.

Is it really necessary to look at broken lines?

JorisGoosen:
[DatasetsCSV.zip](https://github.com/jasp-stats/jasp-desktop/files/5371036/DatasetsCSV.zip)
Here are the testfiles I used.

As you can see with the spacesOrCommas: It will still use commas if possible even a bit.

JorisGoosen:
[ExpectationsMaskWearing.zip](https://github.com/jasp-stats/jasp-desktop/files/5371045/ExpectationsMaskWearing.zip)
EJs single column csv.

JorisGoosen:
Aah yeah so EJ's example is now not fixed...

JorisGoosen:
So I did the special case that @TimKDJ and @vandenman seemed to like so much because I was running into too many problems with my attempts at a glorious algorithm.

boutinb (approved):

PR #4357, opened 2020-10-13
Reliability: copied R and qml files from dynamic module

### LINKED ISSUE #999: Reliability module very slow - files huge
* JASP version: 0.13.1

* OS name and version: both Windows 10 and macOS
* Analysis: Reliability
* Bug description: Performance very slow
* Expected behaviour: I am new to JASP, and very impressed. If I can help its development in some way, I'd be happy to. I am submitting this report after using JASP Reliability with some of my favourite data sets, looking for an easier way to get McDonald omega reliability estimates than having to work through R (also thinking of my students; I am able to use R at least to some extent, but prefer not to especially with beginning postgraduate students).

Here's what I have found:

Processing results from cognitive tests / exams can go pretty well, reasonably fast EXCEPT for a large dataset having results from about 2,500 students who sat an exam with 100 multiple-choice items. I've run it through JASP using a Windows 10 laptop (fairly recent, 8M RAM, 256 GB, i5 processor), and also a MacBook Pro (2018 model).

On both laptops it took JASP over **60 minutes** to process under Windows 10, and over **90** on the MacBook Pro.

Not only does processing take time, saving the JASP file (either to local hard disk or to my OSF account) takes many, many more minutes, at least 30.

This is in contrast to other datasets. One (for example) involves 500 students responding to a 60-item test -- no problem, JASP works fast. Today I'll be running with results from 5,500 students who sat an exam with 50 items and will see how JASP performs.

I have run many datasets through the 'psych' package in R, and that goes without problem, and is of course fast. But I would like to use JASP if possible, and I imagine some of our postgraduates will too.

Regards from Western Australia, Larry Nelson and a PS: there's another GitHub topic that says you may add omega heirarchical to the Reliability module. I will be keen to test that; hope it might be coming along soon.

vandenman:
fixes jasp-stats/jasp-test-release#1035
fixes jasp-stats/jasp-test-release#1036

See https://github.com/jasp-stats/jaspReliability/pull/6, I couldn't verify that this fixes https://github.com/jasp-stats/jasp-issues/issues/999 so it's probably better to leave that issue open for now.

[diff without whitespace changes](https://github.com/jasp-stats/jasp-desktop/pull/4357/files?diff=split&w=1)
[diff in jaspReliability](https://github.com/jasp-stats/jaspReliability/pull/6/files?diff=split&w=1)

### DIFF
--- a/JASP-Engine/JASP/R/reliabilityAnalysisBayesian.R
+++ b/JASP-Engine/JASP/R/reliabilityAnalysisBayesian.R
@@ -38,11 +38,11 @@ reliabilityBayesian <- function(jaspResults, dataset, options) {
       tables = c("McDonald's \u03C9", "Cronbach's \u03B1", "Guttman's \u03BB2", "Guttman's \u03BB6",
                  "Greatest Lower Bound", "Average interitem correlation", "mean", "sd"),
       tables_item = c("McDonald's \u03C9", "Cronbach's \u03B1", "Guttman's \u03BB2", "Guttman's \u03BB6",
-                      "Greatest Lower Bound", "Item-rest correlation", "mean", "sd"),
+                      gettext("Greatest Lower Bound"), gettext("Item-rest correlation"), gettext("mean"), gettext("sd")),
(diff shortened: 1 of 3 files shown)


vandenman:
uh oh I notice some errors already

PR #4363, opened 2020-10-13
Take very low custom scale/ordinal threshold into account

JorisGoosen:
- Fixes https://github.com/jasp-stats/jasp-test-release/issues/1049


Kucharssim:
Yes, this does it (i.e., it works fine now). I think it can be useful to specify even such a low threshold, so I prefer this solution over the alternative of setting a minimum = 2.

boutinb (approved):

PR #4365, opened 2020-10-14
Variables List does not scale in the right way when zooming

boutinb:
The calculation of the size of the Variables Lists inside a
VariablesForm depends on the height of the VariablesForm. It was
depending on the istWidth, depending on the width, that was updated
apparently before the height was updated. So the signal that should be
used should be directly the heightChanged.
Fixes jasp-stats/jasp-test-release#1044


JorisGoosen:
You are fast!

PR #4219, opened 2020-08-05
[DO NOT MERGE] Custom font for plots

vandenman:
A draft for using different fonts to ensure that we support more Unicode characters. 

The code as is doesn't do anything (and thus the unit tests are happy). To make it work, add:
```r
JASPgraphs::setJASPFont("~/github/jasp-desktop/JASP-Desktop/html/font/Lato-Regular.ttf")
``` 
before creating a plot (or add it once in `runJaspResults` to make the effect global). Then all ggplots that use `themeJasp` or `themeJaspRaw` will use this font automatically (for qgraph we may need adjustments to each plot).

In addition, the following R packages are needed:

- showtext [CRAN](https://cran.r-project.org/package=showtext)
- sysfonts [CRAN](https://cran.r-project.org/package=sysfonts)

Something that may be a dealbreaker is that these packages have dependencies:

SystemRequirements: zlib, libpng, FreeType

I don't know if these dependencies are needed if we just use the binaries or if they are required for the few functions I use from these packages.

Stable                     |  This PR (Lato)
:-------------------------:|:-------------------------:
  |  

The difference is most clear for the letter g. Note that `ragg` wasn't necessary for this in the end.

Feedback/ comments are welcome! It would also be great if somebody could test this on macOS (I've only tested this on Ubuntu so far).

Edit: in `Makevars.win` inside `sysfonts` there is:

```
PKG_CPPFLAGS = -I./include -I./include/freetype
PKG_LIBS = -L.${R_ARCH} -lfreetype -lpng -lz


.PHONY: all deps clean

all: deps $(SHLIB)

deps: sysfonts-dep-win-biarch.tar.gz
	tar xzf sysfonts-dep-win-biarch.tar.gz

sysfonts-dep-win-biarch.tar.gz:
	${R_HOME}/bin$(R_ARCH_BIN)/Rscript -e "u = 'https://raw.githubusercontent.com/yixuan/sysfonts-dep/master/sysfonts-dep-win-biarch.tar.gz'; utils::download.file(u, basename(u), mode = 'wb')"

clean:
	$(RM) *.o
	$(RM) *.dll
	$(RM) *.so
	$(RM) *.dylib

```

that download looks a little bit fishy but it refers to this repo: https://github.com/yixuan/sysfonts-dep

lucky for us, that repo contains all dependencies.

PR #4388, opened 2020-11-03
Custom fonts for figures

### LINKED ISSUE #363: Ability to change the font in the results (tables and plots)
It would help a lot to have an option to display results in other font families, particularly Times New Roman. Most of us are psychologists here, and we mostly submit in APA format which requires Times New Roman. I've spent a lot of time figuring out how to get my figures in R showing up in Times New Roman (it's easy now that I know how to do it) but I like using JASP for a mini-meta I'm working on and the results figures can't really be submitted as is in their current font. 

Add a Preference to choose the font family. Ideally one preference for the UI font, and another for the results font; I like working in a sans-serif font, but I need my results to show up in TNR, especially figures.

**Describe alternatives you've considered**
The only alternative I've found so far is to make a table in Excel with the right font and then paste a cropped version of the figure onto that table. Pretty insane and not a long term solution at all. This really stresses me out :s

vandenman:
Fixes https://github.com/jasp-stats/jasp-issues/issues/363 
Continuation of https://github.com/jasp-stats/jasp-desktop/pull/4219

### Message of previous PR

A draft for using different fonts to ensure that we support more Unicode characters. 

The code as is doesn't do anything (and thus the unit tests are happy). To make it work, add:
```r
JASPgraphs::setJASPFont("~/github/jasp-desktop/JASP-Desktop/html/font/Lato-Regular.ttf")
``` 
before creating a plot (or add it once in `runJaspResults` to make the effect global). Then all ggplots that use `themeJasp` or `themeJaspRaw` will use this font automatically (for qgraph we may need adjustments to each plot).

In addition, the following R packages are needed:

- showtext [CRAN](https://cran.r-project.org/package=showtext)
- sysfonts [CRAN](https://cran.r-project.org/package=sysfonts)

Something that may be a dealbreaker is that these packages have dependencies:

SystemRequirements: zlib, libpng, FreeType

I don't know if these dependencies are needed if we just use the binaries or if they are required for the few functions I use from these packages.

Stable                     |  This PR (Lato)
:-------------------------:|:-------------------------:
  |  

The difference is most clear for the letter g. Note that `ragg` wasn't necessary for this in the end.

Feedback/ comments are welcome! It would also be great if somebody could test this on macOS (I've only tested this on Ubuntu so far).

Edit: in `Makevars.win` inside `sysfonts` there is:

```
PKG_CPPFLAGS = -I./include -I./include/freetype
PKG_LIBS = -L.${R_ARCH} -lfreetype -lpng -lz


.PHONY: all deps clean

all: deps $(SHLIB)

deps: sysfonts-dep-win-biarch.tar.gz
	tar xzf sysfonts-dep-win-biarch.tar.gz

sysfonts-dep-win-biarch.tar.gz:
	${R_HOME}/bin$(R_ARCH_BIN)/Rscript -e "u = 'https://raw.githubusercontent.com/yixuan/sysfonts-dep/master/sysfonts-dep-win-biarch.tar.gz'; utils::download.file(u, basename(u), mode = 'wb')"

clean:
	$(RM) *.o
	$(RM) *.dll
	$(RM) *.so
	$(RM) *.dylib

```

that download looks a little bit fishy but it refers to this repo: https://github.com/yixuan/sysfonts-dep

lucky for us, that repo contains all dependencies.

### To get this to work in Jasp, do
  1. Install `showtext` and `sysfonts`.
  2. Checkout this PR
  3. Checkout https://github.com/jasp-stats/jaspGraphs/pull/2 of jaspGraphs
  4. In `jaspBase/R/common.R` add something like:
```r
  # adjust this to wherever you have a cool font
  jaspGraphs::setJaspFont("~/github/jasp-desktop/Desktop/html/font/Lato-Regular.ttf")
  # other code
  oldGraphOptions <- jaspGraphs::graphOptions()
  on.exit(jaspGraphs::graphOptions(oldGraphOptions), add = TRUE)

```
inside `runJaspResults`.

### DIFF
--- a/R-Interface/jaspResults/R/writeImage.R
+++ b/R-Interface/jaspResults/R/writeImage.R
@@ -19,11 +19,21 @@ getImageLocation <- function() {
   )
 }
 
-openGrDevice <- function(...) {
+ openGrDevice <- function(..., dpi) {
   #if (jaspResultsCalledFromJasp())
   #  svglite::svglite(...)
   #else
+
+  # showtext divides by 72 internally, see ?showtext::showtext_opts
+  # 1 + 1 / 3 is a magic number to compensate font size scaling
+  showtext::showtext_opts(dpi = dpi / (1 + 1 / 3))
   grDevices::png(..., type = ifelse(Sys.info()["sysname"] == "Darwin", "quartz", "cairo"))
+  showtext::showtext_begin()
+}
+
+closeGrDevice <- function() {
+  showtext::showtext_end()
+  dev.off()
 }
 
 writeImageJaspResults <- function(width=320, height=320, plot, obj=TRUE, relativePathpng=NULL, ppi=300, backgroundColor="white", location=getImageLocation())
@@ -59,21 +69,12 @@ writeImageJaspResults <- function(width=320, height=320, plot, obj=TRUE, relativ
 
   plot2draw <- decodeplot(plot)
 
+  openGrDevice(file = relativePathpng, width = width, height = height, res = 72 * (ppi / 96), bg = backgroundColor, dpi = ppi)
+  on.exit(closeGrDevice())
+
   if (ggplot2::is.ggplot(plot2draw) || inherits(plot2draw, c("gtable"))) {
 
-    # TODO: ggsave adds very little when we use a function as device...
-    ggplot2::ggsave(
-      filename  = relativePathpng, 
-      plot      = plot2draw,
-      device    = grDevices::png,
-      dpi       = ppi,
-      width     = width,
-      height    = height,
-      bg        = backgroundColor,
-      res       = 72 * (ppi / 96),
-      type      = ifelse(Sys.info()["sysname"] == "Darwin", "quartz", "cairo"),
-      limitsize = FALSE # only necessary if users make the plot ginormous.
-    )
+   grid::grid.draw(plot2draw) # inherited from ggplot2::ggsave
 
     #If we have jaspGraphs available we can get the plotEditingOptions for this plot
     if(requireNamespace("jaspGraphs", quietly = TRUE))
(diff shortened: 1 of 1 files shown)


TimKDJ:
I haven't looked at any code or tested it yet, but a quick thought:
Can we make this work with the font changer that @boutinb implemented?

Screenshot 2020-11-03 at 14 15 17

It would be nice if the plots played ball here

vandenman:
@TimKDJ yes, that should work. This PR only implements the backend part to make it work for plots, there's nothing in place (yet) that allows users to specify what font they want.

JorisGoosen:
> @TimKDJ yes, that should work. This PR only implements the backend part to make it work for plots, there's nothing in place (yet) that allows users to specify what font they want.

Pretty sure we should just use the same font as the rest of the results right?

I do agree that the download link there is very fishy.
And in fact it will break entirely on flatpak...
(because flathub, for obvious reasons, does not want to have unchecked downloads occurring during building so it blocks all network connections)

JorisGoosen:
I think we should do something about that sysfonts thing though, I don't think it is a good idea to integrate such a shaky compile into our buildprocess.

vandenman:
Okay, so what should we do about sysfonts? I looked a little bit into the code and frankly, it's a ton of c code that we'd have to move to jaspGraphs (not just from [sysfonts](https://github.com/yixuan/sysfonts/tree/master/src) but also from [showtext](https://github.com/yixuan/showtext/tree/master/src)). 

@JorisGoosen I'm not very good with makefiles and all that, but shouldn't it be possible to rewrite the makefile of sysfonts so that it doesn't download the dependencies if a user (we) specifies the paths to those dependencies?

JorisGoosen:
Well, actually, if I look at the makefile it might already be set up properly for that.
If we manually download `sysfonts-dep-win-biarch.tar.gz` and put it in the source-directory/tarball and then install from that it probably doesnt download it again.

Should be einfach to test.

PR #4390, opened 2020-11-03
Bayesian wilcoxon signed rank - flip the samples and use testValue

JohnnyDoorn:
Fix https://github.com/jasp-stats/jasp-issues/issues/1038
and use testvalue

### DIFF
--- a/JASP-Engine/JASP/R/ttestbayesianonesample.R
+++ b/JASP-Engine/JASP/R/ttestbayesianonesample.R
@@ -121,7 +121,7 @@ TTestBayesianOneSample <- function(jaspResults, dataset, options, state = NULL) 
           
           .setSeedJASP(options)
           r <- try(.signRankGibbsSampler(
-            x = x, nSamples = options[["wilcoxonSamplesNumber"]], testValue = 0, nBurnin = 0,
+            x = x, nSamples = options[["wilcoxonSamplesNumber"]], testValue = options[["testValue"]], nBurnin = 0,
             cauchyPriorParameter = options[["priorWidth"]]
           ))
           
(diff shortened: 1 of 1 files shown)


AlexanderLyNL:
All plots seem to fail, @vandenman how problematic is this? Is this related to the jaspGraphs upgrade?

JASP-Engine/JASP/R/ttestbayesianonesample.R
```diff
@@ -121,7 +121,7 @@ TTestBayesianOneSample <- function(jaspResults, dataset, options, state = NULL)
           
           .setSeedJASP(options)
           r <- try(.signRankGibbsSampler(
```
AlexanderLyNL, line 123: The name `r` is not so informative, perhaps use a more descriptive name

JASP-Engine/JASP/R/ttestbayesianonesample.R
```diff
@@ -134,11 +134,11 @@ TTestBayesianOneSample <- function(jaspResults, dataset, options, state = NULL)
 
           } else {
             
-            ttestResults[["delta"]][[var]]  <- r[["deltaSamples"]] * -1
+            ttestResults[["delta"]][[var]]  <- r[["deltaSamples"]]
             ttestResults[["rHat"]][[var]]  <- r[["rHat"]]
             
             bf.raw <- .computeBayesFactorWilcoxon(
```
AlexanderLyNL, line 140: bfRaw would be better

AlexanderLyNL (approved):
This seems fine. I added some style comments that should be looked into next year.

vandenman:
> All plots seem to fail, @vandenman how problematic is this? Is this related to the jaspGraphs upgrade?

Yes, I kinda didn't consider that we want the unit tests to keep working here. Either way, if @JohnnyDoorn can also add a PR to jasp-stats/jaspTTest then we can see if the unit tests pass there.

JohnnyDoorn:
@vandenman the changes are already in TTest, because I think I had accidentally checked out the upstream branch and pushed, so it was added without a PR - gonna make sure it doesn't happen in the future.

vandenman (approved):
changes are the same in jaspTTests and there the unit tests pass

PR #4395, opened 2020-11-04
Grand update

JorisGoosen:
JASP now supports a variable amount of engines
- This will also allow users to set a max amount of engines and just be more intelligent about it in general https://github.com/jasp-stats/INTERNAL-jasp/issues/568
- It starts with a single engine.
- When a module request, analysis or rscript cannot find an idle engine and there are less workers than allowed engines a new engine is started every so many secs (10 now, in a define)
- When an engine has been idle for 10m it shuts down.

Besides I removed a whole bunch of stuff that we don't need anymore.
- Init Initing etc is not used so removed
- description.json is not used so removed
- usesJaspResults in qml and other places has been removed

Several other fixes
- Analyses loaded from a jaspfile now start running if they werent Complete or FatalError on saving.
- Memory is now cleaned up a bit after an analysis through some functions I added to jaspBase
- Make sure a development module of a jaspModule also replaces the loaded module


JorisGoosen:
I've noticed it needs a little tweak. I will be pushing this somewhere today.

JorisGoosen:
And now that tweak has been applied.

boutinb:
I think this is OK. I have found some crashes on Windows, but apparently it has nothing to do with this PR.

PR #4397, opened 2020-11-09
fixing marginal custom CI in predictive distributions

FBartos:
the else statement was supposed to be tied to the first if instead of the second - automatically deleting the passed data - fixed by adding curly brackets

### DIFF
--- a/JASP-Engine/JASP/R/LSbinomialcommon.R
+++ b/JASP-Engine/JASP/R/LSbinomialcommon.R
@@ -594,20 +594,18 @@
 }
 .marginalCustomBinomialLS   <- function(density, spikes, lCI, uCI, densityDiscrete = FALSE){
   
-  if (!is.null(density))
-    if (!densityDiscrete)density$y <- density$y/nrow(density)    
-    else
-      density <- data.frame("y" = NULL, "x" = NULL)
-    
-    if (length(spikes) != 0){
-      for(i in 1:length(spikes)){
-        density <- rbind(density[density$x <= spikes[[i]]$x,], spikes[[i]], density[spikes[[i]]$x < density$x,])
-      }
-    }
-    
-    coverage <- sum(density$y[density$x >= lCI & density$x <= uCI])
-    
-    return(cbind.data.frame(xStart = lCI, xEnd = uCI, g = "custom", coverage = coverage, parameter = "theta"))
+  if (!is.null(density) && !densityDiscrete)
+      density$y <- density$y/nrow(density)    
+  if (is.null(density))
+    density <- data.frame("y" = NULL, "x" = NULL)
+  
+  for(i in seq_along(spikes)) {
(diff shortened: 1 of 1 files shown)


JASP-Engine/JASP/R/LSbinomialcommon.R
```diff
@@ -594,20 +594,21 @@
...
+  if (!is.null(density)){
+    if (!densityDiscrete)
+      density$y <- density$y/nrow(density)    
+  }else
+    density <- data.frame("y" = NULL, "x" = NULL)
+  
+  if (length(spikes) != 0){
+    for(i in 1:length(spikes)){
+      density <- rbind(density[density$x <= spikes[[i]]$x,], spikes[[i]], density[spikes[[i]]$x < density$x,])
     }
```
Kucharssim:
```suggestion
  for(i in seq_along(spikes)) {
    density <- rbind(density[density$x <= spikes[[i]]$x,], spikes[[i]], density[spikes[[i]]$x < density$x,])
  }
```

JASP-Engine/JASP/R/LSbinomialcommon.R
```diff
@@ -594,20 +594,21 @@
...
-      density <- data.frame("y" = NULL, "x" = NULL)
-    
-    if (length(spikes) != 0){
-      for(i in 1:length(spikes)){
-        density <- rbind(density[density$x <= spikes[[i]]$x,], spikes[[i]], density[spikes[[i]]$x < density$x,])
-      }
+  if (!is.null(density)){
+    if (!densityDiscrete)
+      density$y <- density$y/nrow(density)    
+  }else
```
Kucharssim: This is still a bit iffy, I believe you it works now but can you simplify the logic a little to remove these nested if statemenets? I won't beat you on this but it's definitely not pretty...

JASP-Engine/JASP/R/LSbinomialcommon.R
Kucharssim: if `length(spikes) == 0`, `seq_along(spikes)` gives you `numeric(0)` and the loop is skipped, no need to nest it like this

Kucharssim (changes requested):

JASP-Engine/JASP/R/LSbinomialcommon.R
FBartos: true, thanks (y)

JASP-Engine/JASP/R/LSbinomialcommon.R
FBartos: yeah, I think that it's a remnant from a time when the densityDiscrete was a part the density object itself. I will change it.

Kucharssim (approved):
Assuming the tests are ok, it's fine to merge

FBartos:
I merged the commits together, but I can't build/test the release version of JASP, I have some problems with the build...

TimKDJ:
Does this close an issue?

FBartos:
it does not, I fixed it directly without creating an issue

PR #4401, opened 2020-11-12
Fixing option encoding

JorisGoosen:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/1163 (ANOVA extra encoding broken)
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/1162 (JAGS gets confused about columnnames)

### DIFF
--- a/Engine/engine.cpp
+++ b/Engine/engine.cpp
@@ -519,14 +519,10 @@ void Engine::_encodeColumnNamesinOptions(Json::Value & options, Json::Value & me
 				_encodeColumnNamesinOptions(options[i], meta[i]);
 		
 		else if(isRCode)
-		{
 			for(size_t i=0; i<options.size(); i++)
 				if(options[i].isString())
 					options[i] = ColumnEncoder::columnEncoder()->encodeRScript(options[i].asString());
-		}
-		else if(encodePlease)
-			ColumnEncoder::columnEncoder()->encodeJson(options, false);
-		
+	
 		return;
 
 	case Json::objectValue:
--- a/Engine/engine.cpp
+++ b/Engine/engine.cpp
@@ -501,31 +501,47 @@ void Engine::_encodeColumnNamesinOptions(Json::Value & options, Json::Value & me
 	if(meta.isNull())
 		return;
 	
-	bool encodePlease = meta.isObject() && meta.isMember("containsColumn") && meta["containsColumn"].asBool();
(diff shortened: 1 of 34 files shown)


boutinb:
I have tested the file in https://github.com/jasp-stats/INTERNAL-jasp/issues/1163 (RepMeasuresANOVA_SPACEY). 
It seems to work in 0.14.
In development I get this error:
Screenshot 2020-11-13 at 14 26 51

With your branch I get:
Screenshot 2020-11-13 at 14 27 04

I'm not sure if this is better!

Furthermore, the encoding is set in BoundQMLListViewTerms which is used by many kinds of controls. For example the ComponentsList used in all MixedModels analyses. And if I test a MixedModels analysis  I get the same error:
`Trying to decode columnName but 'Kangaroo Testicle' is not an encoded columnName!'`

Maybe it's better to wait for the Control refactor project is finished, and then to handle this problem.

JorisGoosen:
Did you do a proper clean and all that?

JorisGoosen:
And please lets merge this first before the whole refactor...
JAGS is fixed for sure btw.

I can check later whether I fixed it properly or not.

But all those `Trying to decode columnName but 'Kangaroo Testicle' is not an encoded columnName!'` errors are what I was trying to fix...

I also don't get what you mean by "tested in `development`", you should test my branch right?

Engine/engine.cpp
```diff
@@ -501,31 +505,48 @@ void Engine::_encodeColumnNamesinOptions(Json::Value & options, Json::Value & me
...
 				_encodeColumnNamesinOptions(options[i], meta[i]);
-		else if(meta.type() == Json::objectValue && meta.get("containsColumn", false).asBool())
+		
+		else if(isRCode)
+		{
+			for(size_t i=0; i<options.size(); i++)
+				if(options[i].isString())
+					options[i] = ColumnEncoder::columnEncoder()->encodeRScript(options[i].asString());
+		}
+		else if(encodePlease)
```
boutinb: This extra `else if(encodePlease)` is quite redundant.

boutinb (changes requested):
Just a small remark.
It seems to work indeed.

Engine/engine.cpp
JorisGoosen: Indeed indeed, ive fixed it now. Not sure how I missed that before but ah well :p

PR #4402, opened 2020-11-12
Remove Required Modules and Packages from Description.qml

JorisGoosen:
- Should finish off https://github.com/jasp-stats/INTERNAL-jasp/issues/1152
- Now the required modules for another module are added to the libPaths
- We will however still have problems installing modules because remotes-pkg loves to ddos github...


vandenman (approved):
I can't really judge the code but the first two points are indeed fixed.

boutinb:
It seems to be OK. Only thing: this PR seems to change some submodules (Jags, TTests).

JorisGoosen:
>  Only thing: this PR seems to change some submodules (Jags, TTests).

I know, but if you just do `git submodule update --remote` it doesnt matter anyway.

PR #4404, opened 2020-11-19
Fix the drawing of base plots after the decoding step

TimKDJ:
The base plots of abtest errored out because `writeImage()` tried to use `plot()` on an object of class `"gTree" "grob"  "gDesc"`

### DIFF
--- a/R-Interface/jaspResults/R/writeImage.R
+++ b/R-Interface/jaspResults/R/writeImage.R
@@ -59,7 +59,7 @@ writeImageJaspResults <- function(width=320, height=320, plot, obj=TRUE, relativ
 
   plot2draw <- decodeplot(plot)
 
-  if (ggplot2::is.ggplot(plot2draw) || inherits(plot2draw, c("gtable"))) {
+  if (ggplot2::is.ggplot(plot2draw) || inherits(plot2draw, c("gtable", "gTree"))) {
 
     # TODO: ggsave adds very little when we use a function as device...
     ggplot2::ggsave(


TimKDJ:
We should probably sort out the duplicated code of the two writeImage's (they're both still used), but that's for another time.

vandenman:
Ideally, we just use the `ggsave` if statement because that already uses `grid.draw` behind the scenes. The `plot` part is really just a fallback. Did you try that?

vandenman:
for example:

```r
  if (ggplot2::is.ggplot(plot2draw) || inherits(plot2draw, c("gtable", "gTree"))) {
```
you can also add the definition of `ggplot2::is.ggplot` to make a single call to `inherits` if you prefer.

vandenman:
Also

> The base plots of abtest errored out because writeImage() tried to use plot() on an object of class "gTree" "grob" "gDesc"

if it's a base plot, then plot should just work. Is their plot not some kind of grob?

TimKDJ:
I'll check if `ggsave` works, because that'd indeed be better.

> if it's a base plot, then plot should just work. Is their plot not some kind of grob?

It's no longer a regular base plot function after the decode step: https://github.com/jasp-stats/jasp-desktop/blob/a9ae27b35b5fdba9f02b431cfdc215e661bf129a/R-Interface/jaspResults/R/writeImage.R#L191

TimKDJ:
It works, so `ggsave` it is :)

vandenman:
I completely forgot that we did that for de/encoding :p this might mean that all the recorded plot stuff can also be deleted, but that's for later

vandenman (approved):

PR #4409, opened 2020-11-25
Change String label to Int label when column is changed with integers

### LINKED ISSUE #1077: Old values not discarded after sync with spreadsheet editor
* JASP version: 0.14
* OS name and version: macOS Big Sur 11.0.1
* Analysis: N/A – data editor
* Bug description: values from spreadsheet are not discarded after sync with spreadsheet editor
* Expected behaviour: after sync, only the up-to-date values should be shown and old values discarded

1. Open test.csv in JASP 0.14 and in a spreadsheet editor. Notice that **group** is a nominal variable.
2. Change the values in the first column from words into numbers.
3. Synchronize the spreadsheet with JASP (or have JASP autosync). JASP should now consider **group** an ordinal or scale variable.
4. After the sync JASP still displays the old values; seemingly, it did not sync.
5. But in fact, JASP did sync the values. The new values become visible when manually changing the variable type.
6. The old values, which have disappeared from the csv file, can still be displayed in JASP when changing **group** back to a nominal variable (perhaps they are cached somewhere?). I think this should not be possible.
7. This bug only occurs when the sync implies a change of data type (e.g., from nominal to ordinal).

[test.csv.zip](https://github.com/jasp-stats/jasp-issues/files/5592088/test.csv.zip)

[JASP_sync_problem.mp4.zip](https://github.com/jasp-stats/jasp-issues/files/5592132/JASP_sync_problem.mp4.zip)

boutinb:
Fixes jasp-stats/jasp-issues#1077

### DIFF
--- a/Common/labels.cpp
+++ b/Common/labels.cpp
@@ -137,10 +137,17 @@ bool Labels::syncInts(const std::set<int> &values)
 {
 	std::set<int>	valuesToAdd		= values,
 					valuesToRemove;
+	bool			isChanged			= false;
 
-	for (const Label & label : _labels)
+	for (Label & label : _labels)
 	{
 		int value = label.value();
+
+		if (!label.hasIntValue())
+		{
+			label.setValue(value);
+			isChanged = true;
+		}
 
 		if(values.count(value) > 0)	valuesToAdd.erase(value);
 		else						valuesToRemove.insert(value);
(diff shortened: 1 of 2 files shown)


Common/labels.cpp
```diff
@@ -138,10 +138,13 @@ bool Labels::syncInts(const std::set<int> &values)
 	std::set<int>	valuesToAdd		= values,
 					valuesToRemove;
 
-	for (const Label & label : _labels)
+	for (Label & label : _labels)
 	{
 		int value = label.value();
 
+		if (!label.hasIntValue())
+			label.setValue(value);
```
JorisGoosen, line 148: The function `bool syncInts` should tell if it changed right?

So what could happen, possibly, is that all the values were maybe all already convertible to the ints that are being synced. Then you do `setValue` which changes it so that it "hasIntValue", Im not sure if that should count as a "change" but I think it wouldn't hurt to do so.

So let the function remember if it did `setValue` on at least one of the labels and `return  setValueAtLeastOnce || valuesToAdd.size() + valuesToRemove.size() > 0;` or something like that.

JorisGoosen (changes requested):

Common/labels.cpp
boutinb: OK I have added your remark.

PR #4410, opened 2020-11-27
updated docs for github actions

vandenman:
I'm not sure if we should update more. There's also the function `jaspTools::runTestsTravis()` which is referenced in here that should be renamed or deprecated I guess.


TimKDJ (approved):
Thanks!

I'll merge `runTestsTravis` into `testAll`

PR #4412, opened 2020-11-27
Better handling of Dynamic modules and (re)loading

JorisGoosen:
- Removed some stuff from Analysis/Analyses for oldstyle modules, as we don't have them anymore
- Made sure reloading Development Module does not crash anymore
-- fixes: https://github.com/jasp-stats/INTERNAL-jasp/issues/1175
-- This was solved by remembering the _slaveFinishedConnection = connect(_slaveProcess, QOverload<int, QProcess::ExitStatus>::of(&QProcess::finished),	this, &EngineRepresentation::processFinished);
-- And removing it when an engine stopped, a race condition occured whereby the new engine started so fast it got the "closed" signal from the previous one and interpreted it as a crash. While in fact it worked fine.
- Added a loading indicator in the ribbonbar for modules while they are loading
- Added an error indicator, but Im not sure in which situation a module would have an error *and remain installed/loaded*. So not actually sure if it makes sense or not. I guess it would be useful in case an icon is missing. But a then a question mark would make more sense.
- Added a few more logging statements to engine startup
- Removed the "not running analyses on engine 0" define in the constructor of EngineRepresentation...

### DIFF
--- a/Desktop/engine/enginerepresentation.cpp
+++ b/Desktop/engine/enginerepresentation.cpp
@@ -9,11 +9,6 @@ EngineRepresentation::EngineRepresentation(IPCChannel * channel, QProcess * slav
 	: QObject(parent), _channel(channel)
 {
 	setSlaveProcess(slaveProcess);
-
-#ifndef JASP_DEBUG
-	if(channelNumber() == 0)
-		setRunsAnalysis(false); // don't perform 'runs' on process 0, "only" inits & filters & rCode & columnComputes & moduleRequests.
-#endif
 }
 
 
@@ -23,7 +18,7 @@ void EngineRepresentation::setSlaveProcess(QProcess * slaveProcess)
 	_slaveProcess = slaveProcess;
 	_slaveProcess->setParent(this);
 
-	connect(_slaveProcess, QOverload<int, QProcess::ExitStatus>::of(&QProcess::finished),	this, &EngineRepresentation::processFinished);
+	_slaveFinishedConnection = connect(_slaveProcess, QOverload<int, QProcess::ExitStatus>::of(&QProcess::finished),	this, &EngineRepresentation::processFinished);
 }
 
 EngineRepresentation::~EngineRepresentation()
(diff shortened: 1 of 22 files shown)


Desktop/engine/enginerepresentation.cpp
```diff
@@ -23,7 +18,7 @@ void EngineRepresentation::setSlaveProcess(QProcess * slaveProcess)
 	_slaveProcess = slaveProcess;
 	_slaveProcess->setParent(this);
 
-	connect(_slaveProcess, QOverload<int, QProcess::ExitStatus>::of(&QProcess::finished),	this, &EngineRepresentation::processFinished);
+	_slaveFinishedConnection = connect(_slaveProcess, QOverload<int, QProcess::ExitStatus>::of(&QProcess::finished),	this, &EngineRepresentation::processFinished);
```
boutinb, line 21: Here you make a copy of the connection: https://stackoverflow.com/questions/14828678/disconnecting-lambda-functions-in-qt5
This won't work.

boutinb (changes requested):

Desktop/engine/enginerepresentation.cpp
JorisGoosen: So, why does it solve the problem?

Also see: https://doc.qt.io/qt-5/qmetaobject-connection.html
It supports both move and copy constructors.

Also it is specifically the object that gets returned by `connect`: https://doc.qt.io/qt-5/qobject.html#connect
And received by `disconnect`: https://doc.qt.io/qt-5/qobject.html#disconnect-4 

Furthermore, in the linked documentation the following is stated literally:
> The function returns a QMetaObject::Connection that represents a handle to a connection if it successfully connects the signal to the slot. The connection handle will be invalid if it cannot create the connection, for example, if QObject is unable to verify the existence of either signal or method, or if their signatures aren't compatible. You can check if the handle is valid by casting it to a bool.

Unless you can actually show me that it doesn't work beyond some vague SO post from a long time ago I won't make any change. 

afbeelding
Not to mention the fact that it is about `lambda connects` and not about a normal QObject to QObject connect.

boutinb: OK. False alarm apparently.

boutinb:
@JorisGoosen There are now conflicts with submodules...

JorisGoosen:
I'll rebase it!

JorisGoosen:
This should be better.

JorisGoosen:
Can you make sure to "Rebase and merge" instead of the other options, that way our commithistory looks nicer.

PR #4413, opened 2020-11-30
Refactor the Source Type for QML Controls and more...

boutinb:
- Fixes some problems with ComboBox. It should also accepts source as LanguageModel.
- Refactor the SourceType so that it becomes an apart class.
- If the source of a control is the variable names, then this case should be treated as a normal kind of source. All special codes that handled the changes of dataset, can be then removed.
- Start making the QML controls independent of the MainWindow. Reference to MainWindow and DataSetPackage should be removed.
- Some code cleaning

### DIFF
--- a/Desktop/analysis/analysisform.cpp
+++ b/Desktop/analysis/analysisform.cpp
@@ -793,7 +793,10 @@ void AnalysisForm::setColumnsModel(QAbstractItemModel *model)
 		{
 			i.next();
 			if (i.value() == "columnName")
+			{
 				_columnsModelRole = i.key();
+				break;
+			}
 		}
 	}
 }
--- a/Desktop/analysis/analysisform.cpp
+++ b/Desktop/analysis/analysisform.cpp
@@ -50,6 +50,9 @@
 
 using namespace std;
 
+QAbstractItemModel* AnalysisForm::columnsModel =  nullptr;
+int AnalysisForm::columnsModelRole = Qt::DisplayRole;
+
 AnalysisForm::AnalysisForm(QQuickItem *parent) : QQuickItem(parent)
 {
 	setObjectName("AnalysisForm");
@@ -217,35 +220,10 @@ void AnalysisForm::_setUp()
 {
 	QList<JASPControl*> controls = _controls.values();
 
+	// set the order of the BoundItems according to their dependencies (for binding purpose)
 	for (JASPControl* control : controls)
-	{
 		control->setUp();
 
-		JASPListControl* listControl = qobject_cast<JASPListControl*>(control);
-		if (listControl)
-		{
-			TextAreaBase* textArea	= qobject_cast<TextAreaBase*>(listControl);
-			if (textArea)
-			{
-				// TODO : Change this special handling for TextArea
-				ListModelTermsAvailable	* availableModel = qobject_cast<ListModelTermsAvailable*>(textArea->model());
-				if (availableModel && textArea->modelNeedsAllVariables())
-					_allAvailableVariablesModels.push_back(availableModel);
-			}
-			else
-			{
-				ListModelTermsAvailable* availableModel = dynamic_cast<ListModelTermsAvailable*>(listControl->model());
-
-				if (availableModel)
-				{
-					if (!listControl->hasSource())	_allAvailableVariablesModels.push_back(availableModel);
-					else							_allAvailableVariablesModelsWithSource.push_back(availableModel);
-				}
-			}
-		}
-	}
-
-	// set the order of the BoundItems according to their dependencies (for binding purpose)
 	for (JASPControl* control : controls)
 	{
 		std::vector<JASPControl*> depends(control->depends().begin(), control->depends().end());
@@ -341,17 +319,6 @@ QString AnalysisForm::msgsListToString(const QStringList & list) const
 	return text;
 }
 
-void AnalysisForm::replaceVariableNameInListModels(const std::string & oldName, const std::string & newName)
-{
-	for (ListModelTermsAvailable * model : _allAvailableVariablesModels)
-	{
-		model->replaceVariableName(oldName, newName);
-		const QList<ListModelAssignedInterface*>& assignedModels = model->assignedModel();
-		for (ListModelAssignedInterface* modelAssign : assignedModels)
-				modelAssign->replaceVariableName(oldName, newName);
-	}
-}
-
 void AnalysisForm::setInfo(QString info)
 {
 	if (_info == info)
@@ -359,33 +326,6 @@ void AnalysisForm::setInfo(QString info)
 
 	_info = info;
 	emit infoChanged();
-}
-
-void AnalysisForm::_setAllAvailableVariablesModel(bool refreshAssigned)
-{
-	if (_allAvailableVariablesModels.size() == 0)
-		return;
-
-	std::vector<std::string> columnNames = DataSetPackage::pkg()->getColumnNames();
-
-
-	for (ListModelTermsAvailable* model : _allAvailableVariablesModels)
-	{
-		model->initTerms(columnNames);
-
-		if (refreshAssigned)
-		{
-			emit model->allAvailableTermsChanged(nullptr, nullptr);
-			const QList<ListModelAssignedInterface*>& assignedModels = model->assignedModel();
-			for (ListModelAssignedInterface* modelAssign : assignedModels)
-				modelAssign->refresh();
-		}
-	}
-
-	if (refreshAssigned)
-		for (ListModelTermsAvailable * model : _allAvailableVariablesModelsWithSource)
-			model->resetTermsFromSourceModels(true);
-
 }
 
 QString AnalysisForm::_getControlLabel(QString controlName)
@@ -436,14 +376,12 @@ void AnalysisForm::bindTo()
 
 	_options->blockSignals(true);
 	
-	_setAllAvailableVariablesModel();	
-
 	std::set<std::string> controlsJsonWrong;
 	
 	for (JASPControl* control : _dependsOrderedCtrls)
 	{
 		BoundControl* boundControl = dynamic_cast<BoundControl*>(control);
-		if (boundControl && control->isBound())
+		if (control->isBound() && boundControl)
 		{
 			std::string name = control->name().toStdString();
 			Option* option   = _options->get(name);
@@ -498,7 +436,7 @@ void AnalysisForm::bindTo()
 			if (availableModel)
 			{
 				if (optionsFromJASPFile != Json::nullValue || _analysis->isDuplicate())
-					availableModel->resetTermsFromSourceModels(false);
+					availableModel->resetTermsFromSources(false);
 				else
 					availableModelsToBeReset.push_back(availableModel);
 			}
@@ -508,7 +446,7 @@ void AnalysisForm::bindTo()
 	}
 
 	for (ListModelAvailableInterface* availableModel : availableModelsToBeReset)
-		availableModel->resetTermsFromSourceModels(true);
+		availableModel->resetTermsFromSources(true);
 	
 	_addLoadingError(tql(controlsJsonWrong));
 
@@ -540,7 +478,7 @@ void AnalysisForm::unbind()
 	for (JASPControl* control : _dependsOrderedCtrls)
 	{
 		BoundControl* boundControl = dynamic_cast<BoundControl*>(control);
-		if (boundControl)
+		if (control->isBound() && boundControl)
 			boundControl->unbind();
 	}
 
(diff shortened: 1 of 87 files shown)


Desktop/analysis/analysisform.cpp
```diff
@@ -50,6 +50,9 @@
 
 using namespace std;
 
+QAbstractItemModel* AnalysisForm::columnsModel =  nullptr;
+int AnalysisForm::columnsModelRole = Qt::DisplayRole;
```
JorisGoosen: Yeah, so I really think directly querying the role like: `AnalysisForm::columnsModel->rolenames()["columnName"]` is much nicer and less chance of bugs because you do not need to remember to always make sure to update `columnsModelRole`.

I think `rolenames` might even be passed by reference, but not sure if possible.

Desktop/mainwindow.h
```diff
@@ -96,6 +96,7 @@ class MainWindow : public QObject
 	bool	welcomePageVisible()	const	{ return _welcomePageVisible;	}
 	bool	checkAutomaticSync()	const	{ return _checkAutomaticSync;	}
 	QString downloadNewJASPUrl()	const	{ return _downloadNewJASPUrl;	}
+	ColumnsModel *	columnsModel()			{ return _columnsModel;			}
```
JorisGoosen: I think it would be nicer to add a `static ColumnsModel` to ColumnsModel and use that as a singleton.
I don't particularly like the whole "let's store everything in a pointer in MainWindow"-idiom we have got going on. It is easier to just directly query ColumnsModel.

Maybe this becomes less practical when we support more than 1 dataset loaded at the same time, but this doesnt seem very likely on the short term. And anyway, with `MainWindow::columnsModel()` there is the same problem.

Desktop/widgets/jasplistcontrol.cpp
```diff
@@ -352,76 +118,69 @@ void JASPListControl::setUp()
...
 
 	return result;
 }
 
+QVector<std::pair<SourceItem*, Terms> > JASPListControl::getTermsPerSource()
+{
+	QVector<std::pair<SourceItem*, Terms> > result; // Do not use a map in order to keep the order of the sources.
+
+	for (SourceItem* sourceItem : _sourceItems)
+		if (sourceItem->combineWithOtherModels())
```
JorisGoosen: Nicer:
```c++
for (SourceItem* sourceItem : _sourceItems)
	result.push_back(std::make_pair(sourceItem, sourceItem->combineWithOtherModels() ? _getCombinedTerms(sourceItem) : sourceItem->getTerms()));
```

Desktop/widgets/listmodel.cpp
```diff
@@ -150,10 +125,11 @@ ListModel *ListModel::getSourceModelOfTerm(const Term &term)
 {
 	ListModel* result = nullptr;
 
-	for (const std::pair<JASPListControl::SourceType *, Terms>& source : listView()->getTermsPerSource())
+	for (const auto& pair : listView()->getTermsPerSource())
```
JorisGoosen: I usually just call the pair in the iterator the names of what they are like:
```c++
for ( const auto & sourceTerms : listView()->getTermsPerSource())
	if(sourceTerms.second.contains(term) && sourceTerms.first->model())
		result = sourceTerms.first->model();
```	

It is nicer right, without the braces and the copy of the pointer?

JorisGoosen:
Bruno tells me this PR fixes: https://github.com/jasp-stats/INTERNAL-jasp/issues/1183

Desktop/widgets/listmodelfiltereddataentry.cpp
```diff
@@ -34,12 +33,13 @@ ListModelFilteredDataEntry::ListModelFilteredDataEntry(TableViewBase * parent, Q
 
 }
 
-void ListModelFilteredDataEntry::dataSetChangedHandler()
+//TODO: This is not called anymore, but should be handled by termsChangedHandler
```
JorisGoosen, line 36: So is this a TODO or not?

JorisGoosen: As in, it certainly looks like dataSetChangedHandler is gone now

Desktop/widgets/listmodelinteractionavailable.cpp
```diff
@@ -20,33 +20,31 @@
...
 	Terms fixedFactors;
 	Terms randomFactors;
 	Terms covariates;
 
-	for (const std::pair<JASPListControl::SourceType *, Terms>& source : listView()->getTermsPerSource())
+	for (const auto& pair : listView()->getTermsPerSource())
 	{
-		ListModel* sourceModel = source.first->model;
-		const Terms& terms = source.second;
+		SourceItem* sourceItem = pair.first;
```
JorisGoosen: Also here it would make the whole code that much nicer to look at.

JorisGoosen: Just give the paris descriptive names instead of copying shit :p

Desktop/widgets/listmodellabelvalueterms.cpp
```diff
@@ -132,23 +116,25 @@ void ListModelLabelValueTerms::_setLabelValues(const JASPListControl::LabelValue
...
 {
-	JASPListControl::LabelValueMap values;
+	JASPListControl::LabelValueMap labelValuePairs;
 
 	if (_listView->addEmptyValue())
-		values.push_back(std::make_pair(_listView->placeholderText(), ""));
+		labelValuePairs.push_back(std::make_pair(_listView->placeholderText(), ""));
 
-	for (const std::pair<JASPListControl::SourceType *, Terms>& source : _listView->getTermsPerSource())
+	for (const auto& pair : listView()->getTermsPerSource())
```
JorisGoosen: Or maybe do it like `void		applyToAll(std::function<void(Analysis *analysis)> applyThis);`
The you could just have it like:
` void forAllSourceTerms(std::function<void(SourceItem * item, const Terms & terms)> callThis);`

That way all of these `for ... copy first and second, do stuff` loops can be so much nicer than what Youve been adding.

JorisGoosen: You could even make one where you check if it has a model and make a `forAllSourceTermsWithModel(std::function<void(SourceItem * item, ListModel * model)> callThis)` or something like that

Desktop/widgets/sourceitem.cpp
```diff
@@ -0,0 +1,406 @@
...
+
+	if (sourceModel)
+	{
+		_model = sourceModel;
+		_listControl->addDependency(_model->listView());
+		ListModel::connect(_model, &ListModel::termsChanged, _listControl->model(), &ListModel::sourceTermsChanged);
+	}
+	else if (_nativeModel)
+	{
+		connect(_nativeModel, SIGNAL(dataChanged(const QModelIndex &, const QModelIndex &, const QVector<int> &)),	this,	SLOT(nativeModelChanged()) );
```
JorisGoosen: New style `connect`'s weren't possible I suppose?

JorisGoosen: (You can refer to functions like `&SuperClass::Function`)

JorisGoosen (changes requested):
Phew, you have been working hard!

I think a few changes could make things a bit better. But I see nothing that would break necessarily, just things that would make the code easier to maintain.

I'll be building it now and testing.

JorisGoosen:
Ok, testing showed my that replacing a variablename in the data through spreadsheeteditor only removes it from variableslists... It does not replace them...

EDIT: that means that if `V1` is used in Descriptives and I rename it to `somethingElse` it gets removed from the list. Not what we want.

Desktop/widgets/listmodellabelvalueterms.cpp
boutinb: Nice. I'll do this

Desktop/widgets/sourceitem.cpp
boutinb: Because of the parameters, I chose this case. But I could use a lambda instead, and adding the sourceItem as context to be sure that the connection is destroyed when the SourceItem is destroyed.

JorisGoosen: But the normal connect also drops the parameters right?
As in, I don't see why this version would be better?

JorisGoosen: By that I mean, why not do:
`connect(_nativeModel, &QAbstractItemModel::dataChanged, this, &SourceItem::nativeModelChanged);` should work I think.

boutinb: That's true. At some point I thought that would not work, but it works indeed!

boutinb:
The problem is that I have now only the ColumnModel, and it does not give such a signal that tells that the column name is changed.
This should be solved with the DataSetPackage solution we discussed.

Desktop/widgets/sourceitem.cpp
JorisGoosen: Yeah, connects automatically drop arguments/parameters from the signal if they are not needed/accepted by the slot.

JorisGoosen:
> 
> 
> The problem is that I have now only the ColumnModel, and it does not give such a signal that tells that the column name is changed.
> This should be solved with the DataSetPackage solution we discussed.

Doesn't it send a `dataChanged` when the columnName changes?
I think so.

Desktop/analysis/analysisform.cpp
```diff
@@ -782,6 +782,22 @@ QString AnalysisForm::metaHelpMD() const
...
+{
+	_columnsModel = model;
+
+	if (model)
+	{
+		QHashIterator<int, QByteArray> i(model->roleNames());
+		while (i.hasNext())
+		{
+			i.next();
+			if (i.value() == "columnName")
```
JorisGoosen, line 795: If you've found it you should `break;` the `while` loop.

JorisGoosen: And... QHash is a lookup table, so you can just do `model->roleNames()["columnName"]` which I think I mentioned before.

Doing that also means you can remove `AnalysisForm::_columnsModelRole` entirely and look it up every time.

JorisGoosen (changes requested):
Looks much better already, but I don't get why you aren't using QHash properly.

Desktop/analysis/analysisform.cpp
JorisGoosen: Ps. see if rolenames can return a `QHash &` because otherwise it is copying the whole thing all the time. Not *so bad* but also not so good.

boutinb: RoleNames is a QHash<int, QByteArrea>, so roleNames()["columnName"] won't work.
I'll add  a break in the loop, but as we can't have the role directly with a lookup, I'll keep the _columnsModelRole. This name don't change.

And columnsModel does not send dataChange signal. I will handle this issue with DataSetPackage rewrite.

JorisGoosen: Aah, well `columnsModel` should send that, but I guess I was lazy when implementing it :p

And ah, yeah, ok that is the wrong way round with the QHash... Ok, then the copy is reasonable.

JorisGoosen:
I'll go test if variablenames are now properly replaced.

Desktop/widgets/sourceitem.cpp
```diff
@@ -102,10 +96,10 @@ void SourceItem::_setUp()
 	}
 	else if (_nativeModel)
 	{
-		connect(_nativeModel, &QAbstractItemModel::dataChanged,			this,	&SourceItem::nativeModelChanged);
-		connect(_nativeModel, &QAbstractItemModel::rowsInserted,		this,	&SourceItem::nativeModelChanged);
-		connect(_nativeModel, &QAbstractItemModel::rowsRemoved,			this,	&SourceItem::nativeModelChanged);
-		connect(_nativeModel, &QAbstractItemModel::modelReset,			this,	&SourceItem::nativeModelChanged);
+		connect(_nativeModel, &QAbstractItemModel::dataChanged,			this,	[this]() { _listControl->model()->sourceTermsChanged(); });
+		connect(_nativeModel, &QAbstractItemModel::rowsInserted,		this,	[this]() { _listControl->model()->sourceTermsChanged(); });
+		connect(_nativeModel, &QAbstractItemModel::rowsRemoved,			this,	[this]() { _listControl->model()->sourceTermsChanged(); });
```
JorisGoosen: Ok, this is not necessarily a complaint, or request for change but I am not sure why you did:
`connect(_nativeModel, &QAbstractItemModel::dataChanged,			this,	[this]() { _listControl->model()->sourceTermsChanged(); });`
When you could also do:
`connect(_nativeModel, &QAbstractItemModel::dataChanged,			this,	_listControl->model(), &ListModel::sourceTermsChanged());` 
(That is, if `_listControl->model()` is a ListModel.

JorisGoosen:
Ok, so replacing variablenames still breaks...

JorisGoosen:
Can this be reviewed again?

boutinb:
Now you can review it. I've changed more things...

PR #4414, opened 2020-12-01
Remove the R unit tests that were moved to the submodules

TimKDJ:


JorisGoosen:
I see you also want to go up in the deletion-rankings :p

PR #4418, opened 2020-12-02
Plot editing v0.1

vandenman:
Something that doesn't work yet is the dropdown in QML for the titleType. There's also a terrible merge conflict, but I figured it might be better to squash and rebase after this is reviewed. Requires the plotediting branch of jaspBase and jaspGraphs.

other PRs
jaspBase: jasp-stats/jaspBase#7
jaspGraphs: jasp-stats/jaspGraphs#4
you need all 3 to actually make it work in jasp.

PR #4420, opened 2020-12-02
Plotediting

JorisGoosen:
I rebased https://github.com/jasp-stats/jasp-desktop/pull/4418 on `development` and squashed all the commits in one.
Also jaspBase and jaspGraphs have been updated to include plotediting changes.

- copy changes from old branch
- Move PlotEditor.qml back to its buddies and make sure it can actually load
- Move stuff to special  PlotEditor function to prepare for separating qml stuff
- Plot is now shown the same as in results when a dark theme is selected
- axes can now be edited in visible tables
- refactored some code
- moved stuff to separate qml for the table
- little problem is that the plot is updated live when editing breaks/labels but changes made while editImage is running are ignored.
  This could be solvved by either only sending an editImage request when user is done editing or tracking requests or so. The last one would be cooler I think.
- Store a request counter in editImage and send again if wrong response
- Solves the problem where an old change was shown
- add support for discrete axes
- broken dropdown ?


JorisGoosen:
More changes incoming btw

TimKDJ:
- [x] I can't make the breaks table show anymore:
Screenshot 2020-12-04 at 13 55 15

- [x] When there is a scrollbar the bottom cuts off:
Screenshot 2020-12-04 at 13 55 27

- [ ] This `Specify breaks:` seems like it could go (not sure though, maybe it is just the missing table problem and it being there does make sense) when there's nothing to specify:
Screenshot 2020-12-04 at 13 34 10

- [ ] Dunno if I should include this here, but @vandenman it doesn't much care for piechart. Also note the`NaN`:
Screenshot 2020-12-04 at 13 34 33

TimKDJ:
You didn't push any changes yet, but I'll review them when you do

JorisGoosen:
(The "more changes incoming" are already there, but your points are very valid and I will fix them)

boutinb (approved):

JorisGoosen:
> This `Specify breaks:` seems like it could go (not sure though, maybe it is just the missing table problem and it being there does make sense) when there's nothing to specify:

Ok, so the breaks aren't entirely straightforward to fix but it should be visible there. I don't know how to recognize it in C++/QML right now.

> Piechart problems

Well, the piechart doesnt have an X or Y axis so no wonder it breaks. 

I've fixed the other issues and will push after fixing the conflicts.
I think after we merge this me, @vandenman and @TimKDJ should have a zoommeeting to talk about next steps. And Tim can tell us all the things he thinks are weird or ugly 'cause I'm sure he noticed some ;)

vandenman:
> This Specify breaks: seems like it could go (not sure though, maybe it is just the missing table problem and it being there does make sense) when there's nothing to specify:

That looks like a bug with discrete axes, but I'm not entirely sure. There should be one axis tick at 1 or 0 with an empty string `""` as a label.

> Well, the piechart doesnt have an X or Y axis so no wonder it breaks.

I still need to write some validation code so that users can't even try to edit things like pie-charts. 

> I think after we merge this me, @vandenman and @TimKDJ should have a zoommeeting to talk about next steps.

Good idea!

JorisGoosen:
Pfoeey!

Ok, here you go guys.
The TitleType dropdown doesnt seem to work perfectly but to be honest: I don't really understand how `DropDown` works.
@boutinb can help out with that later.

I did change the CPP code a bit more to make sure we use Enums in QML instead of random strings.

JorisGoosen:
Ok, so Bruno fixed the dropdown I believe. It can be reviewed again!

Or at least tested by someone else than me.

Bruno wants to join for the zoom, so let's do that next week. On a tuesday or wednesday for instance?

vandenman:
I tested it a bit today and yesterday and it works pretty well. There is a lot that I want to add/ fix on the R side though. IMO this can be merged unless someone else still wants to review it?

JorisGoosen:
@boutinb said he would have a look and I would certainly appreciate that.

PR #4428, opened 2020-12-10
 Changed column header from mean difference to sample mean and location estimate

AlexanderLyNL:


boutinb:
Can we merge it?

AlexanderLyNL:
Yes, go for it.

PR #4432, opened 2020-12-14
Remove DataSetPackage dependencies in QML Controls

boutinb:
Set properly containsVariables & containsInterations in models
Fixes jasp-stats/INTERNAL-jasp#1181

### DIFF
--- a/Desktop/analysis/analysisform.cpp
+++ b/Desktop/analysis/analysisform.cpp
@@ -18,39 +18,22 @@
 
 #include "analysisform.h"
 #include "knownissues.h"
-
-#include <boost/bind.hpp>
-
 #include "options/boundcontrol.h"
 #include "utilities/qutils.h"
+#include "widgets/listmodeltermsavailable.h"
+#include "widgets/jasplistcontrol.h"
+#include "widgets/expanderbuttonbase.h"
+#include "log.h"
+#include "jaspcontrol.h"
+#include "data/columnsmodel.h"
 
 #include <QQmlProperty>
 #include <QQmlContext>
-
-#include "widgets/listmodeltermsavailable.h"
-
-#include "utils.h"
-#include "dirs.h"
-#include "utilities/settings.h"
-#include "gui/messageforwarder.h"
-#include "mainwindow.h"
-#include "log.h"
-#include "jaspcontrol.h"
-#include "widgets/comboboxbase.h"
-#include "widgets/textareabase.h"
-#include "widgets/boundcontrolterms.h"
-#include "widgets/boundcontrolmeasurescells.h"
-#include "widgets/boundcontrollayers.h"
-#include "widgets/inputlistbase.h"
-#include "widgets/repeatedmeasuresfactorslistbase.h"
-#include "widgets/factorsformbase.h"
-#include "widgets/tableviewbase.h"
-#include "widgets/expanderbuttonbase.h"
-
+#include <QTimer>
 
 using namespace std;
 
-QAbstractItemModel* AnalysisForm::_columnsModel		=  nullptr;
+ColumnsModel*		AnalysisForm::_columnsModel		=  nullptr;
 int					AnalysisForm::_columnsModelRole = Qt::DisplayRole;
 
 AnalysisForm::AnalysisForm(QQuickItem *parent) : QQuickItem(parent)
@@ -64,12 +47,18 @@ AnalysisForm::AnalysisForm(QQuickItem *parent) : QQuickItem(parent)
 
 QVariant AnalysisForm::requestInfo(const Term &term, VariableInfo::InfoType info) const
 {
+	if (!_columnsModel) return QVariant();
+
 	try {
+		QModelIndex index = _columnsModel->index(_columnsModel->getColumnIndex(term.asString()), 0);
 		switch(info)
 		{
-		case VariableInfo::VariableType:		return int(					DataSetPackage::pkg()->getColumnType(term.asString()));
-		case VariableInfo::VariableTypeName:	return columnTypeToQString(	DataSetPackage::pkg()->getColumnType(term.asString()));
-		case VariableInfo::Labels:				return						DataSetPackage::pkg()->getColumnLabelsAsStringList(term.asString());
+		case VariableInfo::VariableType:			return _columnsModel->data(index, ColumnsModel::TypeRole);
+		case VariableInfo::VariableTypeName:		return _columnsModel->data(index, ColumnsModel::TypeNameRole);
+		case VariableInfo::VariableTypeIcon:		return _columnsModel->data(index, ColumnsModel::IconSourceRole);
+		case VariableInfo::VariableTypeDisabledIcon: return _columnsModel->data(index, ColumnsModel::DisabledIconSourceRole);
+		case VariableInfo::VariableTypeInactiveIcon: return _columnsModel->data(index, ColumnsModel::InactiveIconSourceRole);
+		case VariableInfo::Labels:					return	_columnsModel->data(index, ColumnsModel::LabelsRole);
 		}
 	}
 	catch(columnNotFound & e) {} //just return an empty QVariant right?
@@ -381,6 +370,8 @@ void AnalysisForm::bindTo()
 	for (JASPControl* control : _dependsOrderedCtrls)
 	{
 		BoundControl* boundControl = dynamic_cast<BoundControl*>(control);
+		JASPListControl* listControl = dynamic_cast<JASPListControl *>(control);
+
 		if (control->isBound() && boundControl)
 		{
 			std::string name = control->name().toStdString();
@@ -419,11 +410,13 @@ void AnalysisForm::bindTo()
 				_options->add(name, option);
 			}
 
+			if (listControl)
+				option->setShouldEncode(listControl->containsVariables());
+
 			boundControl->bindTo(option);
 			_optionControlMap[option] = control;
 		}
 
-		JASPListControl* listControl = dynamic_cast<JASPListControl *>(control);
 		if (listControl && listControl->hasSource())
 		{
 			ListModelAvailableInterface* availableModel = qobject_cast<ListModelAvailableInterface*>(listControl->model());
@@ -782,7 +775,7 @@ QString AnalysisForm::metaHelpMD() const
 	return "---\n# " + tr("Output") + "\n\n" + metaMDer(_analysis->meta(), 2);
 }
 
-void AnalysisForm::setColumnsModel(QAbstractItemModel *model)
+void AnalysisForm::setColumnsModel(ColumnsModel *model)
 {
 	_columnsModel = model;
 
--- a/Desktop/analysis/analysisform.cpp
+++ b/Desktop/analysis/analysisform.cpp
@@ -33,9 +33,6 @@
 
 using namespace std;
 
-ColumnsModel*		AnalysisForm::_columnsModel		=  nullptr;
-int					AnalysisForm::_columnsModelRole = Qt::DisplayRole;
-
 AnalysisForm::AnalysisForm(QQuickItem *parent) : QQuickItem(parent)
 {
 	setObjectName("AnalysisForm");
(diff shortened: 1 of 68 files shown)


Desktop/analysis/options/optionterms.h
```diff
@@ -27,7 +27,6 @@ class OptionTerms : public OptionI<std::vector<std::vector<std::string> > >
 public:
 			OptionTerms() : OptionI(true), _onlyOneComponent(false), _onlyOneTerm(false) 
 			{
-				setShouldEncode(true); //Maybe we should check if the terms are in fact columnnames or not? But, for now I will just assume that terms are very likely to contain columnnames. And otherwise Ill have to override asMetaJson and step through all this like in OptionTerms::asJSON() and simply check it
```
JorisGoosen, line 30: Did you take the meta into account after removing this?

boutinb: OptionTerms is never used directly, only as subclass of OptionVariables (which is always encoded), or of OptionTerm. OptionTerm is used in CustomContrast table, JAGS and ComponentList for values that are not variables, and in BoundControlTerms where I have added code so that it sets the encoding only when the model contains variables.
I hope to remove all these option classes soon...

Desktop/analysis/analysisform.cpp
```diff
@@ -18,39 +18,22 @@
...
+#include "data/columnsmodel.h"
 
+#include <QQmlProperty>
+#include <QQmlContext>
+#include <QTimer>
 
 using namespace std;
 
-QAbstractItemModel* AnalysisForm::_columnsModel		=  nullptr;
+ColumnsModel*		AnalysisForm::_columnsModel		=  nullptr;
```
JorisGoosen: I would suggest either making `ColumnsModel` a singleton so you can just refer to `ColumnsModel::colModel()` or something like that.
Or you store the columnsmodel per AnalysisForm. But having a static reference to a ColumnsModel in AnalysisForm seems a bit impractical.

Desktop/analysis/analysisform.h
```diff
@@ -138,8 +135,8 @@ public slots:
 	QString		warnings()			const { return msgsListToString(_formWarnings);	}
 	QVariant	analysis()			const { return QVariant::fromValue(_analysis);	}
 
-	static void					setColumnsModel(QAbstractItemModel* model);
-	static QAbstractItemModel*	getColumnsModel()		{ return _columnsModel;		}
+	static void					setColumnsModel(ColumnsModel* model);
+	static ColumnsModel*		getColumnsModel()		{ return _columnsModel;		}
 	static int					getColumnsModelRole()	{ return _columnsModelRole;	}
```
JorisGoosen: And if you are using `ColumnsModel` specifically anyway you do not need to store the Role at all anymore right?

Desktop/analysis/options/optionterms.h
JorisGoosen: And have you also taken into account the "extra encodings" as needed for the levels-thing in ANOVA RM?

Desktop/data/datasettablemodel.cpp
```diff
@@ -21,6 +21,9 @@
 DataSetTableModel::DataSetTableModel() : DataSetTableProxy(parIdxType::data)
 {
 	connect(DataSetPackage::pkg(),	&DataSetPackage::columnsFilteredCountChanged,	this, &DataSetTableModel::columnsFilteredCountChanged	);
+	connect(DataSetPackage::pkg(),	&DataSetPackage::columnDataTypeChanged,			this, &DataSetTableModel::columnTypeChanged				);
+	connect(DataSetPackage::pkg(),	&DataSetPackage::labelChanged,					this, &DataSetTableModel::labelChanged					);
```
JorisGoosen, line 25: I assume this signal is sent in addition to `dataChanged` and not as a replacement?

Desktop/data/datasettablemodel.h
```diff
@@ -45,11 +45,19 @@ class DataSetTableModel : public DataSetTableProxy
...
+	QStringList				getColumnLabelsAsStringList(int col)	const				{ return DataSetPackage::pkg()->getColumnLabelsAsStringList(col);					}
+	size_t					getMaximumColumnWidthInCharacters(int index) const			{ return DataSetPackage::pkg()->getMaximumColumnWidthInCharacters(index);			}
+	QModelIndex				parentModelForType(parIdxType type, int column = 0)	const	{ return DataSetPackage::pkg()->parentModelForType(type, column);					}
+
 				bool		showInactive()							const				{ return _showInactive;	}
 
 signals:
 				void		columnsFilteredCountChanged();
 				void		showInactiveChanged(bool showInactive);
+				void		columnTypeChanged(QString colName);
```
JorisGoosen, line 58: Or maybe you left it at the `resetModel` style instead of fixing `dataChanged`?

Desktop/widgets/boundcontrolmeasurescells.cpp
```diff
@@ -77,8 +75,5 @@ void BoundControlMeasuresCells::updateOption()
 	const Terms& terms = _measuresCellsModel->terms();
 	
 	if (_boundTo)
-	{
-		_boundTo->setShouldEncode(true);
```
JorisGoosen, line 81: Where did this go then? Superclass?

Desktop/widgets/boundcontrolterms.cpp
```diff
@@ -130,14 +130,14 @@ void BoundControlTerms::bindTo(Option *option)
...
+	if (_listView->hasRowComponent() || _listView->containsInteractions())
 	{
 		Options* templote = new Options();
 
-		if (_termsModel->areTermsInteractions())
+		if (_listView->containsInteractions())
 		{
 			OptionTerm * newOptTerm = new OptionTerm();
-			newOptTerm->setShouldEncode(true);
+			newOptTerm->setShouldEncode(_listView->containsVariables());
```
JorisGoosen, line 140: I assume this is that superclass then (It isnt easy to see here in the review part of github sorry)

Desktop/analysis/options/optionterms.h
JorisGoosen: https://github.com/jasp-stats/INTERNAL-jasp/issues/1163

Desktop/widgets/comboboxbase.cpp
```diff
@@ -116,34 +120,40 @@ void ComboBoxBase::termsChangedHandler()
...
+
+	if (initialized())
 	{
-		QString label = term.asQString();
-		QString value = _model->getValue(label);
-		options.push_back(value.toStdString());
-		if (value == _currentValue)
-			index = counter;
-		counter++;
+		int counter = 0;
```
JorisGoosen: Why not just do `for(size_t counter=0; counter < terms.size(); counter++)`  then?

Desktop/widgets/listmodel.h
```diff
@@ -131,8 +137,7 @@ public slots:
 
 			JASPListControl*				_listView = nullptr;
 			Terms							_terms;
-			bool							_areTermsVariables		= true;
-			bool							_areTermsInteractions	= false;
+			QStringList						_columnsUsedForLabels;
```
JorisGoosen, line 140: It seems to me you only use this to look up things, I would suggest you turn it into a `set` then. 

You could add the inverse of https://github.com/jasp-stats/jasp-desktop/blob/cad06e46416bec1de0183cdcdeba06877ef844d8/Desktop/utilities/qutils.h#L43 there for it

Desktop/widgets/listmodelfiltereddataentry.cpp
```diff
@@ -331,36 +332,37 @@ QVariant ListModelFilteredDataEntry::data(const QModelIndex &index, int role) co
...
 
-	if(!DataSetPackage::pkg()->hasDataSet() || column > _colNames.size() || column < 0)
+	if(getDataSetRowCount() == 0 || column > _colNames.size() || column < 0)
 		return QVariant();
 
 	std::string colName = _colNames[column].toStdString();
 	size_t rowData		= _filteredRowToData[static_cast<size_t>(row)];
 
-	DataSetPackage * package = DataSetPackage::pkg();
+	ColumnsModel* columnsModel = AnalysisForm::getColumnsModel();
```
JorisGoosen: See, if you had something like `ColumnsModel::colModel()` you could just use that directly!

JorisGoosen: And if it was specific to each AnalysisForm it ought to not be static.

Desktop/widgets/variableslistbase.cpp
```diff
@@ -306,6 +310,70 @@ void VariablesListBase::moveItems(QList<int> &indexes, ListModelDraggable* targe
...
+
+void VariablesListBase::_setAllowedVariables()
+{
+	if (suggestedColumns().empty() && !allowedColumns().empty())
+		setSuggestedColumns(allowedColumns());
+	else if (allowedColumns().empty() && !suggestedColumns().empty())
+	{
+		QStringList newAllowedColumns = suggestedColumns();
+		if (suggestedColumns().contains("scale"))
+		{
```
JorisGoosen, line 348: I think it looks nicer as:
```
			if (!newAllowedColumns.contains("nominal")) newAllowedColumns.push_back("nominal");
			if (!newAllowedColumns.contains("ordinal"))   newAllowedColumns.push_back("ordinal");
```

But even better would be to turn the list into a set and simply do: `newAllowedColumns.insert("nominal")` 
Because there really is not reason to keep order for a list of allowed columns is there?

JorisGoosen (changes requested):
Mostly looks great!

I don't know why you always use a list, even in places where some kind of set would be more appropriate. So I'd suggest you change those.

Furthermore I would hope you make sure the "extra encodings" as required by ANOVA RM and the linked issue still works.

And please either make ColumnsModel non-static or move it to ColumnsModel itself.

Desktop/analysis/options/optionterms.h
boutinb: Yes the JASP file in this issue still works.

Desktop/data/datasettablemodel.cpp
boutinb: yes

Desktop/widgets/comboboxbase.cpp
boutinb: OK

Desktop/analysis/analysisform.cpp
boutinb: Yes, good idea

Desktop/analysis/analysisform.h
boutinb: Yes, I'll remove it

Desktop/data/datasettablemodel.h
boutinb: I don't understand... with columnTypeChanged I have precisely the information I need to change the type. I don't hae to refresh all models.

Desktop/widgets/boundcontrolmeasurescells.cpp
boutinb: this is set automatically because this control has a source that contains variables.

Desktop/widgets/boundcontrolterms.cpp
boutinb: yep

boutinb:
@JorisGoosen I have made ColumnsModel a singleton and remove the columnModel pointer and role in AnalysisForm.
And some small changes. 
For you remark about the abusive use of QList, If I have a look to the header changes, I have only added QStringList  because it is a type that can be set as Q_PROPERTY.
Can you have a look, and if it's OK merge this PR. Koen is waiting for it.

JorisGoosen:
> 
> 
> @JorisGoosen I have made ColumnsModel a singleton and remove the columnModel pointer and role in AnalysisForm.
> And some small changes.
> For you remark about the abusive use of QList, If I have a look to the header changes, I have only added QStringList because it is a type that can be set as Q_PROPERTY.
> Can you have a look, and if it's OK merge this PR. Koen is waiting for it.

Yeah but on several types of changes it goes through the whole list every time right? And on the get-function for the property you could just call `tql(std::set<std::string>)`. BUt Admittedly all these lists will probably be small so it probably doesnt matter.

Desktop/data/datasettablemodel.h
JorisGoosen: Well, to have DataSetPackage as the root of all models all changes should ideally go through that. But I see that it comes from there so it is fine.

Desktop/data/columnsmodel.cpp
```diff
@@ -1,15 +1,19 @@
...
-   connect(_tableModel, &DataSetTableModel::labelsReordered,		this, &ColumnsModel::labelsReordered	);
+	connect(_tableModel, &DataSetTableModel::headerDataChanged,		this, &ColumnsModel::onHeaderDataChanged);
+	connect(_tableModel, &DataSetTableModel::dataChanged,			this, &ColumnsModel::onDataChanged		);
+	connect(_tableModel, &DataSetTableModel::modelAboutToBeReset,	this, &ColumnsModel::beginResetModel	);
+	connect(_tableModel, &DataSetTableModel::modelReset,				this, &ColumnsModel::endResetModel			);
+	connect(_tableModel, &DataSetTableModel::columnTypeChanged,		this, &ColumnsModel::columnTypeChanged	);
+	connect(_tableModel, &DataSetTableModel::labelChanged,			this, &ColumnsModel::labelChanged		);
+	connect(_tableModel, &DataSetTableModel::labelsReordered,		this, &ColumnsModel::labelsReordered	);
+
+	if (_singleton == nullptr) _singleton = this;
```
JorisGoosen, line 16: Now that you set it, you should also unset it in the destructor!

JorisGoosen (changes requested):
Sorry one more change, you really need to unset the _singleton variable because that could lead to crashes that are hard to debug.

At least if we get a crash because _singleton == nullptr we can see it was destroyed.

After that it can totally be merged.

PR #4437, opened 2021-01-05
Change ColumnsModel into a QTransposeProxyModel

boutinb:
fixes jasp-stats/INTERNAL-jasp#1206
Fixes also label changes for Custom Contrast.

### DIFF
--- a/Desktop/data/columnsmodel.cpp
+++ b/Desktop/data/columnsmodel.cpp
@@ -16,7 +16,7 @@ ColumnsModel::ColumnsModel(DataSetTableModel *tableModel) : QTransposeProxyModel
 
 QVariant ColumnsModel::data(const QModelIndex &index, int role) const
 {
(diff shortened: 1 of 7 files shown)


Desktop/data/columnsmodel.cpp
```diff
@@ -16,7 +16,7 @@ ColumnsModel::ColumnsModel(DataSetTableModel *tableModel) : QTransposeProxyModel
 
 QVariant ColumnsModel::data(const QModelIndex &index, int role) const
 {
-	if(index.row() < 0) return QVariant() ; // index.row() >= rowCount()): the row can be greater than the rowCount when a data of a column is asked.
+	if(index.row() < 0 && index.row() >= rowCount()) return QVariant();
```
JorisGoosen: How could that ever be true?

JorisGoosen (changes requested):

Desktop/data/columnsmodel.cpp
boutinb: net aangepast voordat ik jouw comment zag...

PR #4438, opened 2021-01-06
Rewrite variables window

JorisGoosen:
Implements a version of the variableswindow aka labeleditor based on our own datasetview.
Also fixes the crash of audit analyses.

I've added Koen as a reviewer seeing as he asks it in the QMLs.
I think those changes should also be incorporated in the Audit-module repo.

If this is approved I will start porting the changes to `development`.


koenderks:
You just changed `GridLayout` to `RowLayout` in the audit files?

koenderks (approved):

JASP-Desktop/data/datasetpackage.cpp
```diff
@@ -271,7 +279,7 @@ QVariant DataSetPackage::data(const QModelIndex &index, int role) const
 		int parColCount = columnCount(index.parent()),
 			parRowCount = rowCount(index.parent());
 
-		if(!_dataSet || index.column() >= parColCount || index.row() >= parRowCount)
+		if(!_dataSet || index.row() >= parRowCount)
```
boutinb, line 282: You assume here that the index always has a parent. Is it really so?

boutinb (commented):
I don't understand why you had to change datasetview for this PR.
I don't understand why you change the selection in the Variables Window with a CheckBox. Personally I prefer the way the selection was done previously: less control/widget seems to me better.

JASP-Desktop/data/datasetpackage.cpp
JorisGoosen: Yes by definition, that is how DataSetPackage has been setup, see the `index() override`. The parent is used to discern whether or not we are in the `root, filter, label or data` subtree. When we are in the `label` subtree there is always a parent, because otherwise we couldn't be in the `label` subtree.

JorisGoosen:
> I don't understand why you had to change datasetview for this PR.

To add a selection role in the first place and also to allow for setting the width of a column to a specific value as opposed to telling datasetview how many characters to expect.

> I don't understand why you change the selection in the Variables Window with a CheckBox. Personally I prefer the way the selection was done previously: less control/widget seems to me better.

I get what you mean and Im not sure about these boxes. But without the extra box the `TextInput` in the Label column takes most of the clickable space and the filter takes the first column. So then the user would have to click either around the textinput or the value. While the selection-move buttons are on the side. SO I figured putting the selection column right on the side would work smooth. And honestly it does.

But we can have a nice discusssion about it with the rest of the JASP team after it is in. It isn't exactly hard to get it out again.

boutinb:
OK, let's implement it, and check it with some people.
I think it could be possible to select the row while being int the TextInput

JorisGoosen:
> OK, let's implement it, and check it with some people.
> I think it could be possible to select the row while being int the TextInput

Of course it is *possible* I just think it isn't very intuitive.

PR #4439, opened 2021-01-07
Fix FactorForms for CFA

boutinb:
The FactorForms did not work anymore after
https://github.com/jasp-stats/jasp-desktop/pull/4413
Also the dropKeys property of VariablesList is handled properly in c++
and is now dynamic.

The CFA changes are done in https://github.com/jasp-stats/jaspFactor/pull/6

### DIFF
--- a/Desktop/analysis/analysisform.h
+++ b/Desktop/analysis/analysisform.h
@@ -94,16 +94,16 @@ signals:
 				void			analysisChanged();
 
 protected:
-				QVariant		requestInfo(const Term &term, VariableInfo::InfoType info) const override;
+				QVariant		requestInfo(const Term &term, VariableInfo::InfoType info)	const override;
 
 public:
-	ListModel			*	getModel(const QString& modelName)								{ return _modelMap.count(modelName) > 0 ? _modelMap[modelName] : nullptr;	} // Maps create elements if they do not exist yet
-	void					addModel(ListModel* model)										{ if (!model->name().isEmpty())	_modelMap[model->name()] = model;			}
-	Options				*	getAnalysisOptions()											{ return _analysis->options();												}
-	JASPControl			*	getControl(const QString& name)									{ return _controls.contains(name) ? _controls[name] : nullptr;				}
+	ListModel			*	getModel(const QString& modelName)								const	{ return _modelMap.value(modelName);	} // Maps create elements if they do not exist yet
+	void					addModel(ListModel* model)												{ if (!model->name().isEmpty())	_modelMap[model->name()] = model;			}
+	Options				*	getAnalysisOptions()													{ return _analysis->options();												}
+	JASPControl			*	getControl(const QString& name)											{ return _controls.contains(name) ? _controls[name] : nullptr;				}
 	void					addListView(JASPListControl* listView, JASPListControl* sourceListView);
-	ExpanderButtonBase	*	nextExpander(ExpanderButtonBase* expander)						{ return _nextExpanderMap[expander];										}
-	JASPControl			*	getControl(Option* option)										{ return _optionControlMap[option];											}
+	ExpanderButtonBase	*	nextExpander(ExpanderButtonBase* expander)								{ return _nextExpanderMap[expander];										}
+	JASPControl			*	getControl(Option* option)												{ return _optionControlMap[option];											}
 
 	Options				*	options() { return _options; }
 	void					addControl(JASPControl* control);
--- a/Desktop/analysis/analysisform.h
+++ b/Desktop/analysis/analysisform.h
@@ -97,13 +97,13 @@ protected:
 				QVariant		requestInfo(const Term &term, VariableInfo::InfoType info) const override;
 
 public:
-	ListModel			*	getRelatedModel(JASPListControl* listView)	{ return _relatedModelMap[listView]; }
-	ListModel			*	getModel(const QString& modelName)		{ return _modelMap.count(modelName) > 0 ? _modelMap[modelName] : nullptr; } // Maps create elements if they do not exist yet
-	Options				*	getAnalysisOptions()					{ return _analysis->options(); }
-	JASPControl			*	getControl(const QString& name)			{ return _controls.contains(name) ? _controls[name] : nullptr; }
+	ListModel			*	getModel(const QString& modelName)								{ return _modelMap.count(modelName) > 0 ? _modelMap[modelName] : nullptr;	} // Maps create elements if they do not exist yet
+	void					addModel(ListModel* model)										{ if (!model->name().isEmpty())	_modelMap[model->name()] = model;			}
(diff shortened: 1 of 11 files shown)


Desktop/analysis/analysisform.h
```diff
@@ -97,13 +97,13 @@ public slots:
 				QVariant		requestInfo(const Term &term, VariableInfo::InfoType info) const override;
 
 public:
-	ListModel			*	getRelatedModel(JASPListControl* listView)	{ return _relatedModelMap[listView]; }
-	ListModel			*	getModel(const QString& modelName)		{ return _modelMap.count(modelName) > 0 ? _modelMap[modelName] : nullptr; } // Maps create elements if they do not exist yet
-	Options				*	getAnalysisOptions()					{ return _analysis->options(); }
-	JASPControl			*	getControl(const QString& name)			{ return _controls.contains(name) ? _controls[name] : nullptr; }
+	ListModel			*	getModel(const QString& modelName)								{ return _modelMap.count(modelName) > 0 ? _modelMap[modelName] : nullptr;	} // Maps create elements if they do not exist yet
```
JorisGoosen: Use `.at(key)` instead of `[key]` to make sure it never creates an empty entry.

JorisGoosen: Then you can even make the function `const`

JorisGoosen: Aah and this isn't even new code. Well, I wont request this as a change then ;)

Desktop/widgets/factorsformbase.cpp
```diff
@@ -47,9 +48,7 @@ void FactorsFormBase::setUpModel()
 	_initNumberFactors = property("initNumberFactors").toInt();
 
 	QQuickItem::connect(this, SIGNAL(titleChanged(int, QString)), _factorsModel, SLOT(titleChangedSlot(int, QString)));
-	QQuickItem::connect(this, SIGNAL(factorAdded(int, QVariant)), _factorsModel, SLOT(factorAddedSlot(int, QVariant)));
-
-	connect(_factorsModel, &ListModelFactorsForm::addListView, this, &FactorsFormBase::addListViewSlot);
+	QQuickItem::connect(this, SIGNAL(factorAdded(int, QVariant)), this, SLOT(factorsAddedSlot(int, QVariant)));
```
JorisGoosen: Why again with the MACROs? Can't you refer to the actual function-definitions like we do everywhere?

Desktop/widgets/factorsformbase.cpp
```diff
@@ -133,19 +132,16 @@ void FactorsFormBase::termsChangedHandler()
...
-	const vector<tuple<string, string, vector<string> > > &factors = _factorsModel->getFactors();
-	QStringList names;
-	for (auto factor : factors)
-		names.push_back(tq(get<0>(factor)));
-	_availableVariablesListItem->setProperty("dropKeys", names);
-	setProperty("dropKeys", _availableVariablesListName);
+	VariablesListBase* listView = qobject_cast<VariablesListBase *>(item.value<QObject *>());
+	if (!listView)
+	{
+		Log::log() << "JASP Control is not a VariablesListBase in factorAdded" << std::endl;
```
JorisGoosen: Maybe also mention the `name` of the control in this (and these kind of) message(s). That way the log is slightly *more* useful.

Desktop/widgets/listmodelfactorsform.cpp
```diff
@@ -98,20 +96,28 @@ void ListModelFactorsForm::initFactors(const vector<tuple<string, string, vector
...
-	for (int i = 0; i < _factors.length(); i++)
+	for (Factor* factor : _factors)
 	{
-		Factor* factor = _factors[i];
 		JASPListControl* listView = factor->listView;
 		if (listView)
-			result.push_back(make_tuple(factor->name.toStdString(), factor->title.toStdString(), listView->model()->terms().asVector()));
+		{
+			Terms terms = listView->model()->terms();
+			result.push_back(make_tuple(factor->name.toStdString(), factor->title.toStdString(), terms.asVector()));
```
JorisGoosen: Not `fq()` ?

Desktop/widgets/listmodelfactorsform.h
```diff
@@ -43,14 +44,15 @@ class ListModelFactorsForm : public ListModel
 	Terms					termsEx(const QString& what)									override;
 	void					initFactors(const std::vector<std::tuple<std::string, std::string, std::vector<std::string> > > &factors);
```
JorisGoosen: I think a typedef for factors vector is in order:
```
typedef FactorVec std::vector<std::tuple<std::string, std::string, std::vector<std::string> > >;
```
Maybe also one for the elements?
Because this is pretty much unreadable.
Maybe a struct is even better, not sure about that though.

JorisGoosen (changes requested):

Desktop/widgets/factorsformbase.cpp
boutinb: factorAdded & titleChanged are QML signals

Desktop/widgets/factorsformbase.cpp
```diff
@@ -112,7 +112,7 @@ bool FactorsFormBase::isJsonValid(const Json::Value &optionValue)
 
 void FactorsFormBase::termsChangedHandler()
 {
-	const vector<tuple<string, string, vector<string> > > &factors = _factorsModel->getFactors();
+	const ListModelFactorsForm::FoctorVec &factors = _factorsModel->getFactors();
```
JorisGoosen: FoctorVec?

Try that one again ;)

Desktop/widgets/listmodelfactorsform.h
```diff
@@ -44,7 +45,7 @@ class ListModelFactorsForm : public ListModel
...
```
JorisGoosen: You can now use `FoctorVec` in here right? Or maybe `FactorVec` is better.

Desktop/widgets/factorsformbase.cpp
JorisGoosen: But it still comes from `factorsformbase` right? You are referring to `this` after all.
You could just make a `Q_INVOKABLE void factorAdded(int QVariant);` signal in `factorsformbase.h` .
The same for `titleChanged`.

That way the signal definition is in C++ and you can use a normal connect that does typechecks etc.

JorisGoosen:
I've tested this with the code https://github.com/jasp-stats/jaspFactor/pull/6 and everything seems to work, so this PR is almost done.

I really think the following changes should still be made though:
-  `FoctorsVec` should be renamed. 
- Those signals should be defined Q_INVOKABLE and in C++. With the connection changed to new-style
- `initFactors` should use the renamed `FactorsVec`

After that both PRs can be merged as far as Im concerned!

boutinb:
OK, I have added your 3 points.

PR #4445, opened 2021-01-15
Make translations work for modules

FransMeerhoff:
- Some code changes in LanguageModel and DynamicModule needed.
- Project files changes: make translation files for each single module.

### DIFF
--- a/Desktop/modules/dynamicmodule.cpp
+++ b/Desktop/modules/dynamicmodule.cpp
@@ -131,9 +131,13 @@ void DynamicModule::initialize()
 
 	//Ok everything seems to be in order, let's load!
 	QString qmlTxt = "", jsonTxt = "";
+	
+	QUrl url;
 	try
 	{
 		QFileInfo descriptionInfo = checkForExistence(getQmlDescriptionFilename(), true);
+		
+		url = QUrl::fromLocalFile(descriptionInfo.absoluteFilePath());
 
 		QFile descriptionFile(descriptionInfo.absoluteFilePath());
 		descriptionFile.open(QFile::ReadOnly);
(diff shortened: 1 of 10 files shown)


Desktop/modules/dynamicmodule.cpp
```diff
@@ -272,9 +281,10 @@ Description * DynamicModule::instantiateDescriptionQml(QString descriptionTxt, Q
 	return description;
 }
 
-void DynamicModule::loadDescriptionQml(QString descriptionTxt)
+
+void DynamicModule::loadDescriptionQml(QString descriptionTxt, QUrl & url)
```
boutinb: You could also set the description as const QString&

Desktop/modules/dynamicmodules.cpp
```diff
@@ -111,21 +111,24 @@ bool DynamicModules::initializeModule(Modules::DynamicModule * module)
...
+			emit loadModuleTranslationFile(module);
+			
 			module->initialize();
 		}
 
 		if(!wasAddedAlready)
 		{
 			emit dynamicModuleAdded(module);
-			emit loadModuleTranslationFile(module);
+			//emit loadModuleTranslationFile(module);
```
boutinb: Why is it commented out? If it is really not necessary, just remove it.

boutinb (changes requested):
Don't forget that we have also a chinese translation.

Desktop/modules/dynamicmodules.cpp
FransMeerhoff: The emit was moved to the  ! initialized but was unnecessary. I've put it back

boutinb: So suppress this line of code. I don't like to see commented out codes when it's not explained

JorisGoosen: Double that ;)

PR #4449, opened 2021-01-20
General updates to language changing in JASP

### LINKED ISSUE #1088: Translation of headers not shown after language switch
* JASP version: 0.14
* OS name and version: macOS Big Sur 11.0.1
* Analysis: any
* Bug description: In the results pane, the translation of the headers is not shown after switching the interface language
* Expected behaviour: Translated headers should be shown after switching the interface language

1. Open a data set.
2. Open 'Descriptives':
before
3. Switch the interface language.
4. The header is actually translated, but the translation is not shown:
after
5. You can see that this is a display issue, not a translation issue, when you switch the interface language again.

[test.csv.zip](https://github.com/jasp-stats/jasp-issues/files/5611897/test.csv.zip)

JorisGoosen:
- Also fix some whitespace
- Fixes https://github.com/jasp-stats/jasp-issues/issues/1088 (analysis results headers dont change on language change)
- Javascript should use just the proper title from Analysis directly instead of the one From R. So jaspResults does not need to send it either.
- pause and resume engines for languagechange to avoid running stuff while it is still going on
- Also added a queue for javascript commands to the results, this because the commands are otherwise silently ignored during the small window of time where the webeengine is reloading during languagechange.
- And removed some legacy JS code, pretty sure this doesn't affect anything but would appreciate if Tim glances at it suavely.

### DIFF
--- a/Desktop/html/js/analysis.js
+++ b/Desktop/html/js/analysis.js
@@ -3,6 +3,7 @@ JASPWidgets.Analysis = Backbone.Model.extend({
 		id: -1,
 		progress: -1,
 		results: {},
+		title: 'Analysis Title',
 		status: 'waiting',
 		optionschanged: [],
 		saveimage: [],
@@ -514,17 +515,10 @@ JASPWidgets.AnalysisView = JASPWidgets.View.extend({
 
 			if (!_.has(results, name))
 				continue
+
 			let data = results[name];
-
-			/*if (meta.type == 'collection' && data.title == "") {  // remove collections without a title from view
-				let collectionMeta = meta.meta;
-				if (Array.isArray(collectionMeta)) { // the meta comes from a jaspResult analysis
-					this.createResultsViewFromMeta(data["collection"], collectionMeta, $result);
-					continue;
-				}
-			}*/
-
 			let itemView = this.createChild(data, this.model.get("status"), meta);
+			
 			if (itemView === null)
 				continue;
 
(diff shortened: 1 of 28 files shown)


JorisGoosen:
Pfoewey, the problems here were quite a bit worse and hard to untangle than I expected when I started.
But I think Ive got it all straightened out now.

To make life easier for you poor reviewers I've turned on some nightlies that are building as I type.

Tim, I added you mostly for the javascript to check that it was ok for me to remove that stuff. Im pretty sure it could do with a bigger cleanup though. Some other day.

Frans, I can take you through the changes and why I made them tomorrow if you like?

Hi Bruno!

JorisGoosen:
I think maybe the progressbar in the results  might need some JS love though.

Desktop/html/js/analysis.js
```diff
@@ -632,15 +593,8 @@ JASPWidgets.AnalysisView = JASPWidgets.View.extend({
...
 
-		var results = this.model.get("results");
-
-		// once everything becomes jaspResults this is always an object and the following can be removed
-		var progress = this.model.get("progress")
-		if (typeof progress == "number")
-			this.model.set("progress", { value: progress, label: "" })
-		else if (!progress)
-			this.model.set("progress", { value: -1, label: "" })
-		// up to here
```
TimKDJ, line 643: I recall from a previous attempt to remove this, that `progress` is not always an object even now with jaspResults. So I'd keep this in here

JorisGoosen: Or I can push a fix to jaspResults to always make it an object

JorisGoosen: I think, if it isn't an object it will be `null` that should be fairly easy to check for.

TimKDJ: We at least have to be mindful of it being null in:
https://github.com/jasp-stats/jasp-desktop/blob/0db025f5aa07b9a192599c60c407f1ec23fd7b74/Desktop/analysis/analysis.h#L80
and in the engine.cpp there seem to be a few `_progress					= -1;` hardcoded, although it could well be that the analysis render is never called in that case.

JorisGoosen: Well, I didn't see those, but maybe I didnt look hard enough. I can just change those "-1" to null and then check for it in js.

JorisGoosen: So I think the new commit should fix the progressbar blemishes.

Desktop/html/js/analysis.js
```diff
@@ -596,9 +596,10 @@ JASPWidgets.AnalysisView = JASPWidgets.View.extend({
...
 
-		if (results == "" || results == null) {
+		if (results === "" || results === null) {
 			progress = this.model.get("progress");
-			if (progress.value > -1)
+
+			if (progress !== null && progress.value > -1)
 				this.updateProgressbarInResults();
 			return this;
 		}
```
TimKDJ, line 605: If you want you can delete these lines, as they won't be used anymore. They belong to the old style analysis only sending progress updates without the results being included. I don't think this is possible anymore with jaspResults

TimKDJ:
So I've done some testing and the language changing seems to work ok, so that's great. I did notice that on a few occasions the ribbon got stuck when messing around in analyses -- sometimes in combination with an analysis itself getting stuck. Not sure if this problem is introduced by the PR though. Will try to get a reproduction path, but messing around with the Bayesian ANOVA and "Model averaged posteriors" seemed to do the trick a few times.
ribbons-stuck

JorisGoosen:
I'll have a look at the ribbon-loading problem.

JorisGoosen:
I think now it won't break anymore and it should all work a bit better.

Desktop/html/js/analysis.js
JorisGoosen: Yeah but https://github.com/jasp-stats/INTERNAL-jasp/issues/1159 could make use of that really well.
It should also avoid some of the slowness thats been observed in the webengine during progressbar phase (because it keeps rerendering the results despite them not changing).

boutinb (approved):

SamanthaS1:
Hi there, I was just wondering how I would be able to find the missing values in the descriptive statistics chart in jasp. So in other words, the problem I've come across is that it's not giving me the options for descriptive statistics in order to analyze values such as median, mode, variance and range. So my question is, how do I add the missing values without the option of adding them? Am I supposed to request this option or some kind of update? The main issue is that it's just not letting me scroll down to where I can check the values that I need for the analysis. If someone could get back to me on this I'd appreciate it,Thanks.

JorisGoosen:
Hello @SamanthaS1 this is not quite the right place to ask these kind of questions, I suggest you open an issue via https://github.com/jasp-stats/jasp-issues/issues/new/choose or ask on the [cogsci forum](https://forum.cogsci.nl/categories/jasp-bayesfactor)

PR #4450, opened 2021-01-21
Plot Editing: Simple & Advanced mode + Undo & Redo

vandenman:
Fixes most of https://github.com/jasp-stats/INTERNAL-jasp/issues/1191
Requires: https://github.com/jasp-stats/jaspGraphs/pull/10

Some things that are not fixed:
- All plots that can't be edited should have a disabled menu entry with a tooltip explaining why it can't be edited.
- Duplicate analysis doesn't copy the state and causes errors.
- Tooltips don't work when a QML thingy is disabled (see the TextField example).

Aside from that, discrete axes needs some work, but that might require a new QML component and/ or different approach.


Desktop/components/JASP/Widgets/PlotEditor/PlotEditor.qml
```diff
@@ -81,61 +81,68 @@ Popup
...
+					
+						PlotEditingAxis
 						{
-							id:				axisSeparator
-							height:			1
-							color:			jaspTheme.uiBorder
+							id:				xAxis
+							title:			qsTr("x-axis")
+							axisModel:		plotEditorModel.xAxis
+							visible:		axisDropDown.value === "x-axis"
```
JorisGoosen: It would be better to add an Enum for the axis and select it based on that.

JorisGoosen: To make the selected axis better available here you could add a property to `PlotEditorModel` called `selectedAxis` and have that be a reference to the `PlotEditorAxisModel*` that is current.

But maybe it is better if I implement that.

Desktop/results/ploteditoraxismodel.cpp
```diff
@@ -399,20 +466,22 @@ std::string		AxisModel::LimitsTypeToString(LimitsType type) const
...
-AxisModel::BreaksType	AxisModel::BreaksTypeFromString(std::string type) const
+AxisModel::BreaksType	AxisModel::BreaksTypeFromString(const std::string& type) const
 {
 	if(type == "manual"	)	return BreaksType::BreaksManual;
+	if(type == "NULL")		return BreaksType::BreaksNull;
 							return BreaksType::BreaksRange; //default
 }
 
-AxisModel::LimitsType	AxisModel::LimitsTypeFromString(std::string type) const
+AxisModel::LimitsType	AxisModel::LimitsTypeFromString(const std::string& type) const
```
JorisGoosen, line 484: Ah yeah, oops

Desktop/results/ploteditormodel.cpp
```diff
@@ -134,6 +143,86 @@ void PlotEditorModel::somethingChanged()
...
+void PlotEditorModel::setUndoEnabled(bool undoEnabled)
+{
+	if (_undoEnabled == undoEnabled)
+		return;
+
+	_undoEnabled = undoEnabled;
+	emit undoEnabledChanged(_undoEnabled);
+}
+
+void PlotEditorModel::setRedoEnabled(bool redoEnabled)
```
JorisGoosen: This whole function isn't necessary and I'll show you later how it can be done more efficiently.

JorisGoosen: (And avoiding double administration, because redo-enabled == redostacksize > 0 )

JorisGoosen (changes requested):
Alright, overall this looks good.
Two things:
- [x] You forgot to add undo/redo buttons to the darkTheme... 
- [x] If you have the "Manual" breaks selected in simple mode and you switch to advanced it selects something else.

Now, I can have a look at that last issue and also clean up the axisselection.
I would like you to add those icons though.

JorisGoosen:
I'll change the axis-enum and the redo enabled later if you like.

JorisGoosen:
(Of course if you do not like I can help you figure how to do those changes)

Desktop/mainwindow.cpp
```diff
@@ -168,7 +168,8 @@ MainWindow::MainWindow(QApplication * application) : QObject(application), _appl
 	qmlRegisterType<JASPDoubleValidator>						("JASP",		1, 0, "JASPDoubleValidator"				);
 	qmlRegisterType<ResultsJsInterface>							("JASP",		1, 0, "ResultsJsInterface"				);
 
-	qmlRegisterUncreatableType<PlotEditor::AxisModel>			("JASP.PlotEditor",		1, 0, "AxisModel",		"Can't make it");
+	qmlRegisterUncreatableType<PlotEditor::AxisModel>			("JASP.PlotEditorAxis",		1, 0, "AxisModel",			"Can't make it");
```
JorisGoosen: Isn't it more logical to put both of these QML items inside the same group called "JASP.PlotEditor" ?

That way you can just do `import JASP.PlotEditor 1.0` and then use `AxisModel` and `PlotEditorModel` directly.
The way you've done it now is rather confusing.

vandenman: that somehow didn't work before, but I'll try again :)

vandenman: and now it does work ...

Desktop/components/JASP/Widgets/PlotEditor/PlotEditor.qml
vandenman: I changed it, let me know if you have any further comments!

JorisGoosen (approved):
Ok, this is starting to look good.
I will be merging it so other people can play around with it.

But there are some improvements we can (and should) still make.

PR #4451, opened 2021-01-21
Single / Double click for select / edit

JorisGoosen:
Instead of the selection column.

If this is merged I can build a new flatpak version and fix the variableswindow issue there.


boutinb (approved):
This is only a temporary fix so that Linux people can change the labels.

PR #4455, opened 2021-01-27
Rewritten variables window

JorisGoosen:
Taken from stable and modified slightly to work with Ctrl+Click for selecting multiple rows. 
And the normal click for a single row.

Shift+click was a nightmare. Ive left it out to get on with my life.

### DIFF
--- a/Desktop/qquick/datasetview.cpp
+++ b/Desktop/qquick/datasetview.cpp
@@ -8,6 +8,7 @@
 #include "log.h"
 #include "gui/preferencesmodel.h"
 #include "qquick/jasptheme.h"
+#include <QScreen>
 
 DataSetView * DataSetView::_lastInstancedDataSetView = nullptr;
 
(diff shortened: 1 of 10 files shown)


Desktop/qquick/datasetview.cpp
```diff
@@ -832,10 +859,11 @@ QQmlContext * DataSetView::setStyleDataItem(QQmlContext * previousContext, bool
 	previousContext->setContextProperty("itemText",			text);
 	previousContext->setContextProperty("itemActive",		active);
 	previousContext->setContextProperty("itemEditable",		isEditable);
-	previousContext->setContextProperty("itemInputType",	itemInputType);
+	previousContext->setContextProperty("itemSelected",		_model->data(idx, _roleNameToRole["selected"]));
+	previousContext->setContextProperty("itemInputType",	_model->data(idx, _roleNameToRole["itemInputType"]));
 	previousContext->setContextProperty("columnIndex",		static_cast<int>(col));
 	previousContext->setContextProperty("rowIndex",			static_cast<int>(row));
-	previousContext->setContextProperty("isDynamic",		true);
```
boutinb, line 838: Why did you change isDynamic to hasContextForm? The TableView controls need this context.

boutinb (changes requested):
I still don't feel that it is really user-friendly. 
If we follow the Excel way, then single-click should select the row and you can already write into the cell. Double-Click should unselect the row, and you can only write into the cell.
Is that doable?

JorisGoosen:
> I still don't feel that it is really user-friendly.
> If we follow the Excel way, then single-click should select the row and you can already write into the cell. Double-Click should unselect the row, and you can only write into the cell.
> Is that doable?

I also don't feel this is userfriendly, but to get it working as you describe I would have to move all the selectioncode to `datasetview` which would require me to reselect everything there somehow after moving (now I just use a set of values).

I am getting quite sick of working on this to be honest. The selection column I added before was simple, worked fine and was easy to understand.
Now we have clicking that selects or not, doubleclicking and who knows what. Im going to simplify it. 

I'll just make it so that clicking on the value or around the textinput selects and clicking on the textinput will allow one to edit the text. Excel can go out of the window.

Desktop/qquick/datasetview.cpp
JorisGoosen: Ah that is a copy mistake!
But I also think it is a bit weird to have a property here that is always the exact same value... What is the point then?
But I'll fix it.

JorisGoosen: Also, the property "tableView" already tells your tableview it is a tableview right?
But still.. Ill fix it.

boutinb:
PR https://github.com/jasp-stats/jasp-desktop/pull/4463 takes it over.

PR #4456, opened 2021-01-27
Add a scrollbar for menu

boutinb:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/1214

### DIFF
--- a/Desktop/components/JASP/Widgets/CustomMenu.qml
+++ b/Desktop/components/JASP/Widgets/CustomMenu.qml
@@ -100,6 +100,8 @@ FocusScope
 		MouseArea
 		{
 			anchors.fill	: parent
+			acceptedButtons	: Qt.NoButton
+			onWheel			: wheel.accepted = true
 		}
 
 		JASPControl.JASPScrollBar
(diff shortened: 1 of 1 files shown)


Desktop/components/JASP/Widgets/CustomMenu.qml
```diff
@@ -94,170 +94,203 @@ FocusScope
...
+			}
+		}
 
-				menuRectangle.width   = Math.max(item.width, menuRectangle.width);
-				menuRectangle.height += (item.height + jaspTheme.menuSpacing)
+		Flickable
+		{
+			id						: itemFlickable
+			anchors.fill			: parent
+			anchors.leftMargin		: jaspTheme.menuPadding / 2
```
JorisGoosen, line 129: You should add a topMargin, because it looks a bit weird now.
See:
image

JorisGoosen (changes requested):
Besides the topMargin you should also add a MouseArea behind the flickable (aka covering the root of the menu).
This mousearea should have `acceptedButtons: Qt.NoButton` and `onWheel: wheel.accepted = true;`.

That way if you reach the top or bottom of the menu it won't start scrolling the dataview (or analyses or whatever) thats underneath the menu. Because I really don't think that is supposed to happen.

JorisGoosen (approved):

PR #4457, opened 2021-02-03
No Options

boutinb:
Remove the Option classes and use directly Json in JASP Controls.
The BoundControl classes are extended to take care about the logic that the Options classes had. The main goal of the BoundControl classes is to take care that an json input sets the right values to the controls or models, and that when some option is changed, a new json string is automatically generated.
The boost signal/connect library is not needed anymore.


Desktop/widgets/boundcontrolfilteredtableview.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+	result.append(row);
+	return result;
+}
+
+void BoundControlFilteredTableView::fillTableTerms(const Json::Value &value, ListModelTableViewBase::TableTerms &tableTerms)
+{
+	if (value.size() > 0)
+	{
+		const Json::Value& firstRow = value[size_t(0)];
```
JorisGoosen: `value[size_t(0)]` is apparently ambiguous on linux/gcc so should be changed to `value[Json::UInt(0)]`

Desktop/widgets/boundcontrolfilteredtableview.cpp
```diff
@@ -0,0 +1,102 @@
...
+		tableTerms.colName	= tq(firstRow["colName"].asString());
+
+		for (const Json::Value& value : firstRow["dataCols"])
+			tableTerms.colNames.push_back(tq(value.asString()));
+
+		const Json::Value& extraCol = firstRow["extraCol"];
+		if (extraCol.isString())
+			tableTerms.extraCol = tq(extraCol.asString());
+		else if (extraCol.isArray() && extraCol.size() > 0)
+			tableTerms.extraCol = tq(extraCol[size_t(0)].asString());
```
JorisGoosen: Same here: `value[size_t(0)]` is apparently ambiguous on linux/gcc so should be changed to `value[Json::UInt(0)]`

JorisGoosen (changes requested):

JorisGoosen:
Ok running the data-library-unittests seems to indicate that all those options mostly load fine. I did have to make those two requested changes to be able to compile on linux so it would be good to make them. After that we can merge this!

The were a bunch of errors in Audit and Bain btw, but I think those arent related to this PR. @koenderks might want to have a look at those though.

JorisGoosen (approved):

PR #4461, opened 2021-02-11
Show a warning why a columntypechange fails

JorisGoosen:
- fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/977

Maybe Bruno can check if the c++ looks ok and Tim can give his verdict on the userfriendliness of the warningmsgs?

I've also asked @EJWagenmakers and he might chime in with good messages as well.

To make it easier to figure out which warnings will be shown, see:
- For a fail from scalar to ordinal: `Could not convert all values to integer because they have decimal values.`
- For a fail from nominalText to scalar: `Could not convert all values to scalar because they contained text.`
- For a fail from nominalText to Ordinal: `Could not convert all values to integer because they contained text.` <- This one could also be from double converted to text which might be slightly confusing but well..

### DIFF
--- a/Common/column.cpp
+++ b/Common/column.cpp
@@ -452,12 +452,12 @@ columnType Column::getColumnType() const
 	return _columnType;
 }
 
-bool Column::_changeColumnToNominalOrOrdinal(enum columnType newColumnType)
+columnTypeChangeResult Column::_changeColumnToNominalOrOrdinal(enum columnType newColumnType)
 {
 	if (_columnType == columnType::nominal || _columnType == columnType::ordinal)
 	{
 		_columnType = newColumnType;
-		return true;
+		return columnTypeChangeResult::changed;
 	}
 
 	if (_columnType == columnType::nominalText)
@@ -494,12 +494,12 @@ bool Column::_changeColumnToNominalOrOrdinal(enum columnType newColumnType)
 		{
 			_labels.clear();
 			setColumnAsNominalOrOrdinal(values, intLabels, newColumnType == columnType::ordinal);
-			return true;
+			return columnTypeChangeResult::changed;
 		}
 
 		// nominalText to nominal: we could not make the values as integers, but
 		// the column can still stay a NominalText, so it is not a failure.
-		return newColumnType == columnType::nominal;
+		return newColumnType == columnType::nominal ? columnTypeChangeResult::changed : columnTypeChangeResult::cannotConvertStringValueToInteger;
 	}
 	else if (_columnType == columnType::scale)
 	{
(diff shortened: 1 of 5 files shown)


boutinb (approved):

TimKDJ:
Ah, giving the user feedback is fantastic. I've gotten way too many emails from users who could not figure out why a conversion didn't work. On that note though, could we be more specific? If a user has 1000 rows and one cell contains something it shouldn't, then finding it is still a bit tricky.

I could think of two ways to make it more specific:
1. Ideally, we'd add a little filter button in the message that, when people press it, crosses out all values except for the ones preventing the conversion. I appreciate this might be a lot of work though.
2. We could list the row numbers of values that are preventing the conversion, granted that it's less than ~ 5 of them, if it's more than that we could just specify the total number of values that are preventing the conversion.

EJWagenmakers:
A tooltip could also be useful perhaps

JorisGoosen:
> 
> 
> Ah, giving the user feedback is fantastic. I've gotten way too many emails from users who could not figure out why a conversion didn't work. On that note though, could we be more specific? If a user has 1000 rows and one cell contains something it shouldn't, then finding it is still a bit tricky.
> 
> I could think of two ways to make it more specific:
> 
>     1. Ideally, we'd add a little filter button in the message that, when people press it, crosses out all values except for the ones preventing the conversion. I appreciate this might be a lot of work though.
> 
>     2. We could list the row numbers of values that are preventing the conversion, granted that it's less than ~ 5 of them, if it's more than that we could just specify the total number of values that are preventing the conversion.

All good points, just like the tooltip suggestion!

But all of these require writing an extra function that checks everything, which is more work.
I've now just went for "Lets get as many issues as possible sort-of fixed" and this one should at least give people *some* feedback on what went wrong. 
We could leave the issue open with the notion that more feedback is warranted.

But I think having this is already a great improvement as people will stop opening issues with "type changing doesnt work!".
And solving this issue properly might actually be a good test-issue for a new c++ programmer.

I'll move the extra requests to the issue.

JorisGoosen:
I'd prefer to spend time on more pressing issues is my general reason to not make this more complicated.

TimKDJ:
Fair enough. 

Then just some suggestions for the sentences; I tried to dumb it down a little and use terms from https://jasp-stats.org/getting-started:
```
Failed to convert column to ordinal: values contain decimals.
Failed to convert column to continuous: values contain text.
Failed to convert column to ordinal: values contain text.
```
Not sure if they're better, but it's never bad to have some alternative options!

Edit: included EJ's suggestion.

EJWagenmakers:
The "not" should be omitted, right?

JorisGoosen:
I'll change em on monday, they do sound better!

PR #4463, opened 2021-02-15
New variables window

boutinb:
Following https://github.com/jasp-stats/jasp-desktop/pull/4455

I have removed the cache of DataSetView because it mixed up the items.
Also I have made the LabelModel a one-column list, and one row build the Filter button, the value text element, and the Input field for the label.
I have also removed the begin/endResetModel calls when a selection is made: it also mixed up the items, disturbing the focus of the Input field. 
I have also added support for CTRL and SHIFT clicks.

### DIFF
--- a/Desktop/components/JASP/Widgets/JASPDataView.qml
+++ b/Desktop/components/JASP/Widgets/JASPDataView.qml
@@ -10,6 +10,10 @@ FocusScope
 	id: __JASPDataViewRoot
 
 				property alias model:					theView.model
+				property alias toolTip:					datasetMouseArea.toolTipText
+				property alias cursorShape:				datasetMouseArea.cursorShape
+				property alias mouseArea:				datasetMouseArea
+				property bool  doubleClickWorkaround:	true
 
 				property alias itemDelegate:			theView.itemDelegate
 				property alias rowNumberDelegate:		theView.rowNumberDelegate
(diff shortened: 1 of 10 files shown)


JorisGoosen (changes requested):
Shift + click doesn't work and Ctrl + click should toggle, not just turn on.

boutinb:
> Shift + click doesn't work and Ctrl + click should toggle, not just turn on.

Shift + click works for me. On which system did you try it?
Ctri + click does not toggle either on Excel.... Only simple click should toggle.

JorisGoosen:
Ok I am not an excel user but every software Ive used uses Ctrl+click to toggle selection. But ok, whatever.

Shift-click should allow you to drag the mouse and select a number of elements right?
Because if I do that I just drag the variableswindow.

Are you going to tell me that excel does not work like that either?
Because if not what is the point of having both ctrl+click and shift+click?

boutinb:
> Ok I am not an excel user but every software Ive used uses Ctrl+click to toggle selection. But ok, whatever.
> 
> Shift-click should allow you to drag the mouse and select a number of elements right?
> Because if I do that I just drag the variableswindow.
> 
> Are you going to tell me that excel does not work like that either?
> Because if not what is the point of having both ctrl+click and shift+click?

Shift-click is to select at once several rows by clicking the first row with a simple click and the last row with a shift-click... It's how it works on Excel... and it's also what I expect to do.

JorisGoosen (approved):
Ok, if the behaviour you described is in fact what excel does then this PR does what it needs to do.

I still think this is a confusing and weird way of selecting cells though and will not hesitate to mention this at any good opportunity.

PR #4464, opened 2021-02-15
Add possibility to have a source for SimpleTableView

boutinb:
The source feature was set in the specialized MultitonimialChi2Test TableView. This possibility should be added for all TableViews.


JorisGoosen:
Ive tested if tableviews still work after this change, they seem to.

Beyond that I guess if the chi2 still works it means this works ^^

JorisGoosen (approved):

PR #4470, opened 2021-02-18
Retrieve information of old plots from ghost results

### LINKED ISSUE #824: Network plot resizing back to default size when changing it's options.
* JASP version: 0.12.2

* Analysis: A minor annoyance, but I feel it's worth bringing it up.
* Bug description:  When in network analysis and you are using the network plot, making changes to the size of the network plot will not stay if you change it's options as it will output in the default size again.
* Expected behaviour: Be able to change the network plot options while still keeping the same size.

vandenman:
Fixes https://github.com/jasp-stats/jasp-issues/issues/824
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/1259
Fixes a part of https://github.com/jasp-stats/INTERNAL-jasp/issues/1257

I'm not entirely sure whether we should also reapply plot editing. If we do want that, I'll add it in a separate PR as it will require a bit more work that should probably go into jaspGraphs.

depends on https://github.com/jasp-stats/jaspBase/pull/17


R-Interface/jaspResults/src/jaspContainer.cpp
```diff
@@ -428,3 +475,25 @@ void jaspContainer::setError(std::string message)
...
 }
+
+void jaspContainer::renderPlotsOfChildren()
+{
+	for(auto & child : _data)
+		switch (child.second->getType())
+		{
+		case jaspObjectType::container:
+			static_cast<jaspContainer*>(child.second)->renderPlotsOfChildren();
+			break;
```
JorisGoosen, line 486: A newline underneath each `break;` would have been nice, to let it breathe a little.

R-Interface/jaspResults/src/jaspObject.cpp
```diff
@@ -415,6 +427,25 @@ void jaspObject::setDeveloperMode(bool developerMode)
 	_developerMode = developerMode;
 }
 
+bool jaspObject::connectedToJaspResults()
+{
+
+	// we could add a private bool to jaspObject to shortcircuit this recursive chain
```
JorisGoosen: And then having to update that and all that?
And waste a whole byte for what is probably a few calls up (who would be nesting jaspObjects so deep that it would ever exceed like 5 or so?)

Embrace the recursion ;)

R-Interface/jaspResults/src/jaspPlot.cpp
```diff
@@ -57,17 +57,55 @@ void jaspPlot::initEnvName()
...
+	if (plotInfoObj.isNULL())
+		return;
+
+	Rcpp::List plotInfo = Rcpp::as<Rcpp::List>(plotInfoObj);
+	Rcpp::RObject obj = plotInfo["obj"];
 
 	if(!obj.isNULL())
 	{
+
+		jaspPrint("Now rendering a plot!");
```
JorisGoosen: It might be helpful to add the name or title to this debugstatement, that way if in the future we get logs from a user where the plotrendering failed we can pinpoint in the logs which plot is breaking it.
Because we can no longer assume the order in which the functions are called are the same order the plots are rendered.

JorisGoosen: It certainly *is* helpful to havve a printstatement here btw!

JorisGoosen (changes requested):
A small change to have a more helpful log later on.

PR #4476, opened 2021-02-23
Add the possibility to set initial values in the FilteredDataEntryModel TableView

boutinb:
With the property initialValuesSource, you can now set which column(s)
should give the initial values of the table.
To test it, add in AuditClassicalWorkflow or AuditBayesianWorkflow
`initialValuesSource:				"monetaryVariable"`
to the performAudit TableView.

### DIFF
--- a/Desktop/widgets/boundcontroltableview.cpp
+++ b/Desktop/widgets/boundcontroltableview.cpp
@@ -136,11 +136,11 @@ Json::Value BoundControlTableView::_defaultValue()
 	JASPControl::ItemType itemType = _tableView->itemType();
 
 	if (itemType == JASPControl::ItemType::Double)
-		defaultValue = _tableView->defaultEmptyValue().toDouble();
+		defaultValue = _tableView->defaultValue().toDouble();
 	else if (itemType == JASPControl::ItemType::Integer)
-		defaultValue = _tableView->defaultEmptyValue().toInt();
+		defaultValue = _tableView->defaultValue().toInt();
 	else
-		defaultValue = fq(_tableView->defaultEmptyValue());
+		defaultValue = fq(_tableView->defaultValue().toString());
 
 	return defaultValue;
 }
--- a/Desktop/widgets/listmodelfiltereddataentry.cpp
+++ b/Desktop/widgets/listmodelfiltereddataentry.cpp
@@ -155,6 +155,24 @@ void ListModelFilteredDataEntry::sourceTermsReset()
 	fillTable();
 }
 
+void ListModelFilteredDataEntry::initialValuesChanged()
+{
+	_initialValues.clear();
+	if (_tableView->initialValuesControl())
+	{
+		const Terms& terms = _tableView->initialValuesControl()->model()->terms();
+		if (terms.size() > 0)
+		{
+			std::string initColName = terms[0].asString();
+			DataSetTableModel* dataSetModel = DataSetTableModel::singleton();
+			int colIndex = dataSetModel->getColumnIndex(initColName);
+			for (int i = 0; i < dataSetModel->rowCount(); i++)
+				_initialValues.push_back(dataSetModel->data(dataSetModel->index(i, colIndex)).toDouble());
+		}
+	}
+	fillTable();
+}
+
 void ListModelFilteredDataEntry::initTableTerms(const TableTerms& terms)
 {
 	//std::cout << "ListModelFilteredDataEntry::initValues(OptionsTable * bindHere)" << std::endl;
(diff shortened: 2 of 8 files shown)


koenderks:
When monetaryvariable is not given, the table now initializes 1's. Can we make that 0's? A zero means correct, while a 1 means incorrect in this audit table. Better to initialize that all transactions are correct.

koenderks (changes requested):
See comment above

JorisGoosen:
> When monetaryvariable is not given, the table now initializes 1's. Can we make that 0's? A zero means correct, while a 1 means incorrect in this audit table. Better to initialize that all transactions are correct.

It would be good to define the value through something like `defaultValue`.

Desktop/widgets/listmodelfiltereddataentry.cpp
```diff
@@ -190,9 +208,12 @@ void ListModelFilteredDataEntry::initTableTerms(const TableTerms& terms)
 	for (int rowIndex : _tableTerms.rowIndices)
 	{
 		size_t row = static_cast<size_t>(rowIndex) - 1;
-
-		_enteredValues[row] = _tableTerms.values[0][valIndex++].toDouble();
+		if (_tableTerms.values[0].size() < valIndex)
+			_enteredValues[row] = _tableTerms.values[0][valIndex].toDouble();
+		else
+			_tableTerms.values[0].push_back(_tableView->defaultValue());
```
JorisGoosen, line 214: Which @boutinb does here! 
(Getting a defaultValue that is, @koenderks have you tried setting that value?)

koenderks: No ;)

Desktop/widgets/listmodelfiltereddataentry.cpp
```diff
@@ -155,6 +155,24 @@ void ListModelFilteredDataEntry::sourceTermsReset()
...
+	if (_tableView->initialValuesControl())
+	{
+		const Terms& terms = _tableView->initialValuesControl()->model()->terms();
+		if (terms.size() > 0)
+		{
+			std::string initColName = terms[0].asString();
+			DataSetTableModel* dataSetModel = DataSetTableModel::singleton();
+			int colIndex = dataSetModel->getColumnIndex(initColName);
+			for (int i = 0; i < dataSetModel->rowCount(); i++)
+				_initialValues.push_back(dataSetModel->data(dataSetModel->index(i, colIndex)).toDouble());
```
JorisGoosen, line 170: Or actually it needs to be checked here if the value from the model is QVariant? Hmm

Desktop/widgets/listmodelfiltereddataentry.cpp
koenderks: It does not seem to know it
image

boutinb: The property is defaultEmptyValue: "0". I have just checked and it works.

boutinb: I should make the defaultEmptyValue property a Variant, so that you can write defaultEmptyValue: 0 (without quote).

Desktop/widgets/listmodelfiltereddataentry.cpp
boutinb: _initialValues is used only in listModelFilteredDataEntry, so it should be anyway a double, and the toDouble just returns 0 if it cannot convert it. Or maybe I don't understand your remark...

JorisGoosen: I meant more, that maybe it was going wrong because a model (DataSetModel in this case) return `QVariant()` if out of bounds. So I thought maybe initializing the initialValues maybe went wrong because the model was being used wrong or something. 

But it seems I just misunderstood the code which is fine. If the `defaultEmptyValue` is interpreted as variant and can be used I am totally happy.

Desktop/widgets/boundcontroltableview.cpp
```diff
@@ -136,11 +136,11 @@ Json::Value BoundControlTableView::_defaultValue()
 	JASPControl::ItemType itemType = _tableView->itemType();
 
 	if (itemType == JASPControl::ItemType::Double)
-		defaultValue = _tableView->defaultEmptyValue().toDouble();
+		defaultValue = _tableView->defaultValue().toDouble();
```
JorisGoosen, line 139: Shouldnt there be a tryCatch or something around these conversions? Or a check if it *can* be converted?

koenderks (approved):

Desktop/widgets/boundcontroltableview.cpp
boutinb: toDouble does not throw an error, it just returns 0 if it cannot convert, which is good enough.

JorisGoosen (approved):

PR #4480, opened 2021-03-01
Custom fonts attempt 3

vandenman:
So the third (and hopefully last) PR for custom fonts for plots. This should also partially fix https://github.com/jasp-stats/INTERNAL-jasp/issues/929 and https://github.com/jasp-stats/jasp-issues/issues/363.

For example the beautiful font "Ani":

image

From the screenshot, you'll notice that `geom_text` in the prior-posterior plot still uses the default font. For some reason, that kind of text doesn't inherit from the theme. I think we can fix this, although it might need to be fixed separately for each individual plot element that draws text (see e.g., https://ggplot2.tidyverse.org/reference/update_defaults.html).

Requires https://github.com/jasp-stats/jaspBase/pull/20 and  R packages `systemfonts` and `ragg`.

Since this PR also implies switching from `grDevices::png` to `ragg::agg_png` it's probably a good idea to test this very thoroughly. To change the font you can just change the font at `Preferences`->`Interface`->`Results & Help`.

### DIFF
--- a/R-Interface/jaspResults/R/writeImage.R
+++ b/R-Interface/jaspResults/R/writeImage.R
@@ -23,7 +23,8 @@ openGrDevice <- function(...) {
   #if (jaspResultsCalledFromJasp())
   #  svglite::svglite(...)
   #else
-  grDevices::png(..., type = ifelse(Sys.info()["sysname"] == "Darwin", "quartz", "cairo"))
+  # grDevices::png(..., type = ifelse(Sys.info()["sysname"] == "Darwin", "quartz", "cairo"))
+  ragg::agg_png(...)
 }
 
 writeImageJaspResults <- function(plot, width = 320, height = 320, obj = TRUE, relativePathpng = NULL, ppi = 300, backgroundColor = "white",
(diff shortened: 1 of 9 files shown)


R-Interface/jaspResults/R/writeImage.R
```diff
@@ -23,7 +23,8 @@ openGrDevice <- function(...) {
   #if (jaspResultsCalledFromJasp())
   #  svglite::svglite(...)
   #else
-  grDevices::png(..., type = ifelse(Sys.info()["sysname"] == "Darwin", "quartz", "cairo"))
+  # grDevices::png(..., type = ifelse(Sys.info()["sysname"] == "Darwin", "quartz", "cairo"))
+  ragg::agg_png(...)
```
TimKDJ, line 27: The line above was introduced in https://github.com/jasp-stats/jasp-desktop/pull/3218; do you know what `ragg` does behind the scenes? I'll verify if it works ok on MacOS soon; Windows seems to be fine at least!

TimKDJ (commented):
I played around with it a bit and it seems to work as intended. Very good addition indeed! I did encounter a bunch of plots that have "static font" elements (as you pointed out), so that will be a bit of work still.

I do wonder if the current approach to setting new defaults is the most optimal (mainly referring to the `registerFonts()` that needs to be placed in the correct spots in https://github.com/jasp-stats/jaspBase/pull/20/files). We'll probably add the ability to change the default plot theme in the not too distant future and I'm sure more settings will follow. With that in mind, it would be nice if the plot defaults are always just up-to-date (e.g., because they are directly queried from Rinside or set through a central R function as soon as settings change). Then at least we don't need to worry that we might forget to call the correct "inform jaspGraphs it needs different defaults" functions before the plotting stuff happens. Maybe @JorisGoosen has an idea here.

R-Interface/jaspResults/R/writeImage.R
vandenman: `ragg` uses its own graphics device which should eliminate the need for these OS-specific workarounds. If you look at https://www.tidyverse.org/blog/2021/02/modern-text-features/#advanced-script-support then there are quite a few comparisons between `ragg`, `Cairo`, `Quartz`, and the windows device. There `ragg` outperforms its competitors, although we should test carefully if their claims hold up in practice.

JorisGoosen:
There are some new problems with initing stuff due to `renv`, see https://github.com/jasp-stats/INTERNAL-jasp/issues/1345

But I think the easiest way to distribute the defaults would be through some property/global-var-function/whatever in `jaspBase`. And then we can change that setting in/through [void Engine::receiveSettings(const Json::Value & jsonRequest)](https://github.com/jasp-stats/jasp-desktop/blob/78c0437b77977793a30565c9c5f12c3bc09c33a1/Engine/engine.cpp#L952), that way it should always be up-to-date.

And that should avoid the `renv` problem, as long as the modules have a `(>= 0.15)` requirement in their dependency on jaspBase I think? Although Im not sure how R reacts if you have a copy in both `R_HOME/library` *and* in `renv-cache/v5/...../jaspBase/`. I would hazard a guess and think that it would be fine.

JorisGoosen (approved):
Seems fine

PR #4481, opened 2021-03-02
R-Interface: put column encoder behind an ifdef so R can still install jaspResults

vandenman:
basically, Github Actions doesn't work anymore because of

```c++
#include "columnencoder.h"

...

std::string encodedUniqueNestedName = ColumnEncoder::columnEncoder()->encodeAll(uniqueNestedName);
```


JorisGoosen (approved):

JorisGoosen:
I'll approve this because it makes sense to do this. However we should at some point make sure we can have the encoding separately or something. Also to support it in R because a lot of stuff will break when we aren't encoding, as Im sure you know.

vandenman:
yeah, this is not a good solution, but the unit tests shouldn't get wrecked.

PR #4483, opened 2021-03-03
Fix the column encoder

TimKDJ:
Fix by @JorisGoosen


JorisGoosen:
It turns out I need to fix some more.

JorisGoosen:
But my computer isn't cooperating and while this PR breaks the nice resize-persistence at least it compiles and works which is better than current `development`

PR #4485, opened 2021-03-04
Make JASP use renv stuff- Part of https://github.com/jasp-stats/INTERNAL-jasp/issues/1237

vandenman:
With the checksums in https://github.com/jasp-stats/jaspBase/pull/22, we no longer need to delete the module folders before writing them. In fact, when you do CLEAN I think the only thing we need to do is delete the folder of a module (e.g., jaspAnova). Even if the checksums match, everything will still be reinstalled when the folder is missing as a failsafe.

The reason why I prefer calling `jaspBase::installJaspModule` is that I can put all the common stuff (e.g., setting the repository, comparing the checksums) in a common function before dispatching to installing based on a lock file vs installing based on the DESCRIPTION.


JorisGoosen:
this was manually cherry-picked

PR #4486, opened 2021-03-05
Use of VariablesForm inside a TabView does not work

boutinb:
This is due to the initialisation of dynamic controls. They should not
be setup immediately when they are created, but when all dynamic
controls are created: when there are some dependencies between some
controls, they can then refer to each other being sure that they are
already created.
The VariablesForm is set as a JASPControl, so that we can beter control
when its initialisation starts: this initialisation sets the
relationships between the VariablesList inside the form, and those
relationships must be set before the VariablesLists self start their
setup.

To test it, just add this in a QML file:
```
		TabView
		{
			name: "modelsList"
			maximumItems: 6
			content: VariablesForm
			{
				height: jaspTheme.smallDefaultVariablesFormHeight
				AvailableVariablesList { name: "availableCovariates"; source: [{name: "allVariablesList", use: "type=scale"}] }
				AssignedVariablesList  { name: "covariates"; title: qsTr("Covariates"); allowedColumns: "scale" }
			}
		}
```

### DIFF
--- a/Desktop/analysis/analysisform.cpp
+++ b/Desktop/analysis/analysisform.cpp
@@ -196,14 +196,8 @@ void AnalysisForm::_setUpModels()
 	}
 }
 
-void AnalysisForm::_setUp()
+void AnalysisForm::sortControls(QList<JASPControl*>& controls)
 {
-	QList<JASPControl*> controls = _controls.values();
-
-	// set the order of the BoundItems according to their dependencies (for binding purpose)
-	for (JASPControl* control : controls)
-		control->setUp();
-
 	for (JASPControl* control : controls)
 	{
 		std::vector<JASPControl*> depends(control->depends().begin(), control->depends().end());
@@ -224,10 +218,21 @@ void AnalysisForm::_setUp()
 		}
 	}
 
-	std::sort(controls.begin(), controls.end(), 
+	std::sort(controls.begin(), controls.end(),
 		[](JASPControl* a, JASPControl* b) {
 			return a->depends().size() < b->depends().size();
 		});
+}
+
+void AnalysisForm::_setUp()
+{
+	QList<JASPControl*> controls = _controls.values();
+
+	// set the order of the BoundItems according to their dependencies (for binding purpose)
+	for (JASPControl* control : controls)
+		control->setUp();
+
+	sortControls(controls);
 
 	for (JASPControl* control : controls)
 	{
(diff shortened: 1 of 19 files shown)


Kucharssim (approved):
Yeah, now it does exactly what I expected in the first place!

JorisGoosen:
> This is due to the initialisation of dynamic controls. They should not
> be setup immediately when they are created, but when all dynamic
> controls are created: when there are some dependencies between some
> controls, they can then refer to each other being sure that they are
> already created.
> The VariablesForm is set as a JASPControl, so that we can beter control
> when its initialisation starts: this initialisation sets the
> relationships between the VariablesList inside the form, and those
> relationships must be set before the VariablesLists self start their
> setup.
> 
> To test it, just add this in a QML file:
> 
> ```
> 		TabView
> 		{
> 			name: "modelsList"
> 			maximumItems: 6
> 			content: VariablesForm
> 			{
> 				height: jaspTheme.smallDefaultVariablesFormHeight
> 				AvailableVariablesList { name: "availableCovariates"; source: [{name: "allVariablesList", use: "type=scale"}] }
> 				AssignedVariablesList  { name: "covariates"; title: qsTr("Covariates"); allowedColumns: "scale" }
> 			}
> 		}
> ```

Do you think it is possible to add a test-case to jaspTestModule?

Desktop/analysis/analysisform.cpp
```diff
@@ -224,10 +218,21 @@ void AnalysisForm::_setUp()
 		}
 	}
 
-	std::sort(controls.begin(), controls.end(), 
+	std::sort(controls.begin(), controls.end(),
 		[](JASPControl* a, JASPControl* b) {
 			return a->depends().size() < b->depends().size();
```
JorisGoosen, line 226: Would this not go wrong if you have:
- D depends on C
- C depends on B and A
- B depends on A

Then the order is [ A, B, D, C ] right?

JorisGoosen: But it should be [ A, B, C, D] no?

JorisGoosen: And if this works, because the `depends()` adds all dependencies recursively so that "D depends on C and thus on B and A" then it would be good to add that as a comment here.

JorisGoosen (changes requested):
The depends sorting is possibly good or bad. But either a fix or a comment would be nice.
The rest of the code seems to look good but it is a bit too much to actually properly review so I'll gladly mark that all as approved ^^

boutinb:
I will add an example in the test module.

JorisGoosen (approved):

PR #4488, opened 2021-03-08
Implement jaspQmlSource in basis of jaspJson instead of jaspTable to allow for random RObject input

JorisGoosen:


boutinb:
That works, but if I set some integers, the json sends always a float number: 2 becomes 2.0. I could not see in the code where it goes wrong exactly: in RObject_to_JsonValue (in jaspJson.h), it calls the template function RVectorEntry_to_JsonValue, and it chooses the REALSXP in place of the INTSXP version. So the object is of this type, but I don't know why.

JorisGoosen:
> That works, but if I set some integers, the json sends always a float number: 2 becomes 2.0. I could not see in the code where it goes wrong exactly: in RObject_to_JsonValue (in jaspJson.h), it calls the template function RVectorEntry_to_JsonValue, and it chooses the REALSXP in place of the INTSXP version. So the object is of this type, but I don't know why.

JSON is JavaScriptObjectNotation and only supports what javascript supports... Which is "Number" not `int`, `double` or whatever...
So there is no way around this i'm afraid.
Just like that `Inf` and `NaN` aren't supported...

boutinb:
> > That works, but if I set some integers, the json sends always a float number: 2 becomes 2.0. I could not see in the code where it goes wrong exactly: in RObject_to_JsonValue (in jaspJson.h), it calls the template function RVectorEntry_to_JsonValue, and it chooses the REALSXP in place of the INTSXP version. So the object is of this type, but I don't know why.
> 
> JSON is JavaScriptObjectNotation and only supports what javascript supports... Which is "Number" not `int`, `double` or whatever...
> So there is no way around this i'm afraid.
> Just like that `Inf` and `NaN` aren't supported...

The problem is that in R `is.integer(1)` returns false: it considers it as a real number, and gives a REALSXP type.  Then the TO_INFINITY_AND_BEYOND macro converts it to a double Json value.
But if I set the value in R to '1L' then it returns an integer. I suppose that the R programmers are aware about this subtility.

boutinb (approved):

JorisGoosen:
> > > That works, but if I set some integers, the json sends always a float number: 2 becomes 2.0. I could not see in the code where it goes wrong exactly: in RObject_to_JsonValue (in jaspJson.h), it calls the template function RVectorEntry_to_JsonValue, and it chooses the REALSXP in place of the INTSXP version. So the object is of this type, but I don't know why.
> > 
> > 
> > JSON is JavaScriptObjectNotation and only supports what javascript supports... Which is "Number" not `int`, `double` or whatever...
> > So there is no way around this i'm afraid.
> > Just like that `Inf` and `NaN` aren't supported...
> 
> The problem is that in R `is.integer(1)` returns false: it considers it as a real number, and gives a REALSXP type. Then the TO_INFINITY_AND_BEYOND macro converts it to a double Json value.
> But if I set the value in R to '1L' then it returns an integer. I suppose that the R programmers are aware about this subtility.

This shouldn't be too bad if they coerce their vectors to something in R right?

PR #4491, opened 2021-03-15
Column width in TableView does not fit the width of the content of the column

boutinb:
In Classical or Bayesian Sampling Workflow, if you use a column with long values (bigger than the column name), then the Table View (in Data Entry) cut the values of this column: the width of the column is too small.


koenderks (approved):
Seems to work nicely!

PR #4492, opened 2021-03-18
Fix synching issue

### LINKED ISSUE #1193: JASP crashes on syncing changed .csv files
I've been having this issue a lot lately where JASP frequently either crashes on startup with "large" .jasp files or does so, when manually synchronizing with the .csv file that the Data is based on. The .csv file in question was changed outside of JASP (as far as I know, thats the only possible way) and when loading it up, JASP seems to detect the changes fine, as i could see from the log file that was created but then crashes without any further info.

Here are the last couple of lines from the .log file (shortened for better reading and due to privacy stuff)
[JASP log file before crash.log](https://github.com/jasp-stats/jasp-issues/files/6091499/JASP.log.file.before.crash.log)

boutinb:
Fixes https://github.com/jasp-stats/jasp-issues/issues/1193

There were 2 issues:  during synchronising, the dataset emits signals
that the data is changed, but the changes in the form should be handled
only by the datasetChanged signal in ColumnModel. All other signal
before should be ignored.
Also the Repeated Measures control does not handle correclty when some columns
are missing.
The RepeatedMeasuresFactorsList has been a bit reshaped to a FactorLevelList so that it becomes a bit more generic.

After this PR, the (Bayesian) Repeated ANOVA must be updated to use FactorLevelList in place of RepeatedMeasuresFactorsList


Desktop/widgets/factorlevellistbase.h
```diff
@@ -16,64 +16,77 @@
...
+	GENERIC_SET_FUNCTION(MinFactors,			_minFactors,			minFactorsChanged,			int			)
+	GENERIC_SET_FUNCTION(MinLevels,				_minLevels,				minLevelsChanged,			int			)
 
 private:
-	ListModelRepeatedMeasuresFactors*	_factorsModel	= nullptr;
+	ListModelFactorLevels*				_factorLevelsModel	= nullptr;
 
-	QStringList							_defaultFactors		=	{ tr("RM Factor 1")					};
-	QStringList							_defaultLevels		=	{ tr("Level 1"), tr("Level 2")	};
+	QString								_factorName			=	tr("Factor");
```
JorisGoosen: This will only be set in the constructor, no?

So when the language changes this isn't in fact updated is it?
Or does that happen because QML is reloaded completely and thus the constructor as well?

Docs/development/jasp-qml-guide.md
```diff
@@ -502,24 +502,30 @@ Properties
 
 </details>
 
-#### RepeatedMeasuresFactorsList
+#### FactorLevelList
+With this component, the user can set names of fcators, and for each factor, the names of its levels. This is used in Repeated Measures ANOVA.
```
JorisGoosen: fcators -> factors

JorisGoosen (commented):
Depending on whether the translation of "Factor" and "Level" is triggered properly this can be merged or not.

Desktop/widgets/factorlevellistbase.h
boutinb: It is even worse: tr() does not work in a header file.
But it is indeed better to set the default value is QML: it will be changed automatically when the language is changed. This won't be the case in the constructor (even in the c++ file).

JorisGoosen (approved):

PR #4493, opened 2021-03-19
Default Value with Binding issue

boutinb:
This issue was discovered when trying to solve
https://github.com/jasp-stats/jasp-issues/issues/1139

Open the ‘Default values with bindings’ analysis from the Test Module.
Assign some variables, uncheck the ‘check’ checkbox, set the `radio`
Radio Button to ‘three’ and set the Dropdown too ‘none’. Save the JASP
file, and load it again: the same values should be set.

### DIFF
--- a/Desktop/analysis/analysis.cpp
+++ b/Desktop/analysis/analysis.cpp
@@ -159,12 +159,6 @@ void Analysis::reload()
 	Analyses::analyses()->reload(this, true);
 }
 
-void Analysis::rebind()
-{
-	if (_analysisForm)
-		_analysisForm->bindTo();
-}
-
 void Analysis::exportResults()
 {
 	emit Analyses::analyses()->analysesExportResults();
--- a/Desktop/analysis/analysis.h
+++ b/Desktop/analysis/analysis.h
@@ -134,7 +134,6 @@ public:
 			void		run();
 			void		refresh();
 			void		reload();
-			void		rebind();
 			void        exportResults();
 			void		remove();
 
--- a/Desktop/analysis/analysisform.cpp
+++ b/Desktop/analysis/analysisform.cpp
@@ -399,7 +399,7 @@ void AnalysisForm::bindTo()
 
 	for (ListModelAvailableInterface* availableModel : availableModelsToBeReset)
 		availableModel->resetTermsFromSources(true);
-	
+
 	_addLoadingError(tql(controlsJsonWrong));
 
 	//Ok we can only set the warnings on the components now, because otherwise _addLoadingError() will add a big fat red warning on top of the analysisform without reason...
@@ -599,6 +599,7 @@ void AnalysisForm::setAnalysisUp()
 	blockValueChangeSignal(false, false);
 
 	_analysis->initialized(this, isNewAnalysis);
+	_initialized = true;
 
 	emit analysisChanged();
 }
--- a/Desktop/analysis/analysisform.h
+++ b/Desktop/analysis/analysisform.h
@@ -106,6 +106,7 @@ public:
 	Q_INVOKABLE void		addFormError(const QString& message);
 	Q_INVOKABLE void		refreshAnalysis();
 	Q_INVOKABLE void		runAnalysis();
+	Q_INVOKABLE bool		initialized()	const	{ return _initialized; }
 
 	void		addControlError(JASPControl* control, QString message, bool temporary = false, bool warning = false);
 	void		clearControlError(JASPControl* control);
@@ -180,7 +181,8 @@ private:
 	QQmlComponent*								_controlErrorMessageComponent = nullptr;
 	QList<QQuickItem*>							_controlErrorMessageCache;
 	bool										_runOnChange	= true,
-												_formCompleted = false;
+												_formCompleted = false,
+												_initialized = false;
 	QString										_info;
 	QMap<QString, QSet<ListModel*> >			_rSourceModelMap;
 	int											_signalValueChangedBlocked = 0;
--- a/Desktop/analysis/boundcontrolbase.h
+++ b/Desktop/analysis/boundcontrolbase.h
@@ -33,9 +33,9 @@ public:
 	virtual				~BoundControlBase()	{}
 
 	Json::Value					createJson()														override { return Json::nullValue;		}
-	void						bindTo(const Json::Value& value)									override { setBoundValue(value, false); }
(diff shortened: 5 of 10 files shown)


JorisGoosen:
Hmm...

Wouldn't it be a bit simpler to use the "QQmlProperty::write(" function mentioned here: https://stackoverflow.com/a/49571391 to break the binding?

So, first load the QML (before even attempting to set values from JASP file). Then check if the value of the JASP file is not the same as in QML, if it isn't use the above function and it will break the binding.

And the iterations you take is because you aren't sure whether all the properties have propagated?

boutinb:
I don't want to break the bindings. 
For example: `CheckBox { checked: myVariables.count > 0 }` This means that the CheckBox is automatically checked when some variables are added in the Variables List, but the user can still unset it if he really wants to. Removing this binding would remove this feature wanted by the developer of the analysis. 
Furthermore, removing the bindings for all components having a value other than the default one, may remove many bindings, changing the behaviour of the interface: an analysis just added in JASP would react in another way than the one loaded from a JASP file.

I need to iterate in case of chain reaction: value of component A depends on the value of component B, which depends on the value of component C. With one iteration, component B is sure to get its value from the JASP file, but not component A.

I had a look in Qt6: it is also here not possible to know which properties (and thus which item) a property depends on. That's a pity, because I could have solved this problem in a nicer way if I'd known this information.

JorisGoosen:
But the user changing the value of a QML element would be breaking the binding right?
Or how can that be different?

boutinb:
> But the user changing the value of a QML element would be breaking the binding right?
> Or how can that be different?

No, setting interactively the value does not break the binding.

JorisGoosen:
So the behaviour of the QML widgets is that they value is set either through a binding or through a manual interaction by the user?
Which means you must have both the value the user enters and the value that comes from the binding right?

So this *can* work without your wait-iteration loop. You should be setting that second user-interactive value from the jasp-file and not the binding-propagated value.

JorisGoosen:
> 
> 
> So the behaviour of the QML widgets is that they value is set either through a binding or through a manual interaction by the user?
> Which means you must have both the value the user enters and the value that comes from the binding right?
> 
> So this _can_ work without your wait-iteration loop. You should be setting that second user-interactive value from the jasp-file and not the binding-propagated value.

If this discussion is getting out of hand I think it might be good to zoom about it.

boutinb:
This issue is still scratching my head...
This is really not intuitive for people to understand, for example, that they should not set the checked property of a CheckBox with an expression.
This time, in place of a loop to force the original value (from the JASP file) to be set, this original value is stored (in _orgValue), and if the value is changed during the initialization of the form, a slot (_resetBindingValue) reset the original value. I think this is more proper way to handle this problem.

JorisGoosen:
Please help me to understand, your comment isn't all clear to me.

Do you mean that you will remove the code from this PR with the loop?
And how exactly would you replace it? Could you expand?

boutinb:
I have pushed new change for this PR.
The loop is removed. I have replaced it by keeping the original value of the JASP file, and with a signal/slot construction, if the value is changed before the analysis is completely initialized, it sets back the original value. It sounds to me more logical.

JorisGoosen:
Wouldn't it be more logical to just wait until it is initialized?
Or does that go wrong because cascading bindings?

And if it goes wrong because of cascading bindings, how does this solve it?

JorisGoosen:
To me it seems more logical to do (now that you are storing the user-supplied-value separately anyway):
- By default store `null` in the usersupplied value.
- Store the "normal" value like normally
- When QML needs the value it gets it from the boundqml-object through a property that either returns the `_orgValue` when it isn't `null` or the `value` when it is.

But that leaves the problem of when to reset `_orgValue` I guess?
How do you do that now?

I still think this issue is going to bite us...

boutinb:
I don't mind to be bitten as long as it makes the ideas clearer.
```
Wouldn't it be more logical to just wait until it is initialized?
Or does that go wrong because cascading bindings?
```
Yes, the value might be changed several times due to cascading bindings (it happens in the ‘Default values with bindings’ analysis of the Test Module). So the value correction can lead to a value change in another control. Circular change cannot happen since circular bindings are already blocked in QML. 
The value has to be set back also in the interface, so it's not enough to keep the original value internally.

JorisGoosen:
> I don't mind to be bitten as long as it makes the ideas clearer.

Is this some french expression I do not know?
At least, I don't know who or what is biting you :p

> Yes, the value might be changed several times due to cascading bindings (it happens in the ‘Default values with bindings’ analysis of the Test Module). So the value correction can lead to a value change in another control. Circular change cannot happen since circular bindings are already blocked in QML.

Ok, clear.

> The value has to be set back also in the interface, so it's not enough to keep the original value internally.

But how do you know when a change is because of a binding and how do you know when the user has changed something after loading? Do you remove all the `_orgValue` settings after you've loaded the JASP-file?

boutinb:
> But how do you know when a change is because of a binding and how do you know when the user has changed something after loading? Do you remove all the `_orgValue` settings after you've loaded the JASP-file?

The correction is done only if the analysis is not completely initialized, that is until all controls are initialized. Afterwards, this does not do anything.

JorisGoosen:
But how do you know it is initialized?

boutinb:
> But how do you know it is initialized?

It's in the code: after the bindTo is called in AnalysisForm, the _initialized boolean is set to true: all controls have got their values, and if they have changed the values of other controls, the _resetBindingValue slot were already called.

JorisGoosen:
> > But how do you know it is initialized?
> 
> It's in the code: after the bindTo is called in AnalysisForm, the `_initialized` boolean is set to true: all controls have got their values, and if they have changed the values of other controls, the `_resetBindingValue` slot were already called.

I just wanted to hear your high-level description of the solution before reading the code, I will do this soon, when I get the chance.

This does sound like a pretty reasonable solution btw!

JorisGoosen:
So, the code looks great and the plan is mostly sound.

I cannot find any documentation that outright states whether the bindings in QML use direct or queued connections, but what I've read seems to indicate that the update are immediate which implies direct connections. If this is so your code should work splendidly.

If it somehow does queued connections it won't work. But given that you would need to [specify having a delay explicitly](https://doc.qt.io/qt-5/qml-qtqml-binding.html#delayed-prop) in a qml `Binding` I think it is safe to say that direct is how it is.

JorisGoosen (approved):

PR #4496, opened 2021-03-24
Use user setting also when starting spreadsheet editor for sav file.

### LINKED ISSUE #1189: JASP opens SAV in JASP despite custom editor setting
* JASP version: 0.14.1
* Bug description: JASP opens SAV in JASP despite custom editor setting
* Expected behaviour: Open the dataset in LibreOffice, as set in the Data Preferences (C:/Users/first.last/.programmes/LibreOfficePortable/LibreOfficeCalcPortable.exe)

#### Steps to reproduce:
1. Make JASP default programme for sav files
2. Open sav file
3. Double-click on data array

boutinb:
Fixes jasp-stats/jasp-issues#1189

When the file is a sav file, JASP used to bypass the user settings, and
start always the default editor associated with this kind of file, with
the presumption that SPSS is this default application for sav file.
But this might be confusing: some users have set JASP as default
application, starting JASP a second time when double-clicking the data
panel

### DIFF
--- a/Desktop/mainwindow.cpp
+++ b/Desktop/mainwindow.cpp
@@ -1406,12 +1406,6 @@ void MainWindow::startDataEditor(QString path)
 	bool useDefaultSpreadsheetEditor = Settings::value(Settings::USE_DEFAULT_SPREADSHEET_EDITOR).toBool();
 	QString appname = Settings::value(Settings::SPREADSHEET_EDITOR_NAME).toString();
 
-	if (QString::compare(fileInfo.suffix(), "sav", Qt::CaseInsensitive) == 0)
-	{
-		if (!useDefaultSpreadsheetEditor && !appname.contains("SPSS", Qt::CaseInsensitive))
-			useDefaultSpreadsheetEditor = true;
-	}
-
 	if (appname.isEmpty())
 		useDefaultSpreadsheetEditor = true;
 
--- a/Desktop/mainwindow.cpp
+++ b/Desktop/mainwindow.cpp
@@ -1424,7 +1424,7 @@ void MainWindow::startDataEditor(QString path)
 	else
 #endif
 		if (!QDesktopServices::openUrl(QUrl::fromLocalFile(path)))
-			MessageForwarder::showWarning(tr("Start Spreadsheet Editor"), tr("No default spreadsheet editor for file %1. Use Preferences to set the right editor.").arg(fileInfo.completeBaseName()));
+			MessageForwarder::showWarning(tr("Start Spreadsheet Editor"), tr("No default spreadsheet editor for file %1. Use Preferences to set the right editor.").arg(fileInfo.fileName()));
 
 }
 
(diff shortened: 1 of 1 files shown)


JorisGoosen:
Does this also solve https://github.com/jasp-stats/jasp-issues/issues/1203 ?

boutinb:
> Does this also solve [jasp-stats/jasp-issues#1203](https://github.com/jasp-stats/jasp-issues/issues/1203) ?

The error message in this issue is quite strange. I'm not sure that it is completely the same problem.

JorisGoosen:
You read the whole issue? Because the OP kind of edited the message, that is probably why it looks weird. 

See: https://github.com/jasp-stats/jasp-issues/issues/1203#issuecomment-796796337

Now with this fix it would use the custom editor set in JASP?
What happens if you do not have an editor that can handle `sav`? 
I think it should then suggest opening a csv or something.

JorisGoosen:
Anyway, I think it would be good to include that as well, would it be ok for you to do that and add it to this PR?

boutinb:
I have added a commit to display the whole name of the file in the error message when no default application exists for the kind of file to be opened.
Fixes jasp-stats/jasp-issues#1203

JorisGoosen:
But actually we also want to have, as a fallback, it opened in csv (or at least attempted). That way a user could open a `sav` file in JASP and then edit it in excel or something.

boutinb:
> But actually we also want to have, as a fallback, it opened in csv (or at least attempted). That way a user could open a `sav` file in JASP and then edit it in excel or something.

You mean: if JASP fails to open a (not csv) file by starting the default application associated with this file, then the error message should have an option to export the data in a csv file, so that it can start afterwards the application associated with csv file. I think this is a nice idea.

JorisGoosen:
Yes exactly, it also what that issue concluded ;)

I think it should be automatic btw, not asking the user. So, when the user double-clicks the data and it is not csv:
- it first tries to open that file and if that fails it convert it to csv and try to open that instead
- if that also fails then show a warning to the user

boutinb:
I really don't think this should be automatic: this happens only when the data file exists, but the extension (sav or other type like odt) of this file has no default application. This is quite a special situation. The user should be aware of that, and be able to choose between generating a csv file, or setting an application that can edit the original data file. If not, he will see for example Excel starting, editing a generated csv file (where?), without knowing that it is not the original data file.

JorisGoosen:
> 
> 
> I really don't think this should be automatic: this happens only when the data file exists, but the extension (sav or other type like odt) of this file has no default application. This is quite a special situation. The user should be aware of that, and be able to choose between generating a csv file, or setting an application that can edit the original data file. If not, he will see for example Excel starting, editing a generated csv file (where?), without knowing that it is not the original data file.

Ok fair enough, I guess a simple yes/no msgbox isn't so bad.

JorisGoosen (approved):

PR #4497, opened 2021-03-25
Key Navigation for RibbonBar and ModulesMenu

akashrajkn:
## Description 

With "Ctrl + A", focus is given to the first element in the RibbonBar. You can switch between modules with Left and Right arrow keys. On pressing "Enter" key, the analysis menu opens (if it exists). "Esc" hides the menu.  The menu can be navigated with Up and Down arrow keys. Pressing "Enter" key starts an analysis. 

Update:
Added navigation in modulesmenu

#### Related Issue: jasp-stats/INTERNAL-jasp#1245

I tested it on Ubuntu 20.10

jasp_navigation_modules

### DIFF
--- a/Desktop/components/JASP/Widgets/CustomMenu.qml
+++ b/Desktop/components/JASP/Widgets/CustomMenu.qml
@@ -43,6 +43,45 @@ FocusScope
 	property var    sourceItem  : null
 	property point	scrollOri	: "0,0" //Just for other qmls to use as a general storage of the origin of their scrolling
 
+	property int	currentIndex: -1
+
+	Keys.onPressed:
+	{
+		if (event.key === Qt.Key_Up)
+		{
+			if (menu.props["navigateFunc"] === undefined || typeof(menu.props["navigateFunc"]) === "undefined")
+				menu.currentIndex = mod(menu.currentIndex - 1, repeater.count)
+			else
+				menu.currentIndex = menu.props["navigateFunc"](currentIndex, -1);
+		}
+		else if (event.key === Qt.Key_Down)
+		{
+			if (menu.props["navigateFunc"] === undefined || typeof(menu.props["navigateFunc"]) === "undefined")
+				menu.currentIndex = mod(menu.currentIndex + 1, repeater.count)
+			else
+				menu.currentIndex = menu.props["navigateFunc"](currentIndex, 1);
+		}
+		else if (event.key === Qt.Key_Return || event.key === Qt.Key_Space)
+		{
+			if (currentIndex > -1)
+			{
+				menu.props['functionCall'](currentIndex)
+				menu.currentIndex = -1;
+			}
+		}
+		else if (event.key === Qt.Key_Escape)
+		{
+			menu.currentIndex = -1;
+			if (menu.sourceItem !== null)
+			{
+				menu.sourceItem.forceActiveFocus();
+				if (menu.sourceItem.myMenuOpen !== undefined && typeof(menu.sourceItem.myMenuOpen) !== 'undefined')
+					menu.sourceItem.myMenuOpen = true;
+			}
+			menu.hide();
+		}
+	}
+
 	onPropsChanged:
 	{
 		hasIcons = (menu.props === undefined || "undefined" === typeof(menu.props["hasIcons"])) ? true : menu.props["hasIcons"]
@@ -188,7 +227,7 @@ FocusScope
 								height: jaspTheme.menuItemHeight
 								color:	!model.isEnabled
 											? "transparent"
-											: mouseArea.pressed
+											: mouseArea.pressed || index == currentIndex
 												? jaspTheme.buttonColorPressed
 												: mouseArea.containsMouse
 													? jaspTheme.buttonColorHovered
--- a/Desktop/components/JASP/Widgets/FileMenu/FileMenu.qml
+++ b/Desktop/components/JASP/Widgets/FileMenu/FileMenu.qml
@@ -9,10 +9,14 @@ FocusScope
 {
 	id: fileMenu
 
-	Keys.onEscapePressed:	fileMenuModel.visible = false;
+	Keys.onEscapePressed:
+	{
+		ribbon.focus			 = true;
+		ribbon.isFileMenuPressed = true;  // Close FileMenu but show the button as pressed
+		fileMenuModel.visible	= false;
+	}
 	Keys.onDownPressed:		fileMenuModel.actionButtons.selectButtonDown();
 	Keys.onUpPressed:		fileMenuModel.actionButtons.selectButtonUp();
-
 
 	function showToolSeperator(typeRole)
 	{
@@ -33,7 +37,11 @@ FocusScope
 	Connections
 	{
 		target:				fileMenuModel
-		function onVisibleChanged() { if (fileMenuModel.visible) actionMenu.forceActiveFocus(); else fileMenu.focus = false; }
+		function onVisibleChanged()
+		{
+			if (fileMenuModel.visible)
+				actionMenu.forceActiveFocus();
+		}
 	}
 
 	Item
@@ -117,7 +125,12 @@ FocusScope
 								left:		itemActionMenu.left
 							}
 
-							Keys.onLeftPressed:		fileMenuModel.visible = false;
+							Keys.onLeftPressed:
+							{
+								fileMenuModel.visible    = false;
+								ribbon.isFileMenuPressed = true;
+								ribbon.focus             = true;
+							}
 							Keys.onRightPressed:	if(hasSubMenuRole) resourceMenu.forceActiveFocus()
 							Keys.onDownPressed:		fileMenuModel.actionButtons.selectButtonDown()
 							Keys.onUpPressed:		fileMenuModel.actionButtons.selectButtonUp()
--- a/Desktop/components/JASP/Widgets/MainWindow.qml
+++ b/Desktop/components/JASP/Widgets/MainWindow.qml
@@ -58,6 +58,28 @@ Window
 		mainWindowRoot.visibility = mainWindowRoot.visibility === Window.FullScreen ? Window.Windowed : Window.FullScreen;
 	}
 
+	function changeFocusToRibbon()
+	{
+		ribbon.focus = true;
+ 		ribbon.focusOnRibbonMenu();
+	}
+
+	function changeFocusToModulesMenu()
+	{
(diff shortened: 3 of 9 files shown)


akashrajkn:
~The submodules got commited. Any idea how to resolve this? I didn't make any changes in them and git diff doesn't show anything on my system~ Fixed it

boutinb (changes requested):
I'd like the following improvements:
. when you are at the last item of the ribbon, then with a right arrow, you go back to the first ribbon item (and if you are on the first item, a left arrow brings the focus to the last ribbon item). 
. when you are on a ribbon item, now you have to use return key to open the menu: it would be nice to have the space key and the bottom arrow to open the menu. If there is no menu (module with only 1 analysis), then only the return and space key starts the analysis (not the bottom arrow key).
. when you are on the analyses menu, also here, if you are at the bottom of the menu, the bottom array key brings you to the first analysis item (and the other way around).

Desktop/components/JASP/Widgets/CustomMenu.qml
```diff
@@ -43,6 +43,74 @@ FocusScope
...
+
+        let analysisName    = "";
+        let nextIndex       = -1;
+        let moveToNextIndex = false;
+
+        if (event.key === Qt.Key_Up) {
+
+            if (menu.currentIndex > 0) {
+
+                // There should be a better way to do this!
```
boutinb: Yes. I think it is better to handle this in the model self. Here you are presuming that this is the menu for the analyses, but the CustomMenu is used in other situation. So let the model decide what to do with up and down action.

boutinb: Or you could add in the props an extra option (like functionCall): upCall, downCall. If they are not present, then a default up and down action can be defined.

akashrajkn:
I've incorporated the changes you mentioned. Can you check it out?

JorisGoosen:
It looks quite good already, I think it also needs some more changes.

For instance, it would be nice if navigation could skip the disabled ribbonbuttons.
Also nice would be to go left to the main menu button, and right to the module menubutton.

And then left from main menu, would be to module menu and vice versa.

JorisGoosen (changes requested):
Those small changes I mentioned, for the rest it works really smoothly btw.

boutinb (approved):

JorisGoosen:
Aah, and @akashrajkn could you make the shortcut something else than "Ctrl+A" because that is typically the "select all" shortcut.
Maybe F2 would be good? Then we can use F1 for help or something.

akashrajkn:
>For instance, it would be nice if navigation could skip the disabled ribbonbuttons.
Also nice would be to go left to the main menu button, and right to the module menubutton.

I've added this in the latest one commit. I've updated the gif in description ;)

> Maybe F2 would be good? Then we can use F1 for help or something.

I wasn't able to change this. I replaced "Ctrl+A" sequence with Qt.Key_F2, but it is not being recognized. I left it as "Ctrl+A" for now.

Also, the existing shortcut for full screen, Key_F11 doesn't work for me. However, if I change focus to some other element, say, Ribbons, I can catch F_key presses there. It only doesn't work in MainWindow.

**Other Changes**
If a ribbon menu is open, Esc puts the focus to ribbon bar. From there another Esc removes the focus from ribbon bar entirely. If filemenubutton or modules menu button is focused, Esc removes focus from the ribbon bar entirely. I made this behaviour because in a future PR, I will be mergining navigation control from ribbon bar to analysis options and results view.

JorisGoosen:
> > For instance, it would be nice if navigation could skip the disabled ribbonbuttons.
> > Also nice would be to go left to the main menu button, and right to the module menubutton.
> 
> I've added this in the latest one commit. I've updated the gif in description ;)

hah, thanks ^^
This is indeed better!

But... In the ribbon I can press "down" and go into the menu. I would expect this to work for main/module menus as well. But it doesn't. Only Enter.

Also, I would expect "escape" to close the menu and allow me to navigate the ribbonbar again (and this is what happens).
But it doesn't show the current button as "active" anymore so it looks a bit weird.
This is not superimportant though.

> 
> > Maybe F2 would be good? Then we can use F1 for help or something.
> 
> I wasn't able to change this. I replaced "Ctrl+A" sequence with Qt.Key_F2, but it is not being recognized. I left it as "Ctrl+A" for now.

Ok, well I really think that we cannot use Ctrl+A. Like really, it is totally unexpected. 
How about Ctrl+M ? for menu? Aaah wait, Ctrl+M is already for fullscreen :p

How about removing `Shortcut { onActivated: fileMenuModel.visible = !fileMenuModel.visible; sequences: ["Home",   Qt.Key_Home, Qt.Key_Menu];				}` and using those keys for `mainWindowRoot.changeFocusToRibbon()` ?

> 
> Also, the existing shortcut for full screen, Key_F11 doesn't work for me. However, if I change focus to some other element, say, Ribbons, I can catch F_key presses there. It only doesn't work in MainWindow.

Does Ctrl+M work for you though?
For me Ctrl+M works and F11 doesnt (on mac anyway)

> 
> **Other Changes**
> If a ribbon menu is open, Esc puts the focus to ribbon bar. From there another Esc removes the focus from ribbon bar entirely. If filemenubutton or modules menu button is focused, Esc removes focus from the ribbon bar entirely. I made this behaviour because in a future PR, I will be mergining navigation control from ribbon bar to analysis options and results view.

akashrajkn:
> Also, I would expect "escape" to close the menu and allow me to navigate the ribbonbar again (and this is what happens).
But it doesn't show the current button as "active" anymore so it looks a bit weird.
This is not superimportant though.

Fixed this

>In the ribbon I can press "down" and go into the menu. I would expect this to work for main/module menus as well.

I thought it was intuitive to have 'down' only if the element has a menu (so down won't work for Descriptives as well). Should I change this behaviour?

>How about removing Shortcut { onActivated: fileMenuModel.visible = !fileMenuModel.visible; sequences: ["Home", Qt.Key_Home, Qt.Key_Menu]; } and using those keys for mainWindowRoot.changeFocusToRibbon() ?

So, on the newer keyboard layouts, home keys are not always present. However the key sequence {"Fn" + Left arrow} is generally mapped to "Home". I think it's a bit counter-intuitive. Should I go ahead with it?

"Ctrl + M" works for me. F_keys don't work in MainWindow.
In many softwares, key navigation starts with a tab key press. That is also an option. What do you think?

JorisGoosen:
> > In the ribbon I can press "down" and go into the menu. I would expect this to work for main/module menus as well.
> 
> I thought it was intuitive to have 'down' only if the element has a menu (so down won't work for Descriptives as well). Should I change this behaviour?

In general regarding "a menu" as opening the "CustomMenu" is perfectly fine.
But the "main/file menu" and "modules menu" are also menus right?

> 
> > How about removing Shortcut { onActivated: fileMenuModel.visible = !fileMenuModel.visible; sequences: ["Home", Qt.Key_Home, Qt.Key_Menu]; } and using those keys for mainWindowRoot.changeFocusToRibbon() ?
> 
> So, on the newer keyboard layouts, home keys are not always present. However the key sequence {"Fn" + Left arrow} is generally mapped to "Home". I think it's a bit counter-intuitive. Should I go ahead with it?

Well, if it *is there* people can use it and I suppose "Fn+something" is not much worse than "Ctrl+something" anyway.

> 
> "Ctrl + M" works for me. F_keys don't work in MainWindow.
> In many softwares, key navigation starts with a tab key press. That is also an option. What do you think?

Hmm, well that is also a good one, but I would expect `tab` to also switch from area in gui to area. But I guess starting in the ribbonbar (if no data/workspace is loaded at least) makes a lot of sense.

For both of those btw, "Home/etc" and "tab-with-welcome" I would expect the focus to jump to the  main menu button.
What do you think?

JorisGoosen:
Aah you didn't do the tab thing right? No matter, that will be in the next PR then?

This is almost perfect!
Just one more thing....

When you are in the CustomMenu and you press escape now you are back in the ribbon and the button is selected again, very nice.
However, this doesnt happen for the FileMenu and ModuleMenu. If that is added then I will stop bothering you and this PR will be golden ^^

boutinb:
Om Mac, I did not know about the Fn-Left Arrow, but maybe this is standard...
One small thing: when I am on a Ribbon item, and I click with the mouse on another item, then the first item is still grayed out as is it still has the focus.

JorisGoosen:
Alright, almost there I think, one problem I saw was the following error:
`qrc:/components/JASP/Widgets/CustomMenu.qml:75: TypeError: Cannot read property 'forceActiveFocus' of null`
I think it could be good to put a little `if(menu.sourceItem !== null)` around:
```
menu.sourceItem.forceActiveFocus();
			if (menu.sourceItem.myMenuOpen !== undefined && typeof(menu.sourceItem.myMenuOpen) !== 'undefined')
				menu.sourceItem.myMenuOpen = true;
```

Then, what Bruno said is true. I think when the user clicks on a menubutton the keyboardfocus should also move there immediately. And then the "keyboardfocus-button" isn't shown as "pressed" while the "mouseclick-button" is *also*  shown as "pressed".

And I think that pressing left/right inside the menu when it is opened should be the same as pressing left/right on a button. But If you or @boutinb disagree with that then Im also fine with not having that.

akashrajkn:
Ah! I am in the process of fixing that. Also (Keyboard + Mouse) interactions are not smooth right now. I'll fix them and request a review

JorisGoosen:
Lekker bezig!

Ok, I will wait for your review-re-request then ^^

akashrajkn:
@JorisGoosen @boutinb I've made changes according to feedback. I've tested most of (key + mouse mix) navigation behaviours. Let me know if I missed any. 

> Aah you didn't do the tab thing right?

I will do this in the next PR. I want to use tab to change focus from one view to another (views being: ribbon bar, analysis options, result view, data view). I haven't thought of a nice flow yet

JorisGoosen:
Hey nice this works almost perfectly!

I just noticed a small problem where (for instance) I open debug and add "Descriptives" (which doesnt have a menu).
It keeps it looking "Pressed" ala:
Screenshot_2021-04-19_140950

JorisGoosen (approved):
Ok awesome, the newest version works as expected and the code looks good, I totally approve.

JorisGoosen:
Ah you might need to rebase it now that I moved the `renv` branch into `development`.
Maybe you could also rebase+squash everything into a single commit?

akashrajkn:
Rebased and tested :)

PR #4498, opened 2021-03-26
fix encoding in jaspSem

vandenman:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/1150

This does 2 things:

1. Encode column names in a lavaan model.
2. Pass found column names to R. That should make the R code a lot simpler. For example, right now jaspSem parses the lavaan model (again) and extracts the column names.

The code is mainly inspired by https://github.com/jasp-stats/jasp-desktop/blob/development/Desktop/widgets/boundcontroljagstextarea.cpp.

Requires https://github.com/jasp-stats/jaspSem/pull/15 to see it in action in JASP.

### DIFF
--- a/Desktop/widgets/textareabase.cpp
+++ b/Desktop/widgets/textareabase.cpp
@@ -72,14 +72,11 @@ void TextAreaBase::setUp()
 
 void TextAreaBase::rScriptDoneHandler(const QString & result)
 {
-	//This is for the lavaan model, but can be used by other type
-	if (result.length() == 0)
+	QString error = _boundControl->rScriptDoneHandler(result);
+	if (error.isEmpty())
 	{
 		setHasScriptError(false);
 		setProperty("infoText", tr("Model applied"));
-
-		_boundControl->rScriptDoneHandler(result);
-
 	}
 	else
 	{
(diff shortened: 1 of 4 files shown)


Desktop/widgets/textareabase.cpp
```diff
@@ -78,7 +78,8 @@ void TextAreaBase::rScriptDoneHandler(const QString & result)
 		setHasScriptError(false);
 		setProperty("infoText", tr("Model applied"));
 
-		_boundControl->setBoundValue(Json::Value(text().toStdString()));
+		_boundControl->rScriptDoneHandler(result);
+
```
boutinb: result is here always empty, so you don't have to send it.
I think it would be more logical to make the check whether the result is OK in the rScriptDoneHandler method (the boundcontrol class makes up the R script, so it knows whether the result is OK), and that this method returns a boolean or better the error message, so that the TextArea class can handle the error:
```
void TextAreaBase::rScriptDoneHandler(const QString & result)
{
	QString error = _boundControl->rScriptDoneHandler(result)
	if (error.isEmpty())
	{
		setHasScriptError(false);
		setProperty("infoText", tr("Model applied"));
	}
	else
	{
		setHasScriptError(true);
		setProperty("infoText", error);
	}
}

```

And in rScriptDoneHandler:
```
QString BoundControlLavaanTextArea::rScriptDoneHandler(const QString &result)
{
  if (!result.isEmpty()) return result;
....
```

boutinb (changes requested):

boutinb (approved):
There is also an upgrade problem here. But this is more general problem: https://github.com/jasp-stats/INTERNAL-jasp/issues/1312

PR #4499, opened 2021-03-29
Make JASP use renv stuff

JorisGoosen:
This needs to be combined with https://github.com/jasp-stats/jaspBase/pull/27
And to build you will need to use either https://github.com/jasp-stats/jasp-required-files/tree/MacOS-renv or https://github.com/jasp-stats/jasp-required-files/tree/Windows-renv

And in fact, on Windows you would also need to install `rtools40` and make a new kit for it in qtcreator and also install `make` in there... Hmm, maybe I should update the building guide for all this... Yes, yes I should.

This all doesn't touch flatpak at all btw but I think this is workable now.
Also the latest `renv` nightlies for download are based on the same code as this but just not squashed, so you can use those to test. Which might be good! 

I had to make the installation process on Windows more complicated...

The commit basically says:
- Part of https://github.com/jasp-stats/INTERNAL-jasp/issues/1237
- Break https://github.com/jasp-stats/INTERNAL-jasp/issues/1227 again
- Fix module request tryCatch thing
- Use checksums magic for avoiding reinstalling jasp* pkgs
- Really dont clean cache on linux
- Current R version should be a string not a double

Some MacOS stuff:
makedmg.sh changes for renv
- it should copy renv-cache
- It should replace all absolute symlinks in the build-folder/Modules by relative links to renv-cache
- it should remove BH https://github.com/jasp-stats/INTERNAL-jasp/issues/1285
- handling the symlinks is done in the new symlinkTools.R
-- it can collect all absolute symlinks in a data.frame (tested only on mac for now)
-- it can also use this data.frame to replace them all by their relative variants. (will probably break if the symlinks do not point to renv-cache btw)
-- call it from make-dmg.sh before copying

Some Windows stuff:
- windows symlinkTools.R script can now store junctions relatively in junctions.rds and another function can restore the entire Modules folder from that
- JASP and JASPEngine should be capable of running symlinkTools.R to collect junctions
- Paths should be under 260 chars otherwise WIX/light chokes, so follow path with ../.. to shave some chars off
- Copying *all* contents of jasp* pkgs is important
- Takeaway is that the installer now restores the modules-junctions. Zip should do this as well, but then on startup

- JAGS is now broken on Mac, see https://github.com/jasp-stats/INTERNAL-jasp/issues/1345
- Language files stuff has been removed because we dont need it anymore

Co-authored-by: vandenman <donvdbergh@hotmail.com>

### DIFF
--- a/Desktop/utilities/processhelper.cpp
+++ b/Desktop/utilities/processhelper.cpp
@@ -1,0 +1,99 @@
+#include "processhelper.h"
+#include "appdirs.h"
+#include "tempfiles.h"
+#include "utilities/qutils.h"
+#include "utils.h"
+#include "log.h"
+
+QProcessEnvironment ProcessHelper::getProcessEnvironmentForJaspEngine(bool withTmpDir, bool forceLC_CTYPE_C)
+{
+	QDir				programDir	= AppDirs::programDir();
+	QString				engineExe	= programDir.absoluteFilePath("JASPEngine");
+	QProcessEnvironment env			= QProcessEnvironment::systemEnvironment();
+
+	if(withTmpDir)
+		env.insert("TMPDIR",						tq(TempFiles::createTmpFolder()));
+	
+	env.insert("R_REMOTES_NO_ERRORS_FROM_WARNINGS", "true"); //Otherwise installing dependencies for modules can crap out on ridiculous warnings
+	env.insert("RENV_PATHS_ROOT",					AppDirs::renvRootLocation());
+	env.insert("RENV_PATHS_CACHE",					AppDirs::renvCacheLocations());
+	
+	//Seems a bit weird but we need to tell this to jaspBase so it can tell renv to run it again because that will be running in a subprocess. 
+	//Which also means we have the following process -> subprocess structure while installing a dynamic module:
+	// jasp -> JASPEngine with R-embedded -> Separate R -> separate instances of JASPEngine...
+	env.insert("JASPENGINE_LOCATION",				engineExe); 
+	
+	QString rHomePath = AppDirs::rHome();
+	QDir rHome(rHomePath);
+
+	QString custom_R_library = "";
+#ifdef JASP_DEBUG
+	// allow an environment variables to specify the location of packages
+	if (env.contains("JASP_R_Library"))
+		custom_R_library = ":" + env.value("JASP_R_Library");
+#endif
+#ifdef _WIN32
+#if defined(ARCH_32)
+#define ARCH_SUBPATH "i386"
+#else
+#define ARCH_SUBPATH "x64"
+#endif
+	
+	auto longToShort = [](QString in) -> QString { return QString::fromStdWString(Utils::getShortPathWin(in.toStdWString())); };
+	
+	QString PATH		= longToShort(programDir.absoluteFilePath("R/library/RInside/libs/" ARCH_SUBPATH)) + ";" + longToShort(programDir.absoluteFilePath("R/library/Rcpp/libs/" ARCH_SUBPATH)) + ";" + longToShort(programDir.absoluteFilePath("R/bin/" ARCH_SUBPATH)) + ";" + longToShort(env.value("PATH")),
+			R_HOME		= longToShort(rHome.absolutePath()),
+			JAGS_HOME	= longToShort(programDir.absoluteFilePath("JAGS/"));
+	
+	Log::log() << "R_HOME set to " << R_HOME << std::endl;
+
+	env.insert("PATH",				PATH);
+	env.insert("R_HOME",			R_HOME);
+	env.insert("JAGS_HOME",			JAGS_HOME);
+	
+#undef ARCH_SUBPATH
+
+	env.insert("R_LIBS",			 R_HOME + "/library");
+
+	env.insert("R_ENVIRON",			"something-which-doesn't-exist");
+	env.insert("R_PROFILE",			"something-which-doesn't-exist");
+	env.insert("R_PROFILE_USER",	"something-which-doesn't-exist");
+	env.insert("R_ENVIRON_USER",	"something-which-doesn't-exist");
+	
+	if(forceLC_CTYPE_C)
+	{
+		Log::log() << "Setting LC_CTYPE to C!" << std::endl;
+		env.insert("LC_CTYPE",			"C"); //To force utf-8 output from gettext et al.
+	}
+	else
+		Log::log() << "Leaving LC_CTYPE at systemdefault" << std::endl;
+					  
+
+#elif __APPLE__
+
+	env.insert("R_HOME",			rHome.absolutePath());
+	env.insert("JASP_R_HOME",		rHome.absolutePath()); //Used by the modified R script in jasp-required-files/Framework/etc/bin to make sure we use the actual R of JASP! (https://github.com/jasp-stats/INTERNAL-jasp/issues/452)
+	env.insert("R_LIBS",			rHome.absoluteFilePath("library") + ":" + programDir.absoluteFilePath("R/library"));
+	env.insert("JAGS_HOME",			programDir.absoluteFilePath("JAGS/"));
+
+	//env.insert("R_ENVIRON",			"something-which-doesnt-exist");
+	//env.insert("R_PROFILE",			"something-which-doesnt-exist");
+	//env.insert("R_PROFILE_USER",	"something-which-doesnt-exist");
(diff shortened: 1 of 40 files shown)


Desktop/utilities/processhelper.cpp
```diff
@@ -0,0 +1,99 @@
...
+	if(withTmpDir)
+		env.insert("TMPDIR",						tq(TempFiles::createTmpFolder()));
+	
+	env.insert("R_REMOTES_NO_ERRORS_FROM_WARNINGS", "true"); //Otherwise installing dependencies for modules can crap out on ridiculous warnings
+	env.insert("RENV_PATHS_ROOT",					AppDirs::renvRootLocation());
+	env.insert("RENV_PATHS_CACHE",					AppDirs::renvCacheLocations());
+	
+	//Seems a bit weird but we need to tell this to jaspBase so it can tell renv to run it again because that will be running in a subprocess. 
+	//Which also means we have the following process -> subprocess structure while installing a dynamic module:
+	// jasp -> JASPEngine with R-embedded -> Separate R -> separate instances of JASPEngine...
```
JorisGoosen, line 23: Not entirely sure about this btw.

R-Interface/jasprcpp.cpp
```diff
@@ -412,12 +425,14 @@ const char*	STDCALL jaspRCPP_evalRCode(const char *rCode) {
...
 	const std::string rCodeTryCatch(""
 		"returnVal = 'null';	"
-		"tryCatch(				"
-		"	suppressWarnings({	returnVal <- eval(parse(text=.rCode))     }),	"
-		"		error	= function(e) { .setRError(toString(e$message))	  }, 	"
-		"		warning	= function(w) { .setRWarning(toString(w$message)) }		"
-		");			"
-		"returnVal	");
+	//	"tryCatch("
+		"  withCallingHandlers(	"
```
JorisGoosen: @vandenman does it make sense what I did here?

JorisGoosen:
Aah I still need to make sure that not everything is copied on Mac.

Edit: I think this used to be done by `create-r-framework.py`, but Im not sure. I cannot now find the code for it, but the installer sure seems bigger.

R-Interface/jasprcpp.cpp
```diff
@@ -159,23 +159,30 @@ void STDCALL jaspRCPP_init(const char* buildYear, const char* version, RBridgeCa
...
-	jaspRCPP_parseEvalQNT("suppressPackageStartupMessages(library(\"modules\"))");
+	
+	jaspRCPP_logString("Initializing jaspBase.\n");
+	jaspRCPP_parseEvalQNT("library(\"jaspBase\")");
+	
+	jaspRCPP_logString("Initializing methods.\n");
+	jaspRCPP_parseEvalQNT("library(\"methods\"))");
+	
+	jaspRCPP_logString("Initializing modules.\n");
+	jaspRCPP_parseEvalQNT("library(\"modules\"))");
```
vandenman: is this necessary? the file that need this should just do `modules::` or `methods::` imo.

R-Interface/jasprcpp.cpp
vandenman: well, a little bit, assuming it's necessary to print the error message as it is thrown in R.

R-Interface/R/symlinkTools.R
```diff
@@ -0,0 +1,302 @@
...
+    )
+  )
+}
+
+get.junction.follow <- function(paths) 
+{
+  sapply(
+    paths, 
+    function(path) normalizePath(path) #Works fine but only if the target exists
+)
```
vandenman: `normalizePaths` is vectorized
```suggestion
  existing <- file.exists(paths)
  if (!all(existing))
    warning("Some paths passed to get.junction.follow did not exist!")
  paths[existing] <- normalizePaths(paths[existing])
)
```

R-Interface/R/symlinkTools.R
```diff
@@ -0,0 +1,302 @@
...
+
+#This returns atwo functions that can be used to convert paths in two direcitons
+determineOverlap <- function(targetRoot, sourceRoot)
+{
+  targetSplit <- splitPath(targetRoot)
+  sourceSplit <- splitPath(sourceRoot)
+  len         <- min(length(targetSplit), length(sourceSplit))
+  overlap     <- 0
+
+  for(idx in seq(len))
```
vandenman: if `len` is ever 0 then you get a loop over 1, 0. 0 will cause issues since you use it as an index. With `seq_len` you'd get an empty vector whenever `len` is 0 and the loop would never be entered.
```suggestion
  for(idx in seq_len(len))
```

R-Interface/R/symlinkTools.R
```diff
@@ -0,0 +1,302 @@
...
+  collectSymlinks <- function(paths)
+    for(path in paths)
+    {
+      #path <- normalizePath(path) This follows the junction immediately...
+      #print(paste0("Checking if path is link: ", path))
+      if(isLink(path))
+      {
+        #print('is symlink')
+        symPath  <- getLink(path)
+        if(substring(symPath, 1, 1) != ".") #if starts with dot it is already relative
```
vandenman:
```suggestion
        if(!startsWith(symPath, ".")) #if starts with dot it is already relative
```

R-Interface/R/symlinkTools.R
```diff
@@ -0,0 +1,302 @@
...
+generatePadFunction <- function()
+{
+  #little helper to make the log output easier to read
+    printSizes <- integer(3)
+    padToMax <- function(str, idx)
+    {
+      printSizes[[idx]] <<- max(nchar(str), printSizes[[idx]])
+      needThisMany      <-  printSizes[[idx]] - nchar(str)
+
+      return(paste0(str, paste0(rep(" ", needThisMany), collapse="", sep="")))
```
vandenman:
```suggestion
      return(paste0(str, strrep(" ", needThisMany)))
```

vandenman (changes requested):
I still need to test this in JASP but some comments below.

R-Interface/jasprcpp.cpp
JorisGoosen: Yeah well, that code was there before. So I left it, but you are right, they should just do that (and apparently the `library("modules")` doesnt do it anymore because I had to add `modules::` anyway. 

The `methods` has been there for ever. I can remove it, but I don't honestly know why it is there in the first place.

JorisGoosen:
> 
> 
> I still need to test this in JASP but some comments below.

I will process your comments, but I also need to add `ragg` and `systemfonts` now that your custom font stuff is in development anyway.. So I'll get on that.

R-Interface/jasprcpp.cpp
vandenman: let's remove it, if we run into errors I think we can always put it back.

R-Interface/R/symlinkTools.R
JorisGoosen: Aah ok then I can just remove the whole function, because where it is used it is certain it exists. And if not it should totally break.

JorisGoosen:
Ive implemented those changes and new nightlies are growing.

boutinb (approved):
So... I've spent 2 hours to understand this code. I have found nothing wrong, but even if I have an idea why the things are done like that, I cannot say that I've understood everything...
It must be thoroughly tested, especially on Windows. I don't have access to a Windows machine this week. Maybe can @vandenman test on Windows. I will test it on Mac.

boutinb:
I have built again the renv branch this morning. I have still this strange (small) issue: the first time the compilation fails because it cannot create the jaspDescriptives subfolder in the Modules folder. If I create it manually it works. If I remove it afterwards, it works also... Not blocking but strange.

JorisGoosen:
> 
> 
> I have built again the renv branch this morning. I have still this strange (small) issue: the first time the compilation fails because it cannot create the jaspDescriptives subfolder in the Modules folder. If I create it manually it works. If I remove it afterwards, it works also... Not blocking but strange.

Maybe you are running something with a different user or something?
Or... Uhm? Yeah weird, I haven't had that happen.

I'll update the `MacOS-renv` branch to R 4.0.5 next week (like I just did windows) and then me and Bruno can coordinate how to drop those signatures... Maybe I should get an unsigned copy of R 4.0.5, that might be easier. Just building it locally perhaps.

JorisGoosen:
Ill move the necessary branches into development straight through git instead of PRs.

PR #4500, opened 2021-04-06
Set wd to old value after plot is written

TimKDJ:
`on.exit(setwd(oldwd))` on line 50 was overwritten by the `on.exit(dev.off())` on line 87 causing unit tests to fail for the Audit module.


vandenman (approved):
Good catch!

PR #4503, opened 2021-04-20
plot editing: use onEditingFinished instead of onValueChanged

vandenman:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/1342

Also fixes:

- after undoing/ redoing we now reset to the last edited axis type/ advanced mode. For example, if you change the x-axis, then navigate to the y-axis, clicking undo would change the x-axis but you would not see that. Now if you click undo the selected axis is changed to the x-axis.

- `renderPlot` may be called multiple times when resizing. For example, when a resize fails we re-render a plot with the original width and height but the error status and message should not persist on the jaspObject. This should actually be in its own PR but the change is very small.


vandenman:
@boutinb or @JorisGoosen any time to review this?

boutinb (approved):

PR #4505, opened 2021-04-28
A minor tweak to get the default mono font on macOS

amirmasoudabdol:
I think this should work on most modern macOS versions.

P.S. I'm just poking around! :-) But this should be relatively safe!

### DIFF
--- a/Desktop/utilities/settings.cpp
+++ b/Desktop/utilities/settings.cpp
@@ -45,10 +45,17 @@ const Settings::Setting Settings::Values[] = {
 	{"interfaceFont",
 #ifdef WIN32
 									"Arial"},	// https://github.com/jasp-stats/INTERNAL-jasp/issues/1146
+#elif defined(Q_OS_MACOS)
+									".AppleSystemUIFont"},
 #else
 									"SansSerif"},
 #endif
-	{"codeFont",					"Fira Code"},
+	{"codeFont",
+#ifndef Q_OS_MACOS
+									"Fira Code"},
+#else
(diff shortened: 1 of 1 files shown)


JorisGoosen:
Nice looking around and in principle a good addition.

I do think we choose the mono-font on purpose, isn't that right @TimKDJ or @vandenman ?

vandenman:
Yeah, I think that was on purpose. I believe van Kesteren originally choose the font though.

amirmasoudabdol:
I guessed so when I saw that you went for the Fira Code. Fira Code is nicely licensed, and it has some nice ligatures that works well with R syntax. However, there is a whole debate regarding the practically and user-friendliness of ligatures, especially outside the tech world. Normal people often freak out when their <= suddenly transform into ≦, or when their <- turns into ←. It's a super cool feature and they may get used to it but from UX perspective, it's a bit aggressive. I often try to go with the native look and feel of the OS to deliver something familiar and less prune to surprises. 

Fira Code is derived from Fira Mono that does not have the ligatures for this reason. Most fonts in fact have less extreme ligatures but they are disabled by default. My favorite one is the Dutch [IJ (digraph)](https://en.wikipedia.org/wiki/IJ_%28digraph%29) that transforms into [“broken U”](https://en.wikipedia.org/wiki/IJ_%28digraph%29#/media/File:Ĳ_uppercase_ligature.svg) which is very cool!

P.S. Either way, as I said, I was just poking around, and the first thing I usually see are fonts! 🙃

JorisGoosen (approved):
I think going for native is a good idea

PR #4506, opened 2021-04-28
Longer tooltip duration, shorter delay and add tooltips for specials

JorisGoosen:
- https://github.com/jasp-stats/jasp-issues/issues/1106


boutinb (approved):
OK, but you have added the links to the submodule in your commit, and there are now conflicts

PR #4507, opened 2021-04-29
GITHUB_PAT stuff

JorisGoosen:
- Write documentation for setting GITHUB_PAT
- Implement settings for GITHUB_PAT for https://github.com/jasp-stats/INTERNAL-jasp/issues/1349


Resources/Help/preferences/prefsadvanced.md
```diff
@@ -19,7 +19,18 @@ Developers modules can be added directly from this folder or modules can be adde
 The CRAN repository URL determines where JASP will try to download the required packages specified in a module from.
 The default is `https://cloud.r-project.org` but a good alternative (when packages seemingly can't be installed for instance) is `cran.r-project.org`.
 
-## Windows Workarounds
+
+### Github personal access tokens
+When a dynamic module is installed JASP uses R internally to get all depencies and it often requests info from https://github.com and often this goes well.
```
boutinb: dependencies

boutinb (changes requested):
- The githubPatSys method is public from the preferencesModel object. So anyone can construct a QML file to ask for the JASP GITHUB_PAD. This is maybe not really dangerous, but it is surely not wanted. So the calls to get the GITHUB_PAD should not be made invokable from QML.
- I don't think the 2 options 'use system defined...' and 'set custom...' should be 2 checkboxes. When the 'set custom...' option is set, the other one disappears (which is quite confusing since it moves the option you are clicking). I would rather do (like for the 'Use default spreadsheet editor' option) only one checkbox 'Use default GITHUB_PAD' (which is the JASP one, or the system one if defined). This checkbox is per default set, and if unset, the TextInput appears with label 'Set custom GITHUB_PAT' (don't need to mention the 'instead of shared JASP one').

JorisGoosen:
>     * The githubPatSys method is public from the preferencesModel object. So anyone can construct a QML file to ask for the JASP GITHUB_PAD. This is maybe not really dangerous, but it is surely not wanted. So the calls to get the GITHUB_PAD should not be made invokable from QML.

It is a PAT (T from token) and I think you are right, in that it would be safer to not have it in QML.
But, I need to have a GITHUB_PAT available in QML, the one the user enters at the least. 

So to work around that I would need to construct a new QML environment/engine just to isolate the preferences from the analyses.
And of course, the R-code always has access to GITHUB_PAT, which is the whole point of it.

This all means that while I agree that it might not be perfect I also think that there is really no point in making it "inaccessible" in QML, because R...

> 
>     * I don't think the 2 options 'use system defined...' and 'set custom...' should be 2 checkboxes. When the 'set custom...' option is set, the other one disappears (which is quite confusing since it moves the option you are clicking).

I can also just disable it, would that be clearer?

> I would rather do (like for the 'Use default spreadsheet editor' option) only one checkbox 'Use default GITHUB_PAD' (which is the JASP one, or the system one if defined). This checkbox is per default set, and if unset, the TextInput appears with label 'Set custom GITHUB_PAT' (don't need to mention the 'instead of shared JASP one').

I guess I can do something like this as well.

JorisGoosen:
@boutinb Ive made changes as you requested, care to have another look?

boutinb:
Still some interface issues:
- the custom GITHUB_PAT should not be in clear (textInput.echoMode: TextInput.Password): in GITHUB this token is never seen in clear.
- There should be consistency with other options: either an option is disabled when it is not settable, either it is not visible. Now the developer folder option is unvisible (when the developer mode is unset), and the 'Set custom GITHUB_PAT' is disabled (when the the Use default PAT  is set), the 'Max logfiles to keep' is disabled (when the Log to file is unset). I think I'm more for the option to be disabled: this is mostly the case in the analyses options, and it does not change abruptely the interface (or there should be a transition if an option becomes visible/unvisible).
- The indentation is also not consistent: the 'Set custom GITHUB_PAT' should be aligned with the other suboptions ('Max logfiles' and 'Select developer folder' options)

JorisGoosen:
Ah I forgot to fix the indent, Ill do it quickly now

JorisGoosen:
Ok, all your points have been applied

boutinb (approved):

PR #4508, opened 2021-05-04
Set right Working Directory in R

boutinb:
Take over PR https://github.com/jasp-stats/jaspBase/pull/28
Probably solves jasp-stats/INTERNAL-jasp#1365

Set the working directory in R to the right temp folder before calling R functions.

### DIFF
--- a/R-Interface/jasprcpp.h
+++ b/R-Interface/jasprcpp.h
@@ -40,7 +40,7 @@ SEXP jaspRCPP_allColumnNamesDataset(SEXP encoded);
 SEXP jaspRCPP_RunSeparateR(SEXP code);
 
 				//Custom parseEvals to make sure sink is set (to capture output)
-void			jaspRCPP_parseEvalQNT(const std::string & code,	bool setWd = false,	bool preface = true);
+void			jaspRCPP_parseEvalQNT(const std::string & code,	bool setWd = true,	bool preface = true);
 RInside::Proxy	jaspRCPP_parseEval(const std::string & code,	bool setWd = true,	bool preface = true);
 void			jaspRCPP_logString(const std::string & code);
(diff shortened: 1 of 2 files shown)


R-Interface/jasprcpp.h
```diff
@@ -40,8 +40,8 @@ SEXP jaspRCPP_allColumnNamesDataset(SEXP encoded);
 SEXP jaspRCPP_RunSeparateR(SEXP code);
 
 				//Custom parseEvals to make sure sink is set (to capture output)
-void			jaspRCPP_parseEvalQNT(const std::string & code,	bool preface = true);
-RInside::Proxy	jaspRCPP_parseEval(const std::string & code,	bool preface = true);
+void			jaspRCPP_parseEvalQNT(const std::string & code,	bool setWd = false,	bool preface = true);
+RInside::Proxy	jaspRCPP_parseEval(const std::string & code,	bool setWd = true,	bool preface = true);
```
JorisGoosen, line 44: I would make the defaults the same for both.

JorisGoosen (changes requested):

R-Interface/jasprcpp.h
boutinb: OK, done

JorisGoosen: Seems like it would have been less work to change the default the other way round, haha.

So, I'll leave it to you. You can either merge like this, or switch the default to `false`

JorisGoosen (changes requested):
Maybe switch the default?

Actually now that I think about it, wouldn't it make sense to *not* set the working dir unless specifically wanted?

Because in general that shouldn't be happening for most R-calls right?

boutinb:
Fixes jasp-stats/jasp-issues#1281

vandenman (approved):

PR #4515, opened 2021-05-17
Replace axes dropdown in PlotEditor by a tab view

boutinb:
As suggested in https://github.com/jasp-stats/INTERNAL-jasp/issues/1191

### DIFF
--- a/Desktop/components/JASP/Widgets/PlotEditor/PlotEditor.qml
+++ b/Desktop/components/JASP/Widgets/PlotEditor/PlotEditor.qml
@@ -89,42 +89,81 @@ Popup
 					{
 						id:					flickChild
 						width:				axesFlickable.width
-						height:				xAxis.y + xAxis.height + jaspTheme.generalAnchorMargin
+						height:				stack.y + stack.height + jaspTheme.generalAnchorMargin
 
-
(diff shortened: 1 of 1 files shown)


JorisGoosen (changes requested):
Almost good, but I think the tabs at the top shouldn't have rounded corners, it looks a bit weird.
image

Actually the best would be to have at least the bottom non-rounded and the top I dont care. So either one of:
image
or
image

boutinb:
I have added the border.

vandenman (approved):
I don't really understand the changes to the QML but it looks very nice in JASP!

image

EJWagenmakers:
yes very nice

TimKDJ:
Looks great already! My suggestion would be to make both tabs distinct. Something like:

tabs-plot-editor


That would also ensure the redo/undo buttons are not inside the y-axis/x-axis bar, which is good IMO as they are unrelated.

JorisGoosen:
(Tims suggestion looks eerily similar to my first handdrawn suggestion ;) )

boutinb:
Is it OK now?

JorisGoosen:
image
Looks great!

JorisGoosen (approved):

PR #4517, opened 2021-05-17
After opening R-commander reload results

JorisGoosen:
- loading the qml apparently messes up webengine, so we reset it
- after a small delay we resend the results to webengine and voila they appear again.
- fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/1280

### DIFF
--- a/Desktop/mainwindow.cpp
+++ b/Desktop/mainwindow.cpp
@@ -546,8 +546,18 @@ void MainWindow::showRCommander()
 	else
 	{
 		Log::log() << "Loading RCommander"  << std::endl;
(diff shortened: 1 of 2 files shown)


boutinb (approved):
It's not wrong. But I wonder what is the cause of the problem...

JorisGoosen:
> 
> 
> It's not wrong. But I wonder what is the cause of the problem...

I think the problem was that loading a new qml-window through `_qml->load` somehow triggered a reload of the `MainPage` or at least the `WebEngineView` in it, thereby losing all the results. I'm not 100% sure about that.

But this solution does work and is not overly heavy anyhow (just sending the results again).

It might remove the notes but this is something to check, but I'll do it after it's merged.

PR #4528, opened 2021-05-27
Set the right type for the default value in TableView

boutinb:
Fixes jasp-stats/jasp-test-release#1229

### DIFF
--- a/Desktop/widgets/tableviewbase.cpp
+++ b/Desktop/widgets/tableviewbase.cpp
@@ -157,20 +157,26 @@ void TableViewBase::termsChangedHandler()
 QVariant TableViewBase::defaultValue() const
 {
 	// Force the QVariant to have the right type
-	if (itemType() == JASPControl::ItemType::Integer)
+	switch (itemType())
+	{
+	case JASPControl::ItemType::Integer:
 	{
 		if (_defaultValue.type() == QVariant::Int)		return _defaultValue;
 		if (_defaultValue.canConvert(QVariant::Int))	return _defaultValue.toInt();
+		break;
 	}
-	else if (itemType() == JASPControl::ItemType::Double)
+	case JASPControl::ItemType::Double:
 	{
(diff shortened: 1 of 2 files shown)


Kucharssim:
Also fixes jasp-stats/jasp-test-release#1230

Desktop/widgets/tableviewbase.cpp
```diff
@@ -153,3 +153,25 @@ void TableViewBase::termsChangedHandler()
 	if (_boundControl)
 		_boundControl->resetBoundValue();
 }
+
+QVariant TableViewBase::defaultValue() const
+{
+	// Force the QVariant to have the right type
+	if (itemType() == JASPControl::ItemType::Integer)
```
JorisGoosen: This should be a switch

JorisGoosen (changes requested):

Desktop/widgets/tableviewbase.cpp
boutinb: OK done

JorisGoosen: You should rerequest a review ;)

Desktop/widgets/tableviewbase.cpp
```diff
@@ -153,3 +153,31 @@ void TableViewBase::termsChangedHandler()
...
+		if (_defaultValue.type() == QVariant::Double)	return _defaultValue;
+		if (_defaultValue.canConvert(QVariant::Double))	return _defaultValue.toDouble();
+		break;
+	}
+	case JASPControl::ItemType::String:
+	{
+		if (_defaultValue.type() == QVariant::String)	return _defaultValue;
+		if (_defaultValue.canConvert(QVariant::String))	return _defaultValue.toString();
+		break;
+	}
```
JorisGoosen, line 179: You could've added a `default` here but I think GCC doesnt trust the switches too much, so this is not so bad.

JorisGoosen (approved):

PR #4531, opened 2021-06-01
Change the tick editor table

boutinb:
Some suggestions in https://github.com/jasp-stats/INTERNAL-jasp/issues/1191#issuecomment-825606708 are implemented:

- Reuse the TableView control to have the same look & feel as in the rest of JASP
- Add header to see which row gives the values and which row gives the labels.
- Add icons to add or remove ticks.
- When the tick value is changed, if the label is the same as the value, then the label is automatically changed.
- Accepts only numbers for the values
- Change the label button "Finish" to "Save and close", and "Original plot" to "Discard changes"


TimKDJ:
The labels following the values if they're the same is a nice touch!

So for others this is what it looks like now:
plot_editor_ticks_table

It definitely makes it much more clear you can add/delete columns now, but I wonder if we can make this a little more subtle?
Like a little hammer icon or downward arrow or something that appears when you hover over a column and after being clicked shows the `Insert column left`, `Insert column right` and `Remove` as a dropdown? Curious what others think of the design.

I did notice some general issues:
- [ ] undo/redo always changes the view to "x-axis" regardless of where the changes were made
- [ ] discard changes causes the advanced settings to be shown even though the checkbox is not ticked
- [ ] discard changes does not reset the manual tick table
- [ ] `Set Manually` should be `Set manually`
- [ ] Shouldn't the default value of "Range" be `Based on ticks`? Otherwise you'll need to dig through advanced to get your plot to go outside its regular boundaries.

Also I'm pretty sure "Save and close" was dropped in favour of "Finish"

JorisGoosen:
I like this design, except for the fact that the icons are underneath the lines...

I think it might make sense to go for slightly more subtle icons, with less bright colors. 
And the plus icons would be better if they were somehow more shaped like a `V` as if they were inserting.

Menus are not a bad addition and Im currently adding some code to make calling a menu in QML a bit less painful (if it works).

TimKDJ:
I'm also curious if we should impose some restrictions, because this is plain wrong:

plot_editor_ticks_order

boutinb:
I have changed the icons in this way:
Screenshot 2021-06-04 at 11 02 58

I have also added a constraint so that you can change a tick bigger than the next one, or smaller the previous one.

> * [ ]  undo/redo always changes the view to "x-axis" regardless of where the changes were made

undo/redo issues will be done in another PR

> * [ ]  discard changes causes the advanced settings to be shown even though the checkbox is not ticked

I cannot reproduce this

> * [ ]  discard changes does not reset the manual tick table

Will be done with the undo/redo PR

> * [ ]  `Set Manually` should be `Set manually`

Done

> * [ ]  Shouldn't the default value of "Range" be `Based on ticks`? Otherwise you'll need to dig through advanced to get your plot to go outside its regular boundaries.

It is so already

> Also I'm pretty sure "Save and close" was dropped in favour of "Finish"

Done

TimKDJ:
Definitely more understated now, but I don't think a horizontal arrow is a good way of indicating that pressing it will create something new, as opposed to performing some form of navigation. Especially because this very arrow is actually used for navigation inside our app.

I can't really think of a good suggestion for the three button design. I'm still partial to using the type of menu many elements have in our interface, it'd be understated and give us the possibility of adding any number of options we want:
add_ticks_plot_editor

JorisGoosen:
Well, I like the fact it shows up above the column selected, it should however not be visible when not having any cell selected.

> Definitely more understated now, but I don't think a horizontal arrow is a good way of indicating that pressing it will create something new, as opposed to performing some form of navigation. Especially because this very arrow is actually used for navigation inside our app.

Totally agree here.
I think something like you had before (with the green-plus red-minus) is not bad, the icons should just be above the header instead of superimposed. Or at least make the `+` and `-` non-see-through!
Nice and clean, same icons and you don't to click on some menu first.

Desktop/components/JASP/Widgets/PlotEditor/PlotEditor.qml
```diff
@@ -310,7 +310,7 @@ Popup
 					bottom:			parent.bottom
 					margins:		jaspTheme.generalAnchorMargin
 				}
-				text:				qsTr("Original plot")
+				text:				qsTr("Discard changes")
```
JorisGoosen, line 313: But, "Discard changes" would imply that:
- I open plot editor, plot in state A, change plot now in state B, close editor
- I open plot editor again, for same plot in state B, make changes and now in state C
- Now I click "Discard changes" and I expect it to go back to B not A.

But now that I test it it doesn't even work at all.
Not in reverting title change and not in reverting changed labels for axis-ticks...

JorisGoosen (changes requested):
It would be nice if "discard changes" worked, but maybe that is more for Dons undo/redo thing?

Desktop/components/JASP/Widgets/PlotEditor/PlotEditor.qml
TimKDJ: In my opinion there is a bit of a problem with the workflow. Stuff is saved without the user knowing when it gets saved -- which is continually, but there is no visual indication that this is happening. But then again there are also sort of "save" moments created by pressing `Finish`, which creates these independent states that do not share the undo stack.

Normally if I muck around in something and want to not save the changes I'll just close it. I get a little prompt if I want the changes to be saved and I say no. But you can't do that here, because the only way to exit is to press `Finish` which does not make it obvious that it really means "Save the current stuff to this new state thing". So then users would first have to press a different button to discard changes to state B (which would have to be different from a button to discard changes to return to state A) and thèn press `Finish` because otherwise they're inadvertently saving the changes.

I think having a dedicated cross at the top which prompts you to save or discard changes when pressed would be quite nice as well as having buttons which either apply the changes or discard them before exiting. E.g.,


```
                                                                       [ x ] 
--
Plot setting | Plot
--

[ OK ] [ Cancel ] [ Revert to original ] [ Export image ]

````

I'm not entirely sure the above would be the way to go, but I think it's good to think about our current implementation.

Comment (JorisGoosen) [reply Desktop/components/JASP/Widgets/PlotEditor/PlotEditor.qml:313]: But maybe we should do  this separately from this PR.

Comment (TimKDJ) [reply Desktop/components/JASP/Widgets/PlotEditor/PlotEditor.qml:313]: Yup, agreed

**Comment by boutinb** [comment]
Here is with the new icons:
Screenshot 2021-06-09 at 11 45 07

**Comment by AlexanderLyNL** [comment]
I'm happy with those colours. Perhaps the + can be a bit more thicker, because it blends in with the cell borders at the moment

**Comment by boutinb** [comment]
Should we merge this PR and I will make another one for the Undo/Dismiss issues?

**Comment by TimKDJ** [comment]
Yes, I think that's a good idea if @JorisGoosen agrees with the code. We can always tweak this a little further if need be, but the basics are solid now

File: Desktop/resources/icons.qrc [hunk 2]
```diff
@@ -210,5 +210,7 @@
         <file>icons/lightTheme/undo.svg</file>
         <file>icons/darkTheme/undo.svg</file>
         <file>icons/darkTheme/redo.svg</file>
+        <file>icons/cross-sign-small-red.svg</file>
```
JorisGoosen: Why not add these double in each theme?
Like every other icon.

boutinb: because they are the same...

boutinb: But maybe another theme could use another icon.... OK, I put them in the theme then.

JorisGoosen: image

They do look ok in dark theme you are right, although I think it would be a little bit better if they were slightly darker.
(And better aligned)

boutinb:
Screenshot 2021-06-10 at 15 57 07
So... here it is.

Desktop/components/JASP/Controls/TableView.qml
```diff
@@ -149,7 +149,7 @@ TableViewBase
 			width:				height
 			buttonPadding:		0
 			color:				"transparent"
-			iconSource:			jaspTheme.iconPath + "/../addition-sign-small-green.svg"
```
JorisGoosen: See aren't you happy you just made duplicates ;)

JorisGoosen (approved):

PR #4532, opened 2021-06-02
Change in Custom contrast does not update the analysis

boutinb:
Fixes jasp-stats/jasp-test-release#1212
Fixes jasp-stats/jasp-test-release#1314


FBartos (commented):
yes, it works

PR #4534, opened 2021-06-03
Implements qml-import paths

JorisGoosen:
- Implements https://github.com/jasp-stats/INTERNAL-jasp/issues/1188
- just add a `inst/qmldir` with a module definition

### DIFF
--- a/Desktop/modules/dynamicmodules.cpp
+++ b/Desktop/modules/dynamicmodules.cpp
@@ -36,6 +36,7 @@
 #include "modules/upgrader/changesetvalue.h"
 #include "gui/preferencesmodel.h"
 #include "dirs.h"
+#include <QRegularExpression>
 
 namespace Modules
 {
@@ -57,7 +58,7 @@ DynamicModules::DynamicModules(QObject *parent) : QObject(parent)
 
 DynamicModules::~DynamicModules()
 {
-	for(auto dynamic : _modules)
+	for(const auto & dynamic : _modules)
 		delete dynamic.second;
 	_modules.clear();
 
@@ -167,7 +168,10 @@ bool DynamicModules::initializeModule(DynamicModule * module)
 			emit dynamicModuleChanged(module);
 			emit loadModuleTranslationFile(module);
 			restartEngines();
+
 		}		
+
+		emit reloadQmlImportPaths();
 
 		return true;
 	}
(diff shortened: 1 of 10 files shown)


Desktop/modules/dynamicmodules.cpp
```diff
@@ -57,7 +58,7 @@ DynamicModules::DynamicModules(QObject *parent) : QObject(parent)
 
 DynamicModules::~DynamicModules()
 {
-	for(auto dynamic : _modules)
+	for(const auto & dynamic : _modules)
```
boutinb: A DynamicModule is a QObject with parent this DynamicModules QObject. So why do you have to delete it. It's even dangerous.

Desktop/modules/dynamicmodules.cpp
```diff
@@ -167,8 +168,11 @@ bool DynamicModules::initializeModule(DynamicModule * module)
 			emit dynamicModuleChanged(module);
 			emit loadModuleTranslationFile(module);
 			restartEngines();
+
```
boutinb, line 169: Why this restartEngines does not have an emit. I know it does not do anything, but it's nice to see that you are using a signal.

Desktop/modules/dynamicmodules.cpp
```diff
@@ -223,9 +227,24 @@ void DynamicModules::registerForInstallingModPkg(const std::string & moduleName)
...
+QStringList DynamicModules::importPaths() const
+{
+	QStringList allImportPaths;
+
+	for(const auto & nameMod : _modules)
+	{
+		QDir moduleFolder(tq(nameMod.second->moduleInstFolder()));
+
+		if(moduleFolder.exists("qmldir")) //So the inst contained a qmldir, and we give the library as an importpath
+			allImportPaths.append(nameMod.second->moduleRLibrary());
```
boutinb, line 242: the name moduleRLibrary is a bit awkward for the absolute path of the module folder.

boutinb (changes requested):

Desktop/modules/dynamicmodules.cpp
JorisGoosen: That is a good point!

JorisGoosen: Remnant from before, when they werent qobjects yet!
I just made the code look a bit better but didnt think about it.

Desktop/modules/dynamicmodules.cpp
JorisGoosen: Yeah, I think they maybe weren't signals to begin with. I'll check em.

Desktop/modules/dynamicmodules.cpp
JorisGoosen: Well, it isn't the absolute path of the module folder, it is the path to the r-library that contains the installed package that is the module folder. And it is particular for the module in question.

JorisGoosen:
Ok changes applied!

boutinb (approved):

PR #4535, opened 2021-06-03
Dropdown does not get the same value when duplicating

boutinb:
Fixes jasp-stats/jasp-test-release#1342


TimKDJ (changes requested):
This does not seem to fix the issue for me. The dropdown still gets the default value rather than the value it was changed to.

TimKDJ (approved):
Scratch that. It was an issue with my build. Problem solved!

PR #4539, opened 2021-06-08
Handling csv encoding on windows

### LINKED ISSUE #975: JASP not syncing after changes in csv (win7) - encoding related
* JASP version: 0.13.1
* OS name and version: Windows 7 (64bit)
* Spreadsheet editor: Microsoft Excel 2016
* Analysis: synchronizing data

* Bug description: 
JASP opens the CSV file (saved from Excel) without problem. But when we try to edit the data (double click on it in JASP - it opens the data in Excel - we make the changes needed - save in Excel - sync in JASP), JASP does not synchronize it. It just says: Synchronizing data...  and nothing happens.

We tried 1) the automatic synchronization. 2) forcing sync typing CTRL-Y. 3) forcing sync thought Sync data / Current and 4) Sync Data / Browse. 5) saving the CSV with another name, and Opening it in JASP, but we can't see the changes in the data.

If we change the preferred spreadsheet editor to Wordpad, and do all the step listed above, the error remains.

Sometimes, when the change is minor in the data (e.g. changing only one value) the sync works just fine, but with "major" changes like sorting data, it never synchronize it.

I know that Win7 is quite old, but I'm teaching statistics in JASP and some of my students still has this operation system, and this bug makes learning statistics and JASP quite hard for them.

Thank you in advance for looking into this problem!
Klara

JorisGoosen:
- Add the BOM back to the debug.csv so that we can actually see the unicode. BOM isnt necessary on mac or linux (or really at all for utf-8) but Windows makes things complicated
- Changed CSV import to assume that not having a BOM while loading a csv on windows means you are loading the native variant. This could go wrong if you copy a csv made on korean locale to a german computer and load it into JASP there.
- Fixes https://github.com/jasp-stats/jasp-issues/issues/975 and possibly https://github.com/jasp-stats/jasp-issues/issues/232
- Adds a setting for windows users to disable the "no bom means native" and disabling it just restores the way it used to work.


JorisGoosen:
I've turned on a nightly for this so you don't need to build it if you dont want to.

JorisGoosen:
Those nightlies are done btw @boutinb

boutinb:
I set my PC in a French locale without UTF-8 , and made an excel sheet with weird french characters.
If I save the file in Excel with Unicode Text format, I can read this file in 0.14.1 and 0.15 (with or without the Windows workaround): JASP recognizes that the file is in UTF-8 and can read it.
If I save the file in Excel with CSV (comma delimited) format, then I can read the file only in 0.15 with the workaround on: in this case the file does not have a BOM, and JASP uses the native locale to read it.
If I save the file in Excel with CSV (MS-DOS) format, JASP cannot read it in 0.14.1 or 0.15 (even with the workaround): in this case file is encoded with Windows specific encoding and JASP does not support this.,

If I set my PC in a French locale with UTF-8, then only Unicode Text format can be read in JASP. If I save the Excel file with CSV (comma delimited) format, then neither JASP 0.14.1 nor 0.15 (even with  the Windows workaround) works. That is still quite strange because I would have hoped that Excel save it in UTF-8 format, and JASP should be able to read this...

[TestWithCSV.zip](https://github.com/jasp-stats/jasp-desktop/files/6637454/TestWithCSV.zip)

JorisGoosen:
> I set my PC in a French locale without UTF-8 , and made an excel sheet with weird french characters.
> If I save the file in Excel with Unicode Text format, I can read this file in 0.14.1 and 0.15 (with or without the Windows workaround): JASP recognizes that the file is in UTF-8 and can read it.

Ok good

> If I save the file in Excel with CSV (comma delimited) format, then I can read the file only in 0.15 with the workaround on: in this case the file does not have a BOM, and JASP uses the native locale to read it.

This is what the PR aims to do and good to hear it works

> If I save the file in Excel with CSV (MS-DOS) format, JASP cannot read it in 0.14.1 or 0.15 (even with the workaround): in this case file is encoded with Windows specific encoding and JASP does not support this.,

There is another CSV format? Goddamn it. 
Ok if [I read this](https://excelribbon.tips.net/T009508_Comma-Delimited_and_MS-DOS_CSV_Variations.html) it seems that the MSDOS version uses a particular codepage, but... It won't have a BOM and there is *no* way for JASP to figure out that the user did this. 
So I suppose I could add another workaround but that becomes quite unwieldy if done via prefs. And adding a menu during loading is not straightforward. So I would suggest that we ignore the MSDOS variant because come on, DOS has been outdated for decades now.

> If I set my PC in a French locale with UTF-8, then only Unicode Text format can be read in JASP. If I save the Excel file with CSV (comma delimited) format, then neither JASP 0.14.1 nor 0.15 (even with the Windows workaround) works. That is still quite strange because I would have hoped that Excel save it in UTF-8 format, and JASP should be able to read this...

If you disable Prefs -> Data -> WindowsNoBomNative then this should work.

boutinb:
> If you disable Prefs -> Data -> WindowsNoBomNative then this should work.

Unfortunately not.... I don't understand why.

But your fix is anyway an improvement.

JorisGoosen:
> 
> 
> > If you disable Prefs -> Data -> WindowsNoBomNative then this should work.
> 
> Unfortunately not.... I don't understand why.
> 
> But your fix is anyway an improvement.

Well, I'd have to have a look at the resulting file. I assumed that if you turn on utf-8 locale and `noBomMeansNative` is off the file would be encoded in utf-8 without a BOM. But maybe not?

I can look at it later but for now I would just like to have this merged because it will at least improve handling csv from excel in normal situations (aka users with a non-utf-8 oldschool windows locale) that save in the "default csv" and the utf8-csv. Which is most of the situations I think users will encounter.

boutinb (approved):

PR #4540, opened 2021-06-08
Make sure not everything sets the wd when running R code

JorisGoosen:
Because it can make reloading a module fail, for instace.
- Might fix https://github.com/jasp-stats/INTERNAL-jasp/issues/1247
- It definitely made my JASP consistently reload the dev module instead of failing and not working anymore...


boutinb:
If I load the jaspDescriptives module (up to date with the upstream), then JASP crashes without any message.

boutinb (approved):

JorisGoosen:
Well I don't get a crash but I do run into: https://github.com/jasp-stats/jasp-test-release/issues/1322
So I'll merge this and then work on that.

PR #4544, opened 2021-06-14
Some fixes to get SEM to load properly and general improvements

JorisGoosen:
- the lavaan textarea does not necessarily need syntax$model and syntax$columns as long as it has modelOriginal. Because the others are derived from it.
- to make that work however rscripts should not be silently dropped when signals from the form are blocked. Instead they are now put in a queue and only sent after signals arent blocked anymore.
- this can however make stuff load repeatedly because we do not, until now, check if a boundvalue actually changed if a value is set. So now before and after is compared and a refresh signal is only sent when the options actually changed. This should save some unnecessary engine
activity
- fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/1399 (SEM backward compat)

### DIFF
--- a/Desktop/analysis/analysisform.cpp
+++ b/Desktop/analysis/analysisform.cpp
@@ -74,8 +74,11 @@ QVariant AnalysisForm::requestInfo(const Term &term, VariableInfo::InfoType info
 
 void AnalysisForm::runRScript(QString script, QString controlName, bool whiteListedVersion)
 {
-	if(_analysis && !_removed && _signalValueChangedBlocked == 0)
-		emit _analysis->sendRScript(_analysis, script, controlName, whiteListedVersion);
+	if(_analysis && !_removed)
+	{
+		if(_signalValueChangedBlocked == 0)	emit _analysis->sendRScript(_analysis, script, controlName, whiteListedVersion);
+		else								_waitingRScripts.push(std::make_tuple(script, controlName, whiteListedVersion));
+	}
 }
 
 void AnalysisForm::refreshAnalysis()
@@ -683,10 +686,22 @@ void AnalysisForm::blockValueChangeSignal(bool block, bool notifyOnceUnblocked)
 	else
 	{
 		_signalValueChangedBlocked--;
-		if (_signalValueChangedBlocked < 0)	_signalValueChangedBlocked = 0;
+		
+		if (_signalValueChangedBlocked < 0)	
+			_signalValueChangedBlocked = 0;
 
-		if (_signalValueChangedBlocked == 0 && notifyOnceUnblocked && _analysis)
-			_analysis->boundValueChangedHandler();
+		if (_signalValueChangedBlocked == 0)
+		{
+			if(notifyOnceUnblocked && _analysis)
+				_analysis->boundValueChangedHandler();
+		
+			while(_waitingRScripts.size() > 0)
+			{
+				const auto & front = _waitingRScripts.front();
+				emit _analysis->sendRScript(_analysis, std::get<0>(front), std::get<1>(front), std::get<2>(front));
+				_waitingRScripts.pop();
+			}
+		}	
 	}
 }
 
@@ -841,10 +856,12 @@ std::vector<std::vector<string> > AnalysisForm::_getValuesFromJson(const Json::V
 	return result;
 }
 
-void AnalysisForm::setBoundValue(const string &name, const Json::Value &value, const Json::Value &meta, const QVector<JASPControl::ParentKey> &parentKeys)
+bool AnalysisForm::setBoundValue(const string &name, const Json::Value &value, const Json::Value &meta, const QVector<JASPControl::ParentKey> &parentKeys)
 {
 	if (_analysis)
-		_analysis->setBoundValue(name, value, meta, parentKeys);
+		return _analysis->setBoundValue(name, value, meta, parentKeys);
+	
+	return false;
 }
 
 std::set<string> AnalysisForm::usedVariables()
--- a/Desktop/analysis/analysisform.cpp
+++ b/Desktop/analysis/analysisform.cpp
@@ -695,13 +695,16 @@ void AnalysisForm::blockValueChangeSignal(bool block, bool notifyOnceUnblocked)
 			if(notifyOnceUnblocked && _analysis)
 				_analysis->boundValueChangedHandler();
 		
-			while(_waitingRScripts.size() > 0)
-			{
-				const auto & front = _waitingRScripts.front();
-				emit _analysis->sendRScript(_analysis, std::get<0>(front), std::get<1>(front), std::get<2>(front));
-				_waitingRScripts.pop();
-			}
-		}	
+			if(_analysis->wasUpgraded()) //Maybe something was upgraded and we want to run the dropped rscripts (for instance for https://github.com/jasp-stats/INTERNAL-jasp/issues/1399)
+				while(_waitingRScripts.size() > 0)
+				{
+					const auto & front = _waitingRScripts.front();
+					emit _analysis->sendRScript(_analysis, std::get<0>(front), std::get<1>(front), std::get<2>(front));
+					_waitingRScripts.pop();
+				}
+			else //Otherwise just clean it up
+				_waitingRScripts = std::queue<std::tuple<QString, QString, bool>>();
+		}
 	}
 }
 
--- a/Desktop/analysis/analysisform.cpp
+++ b/Desktop/analysis/analysisform.cpp
@@ -695,7 +695,7 @@ void AnalysisForm::blockValueChangeSignal(bool block, bool notifyOnceUnblocked)
 			if(notifyOnceUnblocked && _analysis)
 				_analysis->boundValueChangedHandler();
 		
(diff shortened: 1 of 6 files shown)


Desktop/analysis/boundcontrolbase.cpp
```diff
@@ -37,28 +37,33 @@ void BoundControlBase::setBoundValue(const Json::Value &value, bool emitChange)
...
 					emit _control->requestComputedColumnDestruction(orgName);
 			}
 		}
 
-		if (_isColumn || _control->encodeValue())	_meta["shouldEncode"] = true;
+		if (_isColumn || _control->encodeValue())	
+			_meta["shouldEncode"] = true;
 
-		form->setBoundValue(getName(), value, _meta, _control->getParentKeys());
+		emitChange = form->setBoundValue(getName(), value, _meta, _control->getParentKeys()) && emitChange;
```
boutinb: This should not be needed: in line 32, it is already tested whether the new bound value is different from the old one.

Desktop/analysis/analysisform.cpp
```diff
@@ -683,10 +686,22 @@ void AnalysisForm::blockValueChangeSignal(bool block, bool notifyOnceUnblocked)
...
+			_signalValueChangedBlocked = 0;
 
-		if (_signalValueChangedBlocked == 0 && notifyOnceUnblocked && _analysis)
-			_analysis->boundValueChangedHandler();
+		if (_signalValueChangedBlocked == 0)
+		{
+			if(notifyOnceUnblocked && _analysis)
+				_analysis->boundValueChangedHandler();
+		
+			while(_waitingRScripts.size() > 0)
```
boutinb: The _waitingRScripts must be run only if the notifyOnceUnblocked is true (maybe this variable can be renamed by 'runOnceUnblocked'). This notifyOnceUnblocked is false when reading a JASP file: the option must be set (and some options generate R script), but the analysis must not be run, it must stay in the state read from the JASP file.

boutinb (changes requested):
The blockValueChangeSignal is called in really specific situations (reading JASP file, drag & drop in Variables List, and for the 'high order interaction' handling), so I wonder in which situation a R script should have been run but was blocked.
What I mean, is that it might arrive in the future a situation when a R script is blocked and must be afterwards still run (so to save the R script in the _waitingRScripts queue is still a good idea), but I don't see what it fixes now for the SEM module.

JorisGoosen:
And @boutinb this fix is to go from:
```
"models" : [
   {
      "modelName" : "Model 1",
      "syntax" : {
         "modelOriginal" : "ind60 =~ x1 + x2 + x3"
      }
   }
],
```
to:
```
"models" : [
   {
      "modelName" : "Model 1",
      "syntax" : {
         "columns" : [
"JaspColumn_.9._Encoded",
"JaspColumn_.10._Encoded",
"JaspColumn_.11._Encoded"
         ],
         "model" : "ind60 =~ JaspColumn_.9._Encoded + JaspColumn_.10._Encoded + JaspColumn_.11._Encoded",
         "modelOriginal" : "ind60 =~ x1 + x2 + x3"
      }
   }
],
```

And doing so by letting the qml-component do it's thing instead of having to write this all in `Upgrades.qml` which would be the alternative.

The component basically works fine like this except for the fact that the rscripts are blocked during loading...

Desktop/analysis/analysisform.cpp
JorisGoosen: But the problem here is that the way we read it we miss some information and the `waiting rscript` is the one that encodes and checks it and returns this missing information. It doesn't need to run the analysis.

JorisGoosen: And also, the state in the jasp file is changed but only augmented really

boutinb:
You can then use the _wasUpgraded analysis property to force the R script to be run in this case.
I know that some analyses in Mixed models use a lot of the 'Formula Type' input text. If we allow all the formula types to be checked, that would send many R scripts to the engines, and also with the risico that the analysis will be rerun, and that is something you really don't want when you load a JASP file.

JorisGoosen:
> You can then use the _wasUpgraded analysis property to force the R script to be run in this case.

I'll have a look.

> I know that some analyses in Mixed models use a lot of the 'Formula Type' input text. If we allow all the formula types to be checked, that would send many R scripts to the engines, and also with the risico that the analysis will be rerun, and that is something you really don't want when you load a JASP file.

But if the results of those rscripts are identical with what was in the options than nothing should be reran right?

Desktop/analysis/boundcontrolbase.cpp
JorisGoosen: ok

JorisGoosen:
This should be better

Desktop/analysis/analysisform.cpp
```diff
@@ -683,10 +686,25 @@ void AnalysisForm::blockValueChangeSignal(bool block, bool notifyOnceUnblocked)
...
+			_signalValueChangedBlocked = 0;
 
-		if (_signalValueChangedBlocked == 0 && notifyOnceUnblocked && _analysis)
-			_analysis->boundValueChangedHandler();
+		if (_signalValueChangedBlocked == 0)
+		{
+			if(notifyOnceUnblocked && _analysis)
+				_analysis->boundValueChangedHandler();
+		
+			if(_analysis->wasUpgraded()) //Maybe something was upgraded and we want to run the dropped rscripts (for instance for https://github.com/jasp-stats/INTERNAL-jasp/issues/1399)
```
boutinb: If it is not upgrading, and the notifyOnceUnblocked is true, the R scripts must be run also.

Desktop/widgets/boundcontrollavaantextarea.cpp
```diff
@@ -117,7 +118,7 @@ QString BoundControlLavaanTextArea::rScriptDoneHandler(const QString & result)
 
 	boundValue["columns"] = columns;
 
-	setBoundValue(boundValue);
+	setBoundValue(boundValue, !_control->form()->analysisObj()->wasUpgraded());
```
boutinb, line 121: Quite awkward here. If the analysis should not be run when wasUpgraded in true, then set this condition directly in the Analysis::boundValueChangedHandler() method.

boutinb (changes requested):
The  BoundControlBase::setBoundValue method tests already whether the new value is different from the old value:
`if (value == boundValue()) return;`
So now this is tested also in Analysis::setBoundValue. This should not be needed....

Desktop/analysis/analysisform.cpp
JorisGoosen: Ok, but if the rscripts run they might change the boundvalue and cause it to run the analysis on return of script?

Desktop/widgets/boundcontrollavaantextarea.cpp
JorisGoosen: which condition?

Desktop/analysis/analysisform.cpp
boutinb: But here you are in the situation that the valueChanged signal was blocked (and the R script was not sent but cached), and that now the valueChanged signal is unblocked: if this function is called with notifyOnceUnblocked true, then you expect normally that it runs the R scripts. This is a case that does not occur in JASP now (this notifyOnceUnblocked is false when setting up analyses from a JASP file), but in the future this function could be used in another way, and you would expect these R scripts to be run.

Desktop/widgets/boundcontrollavaantextarea.cpp
boutinb: The condition 'wasUpgraded is true'

JorisGoosen: But if I put it there then nothing is ever run, which means _wasUpgraded is never set to false... Which means it won't work.

JorisGoosen: As in:
```
void Analysis::boundValueChangedHandler()
{
	incrementRevision(); // To make sure we always process all changed options we increment the revision whenever anything changes

	Log::log() << "Option changed for analysis '" << name() << "' and id " << id() << ", revision incremented to: " << _revision << std::endl;

	if (_refreshBlocked)
		return;

	if (form() && (form()->hasError() || !form()->runOnChange()))
		return;

	run();
}
```

So I could put that condition there, but it looks like it won't get to `run` then whenever `_wasUpgraded == true`.

But it is only set to `false` in `setResults` and `setVersion`, which means that it first needs to run, no?

Desktop/analysis/analysisform.cpp
JorisGoosen: Ok, I've fixed this

Desktop/widgets/boundcontrollavaantextarea.cpp
JorisGoosen: I mean, I totally get why you find the way I've written it now "awkward" but I thought it would be good to have the decision on whether rscripts should trigger valueChanged or not to be in the qml widget.

I could forward the `wasUpgraded()` function to `BoundControlBase` or something. That might make it look better at least.

boutinb (approved):

PR #4545, opened 2021-06-14
Prevent infinite loop when creating or change name of a computed column

boutinb:
When overwriting the column, the labels were first removed, and then it was checked whether column is changed: which was always true. Removing the labels is not necessary, since it synchronizes anyway the old labels with the new ones.

Fixes jasp-stats/INTERNAL-jasp#1406

### DIFF
--- a/Desktop/widgets/tableviewbase.cpp
+++ b/Desktop/widgets/tableviewbase.cpp
@@ -216,6 +216,19 @@ QVariant TableViewBase::defaultValue() const
 	return _defaultValue;
 }
 
+std::vector<std::string> TableViewBase::usedVariables() const
+{
+	std::vector<std::string> result;
+
+	if (_tableModel && _tableModel->areColumnNamesVariables())
+	{
+		for (int i = 0; i < _tableModel->columnCount(); i++)
+			result.push_back(fq(_tableModel->headerData(i, Qt::Horizontal).toString()));
+	}
+
+	return result;
+}
+
 void TableViewBase::setItemTypePerRow(QVariantList list)
 {
 	QList<JASPControl::ItemType> typeList;
--- a/Common/column.cpp
+++ b/Common/column.cpp
@@ -608,8 +608,6 @@ bool Column::overwriteDataWithScale(std::vector<double> scalarData)
 
 bool Column::overwriteDataWithOrdinal(std::vector<int> ordinalData, std::map<int, std::string> levels)
 {
-	labels().clear();
-
 	size_t setVals = ordinalData.size();
 
 	if(ordinalData.size() != rowCount())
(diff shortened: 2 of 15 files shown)


JorisGoosen:
Ok, but I did add those specifically to make things work: https://github.com/jasp-stats/jasp-desktop/commit/7f25faa70ff1dd99f8367b98e5e44f0d722673b4

Although I shot myself in the foot here by not making an issue and describing it a bit more...
I will test it out a bunch to see if I can run it into some edge-case.

(Because I don't remember what was the exact problem there)

JorisGoosen:
Ok, well I think I was probably just overzealous with the `labels.clear()`  thing, so lets just leave that so.

Your fix however, doesn't work in certain circumstances such as:
Screenshot_2021-06-21_173000

I think another problem could be that perhaps your are returning a column that is added in an analysis as "used in analysis" which might trigger rerunning? (Aka what used to be `usedVariables`)

Because in this classification case the analysis clearly outputs random things and this kind of makes sense as the data is quite unreasonable. But that means that your *fix* here isn't enough.

JorisGoosen (changes requested):
Well, it doesn't work yet (I needed to "comment" to request changes)

boutinb:
@JorisGoosen Maybe you can help here. What I see is that the computed column generates new values each time. In fact this should be possible, but the handling of the engine response forces the desktop to run the analysis again:

- it processes the analysis reply in EngineRepresentation::processAnalysisReply
- it sees that there is a computed Column and calls the EngineRepresentation::checkForComputedColumns
- this emits the EngineSync::computeColumnSucceeded signal with a dataChanged is true (the computed column gets each time new value).
- the signal is received by the ComputedColumnsModel::computeColumnSucceeded that, as dataChanged is true, calls the checkForDependentColumnsToBeSent
- this calls the ComputedColumnsModel::checkForDependentAnalyses: as the same analysis is dependent on this computed column, this analysis is refreshed.

I would say then the checkForDependentAnalyses should not refresh the analysis from whom it gets this call. But the ComputedColumnsModel has no idea which analysis he's handling... Or there is way to know it? Or I need to pass the analysis object as argument to the ComputedColumnsModel methods?

JorisGoosen:
Well...

The analysis shouldnt be dependent on the column right? so `checkForDependentAnalyses` gives back the wrong info.

As in, for workflow in audit the analysis IS dependent but there the loop is solved because the data doesnt change.

But in ML there is only a column added and not used right? So why is the analysis depending on the column it itself created?
The analysis knows this right?
What I think is that the computed column shouldnt be returned by `Analysis::usedVariables()`, because it isnt "used" it is "created".

boutinb:
OK it was more work than I thought. But now is the usedVariables method clearer.

JorisGoosen:
I'll check it out later

Desktop/widgets/tableviewbase.cpp
```diff
@@ -216,6 +216,19 @@ QVariant TableViewBase::defaultValue() const
 	return _defaultValue;
 }
 
+std::vector<std::string> TableViewBase::usedVariables() const
+{
+	std::vector<std::string> result;
+
+	if (_tableModel && _tableModel->areColumnNamesVariables())
+	{
```
JorisGoosen, line 224: these braces are a waste of space

JorisGoosen (approved):

PR #4546, opened 2021-06-14
Change language to Nederlands does not change the SEM menu

boutinb:
This is because the SEM title does not change when the language is changed, and then the menu items.
Fixes jasp-stats/jasp-test-release#1340

### DIFF
--- a/Desktop/modules/description/description.cpp
+++ b/Desktop/modules/description/description.cpp
@@ -4,6 +4,7 @@
 #include "entrybase.h"
 #include "../analysisentry.h"
 #include "../dynamicmodule.h"
+#include "utilities/languagemodel.h"
 
 namespace Modules
 {
@@ -66,6 +67,7 @@ void Description::connectChangesToDelay()
 	connect(this, &Description::nameChanged,			this, &Description::delayedUpdate);
 	connect(this, &Description::requiresDataDefChanged,	this, &Description::delayedUpdate);
 	connect(this, &Description::dynModChanged,			this, &Description::delayedUpdate);
+	connect(LanguageModel::lang(), &LanguageModel::languageChanged, this, &Description::delayedUpdate);
 }
 
 void Description::addChild(DescriptionChildBase * child)
--- a/Desktop/modules/description/description.cpp
+++ b/Desktop/modules/description/description.cpp
@@ -4,7 +4,6 @@
 #include "entrybase.h"
 #include "../analysisentry.h"
 #include "../dynamicmodule.h"
-#include "utilities/languagemodel.h"
 
 namespace Modules
 {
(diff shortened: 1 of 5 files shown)


JorisGoosen (changes requested):
Uhm, instead of doing this the rough way, why not make sure that changes to the menu propagate?

The whole point of delayedUpdate is to group updates via QML.

So if none of the entries in the menu nor the title changes, then there needs to be no update to Description.

But if the problem is that the menu DOES change but it isn't visible in JASP... Well, that means that the menuchange ought to trigger this.

boutinb:
> But if the problem is that the menu DOES change but it isn't visible in JASP... Well, that means that the menuchange ought to trigger this.

OK, I use now the signals of the entryMenu.

Desktop/modules/description/description.h
```diff
@@ -86,7 +86,7 @@ public slots:
 	void iShouldBeUpdated(			Description	  *	desc		);
 
 private slots:
-	void childChanged(DescriptionChildBase * ) { delayedUpdate(); }
+	void childChanged() { delayedUpdate(); }
```
JorisGoosen: But this slot already seems to call delayedUpdate?
So why do we need the connection? 

(And if we do not need it shouldn't it be a signal?)

JorisGoosen (commented):
> > But if the problem is that the menu DOES change but it isn't visible in JASP... Well, that means that the menuchange ought to trigger this.
> 
> OK, I use now the signals of the entryMenu.

Can you "rerequest a review" next time? 
Because otherwise I see "changes requested" in the PR menu and think it won't be necessary to look at it. Luckily I remembered the email.

I'll go check it out now though ^^

Desktop/modules/description/description.h
JorisGoosen: Aah did I not post this before?

Desktop/modules/description/description.cpp
```diff
@@ -66,6 +66,7 @@ void Description::connectChangesToDelay()
 	connect(this, &Description::nameChanged,			this, &Description::delayedUpdate);
 	connect(this, &Description::requiresDataDefChanged,	this, &Description::delayedUpdate);
 	connect(this, &Description::dynModChanged,			this, &Description::delayedUpdate);
+	connect(this, &Description::childChanged,			this, &Description::delayedUpdate);
```
JorisGoosen, line 69: This is double isn't it?

like below

JorisGoosen (changes requested):
`Description::childChanged` is a slot, but it is connected to `delayedUpdate` and it also calls it in the slot...

This doesn't make sense at all. Either it is a signal that triggers `delayedUpdate` or it is a slot and it calls it itself but then the connection should be removed.

I think I have a preference for the making `childChanged` a signal, ow that you've removed all the arguments that can work quite cleanly.

JorisGoosen:
I did test it and it does work so once that change has been made it can be merged ^^

PR #4548, opened 2021-06-16
Update unit testing documentation for vdiffr 1.0

TimKDJ:
To accompany https://github.com/jasp-stats/jaspTools/pull/23

### DIFF
--- a/Docs/development/r-unit-test-guide.md
+++ b/Docs/development/r-unit-test-guide.md
@@ -37,7 +37,7 @@ Your module should include the following structure:
 An empty folder, until you add tests, at that point it will contain a number of test files as described in [Creating a test file](#creating-a-test-file).
 
 ### _snaps
-An empty folder, will be automatically filled later, see [Adding plot tests](#adding-plot-tests).
+An empty folder, will be automatically created and filled later, see [Adding plot tests](#adding-plot-tests).
 
 ### testthat.R
 ```
--- a/Docs/development/r-unit-test-guide.md
+++ b/Docs/development/r-unit-test-guide.md
@@ -6,8 +6,8 @@ Note that it is required to have the R package [jaspTools](https://github.com/ja
 
 - [Unit Tests](#unit-tests)
   - [Folder structure](#folder-structure)
-    - [figs](#figs)
     - [testthat](#testthat)
+    - [_snaps](#_snaps)
     - [testthat.R](#testthatr)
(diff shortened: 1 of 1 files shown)


Docs/development/r-unit-test-guide.md
```diff
@@ -23,23 +23,22 @@ Note that it is required to have the R package [jaspTools](https://github.com/ja
...
   - [.github/workflows/unittests.yml](#githubworkflowsunittestsyml)
-  
-### figs
-An empty folder, will be automatically filled later, see [Adding plot tests](#adding-plot-tests).
 
 ### testthat
 An empty folder, until you add tests, at that point it will contain a number of test files as described in [Creating a test file](#creating-a-test-file).
 
+### _snaps
+An empty folder, will be automatically filled later, see [Adding plot tests](#adding-plot-tests).
```
vandenman:
```suggestion
An empty folder, will be automatically created and filled later, see [Adding plot tests](#adding-plot-tests).
```

Docs/development/r-unit-test-guide.md
```diff
@@ -211,13 +210,13 @@ test_that("Binomial Test table results match", {
 test_that("0 plot matches", {
 	plotName <- results[["results"]][["containerPlots"]][["collection"]][["containerPlots_contBinom"]][["collection"]][["containerPlots_contBinom_0"]][["data"]]
 	testPlot <- results[["state"]][["figures"]][[plotName]][["obj"]]
-	jaspTools::expect_equal_plots(testPlot, "0", dir = "BinomialTest")
+	jaspTools::expect_equal_plots(testPlot, "0")
```
vandenman, line 213: hmm... is "0" a good filename? It probably works fine but maybe in the example, we should be a bit more descriptive?

vandenman (changes requested):
Looks good, two minor comments.

Docs/development/r-unit-test-guide.md
TimKDJ: No, it's a terrible filename, so on line 223 we have
```
You will be able to copy-paste the output directly into a test file. Just make sure that each test/expectation title makes some sense (e.g., "Binomial Test table results match" makes sense but "1 plot matches" could use some work).
In the example above we should adjust the two plot tests, because "0" and "1" are not very informative:
```

This is hidden by the diff viewer though.

PR #4550, opened 2021-06-22
Fix meta encoding level

JorisGoosen:
This changes the way the `meta` is generated for options and makes sure it includes the levels and factors from RM ANOVA.

It also sends this meta to the engine for editImage and saveImage and makes sure that the levels shown there are decoded as well. There is a weird glitch though but Im not sure if that is my fault. 
The plot resizes, but always with a delay. Unless Don can see what is going there I suggest we make an issue after this is merged because with this PR at least the analysis works

### DIFF
--- a/Desktop/widgets/jasplistcontrol.h
+++ b/Desktop/widgets/jasplistcontrol.h
@@ -89,7 +89,7 @@ public:
 			const QString&		valueRole()					const			{ return _valueRole;			}
 			bool				containsVariables()			const			{ return _containsVariables;	}
 			bool				containsInteractions()		const			{ return _containsInteractions;	}
-			bool				encodeValue()				const override	{ return containsVariables();	}
+			bool				encodeValue()				const override	{ return containsVariables() || containsInteractions();	}
 			bool				useSourceLevels()			const			{ return _useSourceLevels;		}
 			void				setUseSourceLevels(bool b)					{ _useSourceLevels = b;			}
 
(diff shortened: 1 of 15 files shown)


Desktop/widgets/jasplistcontrol.h
```diff
@@ -89,7 +89,7 @@ class JASPListControl : public JASPControl
 			const QString&		valueRole()					const			{ return _valueRole;			}
 			bool				containsVariables()			const			{ return _containsVariables;	}
 			bool				containsInteractions()		const			{ return _containsInteractions;	}
-			bool				encodeValue()				const override	{ return containsVariables();	}
+			bool				encodeValue()				const override	{ return containsVariables() || containsInteractions();	}
```
boutinb, line 92: Why did you add this?

boutinb (approved):

Desktop/widgets/jasplistcontrol.h
JorisGoosen: Otherwise interactions between levels with encodable entries wouldn't be encoded. And thus RM ANOVA wouldnt work.

`withinModelTerms` specifically

boutinb: I would have then preferred to overwrite encodeValue() for the specific class...

JorisGoosen: Well, you wrote quite a lot of code, I thought this was the most specific place I could get.

But are there lists where the interactions don't need to be encoded?

PR #4554, opened 2021-06-28
Set languages in alphabetic order

boutinb:


Desktop/utilities/languagemodel.cpp
```diff
@@ -336,7 +331,9 @@ void LanguageModel::findQmFiles(QString qmlocation)
 		{
 			Log::log() << "Language (" << loc.language() << ") not registered in LanguageModel, adding it now" << std::endl;
 
-			_languages.push_back(loc.language());
+			int i = 0;
+			while (i < _languages.size() && loc.nativeLanguageName().compare(_languagesInfo[_languages[i]].nativeLanguageName, Qt::CaseInsensitive) > 0)	i++;
```
JorisGoosen: Why are you using a while loop to add random empty languages to the list? 

Why not make _languages a more sensible datatype...

JorisGoosen: Or just throw it away and use the keys as they are in _languagesInfo?

JorisGoosen: In any case, make the while loop into a for loop.

boutinb: I have restructured the LanguageModel class:

- There is no currentIndex anymore, only a languageName
- _languages map as for key the value displayed in the dropdown. This key is composed by the language code + the native name of the language. So the order of the language code will be used. And as the native language name is used, thanks to the language code, it is easier to know which language the name means (I have no clue to distinguish Japanese from Chinese).
- The locale country is also used: we can now differentiate a language spoken in different country (I know that the portuguese is now mainly Brazilan Portuguese, but in the future Portugal Portuguese could be added).
- Obsolete codes are removed.

Desktop/utilities/languagemodel.h
```diff
@@ -45,71 +33,58 @@ class LanguageModel : public QAbstractListModel
...
-	QString currentLanguageCode()	const;
-	QLocale::Language currentLanguage()	const {return currentLanguageInfo().language; };
+	QString languageCode(const QLocale& locale)		const;
+	QString languageCode(const QString& language)	const;
+	QString	languageName(const QLocale& loc)		const;
+	QString currentLanguageCode()					const	{ return languageCode(_currentLanguage); }
+	QString currentLanguage()						const	{ return _currentLanguage; }
 
+	static	bool	hasDefaultLanguage()		{ return _singleton->currentLanguage() == _singleton->languageName(_defaultLocale); }
 	//This function (currentTranslationSuffix) should be made obsolete through the abolishment of all the _nl etc files:
```
JorisGoosen, line 51: Amen!

Desktop/utilities/settings.cpp
```diff
@@ -38,6 +38,7 @@ const Settings::Setting Settings::Values[] = {
 	{"cranRepositoryURL",			"https://cloud.r-project.org"},
 	{"userHasGitHubAccount",		false},
 	{"preferredLanguage",			QLocale::English},
+	{"preferredCountry",			QLocale::UnitedKingdom},
```
JorisGoosen, line 41: Would @EJWagenmakers prefer `QLocale::UnitedStates`?

JorisGoosen (approved):
That was a delight to review!

boutinb:
> That was a delight to review!

Thanks!

PR #4555, opened 2021-06-29
[DO NOT MERGE] Changes for flatpak

vandenman:
Some changes required for flatpak

I hardcoded most of the paths for flatpak. We could use environment variables instead of hardcoded paths (such as `MODULES_RENV_ROOT = /app/lib64/renv-root`) but flatpak didn't seem to always pick up the environment variables and modifying an environment variable deletes all cached modules (although I suppose you can avoid that by making those environment variables specific to particular modules).

In addition, people can now also set a cran repository (if they are building) by specifying an environment variable (`JASP_R_REPOSITORY`). This is also the only environment variable I use on flatpak atm.

### DIFF
--- a/JASP.pri
+++ b/JASP.pri
@@ -8,16 +8,17 @@ JASP_R_INTERFACE_MINOR_VERSION =  2   # Code changes
 
 JASP_R_INTERFACE_NAME = $$JASP_R_INTERFACE_TARGET$$JASP_R_INTERFACE_MAJOR_VERSION'.'$$JASP_R_INTERFACE_MINOR_VERSION
 
-!exists(/app/lib/*){
+GITHUB_PAT_DEF=$$(GITHUB_PAT_DEF)						#First check if we set this special env var (on buildbot)
+isEmpty(GITHUB_PAT_DEF): GITHUB_PAT_DEF=$$(GITHUB_PAT)	#Otherwise use the one we are using
+exists(/app/lib/*){
+	# Note: this crashes is the env variables aren't base64 encoded
+	GITHUB_PAT_DEF	= $$system(echo $${GITHUB_PAT_DEF}	| base64 -d)
+	GITHUB_PAT		= $$system(echo $${GITHUB_PAT}		| base64 -d)
+}
 
-	GITHUB_PAT_DEF=$$(GITHUB_PAT_DEF)						#First check if we set this special env var (on buildbot)
-	isEmpty(GITHUB_PAT_DEF): GITHUB_PAT_DEF=$$(GITHUB_PAT)	#Otherwise use the one we are using
-	isEmpty(GITHUB_PAT_DEF){
-		error("No GITHUB_PAT or GITHUB_PAT_DEF found!  Follow the steps at https://github.com/jasp-stats/jasp-desktop/blob/development/Docs/development/jasp-building-guide.md$${LITERAL_HASH}github_pat")
-
-		GITHUB_PAT_DEF="NO_GITHUB_PAT_FOUND"	#To indicate a clearer error
-	}
-
+isEmpty(GITHUB_PAT_DEF){
+	error("No GITHUB_PAT or GITHUB_PAT_DEF found!  Follow the steps at https://github.com/jasp-stats/jasp-desktop/blob/development/Docs/development/jasp-building-guide.md$${LITERAL_HASH}github_pat")
+	GITHUB_PAT_DEF="NO_GITHUB_PAT_FOUND"	#To indicate a clearer error
 }
 #R settings
 CURRENT_R_VERSION = "4.1"
--- a/JASP.pri
+++ b/JASP.pri
@@ -8,12 +8,16 @@ JASP_R_INTERFACE_MINOR_VERSION =  2   # Code changes
 
 JASP_R_INTERFACE_NAME = $$JASP_R_INTERFACE_TARGET$$JASP_R_INTERFACE_MAJOR_VERSION'.'$$JASP_R_INTERFACE_MINOR_VERSION
 
-GITHUB_PAT_DEF=$$(GITHUB_PAT_DEF)						#First check if we set this special env var (on buildbot)
-isEmpty(GITHUB_PAT_DEF): GITHUB_PAT_DEF=$$(GITHUB_PAT)	#Otherwise use the one we are using
-isEmpty(GITHUB_PAT_DEF){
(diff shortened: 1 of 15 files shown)


vandenman:
oh and the GitHub diff is messed up, I think this is because
https://github.com/jasp-stats/jasp-desktop/blob/d30f158d0d86a6d896be92703f1b9bd15104b835/.editorconfig#L16-L18
doesn't list .pri and .pro files.

JASP.pri
```diff
@@ -8,12 +8,16 @@ JASP_R_INTERFACE_MINOR_VERSION =  2   # Code changes
 
 JASP_R_INTERFACE_NAME = $$JASP_R_INTERFACE_TARGET$$JASP_R_INTERFACE_MAJOR_VERSION'.'$$JASP_R_INTERFACE_MINOR_VERSION
 
-GITHUB_PAT_DEF=$$(GITHUB_PAT_DEF)						#First check if we set this special env var (on buildbot)
-isEmpty(GITHUB_PAT_DEF): GITHUB_PAT_DEF=$$(GITHUB_PAT)	#Otherwise use the one we are using
-isEmpty(GITHUB_PAT_DEF){
-	error("No GITHUB_PAT or GITHUB_PAT_DEF found!  Follow the steps at https://github.com/jasp-stats/jasp-desktop/blob/development/Docs/development/jasp-building-guide.md$${LITERAL_HASH}github_pat")
-	
-	GITHUB_PAT_DEF="NO_GITHUB_PAT_FOUND"	#To indicate a clearer error
+!exists(/app/lib/*){
```
JorisGoosen: Why do you do this?

vandenman: well, we cannot set a GITHUB_PAT on flatpak, but we also don't need one.

JorisGoosen: we *do* need one, to install modules... I already ran into that error (but only after also missing `gcc` and `gfortran` so I didnt mention it)

vandenman: okay, then I'll look into some other way to do this!

vandenman: I went for the base64 idea, let's hope GitHub won't notice

Tools/flatpak/setup-rpkgs/R/flatpakGeneratePkgsList.R
```diff
@@ -0,0 +1,60 @@
+# This file needs internet!
+
+# TODO: look at https://github.com/andrie/miniCRAN/issues/50#issuecomment-374624319
+# it would be nice to be able to outsource this to miniCRAN which has a method for adding local pkgs to a repository
+# the only downside is that it looks like miniCRAN is not actively maintained anymore.
```
JorisGoosen, line 5: fork it?

Tools/flatpak/setup-rpkgs/R/functions.R
```diff
@@ -0,0 +1,775 @@
...
+}
+
+uploadTarArchive <- function(archivePath = "archives/flatpak_archive.tar.gz") {
+
+  if (!file.exists(archivePath))
+    stop("Archive does not exist")
+
+  archivePath <- normalizePath(archivePath)
+
+  cmd <- sprintf("scp %s jonathonlove@static.jasp-stats.org:static.jasp-stats.org/flatpak_archive.tar.gz", archivePath)
```
JorisGoosen: Aah yeah as mentioned above: maybe add like the commit from jasp-desktop to the name or something? So we can have multiple versions on the server.

Tools/flatpak/setup-rpkgs/R/flatpakGeneratePkgsList.R
```diff
@@ -0,0 +1,60 @@
...
+copyRfiles(dirs)
+
+# debugonce(createTarArchive)
+info <- createTarArchive(dirs, verbose = FALSE, compression = "best")
+
+# update Rpackages.json & install build flatpak
+writeRpkgsJson("~/github/flatpak/org.jaspstats.JASP/RPackages.json", info)
+
+# IF you have ssh setup this will upload the tar.gz to static-jasp. It's nicer to do this via a terminal because there you see a progress bar
+# uploadTarArchive()
```
JorisGoosen: Maybe always do the upload?

The  use-case of this script is that we want to upload this thing right?

Also, does this override the previously used zip on static?
Because that would make it impossible to rebuild an older version later on. 

I checked functions.R and it doesnt so would be good to change that.

JorisGoosen (changes requested):
So `functions.R` is a little bit too big for me to check properly.

but these changes do get flatpak to build so that is pretty awesome, I would like to see the flatpakhelper zip get a unique name though.

Tools/flatpak/setup-rpkgs/R/flatpakGeneratePkgsList.R
vandenman: > Also, does this override the previously used zip on static?
> Because that would make it impossible to rebuild an older version later on.

Good point, I overwrote the older one to avoid spamming the server but I think I can also just put in the JASP commit in the name so that we have a unique zip per version.

Tools/flatpak/setup-rpkgs/R/flatpakGeneratePkgsList.R
vandenman: We could do that, but in hindsight, I'm not 100% sure how `miniCRAN` interacts with `renv` because that always tries to access GitHub if the DESCRIPTION specifies it in its remotes. Instead, I created a fake local GitHub website for renv by redirecting the URLs it wants to "download" to valid local files (in a way that both https://api.github.com/repos/jasp-stats/jaspAnova and https://api.github.com/repos/jasp-stats/jaspAnova/commits/master are valid files).

The nice thing about using something like `miniCRAN`, however, is that we could outsource creating a valid CRAN repository. For example, there are a few exceptions I had to fix manually (e.g., downloading recommended R packages in a separate step) and `miniCRAN` *should* fix that for us.

vandenman:
functions.R is indeed a bit too big. I'm not really sure how to fix that though. I'll implement the feedback and rerequest a review once I'm ready!

Tools/flatpak/setup-rpkgs/R/flatpakGeneratePkgsList.R
JorisGoosen: Good to keep in mind then

JorisGoosen:
Sometimes files are big...

And great, I'll check it out when you are done.

vandenman:
@JorisGoosen the changes in the last commit make sure that
- the name of the archive changes based on the commit hash/ tag of the checked-out jasp-desktop branch.
- the tar.gz file is always uploaded.

Depending on your internet speed, I'd still do the uploading manually in a terminal because you won't get a progress bar in R.

JorisGoosen (approved):
I'll be testing it later when we make the next beta version

PR #4559, opened 2021-07-05
Make sentence about Linux translatable

JoKeyser:
Currently, the hint about spreadheets in  Linux can't be translated. Also, Linux is a name, so it should be capitalized :).


JorisGoosen (approved):

JorisGoosen:
Merci!

PR #4560, opened 2021-07-05
Remove the Undo/Redo functions in the plot editor, and make a proper Cancel function

boutinb:
- The buttons 'Finish' and 'Discard changes' are quite unusual and can be confusing. They are replaced by the more conventional 'OK' and 'Cancel' buttons.
- The Undo/Redo buttons do not work well. Just remove them, and this function can be set in another release.
- In the plot editing, the plot is displayed to fit the height of the popup window.
- The width of the settings panel is now proportional with the zoom setting.
- The OK/Finish button did not work well: the settings were not set properly in the results.


Desktop/analysis/analysis.cpp
```diff
@@ -137,13 +137,18 @@ void Analysis::remove()
...
-	_results	= results;
 	_progress	= progress;
-	_resultsMeta		= _results.get(".meta", Json::arrayValue);
 
 	setStatus(status);
 
-	emit resultsChangedSignal(this);
+	if (_perform != performType::editImg)
+	{
+		// When image edit is processed by the engine, it returns a temporary result with old values...
```
JorisGoosen: Would it not make more sense to **not** return a temporary result with old values?

it doesn't seem very transparent that a result would be passed all the way here and then be ignored. Probably better is: send `Json::null` as results and then here check if it isn't null and if so apply otherwise ignore.

Or something. Because suppose in the future we want results from editImage, how long will it take before someone finds out that the results are simply ignored here in Analysis?

Like, why is there the temporary old result coming in?

TimKDJ:
Can you add (back) the `Restore original` button (previously a.k.a. `Reset plot` then `Original plot` and then `Discard changes` ^.^)?
It should be possible to return the original plot as well as cancel the last set of changes.
See also https://github.com/jasp-stats/INTERNAL-jasp/issues/1191#issuecomment-857519960

JorisGoosen (changes requested):

boutinb:
This PR must be implemented with PR's:

- https://github.com/jasp-stats/jaspBase/pull/37
- https://github.com/jasp-stats/jaspResults/pull/1

PR #4563, opened 2021-07-07
Fix question punctuation: Remove "." after "?"

JoKeyser:
Hi, just another small drive-by fix...


boutinb (approved):

boutinb (changes requested):
I'm afraid that the translations made will be lost afterwards. You can directly change the po and pot files (in Desktop/po folder), to avoid the current translation to be removed.
I can do that also if you prefer.

JoKeyser:
@boutinb , oh yes, I had not thought about that problem. I will try to add these changes into the `.po` and `.pot` files later.

JoKeyser:
@boutinb , please have another look.

boutinb (approved):

PR #4567, opened 2021-07-15
Use a bigger buffer to read CSV

boutinb:
Solves jasp-stats/jasp-issues#1359


JorisGoosen (approved):

JorisGoosen:
Not superfutureproof, could have been nice to use a std::vector and resize as needed but this will work for now.

PR #4571, opened 2021-07-16
Plot Editor Last Changes?

boutinb:
Fix one bug: when changing the Y axe, the X axe is then displayed.
Implement suggestions of https://github.com/jasp-stats/INTERNAL-jasp/issues/1191#issuecomment-879803307

- Make the buttons rounder
- Add the hand cursor when the pointer hovers a control/button
- Rename “Save plot as” to “Save image as” 

For the last suggestion 'add padding to the left and right edges': I'm not convinced... but if the majority wants it, I'll do it.

### DIFF
--- a/Desktop/components/JASP/Widgets/PlotEditor/PlotEditor.qml
+++ b/Desktop/components/JASP/Widgets/PlotEditor/PlotEditor.qml
@@ -385,7 +385,7 @@ Popup
 					Rectangle
 					{
 						id:					plotImgRect
-						color:				jaspTheme.white
+						color:				jaspTheme.themeName === "darkTheme" ? "black" : jaspTheme.white
 						border.color:		jaspTheme.uiBorder
 						border.width:		1
 						anchors.fill:		parent
--- a/Desktop/components/JASP/Widgets/PlotEditor/PlotEditor.qml
+++ b/Desktop/components/JASP/Widgets/PlotEditor/PlotEditor.qml
@@ -38,6 +38,7 @@ Popup
 			border.color:	jaspTheme.uiBorder
 			border.width:	preferencesModel.uiScale
 
+			/* This prevents the cursor to get its right shape.
 			MouseArea
 			{
 				id:				focusCatcher	//Because then people can click away from the axistableview or something
@@ -46,6 +47,7 @@ Popup
 				z:				100
 				
 			}
+			*/
 
 			JASPC.Text
 			{
@@ -138,6 +140,13 @@ Popup
 										horizontalAlignment: Text.AlignHCenter
 										verticalAlignment:	Text.AlignVCenter
 										opacity:			checked ? 1 : .6
+									}
+
+									MouseArea
+									{
+										anchors.fill	: parent
+										cursorShape		: checked ? Qt.ArrowCursor : Qt.PointingHandCursor
+										acceptedButtons	: Qt.NoButton
 									}
 								}
 							}
@@ -325,7 +334,7 @@ Popup
 					bottom:			parent.bottom
 					margins:		jaspTheme.generalAnchorMargin
 				}
-				text:				qsTr("Save plot as")
+				text:				qsTr("Save image as")
 				onClicked:			plotEditorModel.savePlot()
 			}
 
@@ -380,7 +389,7 @@ Popup
 						// Later we can make it so that it only shows the pointing hand when it overlaps a clickable element (which informs the user and makes it feel a bit more interactive)
 						id:				elementPicker
 						anchors.fill:	parent
-						cursorShape:	Qt.PointingHandCursor
+						cursorShape:	Qt.ArrowCursor
 						onClicked:
 						{
 							forceActiveFocus();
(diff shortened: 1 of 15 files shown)


JorisGoosen:
You've got to think of dark mode too: 
afbeelding

Desktop/components/JASP/Widgets/PlotEditor/PlotEditor.qml
```diff
@@ -38,6 +38,7 @@ Popup
 			border.color:	jaspTheme.uiBorder
 			border.width:	preferencesModel.uiScale
 
+			/* This prevents the cursor to get its right shape.
```
JorisGoosen: Sure but now you can no longer click out of the tableview... 
Which is quite normal in my opinion, wouldn't it help to just put the z different?

JorisGoosen: Ok, I fixed it like:
```
Item
					{
						id:					flickChild
						width:				axesFlickable.width
						height:				stack.y + stack.height + jaspTheme.generalAnchorMargin
						
						MouseArea
						{
							id:				focusCatcher	//Because then people can click away from the axistableview or something
							onPressed:		{ forceActiveFocus(); mouse.accepted = false; }
							anchors
							{
								top:		parent.top
								left:		parent.left
								right:		parent.right
							}
							height:			Math.max(flickChild.height, axesFlickable.height)
							z:				-100
						}
```

The flickChild thing is just there to show the location where it ought to be

JorisGoosen (changes requested):
The background should fit with the theme (Im guessing its anchors because the `jaspTheme.uiBackground` ought to be good)

Also being able to "undo" editing a cell in the manual-table buy clicking away should remain possible so Ive shown you how to fix it.

Desktop/results/ploteditormodel.cpp
```diff
@@ -29,43 +29,71 @@ PlotEditorModel::PlotEditorModel()
...
+		return;
+	}
+
+	Json::Value	xAxis	=	editOptions.get(	"xAxis",		Json::objectValue),
+				yAxis	=	editOptions.get(	"yAxis",		Json::objectValue);
+
+	_xAxis->setAxisData(xAxis);
+	_yAxis->setAxisData(yAxis);
+
+	_coordinates.loadCoordinates(editOptions.get("coordinates", Json::objectValue)); // To Do Vincent Pedata: is this the right json object?
```
JorisGoosen: I do not think vincent will be working on it anymore :p

TimKDJ:
First off, I think it looks very nice. I've voiced my misgivings about making "Show title" and "Show axis" so prominent already, so I won't go there again. But the following would be good to address:

- [x] As Joris also noted earlier, pressing outside of textfields does nothing anymore when you're in the process of editing them.
- [ ] If I create a Distribution plot and immediately click on "Reset defaults" then the y-axis title disappears...
- [x] If I click quickly on the "plus" to add two more columns in the Ticks table, then the `+ x +`-symbols will jump first a few columns ahead and then jump to the column previous to the one I was clicking on.

And I wonder if `Treat title as R expression` or `Parse title as R expression` explains better what it does than `Use R expression in title`?

boutinb:
> * If I create a Distribution plot and immediately click on "Reset defaults" then the y-axis title disappears...

I could not reproduce this problem.
But maybe it is solved with my last commit

JorisGoosen:
Im testing now and Ive got one small nitpick:
afbeelding

It would be better if the plot either has a black/white background (instead of uiBackground) *or* a border.
Now it just looks a little bit off.

However like so:
afbeelding

is better right?

Just change line 388 to `color:				jaspTheme.themeName === "darkTheme" ? "black" : jaspTheme.white`

TimKDJ (approved):
> I could not reproduce this problem.
But maybe it is solved with my last commit

I can no longer reproduce this either, so let's assume it's fixed.
The other two issues are definitely fixed.

I did find a different issue (https://github.com/jasp-stats/jasp-test-release/issues/1385), but I don't think that necessarily needs to be included in this PR.

So for me it'd be ok to merge this.

JorisGoosen:
Also I just closed JASP and I got this crash:
```
Thread 0 Crashed:: CrBrowserMain  Dispatch queue: com.apple.main-thread
0   JASP                          	0x000000010b4e7498 Analysis::setStatus(Analysis::Status) + 168 (analysis.cpp:526)
1   JASP                          	0x000000010b4e8475 Analysis::editImage(Json::Value const&) + 37 (analysis.cpp:197)
2   JASP                          	0x000000010b92cb49 PlotEditor::PlotEditorModel::updatePlot(Json::Value&) + 185 (ploteditormodel.cpp:136)
3   JASP                          	0x000000010b92ca86 PlotEditor::PlotEditorModel::cancelPlot() + 86 (ploteditormodel.cpp:84)
4   JASP                          	0x000000010bad548d PlotEditor::PlotEditorModel::qt_static_metacall(QObject*, QMetaObject::Call, int, void**) + 989 (moc_ploteditormodel.cpp:280)
5   JASP                          	0x000000010bad67ad PlotEditor::PlotEditorModel::qt_metacall(QMetaObject::Call, int, void**) + 125 (moc_ploteditormodel.cpp:485)
...
```

But I cannot reproduce it anymore so I don't know what happened there...
I'll keep testing though

JorisGoosen:
For the rest everything seems fine!

The crash seems like a weird occurence and only happened after I closed JASP anyhow so we can ignore that for now.
If you could apply that backgroundcolour change I suggested I am ok with merging this.

PR #4574, opened 2021-07-20
Fixes levels/factors in ANOVA Bay RM

JorisGoosen:
- Fixes https://github.com/jasp-stats/jasp-test-release/issues/1344 and https://github.com/jasp-stats/jasp-test-release/issues/1311
- Partially this was caused by the anova code removing the dots in the "JaspExtraOptions_.1._Encoded" strings, which broke decoding.. So no more dots there
- And the order of binding and getting the factors in FactorLevelListBase was wrong which broke it after reloading a jasp-file


TimKDJ (changes requested):
Never mind me, I hadn't run submodule update.
All issues are fixed.

TimKDJ (approved):

Desktop/widgets/factorlevellistbase.cpp
```diff
@@ -56,6 +54,8 @@ void FactorLevelListBase::bindTo(const Json::Value& value)
 	}
 	
 	_factorLevelsModel->initFactors(factors);
+	
+	BoundControlBase::bindTo(value);
```
boutinb, line 58: Why does it help?

JorisGoosen: the `bindTo` call ultimately triggers `Json::Value FactorLevelListBase::createMeta()` in calculating the `boundValue` for this control. But it only does so once (unless the value changes of course) and is then cached.

See:
```
Json::Value FactorLevelListBase::createMeta()
{
	Json::Value meta(BoundControlBase::createMeta()); //Work from parent to get everything it defines
	
	auto factors = _factorLevelsModel->getFactors();
	
	if(factors.size() == 0)
		return meta;
	
	meta["encodeThis"] = Json::arrayValue;
	for (const auto & factor : factors)
	{
		meta["encodeThis"].append(factor.first);
		
		for(const auto & level : factor.second)
			meta["encodeThis"].append(level);
	}
	
	return meta;
}
```

If `bindTo` is called before `initFactors` the `auto factors = _factorLevelsModel->getFactors();` is pretty empty.

PR #4578, opened 2021-07-22
Fixes the eye in the filter

JorisGoosen:
- fixes https://github.com/jasp-stats/jasp-test-release/issues/1388

Also, @boutinb why did you add that function in the first place? 
Because `DataSetTableModel` is a `DataSetTableProxy` all the filtering (and sorting) is handled by the class itself, `data` is already defined and works fine. Checkout https://doc.qt.io/qt-5/qsortfilterproxymodel.html


boutinb (approved):

boutinb:
> Also, @boutinb why did you add that function in the first place?

I think it's obsolete indeed.

PR #4579, opened 2021-07-22
JASPWidgets["title"] does not exist anymore

boutinb:
Solves jasp-stats/jasp-test-release#1380
Solves jasp-stats/jasp-test-release#1379
Solves jasp-stats/jasp-test-release#1377
Solves jasp-stats/jasp-test-release#1371

I'm not at all sure if this is the right thing to do! 
I just saw that the "title" type is called and that JASPWidgets does not have this type....

### DIFF
--- a/Desktop/html/js/object.js
+++ b/Desktop/html/js/object.js
@@ -18,7 +18,7 @@ JASPWidgets.objectConstructor = function (results, params, ignoreEvents) {
 	if (metaData.type)
 		type = metaData.type;
 
-	if (type === "qmlSource")
+	if (type === "qmlSource" || type === "title")
 		return null;
 
 	if(_.has(results, "title") && results.title === "" && type === "collection")


JorisGoosen:
But this "title" thing isn't called for newer files right?

So shouldn't we update the incoming "wrong" data then? Instead of changing the javascript?

boutinb:
> But this "title" thing isn't called for newer files right?

Apparently, but I could not see in the log when this has happened, so I wondered if you or @TimKDJ knew why this has changed.

JorisGoosen:
Hmm, well, I will have a look at the original issue later and see what I think of it.

JorisGoosen:
Well, it solves the problem and after refreshing in the 0.15 version it also works without the fix so this is good to go.

JorisGoosen (approved):

JorisGoosen:
I am adding a comment though

PR #4588, opened 2021-08-18
JAGS: Encode the column names found in the model before sending them to R

vandenman:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/1468

So I looked at this and noticed this:

image

it's trying to read in `contBinom` but R should have gotten the encoded name, not `contBinom`.

I think that it used to be that after
```c++
_textEncoded = tq(ColumnEncoder::columnEncoder()->encodeRScript(stringUtils::stripRComments(fq(text)), &_usedColumnNames));
```
`_usedColumnNames` contained encoded column names. However, now it contains the decoded names. So this PR just encodes those. It's also possible that this always returned decoded names, in that case, JAGS has been broken for a long, long time. I'm not sure how nobody noticed that during the last testing round though.

### DIFF
--- a/Desktop/widgets/boundcontroljagstextarea.cpp
+++ b/Desktop/widgets/boundcontroljagstextarea.cpp
@@ -101,7 +101,7 @@ void BoundControlJAGSTextArea::checkSyntax()
 	boundValue["model"] = _textEncoded.toStdString();
 	Json::Value columns(Json::arrayValue);
 	for (const std::string& column : _usedColumnNames)
-		columns.append(column);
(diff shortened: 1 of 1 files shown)


JorisGoosen:
The `usedColumns` as returned from `encodeRScript` have probably always been the decoded ones, because they are the same that are used to determine which columns are included in (originally) the filter so that the engine knew which columns to request for the dataset.

Not sure whether it worked before but if it does after this fix all is well.

Your solution seems straightforward enough!

JorisGoosen (approved):

PR #4593, opened 2021-08-23
Make sure help is shown/unshown correctly for both markdown-generated…

JorisGoosen:
…/helpfile analyses

- fixes https://github.com/jasp-stats/jasp-test-release/issues/1501

### DIFF
--- a/Desktop/utilities/helpmodel.cpp
+++ b/Desktop/utilities/helpmodel.cpp
@@ -53,19 +53,18 @@ void HelpModel::setMarkdown(QString markdown)
 {
 	if (_markdown == markdown)
 		return;
-
+	
+	if(markdown != "")
+		_pagePath = "";
+	
 	_markdown = markdown;
+	
 	emit markdownChanged(_markdown);
 }
 
 void HelpModel::setPagePath(QString pagePath)
-{
-	setMarkdown("");
-	_analysis = nullptr;
-	
-	pagePath = convertPagePathToLower(pagePath); //Otherwise we get into to trouble on systems that discern between cases in the filesystem. Also means we should make sure all documentation html are named lowercase!
-
-	_pagePath = pagePath;
+{	
+	_pagePath = convertPagePathToLower(pagePath); //Otherwise we get into to trouble on systems that discern between cases in the filesystem. Also means we should make sure all documentation html are named lowercase!
 	emit pagePathChanged(_pagePath);
 }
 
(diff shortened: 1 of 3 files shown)


boutinb:
Hou have added the jaspRegression & jaspSummayStatistics modules in your commit.

Desktop/utilities/helpmodel.cpp
```diff
@@ -134,36 +133,36 @@ void HelpModel::showOrTogglePage(QString pagePath)
...
-		return;
-	}
-	
 	QString renderFunc, 
 			contentMD,
 			pagePath = convertPagePathToLower(helpPage == "" ? analysis->helpFile() : analysis->fullHelpPath(helpPage));
 	
-	
-	if(analysis == _analysis && pagePath == _pagePath && _visible)
+	if(!analysis || (analysis == _analysis && pagePath == _pagePath && _visible))
```
boutinb: You test here whether the analysis pointer is not null, but you use this analysis pointer in the line before...

boutinb: pagePath should be set in the else expression

boutinb (changes requested):

Desktop/utilities/helpmodel.cpp
JorisGoosen: Oops

JorisGoosen:
> Hou have added the jaspRegression & jaspSummayStatistics modules in your commit.

This was sort of on purpose aas they were part of the fix but are now in development anyhow.

Desktop/utilities/helpmodel.cpp
JorisGoosen: I just wanted to remove the duplication...

JorisGoosen:
It should be good now

boutinb:
If I test it from your branch, I still get same problem as EJ: If I click on the help of the Bayesian Log-Linear Regression, I first get a (really short) help, the secind time I don't get anything. If I then go to another analysis, the first time I click on the icon I get the help, I close the help pop-up, then I click again on the help icon: the pop-up does not show up. If I click again, it does show up.

JorisGoosen:
> get a (really short) help

Did you do `git submodule update` ?

Also, I just tested this code here and it works. I can only suspect you either got the wrong branch/didnt-build properly.

Maybe you could share the output of `git log` and `git status` here?

Edit: I've doublechecked it and it does in fact fix the problem for me...

boutinb (approved):

PR #4595, opened 2021-08-24
plot editing: added help button + first draft at helpfile

vandenman:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/1474

Here's the help button in the top right:
image

Suggestions for different locations are welcome! I do think it makes sense to reuse the same button we use in the analyses (even though we don't have any other buttons in plot editing).

Here's a web link to the help file: https://htmlpreview.github.io/?https://github.com/vandenman/jasp-desktop/blob/plotEditingHelpfile/Resources/Help/other/plotediting.html

@JorisGoosen so I used the \<code>\</code> tags because they look nice in the browser, but I noticed they don't show up at all in the help file viewer in JASP... any suggestions for alternatives?

### DIFF
--- a/Resources/Help/other/plotediting.html
+++ b/Resources/Help/other/plotediting.html
@@ -1,0 +1,105 @@
+<h1>Plot Editing</h1>
+
+<p>
+	Although we try our best, sometimes plots created by JASP need some tweaks.
+	For that purpose, we provide some rudimentary plot editor.
+	The current editor allows you to modify the axes of a plot, but future versions of JASP should provide more options.
+</p>
+
+<h3>Basic options</h3>
+
+<p>
+	In the top left you can select which axis you want to modify.
+	You can select either the x-axis or the y-axis.
+	All of the options below modify either the x-axis or the y-axis, depending on which is currently selected.
+</p>
+
+<ul>
+	<li><code>Show title.</code> Should the title of the selected axis be shown?</li>
+	<li><code>Title.</code> Specify what to put on the axis title.</li>
+</ul>
+
+<h4>Ticks</h4>
+<p>
+	Aside from a title, an axis also has ticks, which consist of two components: positions and labels.
+	Very often, but not always, the labels simply show the positions, for example, a position at -4 is indicated with the label "-4".
+</p>
+
+<p>
+	If you want that the labels and positions are the same, it is convenient to select <code>Specify sequence</code>.
+	If you select <code>Specify sequence</code> then the ticks are fully determined by three numbers.
+	The left and right endpoints are determined by <code>from</code> and <code>to</code> respectively.
+	After every <code>steps</code> numbers, there is a label.
+	For example if <code>from</code> is -4, <code>to</code> is 4, and <code>steps</code> is 2, then you would get ticks at -4, -2, 0, 2, 4.
+</p>
+
+<p>
+	Sometimes you may want that the labels are different from the positions, or maybe you do not want equally spaced labels.
+	In that case, you can select <code>Set manually</code>.
+	Once selected, a table is shown that allows for individually modifying the positions and labels.
+	If you left-click on a column, two green + icons appear and one red cross.
+	The left and right + icons allow you to insert a new position and label on the left and right of the selected column.
+	The red cross allows you to delete the selected column.
+</p>
+
+
+<p>Note that for discrete axes, the ticks cannot be adjusted.</p>
+
+
+<h3>Advanced options</h3>
+
+<code>Parse title as R expression.</code>
+Some figures have complicated titles, which cannot be typeset in plain text.
+For example, to put the symbol δ (delta) on an axis you can type "delta" and enable this option.
+However, you can also create more complicated equations, such as <code>over(alpha, beta)</code>.
+For more details, see <a href="https://stat.ethz.ch/R-manual/R-devel/library/grDevices/html/plotmath.html">here</a> for more details.
+
+<h4>Limits</h4>
+
+The limits consist of two values, called <code>Upper limit</code> and <code>Lower limit</code>, which determine the outer end points of an axis or plotable area.
+For example, even though the axis positions are going from 80 to 100 in steps of 10, it might be that there is a data point at 102.
+The limits provide some leeway to show all data without being forced to increase <code>to</code> to 105.
+
+If there is 
+
+<ul>
+	<li>
+		<code>Based on data.</code>
+		Uses ggplot2's default choices for the limits which usually means that all data is shown, despite the choice ticks.
+	</li>
+	<li>
+		<code>Based on ticks.</code>
+		Automatically rescale the limits with the ticks. 
+		For example, if you increase <code>to</code> to some valuer larger than the <code>Upper limit</code>, this will automatically set <code>Upper limit</code> to <code>to</code>.
+	</li>
+	<li>
+	<code>Set manually.</code>
+	Specify two values for the <code>Upper limit</code> and <code>Lower limit</code> manually.
+	</li>
+</ul>
+When adjusting the limits manually, take these two warnings to heart:
+
+<ol>
+	<li>Any data that falls outside of the limits is not shown.</li>
+	<li>
+		To adjust the <code>Ticks</code> you now also must adjust the <code>Limits</code>.
+		For example, if the <code>Upper limit</code> is set to 4, and you now set <code>to</code> to 8, then no ticks after 4 are shown.
+	</li>
+</ol>
+ 
+
+<h3>References</h3>
+
+<p>
+	Almost all plots in JASP use the R package ggplot2.
+	This package is an implementation of the grammer of graphics.
+	For more information, see
+</p>
+
+<ul>
+	<li>Wickham, H. (2016). <i><a href="https://ggplot2-book.org/">ggplot2: Elegant Graphics for Data Analysis</a></i> (3rd ed.). Springer-Verlag.
+	<li><a href="https://ggplot2.tidyverse.org/index.html">https://ggplot2.tidyverse.org/index.html</a></li>
+	<li>Wilkinson, L. (2005). <i><a href="https://dx.doi.org/10.1007/0-387-28695-0">The grammar of graphics</a></i> (2nd ed.). Springer.</li>
+	<li><a href="https://stat.ethz.ch/R-manual/R-devel/library/grDevices/html/plotmath.html">https://stat.ethz.ch/R-manual/R-devel/library/grDevices/html/plotmath.html</a></li>
+</ul>
+
--- a/Resources/Help/other/plotediting.html
+++ b/Resources/Help/other/plotediting.html
@@ -89,7 +89,7 @@ When using <code>Set manually</code>, take these two warnings to heart:
 	<li>Any data that falls outside of the limits is not shown.</li>
 	<li>
 		To adjust the <code>Ticks</code> you now also must adjust the <code>Limits</code>.
-		For example, if the <code>Upper limit</code> is set to 4, and you now set <code>to</code> to 8, then no ticks after 4 are shown.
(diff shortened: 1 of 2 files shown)


EJWagenmakers:
We would usually have that icon in the GUI itself, but in this case we have two tabs. From the user's point of view, I do think it would be more obvious if it was next to the y-axis tab instead of all the way to the right where it may easily escape attention.

JorisGoosen:
Well, it *is* in the gui right?

My 2c is that this is a good place.

vandenman:
@EJWagenmakers do you mean something like this?

image

if not, please draw a preferred location with paint.

A downside of this location is that it will most likely move as more tabs are added.

EJWagenmakers:
Yes, I like that better

JorisGoosen:
> A downside of this location is that it will most likely move as more tabs are added.

It's not hardcoded right?
Because translations.

vandenman:
> It's not hardcoded right?

Nope, see the pushed changes. But it's likely that there will be at least 4 more tabs (top axis, bottom axis, layers, theme), so it might get a little crowded.

JorisGoosen:
Great, and if it gets crowded we can always move it somewhere else.

Resources/Help/other/plotediting.html
```diff
@@ -0,0 +1,105 @@
+<h1>Plot Editing</h1>
+
+<p>
+	Although we try our best, sometimes plots created by JASP need some tweaks.
+	For that purpose, we provide some rudimentary plot editor.
```
TimKDJ:
```suggestion
```

Resources/Help/other/plotediting.html
```diff
@@ -0,0 +1,105 @@
...
+
+<h3>Basic options</h3>
+
+<p>
+	In the top left you can select which axis you want to modify.
+	You can select either the x-axis or the y-axis.
+	All of the options below modify either the x-axis or the y-axis, depending on which is currently selected.
+</p>
+
+<ul>
```
TimKDJ, line 17:
```suggestion
<h4>Title</h4>
<ul>
```

Resources/Help/other/plotediting.html
```diff
@@ -0,0 +1,105 @@
...
+	<li><code>Show title.</code> Should the title of the selected axis be shown?</li>
```
TimKDJ: For all code blocks I would change `.</code>` to `</code>:`

Resources/Help/other/plotediting.html
```diff
@@ -0,0 +1,105 @@
...
+	<li><code>Title.</code> Specify what to put on the axis title.</li>
```
TimKDJ: I think it'd be good to refer to the `Parse title as R expression` option here

Resources/Help/other/plotediting.html
```diff
@@ -0,0 +1,105 @@
...
+</ul>
+
+<h4>Ticks</h4>
+<p>
+	Aside from a title, an axis also has ticks, which consist of two components: positions and labels.
```
TimKDJ:
```suggestion
	Ticks consist of two components: positions and labels.
```

Resources/Help/other/plotediting.html
```diff
@@ -0,0 +1,105 @@
...
+	Very often, but not always, the labels simply show the positions, for example, a position at -4 is indicated with the label "-4".
+</p>
+
+<p>
+	If you want that the labels and positions are the same, it is convenient to select <code>Specify sequence</code>.
```
TimKDJ:
```suggestion
	If you want the labels and positions to be the same, it is convenient to select <code>Specify sequence</code>.
```

Resources/Help/other/plotediting.html
```diff
@@ -0,0 +1,105 @@
...
+	If you select <code>Specify sequence</code> then the ticks are fully determined by three numbers.
+	The left and right endpoints are determined by <code>from</code> and <code>to</code> respectively.
+	After every <code>steps</code> numbers, there is a label.
```
TimKDJ:
```suggestion
	After every <code>steps</code> numbers, there will be a label.
```

Resources/Help/other/plotediting.html
```diff
@@ -0,0 +1,105 @@
...
+	For example if <code>from</code> is -4, <code>to</code> is 4, and <code>steps</code> is 2, then you would get ticks at -4, -2, 0, 2, 4.
+</p>
+
+<p>
+	Sometimes you may want that the labels are different from the positions, or maybe you do not want equally spaced labels.
```
TimKDJ:
```suggestion
	Sometimes you may want the labels to differ from the positions, or maybe you do not want equally spaced labels.
```

Resources/Help/other/plotediting.html
```diff
@@ -0,0 +1,105 @@
...
+	In that case, you can select <code>Set manually</code>.
+	Once selected, a table is shown that allows for individually modifying the positions and labels.
+	If you left-click on a column, two green + icons appear and one red cross.
```
TimKDJ:
```suggestion
	If you left-click on a column, two green plus icons appear and one red cross.
```

Resources/Help/other/plotediting.html
```diff
@@ -0,0 +1,105 @@
...
+	The left and right + icons allow you to insert a new position and label on the left and right of the selected column.
```
TimKDJ:
```suggestion
	The left and right plus icons allow you to insert a new position and label on the left and right of the selected column.
```

Resources/Help/other/plotediting.html
```diff
@@ -0,0 +1,105 @@
...
+
+<code>Parse title as R expression.</code>
+Some figures have complicated titles, which cannot be typeset in plain text.
+For example, to put the symbol δ (delta) on an axis you can type "delta" and enable this option.
+However, you can also create more complicated equations, such as <code>over(alpha, beta)</code>.
+For more details, see <a href="https://stat.ethz.ch/R-manual/R-devel/library/grDevices/html/plotmath.html">here</a> for more details.
+
+<h4>Limits</h4>
+
+The limits consist of two values, called <code>Upper limit</code> and <code>Lower limit</code>, which determine the outer end points of an axis or plotable area.
```
TimKDJ:
```suggestion
The limits consist of two values: <code>Upper limit</code> and <code>Lower limit</code>. These values determine the outer end points of an axis or plotable area.
```

Resources/Help/other/plotediting.html
```diff
@@ -0,0 +1,105 @@
...
+For example, even though the axis positions are going from 80 to 100 in steps of 10, it might be that there is a data point at 102.
+The limits provide some leeway to show all data without being forced to increase <code>to</code> to 105.
```
TimKDJ, line 66: Should we add something here that explains https://github.com/jasp-stats/jasp-desktop/pull/4607?

Resources/Help/other/plotediting.html
```diff
@@ -0,0 +1,105 @@
...
+
+If there is
```
TimKDJ:
```suggestion
You can select one of the following options
```

Resources/Help/other/plotediting.html
```diff
@@ -0,0 +1,105 @@
...
+The limits consist of two values, called <code>Upper limit</code> and <code>Lower limit</code>, which determine the outer end points of an axis or plotable area.
+For example, even though the axis positions are going from 80 to 100 in steps of 10, it might be that there is a data point at 102.
+The limits provide some leeway to show all data without being forced to increase <code>to</code> to 105.
+
+If there is 
+
+<ul>
+	<li>
+		<code>Based on data.</code>
+		Uses ggplot2's default choices for the limits which usually means that all data is shown, despite the choice ticks.
```
TimKDJ:
```suggestion
		Use an automatic default for the limits that is determined based on the data instead of following the value of the ticks.
```
Not really sure, but I wouldn't mention ggplot2 and it's good to touch on the consequences of choosing this

Resources/Help/other/plotediting.html
```diff
@@ -0,0 +1,105 @@
...
+	</li>
+	<li>
+		<code>Based on ticks.</code>
+		Automatically rescale the limits with the ticks. 
+		For example, if you increase <code>to</code> to some valuer larger than the <code>Upper limit</code>, this will automatically set <code>Upper limit</code> to <code>to</code>.
```
TimKDJ:
```suggestion
		For example, if you increase ticks' <code>to</code> to some value larger than the <code>Upper limit</code>, this will automatically set <code>Upper limit</code> to <code>to</code>.
```

Resources/Help/other/plotediting.html
```diff
@@ -0,0 +1,105 @@
...
+	Specify two values for the <code>Upper limit</code> and <code>Lower limit</code> manually.
+	</li>
+</ul>
+When adjusting the limits manually, take these two warnings to heart:
+
+<ol>
+	<li>Any data that falls outside of the limits is not shown.</li>
+	<li>
+		To adjust the <code>Ticks</code> you now also must adjust the <code>Limits</code>.
+		For example, if the <code>Upper limit</code> is set to 4, and you now set <code>to</code> to 8, then no ticks after 4 are shown.
```
TimKDJ: I don't understand this one

TimKDJ (commented):
Great that you added documentation! Super useful.

Personally, I'd prefer the help icon to be inside the input options panel I think (top right).
For aesthetic reasons, but also so we can then make it scroll / show output related to a specific tab once we add more.
If you guys prefer the current position that's also fine though.

I've also added some suggestions to the text.

Resources/Help/other/plotediting.html
```diff
@@ -0,0 +1,105 @@
...
+	The red cross allows you to delete the selected column.
+</p>
+
+
+<p>Note that for discrete axes, the ticks cannot be adjusted.</p>
+
+
+<h3>Advanced options</h3>
+
+<code>Parse title as R expression.</code>
```
TimKDJ: Oh and if you decide to add a header above the other title option (https://github.com/jasp-stats/jasp-desktop/pull/4595/files#r701049605), then it'd be good to also do that here

Resources/Help/other/plotediting.html
vandenman: I hope it's more clear now :sweat_smile:

Resources/Help/other/plotediting.html
vandenman: The behavior we should go for is not 100% clear to me. That pr basically avoids people from specifying that from < to unless steps is negative. But I'm not fully understanding why steps would ever be negative, and it would be easier to avoid that altogether.

vandenman:
@TimKDJ I implemented all the resolved suggestions

> Personally, I'd prefer the help icon to be inside the input options panel I think (top right).

An alternative is to anchor it to the scroll bar. But since very few other people gave any feedback I'm not sure what other people want.

vandenman:
@JorisGoosen @TimKDJ should we merge this or do you still want to take another look?

TimKDJ:
I think this can be merged, we can always make adjustments later

amirmasoudabdol:
@EJWagenmakers, I would actually leave the Help button where @vandenman proposed initially. Leaving it next to the x-axis and y-axis tab makes the impression that the help button wants to tell you something about those tabs, or the axis. In contrast, leaving it at the top indicates that it wants to tell you something about the overall plot editing functionality. 

As an alternative, if you want it to be more apparent, you can make it into a proper button, and put it on the left side of the Cancel button. This is a common practice, in slightly different configuration, but it's still an acceptable alternative here.

P.S. As a general note/observation, I see some inconsistencies in placement of these help buttons and I take a note to later be able to standardise them.

JorisGoosen (approved):
I agree with Amir that the original placement is good, but I already mentioned that.

Anyhow, regardless of placement I think this can be merged if you @vandenman feel like it.

PR #4607, opened 2021-08-31
Handle better the constraints inside the Plot Editor

boutinb:
The constraints of the 'from' and 'to' values in the ticks settings are reinforced so that the 'from' value is always smaller than the 'to' value (with exception when the steps is negative: this is a strange default behaviour of the Raincloud plots).
Also when setting manually the ticks, it is not anymore possible to set the value of position smaller than its left value and bigger than its right value.

When the ticks are changed (either by specifying the sequence or setting them manually), the manual Lower & Upper limits (in Advanced) are changed also automatically as long as these limits are not previously changed manually (in fact as long as the upper & lower limits correspond to the 'from' and 'to' values).
The limit type (based on data, ticks or set manually) is not changed anymore.
This solves https://github.com/jasp-stats/jasp-test-release/issues/1502
and part of https://github.com/jasp-stats/jasp-test-release/issues/1385

### DIFF
--- a/Common/utils.cpp
+++ b/Common/utils.cpp
@@ -442,6 +442,19 @@ void Utils::convertEscapedUnicodeToUTF8(std::string& inputStr)
 	}
 }
 
+bool Utils::isEqual(const double a, const double b)
+{
+	if (isnan(a) || isnan(b)) return false;
+
+	return (fabs(a - b) <= ( (fabs(a) < fabs(b) ? fabs(b) : fabs(a)) * DBL_EPSILON));
+}
+bool Utils::isEqual(const float a, const float b)
+{
+	if (isnan(a) || isnan(b)) return false;
+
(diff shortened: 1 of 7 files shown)


vandenman (commented):
This behavior for the limits is very nice!
But `Reset defaults` still doesn't always work. To replicate:

- Descriptives -> contNormal -> check Distribution plots.
- Set `from` to 0.
- Set `steps` to -1.
- Click ok on the error message.
- Set `to` to -2.

After the last step plot editing "worked" again (there was no error message). But if you now click on `Reset defaults` you go back to the last situation before the error message. I'm not 100% sure whether this goes wrong in R or in the GUI though. If it goes wrong in R then I should probably take a look at this in a separate PR.

boutinb:
> But `Reset defaults` still doesn't always work.

Yes I know, I want to make another PR for this problem. I think the problem lies somewhere else in the code.

vandenman (approved):

PR #4614, opened 2021-09-08
Replacing limit macros with numeric limits

amirmasoudabdol:
This is trying to address jasp-stats/jasp-issues#1410 issue. I basically replaced all the limit macros with their equivalent `numeric_limits` from `<limits>` header. It's somewhat a moderner C++ I guess 🤷🏻‍♂️ and it should be portable as long as we are building with C++11. 

P.S. I truly hope that I haven't messed up the submodule syncing. I think I handled it correctly, but I really cannot tell!

### DIFF
--- a/Common/column.cpp
+++ b/Common/column.cpp
@@ -65,7 +65,7 @@ void Column::_convertVectorIntToDouble(vector<int> &intValues, vector<double> &d
 	{
 		const int &intValue = *it;
 		double doubleValue = intValue;
-		if (intValue == std::numeric_limits<int>::min())
+		if (intValue == std::numeric_limits<int>::lowest())
 			doubleValue = NAN;
 		doubleValues.push_back(doubleValue);
 	}
@@ -87,7 +87,7 @@ bool Column::_resetEmptyValuesForNominal(std::map<int, string> &emptyValuesMap)
 	for (; ints != end; ints++)
 	{
 		int intValue = *ints;
-		if (intValue == std::numeric_limits<int>::min() && hasEmptyValues)
+		if (intValue == std::numeric_limits<int>::lowest() && hasEmptyValues)
 		{
 			auto search = emptyValuesMap.find(row);
 			if (search != emptyValuesMap.end())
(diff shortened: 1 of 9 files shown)


JoKeyser:
Compiling works with GCC 11.1.0! (I can't comment about anyhing else.)

Common/column.cpp
```diff
@@ -1014,8 +1014,8 @@ string Column::_getScaleValue(int row)
 {
 	double v = AsDoubles[row];
 
-	if (v > DBL_MAX)					return string({ (char)0xE2, (char)0x88, (char)0x9E, 0 });
-	else if (v < -DBL_MAX)				return string({ (char)0x2D, (char)0xE2, (char)0x88, (char)0x9E, 0 });
+	if (v > std::numeric_limits<double>::max())					return string({ (char)0xE2, (char)0x88, (char)0x9E, 0 });
+	else if (v < std::numeric_limits<double>::min())			return string({ (char)0x2D, (char)0xE2, (char)0x88, (char)0x9E, 0 });
```
vandenman:
```suggestion
	else if (v < std::numeric_limits<double>::lowest())			return string({ (char)0x2D, (char)0xE2, (char)0x88, (char)0x9E, 0 });
```
see https://en.cppreference.com/w/cpp/types/numeric_limits. `lowest()` returns the largest negative value (or zero) whereas `min()` returns the smallest positive value. Although `std::numeric_limits<int>::min() == std::numeric_limits<int>::lowest()`, it could make sense to use `std::numeric_limits<int>::lowest()` everywhere since people are guaranteed to use `std::numeric_limits<double>::min()` by mistake if they see `std::numeric_limits<int>::min()` being used.

vandenman:
I don't think it's necessary to update the submodules. There is a GitHub actions script that should do that automatically once a week (and the nightlies take the latest master branch anyway).

Common/column.cpp
amirmasoudabdol: You are 100% right! 👍🏼

JorisGoosen:
The replaces look fine (although my heart bleeds for the distortion in alignment) but the submodules should be removed from the PR. It makes merging etc much cleaner.

Also "merg remote-tracking" is usually not the best way to get stuff into your branch, it is better to rebase on remotes.

amirmasoudabdol:
I think I need an intro to your spacing preferences! 😅 For now, I see if I can align things a bit better.

I'm quite confused by the submodule behaviour. It's likely that I've messed up something and my local repo that does this for every branch! I usually try to rebase, but sometimes when I'm on GitHub, I trust it to Fetch and Rebase, and it seems that it does it in a weird way. I noticed it in another PR too.

JorisGoosen:
I mostly like 2D layouts like:
Screenshot_2021-09-13_144006

But as far as the submodules goes (and other rebasing stuff) I think doing an interactive rebase and dropping your "updating submodules" and " merge remote-tracking" should probably fix things?

amirmasoudabdol:
I indeed screwed up this commit I believe! I think I did one wrong rebase and never recovered from it! If this is a mess, I'm sorry! 🤦🏻‍♂️

amirmasoudabdol:
I think I salvaged the situation! 🥲 🤞🏼

JorisGoosen (approved):
Looks fine now

PR #4616, opened 2021-09-09
rSource provokes running engine endlessly

boutinb:
This bug is accidentally revealed when using a rSource in QML as found by @FBartos.
But it is a more general bug: when a value of a control is changed, it emits a signal so that the analysis must be rerun. This signal is sometimes blocked when some block code may change many values (initializing of analysis for example), and it is unblocked at the end of the block. When unblocked the signal was always emitted even if the signal was not called during the block code. But this signal must be emitted only if the signal was called during the block code.

### DIFF
--- a/Desktop/analysis/analysisform.cpp
+++ b/Desktop/analysis/analysisform.cpp
@@ -574,6 +574,8 @@ void AnalysisForm::boundValueChangedHandler(JASPControl *)
 {
 	if (_signalValueChangedBlocked == 0 && _analysis)
 		_analysis->boundValueChangedHandler();
+	else
+		_signalValueChangedWasEmittedButBlocked = true;
 }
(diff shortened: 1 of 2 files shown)


FBartos:
Could you build me a nightly please?

boutinb:
> Could you build me a nightly please?

I see just a new one on Mac.

JorisGoosen:
I'll let windows buildbot do its thang

JorisGoosen (approved):
Code looks fine

JorisGoosen:
After the nightly is built Frantisek can test it and give final approval

JorisGoosen:
@FBartos it is up now

FBartos (approved):
seems to be fine now, thanks!

PR #4617, opened 2021-09-10
Using the Actual R Logo + NOTICE.txt

amirmasoudabdol:
I ran into this logo, and I thought unless we have any emotional attachment to it, we can probably use the R logo instead! 😅 

Screenshot 2021-09-09 at 15 24 44

So, this PR:

- Replaces _the_ R with the actual R logo. There is a slight adjustments to the logo to make it fit the dark mode better. This is allowed, and we just need to mention it. I did changed the colour ever so slightly, and there is still room to edit it a bit more if we want a better contrast.

Screenshot 2021-09-09 at 17 00 25

Screenshot 2021-09-09 at 17 25 17 Screenshot 2021-09-09 at 17 25 35 1 

- It adds a new files to the repo, `NOTICE.txt` that contains information about the open source software and materials that we are using. In addition, the About form now includes a link to this file.

Screenshot 2021-09-10 at 09 41 21

- I think having the `NOTICE.txt` is a good idea anyway, especially as we are adding more third-party modules, and more people are starting to contribute.

P.S. GitHub did the last commit, merging development into this one! I hope it didn't messed up anything! 🤞🏼
P.P.S. I'm also not sure what's happening with my submodules, it feels that it's included in every commit! This is not normal, right? Any idea what is causes it? I can of course nuke the git repo [like a pro](https://xkcd.com/1597/)!


EJWagenmakers:
+1

JorisGoosen (changes requested):
It would be good if you could clean up the commit-history, I can help you out with that if needed. 

Also, I think that the contrast on the darktheme one is a bit weird. It would probably look better if the R was a bit darker, so it contrasts stronger with the gray ellipse?

JorisGoosen:
To fix the commits I would suggest doing an interactive rebase `git rebase -I HEAD~7` probably. Then squash the commits you want to keep together and dropping the merge one and the submodules one.

amirmasoudabdol:
> I think that the contrast on the darktheme one is a bit weird. 

I agree, I am going to take care of it. I think I will go with overall brighter R, and brighter curve to make it stand out a bit further. Something like this, but I'm going to test it first.

Screenshot 2021-09-20 at 09 17 22

amirmasoudabdol:
Ok, we go from the left one to the right one. Slightly brighter R, and quite a bit brighter ring. The contrast between the ring and R is still not great, but unfortunately that's the case for the original R logo as well. I can still fix it by making the ring even whiter, but I think this is alright.

 Screenshot 2021-09-20 at 10 40 18

JorisGoosen (approved):

PR #4625, opened 2021-09-15
Add a link to the Guideline for translating JASP in the help file

boutinb:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/1537

### DIFF
--- a/Resources/Help/preferences/prefsui.md
+++ b/Resources/Help/preferences/prefsui.md
@@ -14,6 +14,8 @@ Here you can specify if you would like to have a "light" or a "dark" theme on th
 ### Preferred Language
 Select the language in which you would like to use JASP.
 
+If you would like to improve our translations, or to add a new language, please visit https://jasp-stats.org/translation-guideline/
+
 ### Zoom
 This number specifies how the JASP interface will be scaled.
 All menus and results are scaled by this factor.
--- a/Resources/Help/preferences/prefsui_nl.md
+++ b/Resources/Help/preferences/prefsui_nl.md
@@ -13,6 +13,8 @@ Hier kan worden gespecificeerd of de gebruiker liever een 'licht' of 'donker' th
 ### Voorkeurstaal
 Hier kan de voorkeurstaal voor JASP worden geselecteerd. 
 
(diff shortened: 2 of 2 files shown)


amirmasoudabdol:
@boutinb, we can maybe add the same message and link under the Choose Language dropdown in the Preferences. This might be helpful bringing new translators to us. It may also save us from some translation-related issues on GitHub when people find something is missing. 

Something like this mockup:

Screenshot 2021-09-15 at 13 47 53

boutinb:
Yes I was thinking of that also, but that would mean that this string has to be also translated in all languages we support now (Dutch, German, Portuguese, Spanish, Galician, Chinese, Japanese) for 0.15... I can ask them to have a last look again this week, but then we should not add strings anymore.

amirmasoudabdol:
That's true! I guess, I just adjust the UI, and we leave it until we have the translation for it.

JoKeyser:
My comment about the link location applies here too, please see https://github.com/jasp-stats/jasp-desktop/pull/4626#issuecomment-920089610

Resources/Help/preferences/prefsui.md
```diff
@@ -14,6 +14,8 @@ Here you can specify if you would like to have a "light" or a "dark" theme on th
 ### Preferred Language
 Select the language in which you would like to use JASP.
 
+If you would like to improve one translation, or to add a new language, please check this document: https://github.com/jasp-stats/jasp-desktop/blob/development/Docs/development/jasp-guideline-translators.md
```
TimKDJ: I think the text used in https://github.com/jasp-stats/jasp-desktop/pull/4625#issuecomment-919948474 works a bit better. Also good to use the jasp-stats.org url here: https://jasp-stats.org/translation-guideline

TimKDJ (changes requested):

Resources/Help/preferences/prefsui.md
AlexanderLyNL: I tend to copy paste this md into https://jasp-stats.org/translation-guidelines

boutinb:
I have made the changes. If @AlexanderLyNL agree, we can merge it

AlexanderLyNL (approved):
Seems good to me. I didn't follow the whole thread, but hopefully TIm is okay with this as well now.

PR #4626, opened 2021-09-15
Added a link in Pref to our translation doc

amirmasoudabdol:
I added a tiny link to the Pref->Language to let users know that they can help us with the translation. I was expecting the `PrefsGroupRect` to deduct its height from its child but it didn't, so I added them myself, and for this reason, I went for a very concise message that we don't run into layout issue when we have this text translate.

→ This is related to #4625, and as @boutinb said, we need to wait until we have this phrase translated as well before merging this.

Screenshot 2021-09-15 at 14 44 07

### DIFF
--- a/Desktop/components/JASP/Widgets/FileMenu/PrefsUI.qml
+++ b/Desktop/components/JASP/Widgets/FileMenu/PrefsUI.qml
@@ -175,10 +175,13 @@ ScrollView
 			id:		languageGroup
 			title:	qsTr("Preferred language")
 
+			height: 85
+
 			ComboBox
 			{
 				id:						languages
 				fieldWidth:				100
+				height:					25
 
 				label:					qsTr("Choose language  ")
 
@@ -189,6 +192,27 @@ ScrollView
 
 				KeyNavigation.tab:		uiScaleSpinBox
 				KeyNavigation.down:		uiScaleSpinBox
+			}
+
+			Text
+			{
+				id:				translationDocLink
+				anchors.top:	languages.bottom
+				anchors.left:	parent.left
+				height:			25
+
+
+				text:			qsTr("Help us translate or improve JASP in your language.")
+				color:			jaspTheme.blue
+				font.underline:	true
+
+				MouseArea
+				{
+					id:				mouseAreaTranslationDocLink
+					anchors.fill:	parent
+					onClicked:		Qt.openUrlExternally("https://github.com/jasp-stats/jasp-desktop/blob/development/Docs/development/jasp-guideline-translators.md")
+					cursorShape:	Qt.PointingHandCursor
+				}
 			}
 
 		}
--- a/Desktop/components/JASP/Widgets/FileMenu/PrefsUI.qml
+++ b/Desktop/components/JASP/Widgets/FileMenu/PrefsUI.qml
@@ -175,13 +175,10 @@ ScrollView
 			id:		languageGroup
 			title:	qsTr("Preferred language")
 
-			height: 85
-
 			ComboBox
 			{
 				id:						languages
 				fieldWidth:				100
-				height:					25
 
 				label:					qsTr("Choose language  ")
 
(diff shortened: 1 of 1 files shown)


JoKeyser:
I think this is a great idea, but I have doubts about the link location: Is it really the best idea to link directly into GitHub?
Currently, the link is https://github.com/jasp-stats/jasp-desktop/blob/development/Docs/development/jasp-guideline-translators.md, and thus depends on that exact repo, branch, and file. What if you want to re-order this in the future?
Alternatively, the link in JASP could point to something like https://jasp-stats.org/translation-guidelines, and you link from there to GitHub - but in the future, you're free to change the content of that page at any time.

boutinb (changes requested):
You don't have to specify the height and to set the anchors: the PrefsGroupRect takes care of that. Also this does not scale. So something like that works beter:

```
		PrefsGroupRect
		{
			id:		languageGroup
			title:	qsTr("Preferred language")

			ComboBox
			{
				id:						languages
				fieldWidth:				100

				label:					qsTr("Choose language  ")

				currentValue:			languageModel.currentLanguage
				onCurrentValueChanged:	languageModel.currentLanguage = currentValue;

				source:					languageModel

				KeyNavigation.tab:		uiScaleSpinBox
				KeyNavigation.down:		uiScaleSpinBox
			}

			Text
			{
				id:				translationDocLink

				text:			qsTr("Help us translate or improve JASP in your language.")
				color:			jaspTheme.blue
				font.underline:	true

				MouseArea
				{
					id:				mouseAreaTranslationDocLink
					anchors.fill:	parent
					onClicked:		Qt.openUrlExternally("https://github.com/jasp-stats/jasp-desktop/blob/development/Docs/development/jasp-guideline-translators.md")
					cursorShape:	Qt.PointingHandCursor
				}
			}

		}
```


Also the comment of @JoKeyser is a really good one (even if I don't know how to do that exactly in GitHub)

amirmasoudabdol:
@boutinb, that's what I was expecting, but it didn't. I'm going to try again and let you know.

I agree with @JoKeyser as well. But I am not sure how to make a page on the website yet. I guess I will use the link he suggested, and we somehow put a webpage there! :-)

JoKeyser:
@amirmasoudabdol , @boutinb glad you agree with the suggested URL, this decoupling will pay off, I'm sure; it's also nice to make the translation guidelines discoverable on the website. One observation: It seems that all links on the website end with a trailing `/`, i.e. they act as folders, see for example https://jasp-stats.org/feature-requests-bug-reports/.
Whoever is responsible for the website could also list the translation guideline among the support topics at https://jasp-stats.org/support/

amirmasoudabdol:
It was a very good idea! Thanks! For the website, as I understand it's Wordpress, and Wordpress should be smart enough to resolve this, but I will test it as soon as we have the page up.

It's a very good idea to make the page discoverable. I would have been fine by it just being there, since I had difficulty with its navigation, but I guess leaving it under Support is a good idea. @AlexanderLyNL will do it after the guideline is approved.

Desktop/components/JASP/Widgets/FileMenu/PrefsUI.qml
```diff
@@ -191,6 +191,23 @@ ScrollView
 				KeyNavigation.down:		uiScaleSpinBox
 			}
 
+			Text
+			{
+                id:                     translationDocLink
+
+				text:			qsTr("Help us translate or improve JASP in your language.")
```
TimKDJ, line 198: I preferred your text from https://github.com/jasp-stats/jasp-desktop/pull/4625#issuecomment-919948474. `or improve JASP in your language` sounds a little odd to me. One possibility to shorten it:

```
If you would like to improve our translations please visit our website.
```
Having only one part of the sentence be a hyperlink instead of the whole thing also helps set it apart more IMO

amirmasoudabdol: I know what you mean, but I think the current sentence is more suitable for the UI. It's more actionable, "Help us ...". In addition, having the second QML element adds extra complication for later translation and alignment. That's why I changed it to be a single sentence.

boutinb (approved):

boutinb:
@TimKDJ If you agree, we can merge this PR for 0.15

JorisGoosen (approved):
We can change the sentence later

PR #4632, opened 2021-09-20
Increment and Decrement of the SpinBox Should be Invoked by ↑ and ↓

amirmasoudabdol:
I'm trying to investigate the source of [this](https://github.com/jasp-stats/INTERNAL-jasp/issues/1482), and I'm noticing some strange interactions. This one is very odd, and it should not be like this. Different platforms are calling this control differently, but they are often falling under the Stepper category. On .NET, we have [NumericUpDown](https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.numericupdown?view=net-5.0), and on iOS and macOS, they are called [Stepper](https://developer.apple.com/design/human-interface-guidelines/macos/selectors/steppers/), and in both systems, the changes are invoked by Up and Down keys. Qt visual implementation of this is the strangest but even they mapped the +/- to ↑/↓. 

So, I changed it to use up/down instead of left/right. In addition, I fixed a bug where the control looked enabled when being disabled.

Screenshot 2021-09-20 at 17 43 25

P.S. I'm trying to bring some logic to the KeyNavigation, so, I have to remove some of these nonstandard navigations here and there. It seems like that there are many competing KeyNavigation in JASP interface.

P.P.S. I see that we went through the pain of implementing this control from scratch. If in the far far future, we want to improve it, we can stack the up/down button and make it looks normal!

### DIFF
--- a/Desktop/components/JASP/Widgets/SpinBox.qml
+++ b/Desktop/components/JASP/Widgets/SpinBox.qml
@@ -44,8 +44,8 @@ FocusScope
 					width:					plus.x + plus.width
 					height:					valueField.height
 
-					Keys.onLeftPressed:		{ minus.clicked(); event.accepted = true; }
-					Keys.onRightPressed:	{ plus.clicked();  event.accepted = true; }
+					Keys.onDownPressed:		{ minus.clicked(); event.accepted = true; }
+					Keys.onUpPressed:		{ plus.clicked();  event.accepted = true; }
 					Keys.onEnterPressed:	valueField.focus = !valueField.focus;
 					Keys.onReturnPressed:	valueField.focus = !valueField.focus;
 					Keys.onEscapePressed:	focus = false;
@@ -113,7 +113,7 @@ FocusScope
 		selectByMouse:				true
 		selectedTextColor:			jaspTheme.white
 		selectionColor:				jaspTheme.itemSelectedColor
-		color:						jaspTheme.textEnabled
+		color:						enabled ? jaspTheme.textEnabled : jaspTheme.textDisabled
 
 		function processInput()
 		{


JorisGoosen:
How does it make sense to have up/down be the keys for a control that is oriented left/right?

amirmasoudabdol:
You go up to increase something, and down to decrease it! It's not always about visual.

Our implementation is not optimal! If I want to guess, for some reason we ended up using/implementing the touchscreen version of the control. Even then, if the keyboard where allowed controlled are linked to up/down. It's the same on Qt's implementation as well. It is like this everywhere. The traditional version of this control is like the Stepper that I liked too.

JorisGoosen:
> You go up to increase something, and down to decrease it! It's not always about visual.
> 
> Our implementation is not optimal! If I want to guess, for some reason we ended up using/implementing the touchscreen version of the control. Even then, if the keyboard where allowed controlled are linked to up/down. It's the same on Qt's implementation as well. It is like this everywhere. The traditional version of this control is like the Stepper that I liked too.

Well I think I made this one because I couldn't find a qt one or wanted it to be horizontal, also it allowed for up/down navigation without getting stuck in it like with the combobox ;)

amirmasoudabdol:
The Qt version has been added since 5.7.

> without getting stuck in it like with the combobox ;)

That brings us to the next issue. I don't think traversing between controls using arrows keys is a good idea, exactly because of this reason. When you reach a control those arrows have their functions and it is always going to cause issue. I was going to discuss this with you, I think we should stick to tab traversing and remove the arrow navigation (at least inside the Prefs* panels). Then, we don't have to deal with some of these issues, or deal with scrolling conflict, etc.

JorisGoosen:
> The Qt version has been added since 5.7.

Well, perhaps we should just use that then?

> 
> > without getting stuck in it like with the combobox ;)
> 
> That brings us to the next issue. I don't think traversing between controls using arrows keys is a good idea, exactly because of this reason. When you reach a control those arrows have their functions and it is always going to cause issue. I was going to discuss this with you, I think we should stick to tab traversing and remove the arrow navigation (at least inside the Prefs* panels). Then, we don't have to deal with some of these issues, or deal with scrolling conflict, etc.

That is quite a good point, at least for most UIs. 

Would it hurt though to have both up/right increment and left/down decrement?

amirmasoudabdol:
> Well, perhaps we should just use that then?

We should, but for now, I would merge this since it is a very minor fix. Then, we can look into replacing it if it caused problem, or when I was fed up with CMake stuff and wanted a breather!  

> Would it hurt though to have both up/right increment and left/down decrement?

Yes! Sometimes removing functionality is a good thing. The situation with Key Navigation happens because of the same reason probably, allowing both Tab and Arrows. At some point the non-standard behaviour starts to have frictions with the rest of the system.

JorisGoosen (approved):

PR #4635, opened 2021-09-22
add new reliability data sets

juliuspfadt:
Before, we only had data sets for the classical reliability analysis. 
This adds two more, one for the Bayesian unidimensional reliability analysis, and one for the classical intraclass correlation. 
I adjusted the "normal" index.json file. Should I also do the German one? Someone would need to do the dutch one then. Or will this process also be adjusted for the future so that things are "automatically" translated by the translators?


Kucharssim:
Yes, at this point, you need to modify `index_de.json` and `index_nl.json`, otherwise I think the files will not be available in those translations... Don likes to use DeepL for that purpose, but sometimes we just leave the text in english for now.

I think the plan was at some point to make the library dynamic so that people can add their files to their module repos without making PRs to `jasp-desktop` (https://github.com/jasp-stats/INTERNAL-jasp/issues/889). I think it makes sense to look at the translations at that point.

juliuspfadt:
Well, I can also just wait with these files and once there is the functionality to add them to the module directly do that. And then also think about the translations. It depends a bit on the timeframe for this, cause I would like to have some more examples for the reliability analysis some time soon-ish

Kucharssim (changes requested):
Could you change the description in the saved files to be more similar to the other examples in the Data Library?
This is the part that I am talking about
Screenshot 2021-09-23 at 14 43 12
Screenshot 2021-09-23 at 14 43 03

Which should be ideally structured like this:
Screenshot 2021-09-23 at 14 43 36

that is:
[
**Description**
What the data set is about

**Variables** 
list of variable names and what they represent

Then one, two sentences about what the jasp file demonstrates

**References**
]

> Well, I can also just wait with these files and once there is the functionality to add them to the module directly do that. And then also think about the translations. It depends a bit on the timeframe for this, cause I would like to have some more examples for the reliability analysis some time soon-ish

Yeah, I don't think that anyone is looking into it right now, so I would not bet that this will be implemented soon-ish... So if you want to add the data sets about now, it's better to do it this way... even though it's a bit annoying...

Resources/Data Sets/index.json
```diff
@@ -760,10 +760,24 @@
...
+			"kind": "file",
+			"associated_datafile": "ASRM - Mania Scale.csv"
+		  },
+		  {
+			"name": "Interrater Data from Shrout and Fleiss (1979)",
+			"path": "Interrater Data from Shrout and Fleiss (1979).jasp",
+			"description": "Hypothetical data set from a paper on ICCs by Shrout and Fleiss (1979). The file demonstrates the use of classical intraclass correlation coefficients",
+			"kind": "file",
+			"associated_datafile": "Interrater Data from Shrout and Fleiss (1979).csv"
+		  }
```
Kucharssim: On github, it seems that the indenting is off by one tab (in all three index files)

Resources/Data Sets/index_nl.json
```diff
@@ -764,6 +764,20 @@
                 "kind": "file",
                 "associated_datafile": "Fear of Statistics.csv"
               }
```
Kucharssim:
```suggestion
              },
```
here needs to be comma for the json to work

Kucharssim (changes requested):
Just two small things in the jsons, otherwise it's good to go

Kucharssim (approved):

Kucharssim:
If @boutinb does not object, I would merge this

boutinb (approved):

PR #4638, opened 2021-09-24
Hide menu options for plots when they can't be used

vandenman:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/1547

The menu for plots always shows these options:

image

But when a plot shows an error message, like so:

image

It makes no sense to
- Copy the image
- Save the image
- Edit the image

This PR disables those options:

image


vandenman:
@boutinb @AlexanderLyNL any comments?

boutinb (approved):

AlexanderLyNL:
Hiding these options make sense to me.

AlexanderLyNL (approved):

PR #4640, opened 2021-09-26
Improvement to TextField Implementation

amirmasoudabdol:
I has some time this morning, and I thought, oh, let's look around the UI a bit. I ended up looking into the `TextField`, `DoubleField`, and some other UI elements, and noticed some inconsistencies, and made some improvements.

So, here is the issues, and it's almost unsolvable, because we cannot properly align them with the current setup. They should be aligned but since each `DoubleField` has its own containers, that's very tricky. 

CleanShot 2021-09-26 at 12 07 32

I made a tiny adjustment, and set a default size to get it right, then it will look like this:

CleanShot 2021-09-26 at 12 07 53

To do this, I replaced the manual anchoring of the `JS.TextField` with a `RowLayout`, and this has some good side-effect in other places; for instance, we go from the left to the right, 

CleanShot 2021-09-26 at 12 08 12 → CleanShot 2021-09-26 at 12 08 34

Notice the improvement in the spacing of **Interval** controls and **to**. This is not perfect, because the implementation of those are different. To get them right every time, we either have a standard, or design a `Range` control or something, we use it frequently enough everywhere anyway. This probably makes the range checking easier in the future, especially now that we are thinking to use the QML on the R side as well.

PS. I created a draft for this since it probably needs more testing, and discussion, but if you feel it's a good idea, we can of course merge it!

### DIFF
--- a/Desktop/components/JASP/Controls/TextField.qml
+++ b/Desktop/components/JASP/Controls/TextField.qml
@@ -24,8 +24,8 @@ import JASP				1.0
 TextInputBase
 {
 	id:					textField
-	implicitHeight:		control.height
-	implicitWidth:		afterLabel.text ? (afterLabelRect.x + afterLabelRect.width) : (control.x + control.width)
+	implicitHeight:		layout.height
+	implicitWidth:		layout.width
 	background:			useExternalBorder ? externalControlBackground : control.background
 	cursorShape:		Qt.IBeamCursor
 	innerControl:		control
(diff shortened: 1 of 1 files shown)


boutinb:
Maybe it does not work on other system Linux, but on my Mac Mojave I get the right alignment (with the original TextField code):
Screenshot 2021-09-27 at 08 52 04
There exists already a mechanism so that if several TextFields are onder a Group, the field part is aligned (cf _alignTextFields function). But I agree this is quite an awkward solution, and apparently it does not work well on all platforms.
But if I try your solution, it does not work either. For example if the label is a bit bigger then it's not aligned anymore:
Screenshot 2021-09-27 at 09 25 59
This works with my awkward solution on my Mac:
Screenshot 2021-09-27 at 09 29 12
So I think we should brainstorm together to see how to solve correctly this problem.

amirmasoudabdol:
Yes, my solution isn't perfect either, it only takes care of one of the case. Like you said, we need to think about this a bit more, and find a general solution for it. Then, we really need to force it that people —outside of us— who design something for JASP, use them and we end up having a consistence interface.

PR #4646, opened 2021-10-04
Disabled the Exponent Aligning and Implemented the Normalized Notation

amirmasoudabdol:
This partially addressed [this](https://github.com/jasp-stats/jasp-issues/issues/1437), and added an option for "normalized notation" as described [here](https://github.com/jasp-stats/INTERNAL-jasp/issues/1533). So, we will have 

CleanShot 2021-10-04 at 13 22 18

CleanShot 2021-10-04 at 11 29 56

and this feature can be activated from the Table Options, 

CleanShot 2021-10-04 at 13 26 15

I think I have tested this enough, and the exporting to HTML, RTF, and PDF works fine in my test. But, please give it a try, and let me know if you see something odd.


djdekker:
Nice to have the option to express 3.4400e+55 as 3.4400×10<sup>+55</sup>, which looks much more familiar to my students. Still one suggestion: dropping the + sign in front of a positive exponent would make it even more understandable. So I'd suggest to write 3.4400×10<sup>55</sup> instead of 3.4400×10<sup>+55</sup>. What would you think of that?

EJWagenmakers:
Great to see this fixed! I have a slight preference for omitting the plus sign, but there might be reasons to keep it (aligment, immediate contrast between numbers with + and -)...at any rate, this improves on what we did previously.

amirmasoudabdol:
Thanks for your feedback @djdekker. I totally agree with you, and in fact, I opened an internal discussion right after pushing this change. I agree with @EJWagenmakers thought that, here, the sign has some purpose, especially in the "e notation", less on the normalized notation.

That being said, I don't think keeping it _only to achieve_ alignment is a good idea, it's prone to break as soon as numbers are factors apart from each other, and as a result the contrast will not be preserved. As I argued [here](https://github.com/jasp-stats/INTERNAL-jasp/issues/1572), if we really want to, we can maybe offer an option like "Print Implicit Plus" which allows the user to ask for the `+` sign to appear on every number in the table. 

A more elaborate version of this option would be to allow for `+` sign in either base, exponent, or both. Let's say, a dropdown menu with three options, "Everywhere", "Only for base", "Only for exponent". But then, I am not a big fan of providing too many options, and I think a good default will go a long way, and in this case, a good default would be to just drop the sign in the "noramlized notation" and always include it in the "e notation". Perhaps we can discuss this further in the open discussion, and find a good middle ground.

EJWagenmakers:
+1

JorisGoosen:
Latex is broken:
2021-10-25-115407_3840x2160_scrot

(I only tried this webinterface, but Im guessing it'll be the same in miktex or whatever)

`3.440×10$^{+56}$` is converted into `3.44010` with `+56` as superscript, but because the 10 is pasted on the beginnumber it becomes weird.

Besides that in JASP itself the natural notation doesn't line up, but maybe you weren't going for that yet?
2021-10-25-115553_3840x2160_scrot

JorisGoosen (changes requested):
Latex should work as well

amirmasoudabdol:
Oh that's curious. I am pretty sure they passed my tests on several LaTeX engines. I am going to check it again. For the alignment, no, we shouldn't fix it by force. I'm going to look into it.

amirmasoudabdol:
I see the issue, they cannot get the `×` right in their engine. That's an unicode char and it should work but I can fix it for good by using `\times`! 👍🏼

amirmasoudabdol:
This fixes that issue, but I still cannot get a build on that website! 🤷🏻‍♂️ That's how the code looks like, and I can build it with pdfLaTeX, LuaLaTeX, and XeLaTeX on my Mac.

```tex
\begin{table}[h]
	\centering
	\caption{Descriptive Statistics}
	\label{tab:descriptiveStatistics}
	{
		\begin{tabular}{lrrrrr}
			\toprule
			 & contNormal & contGamma & contBinom & contExpon & contWide  \\
			\cmidrule[0.4pt]{1-6}
			Valid & $100$ & $100$ & $100$ & $100$ & $100$  \\
			Missing & $0$ & $0$ & $0$ & $0$ & $0$  \\
			Mean & $-0.19$ & $2.03$ & $0.42$ & $3.44\times10^{+55}$ & $-5.81\times10^{+97}$  \\
			Std. Deviation & $1.06$ & $1.53$ & $0.50$ & $3.44\times10^{+56}$ & $5.14\times10^{+99}$  \\
			Minimum & $-3.02$ & $0.20$ & $0.00$ & $1.74\times10^{-49}$ & $-8.96\times10^{+99}$  \\
			Maximum & $3.36$ & $8.77$ & $1.00$ & $3.44\times10^{+57}$ & $8.93\times10^{+99}$  \\
			\bottomrule
		\end{tabular}
	}
\end{table}
```

CleanShot 2021-10-25 at 18 47 14

I have tried to improve the `formatCellforLaTeX` implementation a bit but I can only do so much. The function should be aware of the type of the text that is receiving. With my changes, it tries to put all the numbers in a math block, and I think it should cover most cases. This addresses this one as well, https://github.com/jasp-stats/INTERNAL-jasp/issues/1572, by fixing the negative sign.

PS. JS is weird! It is fine for the most part, but then its type system is utterly broken! You can tell if something is a number by using `isNaN`!

AlexanderLyNL:
@akashrajkn could you provide some insights?

akashrajkn:
@amirmasoudabdol Your fix of using `$times` seems to work. On that website, there should be a expression `$$` somewhere in the line after `\toprule` (I don't know why), with that it renders
Screenshot 2021-10-26 at 10-37-22 https quicklatex com

Desktop/components/JASP/Widgets/FileMenu/PrefsResults.qml
```diff
@@ -37,6 +37,17 @@ ScrollView
...
 				onCheckedChanged:		preferencesModel.exactPValues = checked
+				KeyNavigation.tab:		useNormalizedNotation
+				KeyNavigation.down:		useNormalizedNotation
+			}
+
+			CheckBox
+			{
+				id:						useNormalizedNotation
+				label:					qsTr("Use normalized notation")
+				labelTextFormat:		Text.RichText
```
JorisGoosen, line 48: Was there a reason to add RichText here?

JorisGoosen (approved):

PR #4650, opened 2021-10-11
Alternative Background for the DMG file

amirmasoudabdol:
Since many users are struggling with installing JASP, I thought let's put the instruction into the DMG file. Hopefully this will make it easier for us/them to deal with this situation until we have a better solution.

@boutinb, if we really want it out, let's push a silent update; otherwise we can wait for 0.16. Also, please check the outcome to make sure that the icons are placed correctly! 🙏🏼

background

It helps with this, https://github.com/jasp-stats/jasp-issues/issues/1426.


djdekker:
Thanks for your work! This should solve one of the issues that Mac users are experiencing when they start JASP.

Two other known Mac issues are:
- JASP is not registered as the default application to open .jasp files;
- JASP is not registered as a recommended application to open .csv files (see [issue 857](https://github.com/jasp-stats/jasp-issues/issues/857)).

Could these also be tackled with the script?

amirmasoudabdol:
@djdekker, this is a workaround and we will go back to the normal background as soon as we have a better solution, but I hope this helps.

As for the other issues, we are aware of them, and I thought it works but it doesn't work on my system, but it does on some others. Has it ever worked for you, across different JASP or macOS versions?

boutinb:
@amirmasoudabdol Did you try in a dmg? Because I get this
Screenshot 2021-10-12 at 14 31 38

Als why did you add the background.svg file?

amirmasoudabdol:
@boutinb, Oh, ok, I messed the placement of that. The problem is, for some reason, my script cannot place the icon at all. You at least get something. Please try this instead:

```
set position of item "'${APP_NAME}'.app" of container window to {400, 660}
```

I included the SVG in case we want to edit it later.

I did try it, but since the JASP logo was not working for me at all, I couldn't test that part of it. I made sure that the Application fits properly and extrapolate form there.

boutinb:
If you set
```
set position of item "'${APP_NAME}'.app" of container window to {150, 640}
set position of item "Applications" of container window to {450, 640}
```
It works well on my machine.

amirmasoudabdol:
I think that should be good! Thanks Bruno!

Tools/macOS/sign-mac.sh
```diff
@@ -180,13 +180,13 @@ tell application "Finder"
...
 		set toolbar visible of container window to false
 		set statusbar visible of container window to false
-		set the bounds of container window to {400, 100, 920, 440}
+		set the bounds of container window to {400, 100, 1000, 900}
 		set viewOptions to the icon view options of container window
 		set arrangement of viewOptions to not arranged
 		set icon size of viewOptions to 72
 		set background picture of viewOptions to file ".background:'${DMG_BACKGROUND_IMG}'"
-		set position of item "'${APP_NAME}'.app" of container window to {160, 205}
-		set position of item "Applications" of container window to {360, 205}
```
boutinb, line 189: Change the position in this way:
```
set position of item "'${APP_NAME}'.app" of container window to {150, 640}
set position of item "Applications" of container window to {450, 640}
```

boutinb (changes requested):

AlexanderLyNL:
Merge this for 0.16

amirmasoudabdol:
@JorisGoosen, shouldn't this go in before the release?

JorisGoosen (approved):

PR #4651, opened 2021-10-11
Improvement to the MessageBoxes

amirmasoudabdol:
I noticed this not so professional message, "There are unapplied changes to your filter; what would you like to do?", and while looking into that, I ended up improving the message box a bit. Because of macOS Guidelines, Qt cannot set the TitleBar text on macOS, and that makes the dialog looks a bit too bold. I don't expect this to break anything on Windows, but it's good to check it anyway.

@boutinb, I have changed a few labels and text, do I have to do anything for this to get into our translation portal?

CleanShot 2021-10-11 at 18 15 47

CleanShot 2021-10-11 at 18 12 05

### DIFF
--- a/Desktop/components/JASP/Widgets/ComputeColumnWindow.qml
+++ b/Desktop/components/JASP/Widgets/ComputeColumnWindow.qml
@@ -367,8 +367,8 @@ FocusScope
 		SaveDiscardCancelDialog
 		{
 			id:			saveDialog
-			title:		qsTr("Computed Column Changed")
-			text:		qsTr("There are unapplied changes to your computed column; what would you like to do?")
+			title:		qsTr("Do you want to save your changes to the Computed Column?")
+			text:		qsTr("Your changes will be lost if you don't save them.")
 			onDiscard:	computedColumnContainer.close()
 			onSave:
 			{
--- a/Desktop/components/JASP/Widgets/ComputeColumnWindow.qml
+++ b/Desktop/components/JASP/Widgets/ComputeColumnWindow.qml
@@ -367,7 +367,7 @@ FocusScope
 		SaveDiscardCancelDialog
 		{
 			id:			saveDialog
-			title:		qsTr("Do you want to save your changes to the Computed Column?")
+			title:		qsTr("Would you like to save your changes to the Computed Column?")
 			text:		qsTr("Your changes will be lost if you don't save them.")
 			onDiscard:	computedColumnContainer.close()
 			onSave:
--- a/Desktop/components/JASP/Widgets/FilterWindow.qml
+++ b/Desktop/components/JASP/Widgets/FilterWindow.qml
@@ -145,7 +145,7 @@ FocusScope
 				{
 					id:		easySaveDialog
 
-					title:	qsTr("Do you want to save your changes to the Filtered Column?")
+					title:	qsTr("Would you like to save your changes to the Filter?")
 					text:	qsTr("Your changes will be lost if you don't save them.")
 
 					property var closeFunc: undefined
--- a/Desktop/components/JASP/Widgets/FilterWindow.qml
+++ b/Desktop/components/JASP/Widgets/FilterWindow.qml
@@ -145,8 +145,8 @@ FocusScope
 				{
 					id:		easySaveDialog
 
-					title:	qsTr("Filter Changed")
-					text:	qsTr("There are unapplied changes to your filter; what would you like to do?")
+					title:	qsTr("Do you want to save your changes to the Filtered Column?")
+					text:	qsTr("Your changes will be lost if you don't save them.")
 
 					property var closeFunc: undefined
 
--- a/Desktop/gui/messageforwarder.cpp
+++ b/Desktop/gui/messageforwarder.cpp
@@ -20,15 +20,25 @@ MessageForwarder * MessageForwarder::_singleton = nullptr;
 
 void MessageForwarder::showWarning(QString title, QString message)
 {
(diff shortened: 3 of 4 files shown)


Desktop/components/JASP/Widgets/ComputeColumnWindow.qml
```diff
@@ -367,8 +367,8 @@ FocusScope
 		SaveDiscardCancelDialog
 		{
 			id:			saveDialog
-			title:		qsTr("Computed Column Changed")
-			text:		qsTr("There are unapplied changes to your computed column; what would you like to do?")
+			title:		qsTr("Do you want to save your changes to the Computed Column?")
```
AlexanderLyNL: Would you like to save your changes to the Computed Column?

Desktop/components/JASP/Widgets/FilterWindow.qml
```diff
@@ -145,8 +145,8 @@ FocusScope
 				{
 					id:		easySaveDialog
 
-					title:	qsTr("Filter Changed")
-					text:	qsTr("There are unapplied changes to your filter; what would you like to do?")
+					title:	qsTr("Do you want to save your changes to the Filtered Column?")
```
AlexanderLyNL: Would you like to save your changes to the Filter?

Desktop/mainwindow.cpp
```diff
@@ -947,7 +947,7 @@ void MainWindow::dataSetIORequestHandler(FileEvent *event)
 			QString title = windowTitle();
 			title.chop(1);
 
-			switch(MessageForwarder::showSaveDiscardCancel(tr("Save File?"), tr("Save changes to the file %1 before closing?\n\nYour changes will be lost if you don't save them.").arg(title)))
+			switch(MessageForwarder::showSaveDiscardCancel(tr("Do you want to save your changes to %1?").arg(title), tr("Your changes will be lost if you don't save them.")))
```
AlexanderLyNL: "Would you like to save your changes to %1 file?").arg(title), tr("Your changes will be lost if you don't save them."

AlexanderLyNL (changes requested):
I think "Would you" is more suitable here. 

https://ell.stackexchange.com/questions/29804/what-is-the-difference-between-do-you-like-and-would-you-like

Desktop/components/JASP/Widgets/ComputeColumnWindow.qml
amirmasoudabdol: @AlexanderLyNL, I agree with this, but for some reason when I probed a few apps I realized that they don't go that far, and they often opt for "Do you", in some cases they even drop the "Do" and go with "Want to ..." (in the case of Microsoft Office apps). I think this is to make it make it more prominent, direct and prominent. But, if you think we should use "Would you", I am fine with that.

AlexanderLyNL:
To me the "Would you" construction works, but if prefer what microsoft does, then that's fine as well.

amirmasoudabdol:
Let's go with the "Would you" then!

AlexanderLyNL (approved):

Desktop/gui/messageforwarder.cpp
```diff
@@ -20,16 +20,26 @@ MessageForwarder * MessageForwarder::_singleton = nullptr;
 
 void MessageForwarder::showWarning(QString title, QString message)
 {
-	QMessageBox::warning(nullptr, title, message);
+    QMessageBox box;
+
+	box.setText(title);
+	box.setInformativeText(message);
+	box.setIcon(QMessageBox::Warning);
```
boutinb, line 27: Why don't you use the warning method of QMessageBox?

amirmasoudabdol: Qt doesn't handle that properly on macOS. It ignores the title on macOS, then we end up with the before screenshot that I shared, everything gets cramped into a pseudo title. This way, on both systems, we get something similar.

> Sets the title of the message box to title. On macOS, the window title is ignored (as required by the macOS Guidelines). -- See setWindowTitle on [QMessageBox](http://doc.qt.io/qt-5/qmessagebox.html#setWindowTitle).

PS. I still cannot run a Windows VM on my Mac. So, I didn't test this on Windows, but it should not cause any issues because of the discrete way that I set the title.

boutinb: On Windows it was like that:
OldMessage
After your changes:
NewMessage

So the 'File has changed' is not any more in the title, and the message is also smaller.

boutinb: On Windows the text is not bold, so the informative text is not different than the text.
Moreover if you set the text in the constructor, Qt resize the Message Box according to the text. With your fix, it does not do that anymore. So I would do rather something like that:

```
QMessageBox box(QMessageBox::Question, title, message, QMessageBox::Yes | QMessageBox::No);
#ifdef Q_OS_MACOS
box.setText(title);
box.setInformativeText(message);
#endif

```

boutinb (approved):

amirmasoudabdol:
@boutinb I see that to fix this properly, I need to change a few strings, and as you said this might make it tricky for the translators. I think I continue working on it, but we merge it after 0.16.1 that there is enough time for the translators to catch up. Is that a good idea?

boutinb:
Yes there is no need to implement this in 0.16.1

JorisGoosen:
Ok I rebased and updated the code to the current JASP version. But Im having trouble pushing to this branch.
So Ill open a new PR.

PR #4653, opened 2021-10-13
Improvements to the Rename Column and Resize Table Forms

amirmasoudabdol:
This got a bit big bigger than I would have liked it to be, but I had to do quite a few things to get to this point. Main changes are:

- Redesigned Rename Column dialog, with a Cancel/Rename button
	- Fixes the bug where the TextInput were growing in size if the user starts entering a long name (I'm not sure anymore, but the TextInput behavior was a bit off)
	- Using our controls instead of TextInput whenever I can, Rounded Buttons, etc.
	- Fixes this: https://github.com/jasp-stats/INTERNAL-jasp/issues/1591 
	- Fixes this: https://github.com/jasp-stats/INTERNAL-jasp/issues/1587
	- Fixes this: https://github.com/jasp-stats/INTERNAL-jasp/issues/1597
- Redesigned Resize Dialog
	- I think we can stick to something more conventional, and ask users about the number of columns and rows. I basically looked up how MS Word does it and replicated that.
	- Again, using our control whenever possible
	- Fixes this: https://github.com/jasp-stats/INTERNAL-jasp/issues/1592
	- Fixes this: https://github.com/jasp-stats/INTERNAL-jasp/issues/1598
	- Removed some of the arrow navigation, I think we should just rely on tabs for navigation, and if there is only two buttons, then Enter/Esc should be enough.
		- I cannot be sure about the behavior of KeyNavigation on my Mac, because apparently it doesn't really work for me! It's a mystery!
- Exposed two new Dialog Boxes
	- Yes/No/Cancel 
	- Yes/No
	- The code was there, I just brought them to the QML realm
- I thought it would be a good idea to explicitly indicates if we are using our own controls, JW for JASP Widgets, and JC for JASP Controls. This makes the code much easier to read and track.

CleanShot 2021-10-13 at 15 24 21 CleanShot 2021-10-13 at 15 24 09

**Know Issues:**

- [x] For some reason, I cannot capture the first column name if I click on right away. Funnily enough, it works fine if I click on a different column, and then click on column one. I think I'm missing something but I cannot see it. I am sure you can pick it up very quickly.
- [x] The same applies to the Resize, for some reason I cannot capture the initial size of the table in some cases. I see that we are using the `Component` mechanism, but I couldn't integrate it into the `contentItem`.
- [ ] `IntegerField`s in the Resize form doesn't behave correctly. They indicates that they are not valid even though they are! 

P.S. While there were some bugs, I thought it's probably better to have it here than sitting on my computer. If we were in the office, I would have just dropped by someone's desk and quickly solve this. So, I'm hopping that we can do the same with drafts! 🙂


amirmasoudabdol:
@AlexanderLyNL Would you please check if [this message](https://github.com/jasp-stats/jasp-desktop/blob/d0e80e5bcae05ec3f2c66481e403d38fa1a6a20f/Desktop/components/JASP/Widgets/ResizeDataDialog.qml#L124) make sense? I have a feeling it could be better.

AlexanderLyNL:
Something like this?

title: qsTr("Would you like to remove some of the rows or columns of %!?").arg(dataSetModel.name())
text: qsTr("The provided rows and columns are smaller than the rows or columns currently present in the data set. This will remove some rows or columns. Please be aware that once the data set is shrunk the lost data cannot be retrieved.")

boutinb:
This has become obsolete

PR #4655, opened 2021-10-15
A bit of a Softer/Brighter Divider

amirmasoudabdol:
I found it very black in the middle of our mostly gray interface. The better solution would be to rely on the GroupBox border, but I don't see our GroupBox having a border.

CleanShot 2021-10-15 at 10 47 42

CleanShot 2021-10-15 at 10 47 46

### DIFF
--- a/Desktop/components/JASP/Widgets/Divider.qml
+++ b/Desktop/components/JASP/Widgets/Divider.qml
@@ -36,7 +36,7 @@ Item
 		border.width: 1
 		height: 2
 		width: parent.implicitWidth
-		border.color: jaspTheme.black
+		border.color: jaspTheme.gray
 	}
 
 	Rectangle


AlexanderLyNL:
Alternatively, you can split this with a section as we do normally and have the section open by default, as in the t-test

amirmasoudabdol:
Sure, but here I literally only looked at the line, and I thought that's a bit too dark. It's been used in other places too, like below, where it creates a very rough separation between the sections. Making it gray still shows different section, but it's not as harsh.

CleanShot 2021-10-15 at 10 34 38

AlexanderLyNL:
So one question is whether lines are the way to go. The idea behind JASP is to not overwhelm users with so many options. Instead, provide them with results quickly based on some nice defaults. If there are options to tune then they should be presented in some orderly fashion grouped together. 

The lines present all the options at once and it's unclear why these belong together. Having sections separates these options and provides the user with insights to what they are tuning.

AlexanderLyNL:
I remember having a discussion regarding option placement in the UI panel with Tim some years ago. It might be good to revisit this discussion and develop a guide for this.

amirmasoudabdol:
Generally, I'm not a big fan of these sections. I don't like interface elements that are hiding things from me, and I have to open them to get to them, and do something with them, and that's the case here. There is always the danger of one user not even realizing it, therefore never finding it. In addition, JASP doesn't remember my choices, and I have to open them every freaking time! They also move which derives me crazy! 😅

I think we have lots and lots of options, and often we cannot do anything about it. For grouping them, I'd have stayed with the standard GroupBox where you get a border around the controls that it is grouping, but they could make your UI feel crowded, and that's probably why we have them setup without border. In some cases thought, border would help, e.g., the meta-analysis screenshot, if those options where inside a Group with a border around them, it'd have been easy to distinguish them. 

I don't see any problem with dividers btw, I think they are used correctly in the Summary Statistics, but not so much in the Meta-Analysis one. There, a group would have been enough, maybe even without the border. Making the title of the group bold would have been sufficient.

AlexanderLyNL:
Sections can be open by default of course. But the basic principle of a JASP analysis should really be providing quick, and informative results. That's why we should provide the user with good default options, but also think about which output we show them by default. In many cases people should not be bothered with tweaking options. 

For someone as technical as you it makes sense that all options are shown, but all these options can easily overwhelm people who are not that technically schooled. If they were fine with all these options they wouldn't be using JASP to begin with.

JorisGoosen:
@amirmasoudabdol have you also checked how this results for dark theme?

amirmasoudabdol:
@JorisGoosen yes, but I forgot to include it! 🤦🏻‍♂️ 

So, here how they look like on dark mode:

CleanShot 2021-10-20 at 15 54 22

JorisGoosen:
Well, it does look nicer I think.

@TimKDJ you agree?

JorisGoosen (approved):

TimKDJ (approved):
Yup looks better

PR #4656, opened 2021-10-15
Enhancement to Column Header Implementation and UI

amirmasoudabdol:
One of the problem that we have with the `DataTableView` is that the column header is getting crowded, and there are a lot of hidden interactions, ie. rename, change label, and these often fight with each other. For instance, there is no indication that clicking on the column name is going to do something, and currently, it sometimes renames, sometimes changes the label.

In addition, as we are adding new features, we need to add more functionality to it, and there is not enough space anymore. So, in a brief talk that I had with Bruno, we thought it would a good idea to move most of these into a tiny button and cleanup the column header. For this to work, I think we can move all those function into a tiny menu on the right of the column header, and remove most of the current interactions, and free it up for selecting the column, like Excel. 

A right-clicked context menu would work too but I think considering our user base, an indicator will not hurt. That being said, I am not the big fan of having a lot of icons there, so, maybe we can do something about the computed, and filtered logo as well. There are two main problems with having too many actionable icons in the column header, 1) it starts to look ugly and crowded, 2) people will have difficulty clicking them, especially if we want to be make JASP more accessible squeezing many options into a small area is not a good idea.

So, this PR is a prototype, and the icon is not great, but I'm thinking if we have a permanent and elegant icon (maybe even hovered enabled/visible), then we can fill it up with some column specific functions, e.g., rename, change label, change type, add filter, create a computed column, add column to left/right, etc; and therefore simplifying the header and remove most of its mouse triggers. 

CleanShot 2021-10-15 at 15 13 43

One option for the icon would be this one, https://fontawesome.com/v5.15/icons/bars?style=solid. The variant of this is being used for this purpose, so some editing is needed, but something in this direction. I will think of a better one if we decide that this is a good direction to go.

Beyond the discussion, this PR:

- Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/1602
- and if we have the menu, it helps with this https://github.com/jasp-stats/INTERNAL-jasp/issues/1588 
- and I think it address this as well, https://github.com/jasp-stats/INTERNAL-jasp/issues/1582
- and again having the menu will make this more intuitive, https://github.com/jasp-stats/INTERNAL-jasp/issues/1577


amirmasoudabdol:
As we discussed, I'm moving some of the functionalities into a menu that appears when the user hovers the mouse over the column header. I think this PR is not a Draft anymore, and it can be reviewed and merged into the dataMode working branch whenever we are picking up the Data Editing. It's not done yet, but it is at the state that we can work on it together at least.

boutinb:
This has become obsolete

PR #4659, opened 2021-10-26
Dynamic options taken from source are not always updated

boutinb:
When the definition of the source self changes, the connection between signal/slot are not set properly
Fixes https://github.com/jasp-stats/jasp-test-release/issues/1582
Fixes https://github.com/jasp-stats/jasp-test-release/issues/1587
Fixes https://github.com/jasp-stats/jasp-test-release/issues/1625


Kucharssim (approved):
This seems to work!

PR #4669, opened 2021-11-05
Making sure that the label scales properly

amirmasoudabdol:
Fixes this, https://github.com/jasp-stats/jasp-desktop/pull/4668#event-5569428307


boutinb:
I solved this problem in this PR: https://github.com/jasp-stats/jasp-desktop/pull/4668
Maybe your solution is more elegant. But you have first to solve the file conflict.

PR #4672, opened 2021-11-10
Bump jaspColumnEncoder + small changes to the flatpak script

vandenman:
Bumps jaspColumnEncoder to the new master (required for flatpak).

The install script should be a bit more robust, but if there is an error it now prints a stack trace and (in non-interactive mode, so `Rscript`) it exits (with status 1):

image

This stack trace could be more pretty but I hope it suffices for now.

The versions of remotes and renv are no longer hardcoded, so that shouldn't cause errors anymore.


JorisGoosen (approved):

AlexanderLyNL:
Will this affect the 0.16 windows and mac builds?

vandenman:
@AlexanderLyNL No.

PR #4673, opened 2021-11-10
Make sure data of circular is copied as well.

JorisGoosen:
Otherwise jaspCircular doesnt work


JorisGoosen:
I had thought about asking @Kucharssim for a review but figured it won't help because he probably doesnt know much about cmd/batch and/or shellscripts right?

And this only comes into play for a nightly, not a local build

boutinb (approved):

boutinb:
Apparently this is the same fix as for viridisLite, so good chance it works... Let's see with the build.

PR #4677, opened 2021-11-16
[WIP] Qt6 - Bringing JASP up to speed

JorisGoosen:
Yeah so they changed a lot of stuff and I wanted to test something with the new Qt.

So this is a corrolary to https://github.com/jasp-stats/jasp-desktop/pull/4658 in such a way that @amirmasoudabdol can get notifications about it.


Desktop/data/datasetpackage.cpp
```diff
@@ -205,22 +267,26 @@ int DataSetPackage::rowCount(const QModelIndex & parent) const
 		int labelSize = col.labels().size();
 		return labelSize;
 	}
-	case parIdxType::filter:
-	case parIdxType::root:		//return int(parIdxType::leaf); Its more logical to get the actual datasize
-	case parIdxType::data:		return !_dataSet ? 0 : _dataSet->rowCount();
+	case parIdxType::filterRoot:	[[clang::fallthrough]];
```
amirmasoudabdol: I think you can just use `[[fallthrough]]`, no?

JorisGoosen: I can definitely try, that would look a lot better.

JorisGoosen: Well it does indeed work, maybe the last time I tried that we hadn't switched to >C++11 and it was missing?
Anyway, much better like this.

amirmasoudabdol: That could very well be. Full C++17 is not very new, especially on GCC.

JorisGoosen: Haha, no I am aware we are living in the future, but it took us a while to switch to >11 even.

amirmasoudabdol:
I'm testing this a bit, and it looks mostly fine, but I see this strange/bad behavior with the help form, and resizing. I guess this only happens on Mac but we need to fix it. There are cases where you cannot see anything on JASP, help, or both windows. There are some random crashes that I cannot reproduce as well. Sometimes the Engine just dies which is weird!



CleanShot 2022-01-07 at 11 26 20@2x

I talked to Tim and Bruno, and I think we can merge this right after releasing the 0.16.1, and provide nightlies for those who are facing the slowness issues. In the meantime, we can refine and test it and prepare it for 0.17. However, if we see it being stable during the testing, we can have a small release and only push this patch out as 0.16.2.

JorisGoosen:
Pfooh, the flickering looks bad!

Did this also occur on normal mac? 

If so I think your suggestion is a good one. But I would still want to put it in 0.16.1 for windows (or at least a quick release) because this bug is now forcing people to work with 0.14 or something ancient like that...

AlexanderLyNL:
I'd prefer to have 0.16.1 out without this. Once that's out we can upgrade to Qt 6 so the upgrade is isolated

JorisGoosen:
> I'd prefer to have 0.16.1 out without this. Once that's out we can upgrade to Qt 6 so the upgrade is isolated

Ok but what about https://github.com/jasp-stats/jasp-issues/issues/1490 https://github.com/jasp-stats/jasp-issues/issues/1431 ?

Also, at least 1 person at SKF has this same problem. 
I do agree that it is nice to deliver the same version to all systems but also: having a usable JASP on all windows machines would be a good thing right? Most users are on windows after all.

EJWagenmakers:
A compromise could be to do 0.16.1 without the upgrade, but a quick 0.16.2 solely with that upgrade. More work though.

JorisGoosen:
> A compromise could be to do 0.16.1 without the upgrade, but a quick 0.16.2 solely with that upgrade. More work though.

More work, but also, the others are right that the changes need to be tested and this way we achieve that very nicely. Also gives users a relatively light way to still use the new stuff in 0.16.1 without Qt6 and the accompanying changes messing it up.

amirmasoudabdol:
I think Qt 6.2.3 has fixed the flickering issue on macOS! 👏🏼

boutinb:
> I think Qt 6.2.3 has fixed the flickering issue on macOS! 👏🏼

That's another really good news!

PR #4678, opened 2021-11-16
Change in a R Source of a ComponentList does not refresh the controls correctly

boutinb:
In the Cochrane Meta Analysis, a ComponentsList of a ComponentsLists of Checkboxes R Sources. These R sources are changed according to some options.
The ComponentsList were not updated correctly each time the R sources changed.
This was due to bad caching (the _rSourceModelMap in AnalysisForm was not updated correctly and is in fact not needed), and by not signaling that the analysis should be rerun again.

This is already tested by @FBartos in the cochraneFix branch.

### DIFF
--- a/Desktop/analysis/analysisform.cpp
+++ b/Desktop/analysis/analysisform.cpp
@@ -561,15 +561,6 @@ void AnalysisForm::setAnalysis(QVariant analysis)
 	setAnalysisUp();
 }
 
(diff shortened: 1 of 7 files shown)


boutinb:
@JorisGoosen could you review (and merge) this PR?

JorisGoosen (approved):
Looks a lot simpler like this!

PR #4680, opened 2021-11-18
Flatpak 0.16 - more edits to setup-rpkgs

vandenman:
This PR contains some fixes that are applied to the tar.gz that is currently used for the flatpak build:

- in the .Rbuildignore for flatpak, do not remove any files that contain "JASP" and end with .R.
- for the v8 R package, always apply the fix to the configure file to both the source package in `renv-root/.../v8_<version>.tar.gz` and to `local-cran/src/contrib/v8_<version>.tar.gz`. That should be a bit more robust when commands are run out of order (or run multiple times).
- delete some unused stuff (e.g., `downloadFakeV8`, we're not doing that anymore).
- started on binary packages. To try it out, uncomment
```r
# options(repos = list(repos = c(RSPM = "https://packagemanager.rstudio.com/all/__linux__/focal/latest")))
# options("binaryPkgs" = TRUE)
```
and then everything else should be automatic. This may speed up local builds in the future.

### DIFF
--- a/Tools/flatpak/setup-rpkgs/R/flatpakGeneratePkgsList.R
+++ b/Tools/flatpak/setup-rpkgs/R/flatpakGeneratePkgsList.R
@@ -37,7 +37,9 @@ source(file.path("R", "validators.R"))
 
 validateSetup(jaspDir, flatpakDir)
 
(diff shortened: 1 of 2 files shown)


JorisGoosen (approved):
looks good

PR #4682, opened 2021-11-19
Add the possibility to duplicate a tab.

boutinb:
Add duplicateWhenAdding property to allow to duplicate the content of the current tab when adding a new one.
VariablesForm could not be set in a TabView because the dropKeys property (this sets the relationship between the available & assigned VariablesLists) was handled statically: it is now handled dynamically so that a VariablesForm can also be created dynamically and not only during the initialization of the Analysis form.
Also clean up the TabView & ComponentsList QML components: they use both the ComponentListBase object so common properties should be set in ComponentListBase.
Also the AssignType is not used anymore (there was long ago the possibility to use an option in the arrow to move variables, so that you could tell which kind of interaction between multiple variables you would like): this is also cleaned up.

### DIFF
--- a/Desktop/widgets/variableslistbase.cpp
+++ b/Desktop/widgets/variableslistbase.cpp
@@ -61,38 +61,11 @@ void VariablesListBase::setUp()
 		}
 	}
 
-	ListModelAssignedInterface* assignedModel = qobject_cast<ListModelAssignedInterface*>(_draggableModel);
+	_setRelations();
 
-	if (assignedModel)
-	{
-		ListModel* relatedModel = getRelatedModel();
-		if (!relatedModel)
-		{
-			if (sourceItems().empty() && !property("debug").toBool())
-				addControlError(tr("Cannot find source for VariableList %1").arg(name()));
-		}
-		else
-		{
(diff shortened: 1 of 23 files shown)


JorisGoosen:
Should this be reviewed and merged for 0.16? By someone?

Also, it would be good to remove the commit to `jaspMetaAnalysis` I think.

Desktop/widgets/variableslistbase.cpp
```diff
@@ -242,18 +215,17 @@ void VariablesListBase::itemsDroppedHandler(QVariant vindexes, QVariant vdropLis
 	
 	_tempDropModel		= dropModel;
 	_tempDropItemIndex	= dropItemIndex;
-	_tempAssignOption	= JASPControl::AssignType(assignOption);
 	// the call to itemsDropped is called from an item that will be removed (the items of the variable list
 	// will be re-created). So itemsDropped should not call _moveItems directly.
 	QTimer::singleShot(0, this, SLOT(moveItemsDelayedHandler()));
```
JorisGoosen, line 220: Queued?

JorisGoosen (commented):
The code looks quite reasonable, but I would rather have liked to be able to test it.

This isn't used anywhere yet is it?

Maybe you could add a dummy analysis or something to `jaspTestModule` ?

PR #4685, opened 2021-11-24
About Box should use the right fonts and be resizable.

boutinb:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/1458: it uses the user selected font and it is also zoomable.
I'm not expert in APA rules, but the citation in the About box is the same as in https://jasp-stats.org/faq/how-do-i-cite-jasp/, so I'm not sure what is wrong here. @vandenman do you know this?

### DIFF
--- a/Desktop/components/JASP/Widgets/AboutWindow.qml
+++ b/Desktop/components/JASP/Widgets/AboutWindow.qml
@@ -58,9 +58,9 @@ Window
 		anchors
 		{
 			left:		parent.left
-			leftMargin: 50
+			leftMargin: 50 * preferencesModel.uiScale
 			top:		aboutInfoBox.top
-			topMargin:	25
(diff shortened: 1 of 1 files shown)


vandenman:
The citation is fine. I did notice two weird things though:

1. Zooming out keeps the window size but adds empty white space: 
image
2. Zooming in resizes the window if it doesn't fit.

boutinb:
Hmm I don't get this problem on Mac. Did this problem exist also on 0.15?

boutinb:
Also fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/738

vandenman:
The problem is also there in 0.16, maybe it's a Linux thing.

vandenman (approved):

PR #4688, opened 2021-11-30
Renamed FileReader to ArchiveReader

JorisGoosen:
Because that is what it is now.
No more weird hybrid of a stream and a c-library.

The one place where the "simple filereading"  was used is replaced by just using a stream directly

Not sure if we should merge it for 0.16.1 or not. I think it is fairly safe. It is only being used used in importing JASP, ODS files.
The simple-file-reading functionality seemed to be only used for storin plot.pngs and state-files of R in the archive (as in, for readin them in to store them).

And quite curious if XCodes static analyzer now stops complaining about it ^^


amirmasoudabdol (approved):
I made a few changes mostly in favor of clarity and readability. Now you may want to review them! 😅

I think overall it is much better than the old version! I'm so happy that you have changed the `isClosed()` to `isOpen()`. I don't think I have ever seen an `is_closed()` function! 🤔

Common/archivereader.cpp
```diff
@@ -0,0 +1,219 @@
...
+#else
+#define ARCHIVE_OPEN_FILENAME_FUNC archive_read_open_filename
+#define PATH_CONVERSION
+#endif
+
+void ArchiveReader::openEntry(const string &archivePath, const string &entryPath)
+{
+	boost::filesystem::path pathArchive = PATH_CONVERSION(archivePath);
+
+	if ((_archiveExists = boost::filesystem::exists(pathArchive)))
```
amirmasoudabdol: I would happily sacrifice a line for clarity here...

amirmasoudabdol:
I think it is going to be safe for 0.16.1 and if it passes a test on Windows at least on your side, I think we can merge it and see if it causes any issues in any of the nightlies, but as you said, I don't expect it to be problematic, and it worked fine on my broken setup! 🤷🏻‍♂️

For the Xcode, I'm going to run their test again as soon as I start merging some of the new Qt6 changes to the CMake branch...

Common/archivereader.h
```diff
@@ -109,22 +89,22 @@ class FileReader
 	void reset();
 
 	/**
-	 * @brief isClosed Checks for file/archive not open.
-	 * @return true if file closed.
+	 * @brief isOpen Checks for file/archive being open.
+	 * @return true if file opened...
 	 */
-	bool isClosed();
+	bool isOpen();
```
amirmasoudabdol: This guy was left empty. I think you are really used to Qt taking care of these! 🙃

JorisGoosen: Oh haha, I guess I messed up the "move to archivereader.h" together with renaming the file or something :p

Merci

Common/archivereader.cpp
JorisGoosen: I mean, it is right above and it just becomes so ugly with those 
```
#ifdef blabla
some code
#else
almost the exact same code just begging for a typo
#endif
```

But then again, I do not see a lot of harm in macros.

JorisGoosen:
> I made a few changes mostly in favor of clarity and readability. Now you may want to review them! 😅

Thanks!

> I think overall it is much better than the old version! I'm so happy that you have changed the `isClosed()` to `isOpen()`. I don't think I have ever seen an `is_closed()` function! 🤔

Haha, yeah what was up with that?

Desktop/data/exporters/jaspexporter.cpp
```diff
@@ -264,14 +264,12 @@ void JASPExporter::saveJASPArchive(archive *a, boost::function<void(int)>)
 			std::vector<std::string> paths = TempFiles::retrieveList(analysisJson["id"].asInt());
 			for (size_t j = 0; j < paths.size(); j++)
 			{
-				boost::nowide::ifstream readTempFile(TempFiles::sessionDirName() + "/" + paths[j], std::ios::in | std::ios::binary);
+				// std::ios::ate seeks to the end of stream immediately after open
```
JorisGoosen, line 267: much nicer

Common/archivereader.cpp
amirmasoudabdol: I was not talking about the MACROs, I was talking about this, `if((_archiveExists = boost::filesystem::exists(pathArchive)))`, that I have made it into two lines, 

```cpp
 _archiveExists = boost::filesystem::exists(pathArchive);

 if (_archiveExists)
```

JorisGoosen: oow yeah, that is not a bad move indeed

PR #4693, opened 2022-01-06
Removing calls to `convertPagePathToLower`

amirmasoudabdol:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/1635

Bain already has its help files named correctly, and it can be used to test this without the hassle of having to update all other modules. 

@vandenman, if you are running Linux, please give this a try. I'm a bit confused whether the helpPath is relative to the Module's path or a full path from the JASP executable. I think it's relative, see `Analysis::fullHelpPath`.

### DIFF
--- a/Desktop/utilities/helpmodel.cpp
+++ b/Desktop/utilities/helpmodel.cpp
@@ -72,7 +72,6 @@ void HelpModel::setMarkdown(QString markdown)
 
 void HelpModel::setPagePath(QString pagePath)
 {	
-	_pagePath = convertPagePathToLower(pagePath); //Otherwise we get into to trouble on systems that discern between cases in the filesystem. Also means we should make sure all documentation html are named lowercase!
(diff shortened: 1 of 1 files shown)


vandenman:
@amirmasoudabdol Bain and Mixed Models work, but other modules (ANOVA, and Regression) fail. The failing modules have incorrect capitalization though.

vandenman (approved):

amirmasoudabdol:
Thanks! That's good enough I guess for now. We can test further after merging those changes, but I think it should be fine.

JorisGoosen:
>  I'm a bit confused whether the helpPath is relative to the Module's path or a full path from the JASP executable. I think it's relative, see `Analysis::fullHelpPath`.

It should now be relative to module path, it used to be relative to jasp exec.

JorisGoosen (approved):

PR #4695, opened 2022-01-10
Replaced the `KeyNavigation.KeyUp` and `.KeyDown` with `tab` and `backTab`

amirmasoudabdol:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/728 and fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/1653

I tweaked the SpinBox just enough that it doesn't mess up with the navigation. It works much more predictable now. There are more to fix there, but I think we can just get rid of it at some point, and switch to the one that comes with Qt out of the box.

Also, in this, I have changed a few things to make the overall experience of the preferences consistent. I notice that we sometimes hide, or disable controls. I think disabling them is better than hiding them, and we do disable them for the most part, so I changed them all to be visible but enabled/disabled based on their values. You can see the changes in the `PrefsData`.

### DIFF
--- a/Desktop/components/JASP/Widgets/FileMenu/Computer.qml
+++ b/Desktop/components/JASP/Widgets/FileMenu/Computer.qml
@@ -60,9 +60,7 @@ Item
 			bottomMargin:	jaspTheme.generalMenuMargin
 		}
 
-		KeyNavigation.down:		browseButton
 		KeyNavigation.tab:		browseButton
-		KeyNavigation.up:		browseButton
 		KeyNavigation.backtab:	browseButton
 	}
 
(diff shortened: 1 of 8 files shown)


boutinb:
All the KeyNavigation,backtab and KeyNavigation.tab can be in fact removed. The QML engine is clever enough to find the next or previous control.

amirmasoudabdol:
That's what I was hoping but I'm not sure if that's the case in different platforms. I still cannot do a lot of the keyboard navigation on my Mac, and if you remember we discussed this when I showed it to you. I don't know what's the case, but it's not working everywhere.

boutinb:
> That's what I was hoping but I'm not sure if that's the case in different platforms. I still cannot do a lot of the keyboard navigation on my Mac, and if you remember we discussed this when I showed it to you. I don't know what's the case, but it's not working everywhere.

I will try it tomorrow on my PC-Windows at the UvA.
I have also removed in another PR the period in 'Help us translate or improve JASP in your language.' You have now conflicts, but you just has to remove the files from your PR.

amirmasoudabdol:
Resolved the conflict! 👍🏼

JorisGoosen (approved):

PR #4696, opened 2022-01-10
Qt Acknowledgement

amirmasoudabdol:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/803

@AlexanderLyNL, if we need to mention anything else, feel free to just add it to the `NOTICE.txt`.


AlexanderLyNL:
Better than nothing, good enough for me. We can always add more to it.

AlexanderLyNL (approved):

PR #4703, opened 2022-01-24
Fixes the missing help files in the preferences panels

amirmasoudabdol:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/1729

- The problem was that `setPagePath` was not actually setting the path. I think there are multiple entry points to this class, and some are not strictly going through the conventional setter/getter.
- I renamed those help files too, for the sake of consistency.


shun2wang:
Hi, @amirmasoudabdol 
I'd like to ask you to confirm if this fix will also work for the image editor's help file, as I found this help documents not available either，but don't see the .md file about the image editor being changed .I guess this shouldn't require reopening a new issue when `setPagePath` work for it. 😀

boutinb (approved):

PR #4708, opened 2022-02-04
The value of a control initialized by a file should not change via binding

boutinb:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/1760

If a value of a control is set with a binding in QML (for example a DoubleField is set per default by the number of variables selected by the user), this value can be changed by the user self to another value. If this value is saved in a JASP file, when reading the file, JASP must prevent the binding to overwrite the value.
This mechanism exists already via the _resetBindingValue JASPControl function. But in a case of an upgrade, an option that did not exist in a previous version, may have a default value set via a binding. In this particular case, the value should not be reset.

### DIFF
--- a/Desktop/analysis/analysisform.cpp
+++ b/Desktop/analysis/analysisform.cpp
@@ -358,6 +358,7 @@ void AnalysisForm::bindTo()
 	
 	for (JASPControl* control : _dependsOrderedCtrls)
 	{
+		bool hasOption = false;
 		BoundControl* boundControl = control->boundControl();
 		JASPListControl* listControl = dynamic_cast<JASPListControl *>(control);
 
@@ -375,6 +376,8 @@ void AnalysisForm::bindTo()
 
 			if (optionValue == Json::nullValue)
 				optionValue = boundControl->createJson();
+			else
+				hasOption = true;
 
 			boundControl->bindTo(optionValue);
 		}
@@ -397,7 +400,7 @@ void AnalysisForm::bindTo()
 			}
 		}
 
-		control->setInitialized();
+		control->setInitialized(hasOption);
 	}
 
 	for (ListModelAvailableInterface* availableModel : availableModelsToBeReset)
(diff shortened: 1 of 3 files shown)


JorisGoosen:
So... "initializedByFile" means "defaultValueFromJaspFile" ? Or what?

Even if I read the problem description and this issue I find it a bit vague to be honest.

boutinb:
> So... "initializedByFile" means "defaultValueFromJaspFile" ? Or what?
> 
> Even if I read the problem description and this issue I find it a bit vague to be honest.

Yes, initializedByFile means that its initialized value is set by a JASP file. I can rename it initializedFromJaspFile.
The issue is indeed quite subtle... It happens because the JASP file made in 0.16 missed an option (Inverse Wishart df) which is added in 0.16.1 with a default value set to the numbers of selected variables (its minimum is also set to the numbers of selected variables, the user can set it to a higher value, but not a smaller one).
A binding overwrite the initial value set by the JASP file. This is usually not wanted: if the user overwrites the value and saves it to a JASP file, he wants to have it back when he reads this JASP file. That's what the function _resetBindingValue do
```
	// If a control gets a value from a JASP file, this value may differ from its default value sets by a QML binding:
	// this QML binding may then change the value during the initialization of the form.
	// In this case, restore the original value.

```
But in the particular case that the option does not exist in the JASP file because it is a new option, then the function _resetBindingValue should not reset to the initial value (which is 0 in this case since it had no value in the JASP file). That's why I need to know whether the initial value is really set by an option in the JASP file or not.

JorisGoosen:
> I can rename it initializedFromJaspFile. 

I think this is a good idea, because there are an awful lot of types of files...

> default value set to the numbers of selected variables

This is the binding then right?

> _resetBindingValue should not reset to the initial value (which is 0 in this case since it had no value in the JASP file).

Why is it setting the value if there was no value in the jasp file?
As in, isnt it easier to do:
`void resetBoundValue() override { if(!_orgValue.isNull()) bindTo(_orgValue); }`

That also not resets when there was no original value right?

And it saves us adding yet another boolean.

boutinb:
I have changed the name of the boolean to initializedFromJaspFile

> Why is it setting the value if there was no value in the jasp file?
> As in, isnt it easier to do:
> `void resetBoundValue() override { if(!_orgValue.isNull()) bindTo(_orgValue); }`

The option must have anyway a default value (and then _orgValue). If there is no binding, then this default value will be used. So your solution won't work.

JorisGoosen (approved):

PR #4716, opened 2022-02-09
Help page for preferences do not show up

boutinb:


boutinb:
This is already covered by https://github.com/jasp-stats/jasp-desktop/pull/4703

PR #4721, opened 2022-02-18
Remove TypeError in CheckBox and RadioButton

boutinb:
With Qt6, the CheckBoxes and RadioButtons generated TypeErrors due to apparently https://forum.qt.io/topic/131823/lots-of-typeerrors-in-console-when-migrating-to-qt6


JorisGoosen (approved):
Would be good if this can be tested on KDE, but my linux is still broken so: https://github.com/jasp-stats/INTERNAL-jasp/issues/1742 Im guessing it will be fine though.

PR #4726, opened 2022-03-03
This makes dynamic modules and engines behave themselves much much better

JorisGoosen:
- adds a "reloadData"  command for the engine so that it does that
- Also tries to be more selective about pausing/resuming engines when loading data. Now it is only done for those that actually need the data.
- The rest are just given a signal to reload after the data has changed (such moduleInstallRequest or so)
- Add better event logging from JorisGoosen/jasp-desktop@slowdownHunter
- preferably kill longest idle engine first
- actually log output from R
- Only loading qml when dynamic module is ready for use solves all the crashing it seems
- show a loader while waiting for qml
- Enginewindow added where you can monitor what all the engines are doing (Ctrl+Shift+Alt+E or via prefs-advanced)

- startup better now make sure that shutdown works
- try and keep IPCChannel slaves finding its objects by giving it time. #It doesnt seem to help though and makes everything more complicated :(
-- This is not enough to solve all problems but allows for working around small timing problems
- to further avoid problems the channels are kept separate from the engine and are no long destroyed ever, this way the communication objects aren't lost.
- and anyhow this is only a problem when the max engine count is changed anyway.
- Make sure channel() isnt used when nullptr

- Engines can be killed in enginewindow with middlemouse
- same engine  for installing a module as running it please. This avoids all those "cannot open connection" errors
- make sure forms are destroyed before the rest of qml is so that we arent inundated with errormessages as it exhales its dying breath
- remove useless "initializing" state from Analysis
- fixes dragging issues
- uninstall developer module if path is changed

### DIFF
--- a/Desktop/analysis/analyses.cpp
+++ b/Desktop/analysis/analyses.cpp
@@ -47,6 +47,14 @@ Analyses::Analyses()
 	new KnownIssues(this);
 }
 
+void Analyses::destroyAllForms() 
+{ 
+	Log::log() << "Analyses::destroyAllForms()" << std::endl;
+	
+	//Destroy all existing forms *before* destroying the rest of QML, to avoid a massive slew of errors
+	applyToAll([](Analysis * a){ if(a->form()) a->destroyForm(); });
+}
+
 
 Analysis* Analyses::createFromJaspFileEntry(Json::Value analysisData, RibbonModel* ribbonModel)
 {
@@ -100,9 +108,6 @@ Analysis* Analyses::createFromJaspFileEntry(Json::Value analysisData, RibbonMode
 		analysis			= create(analysisData, analysisEntry, id, status, false, title, analysisData["dynamicModule"]["moduleVersion"].asString(), &optionsJson);
 	}
 
-	if(analysisEntry && analysisEntry->dynamicModule() && !analysisEntry->dynamicModule()->loaded())
-		analysisEntry->dynamicModule()->setLoadingNeeded();
-
 	if(wasUpgraded)
 		analysis->setUpgradeMsgs(msgs);
 
@@ -147,7 +152,7 @@ void Analyses::storeAnalysis(Analysis* analysis, size_t id, bool notifyAll)
 	{
 		emit analysisAdded(analysis);
 		setCurrentAnalysisIndex(_orderedIds.size() - 1);
-		showAnalysisInResults(id);
+		emit showAnalysisInResults(id);
 	}
 }
 
@@ -165,20 +170,7 @@ void Analyses::bindAnalysisHandler(Analysis* analysis)
 	connect(analysis,	&Analysis::titleChanged,						this, &Analyses::somethingModified					);
 	connect(analysis,	&Analysis::imageChanged,						this, &Analyses::somethingModified					);
 	connect(analysis,	&Analysis::userDataChangedSignal,				this, &Analyses::analysisOverwriteUserdata			);
-
-	if (Settings::value(Settings::DEVELOPER_MODE).toBool())
-	{
-		QString filePath = tq(analysis->qmlFormPath());
-		
-		if (filePath.startsWith("file:"))
-			filePath.remove(0,5);
-		if (!_QMLFileWatcher.files().contains(filePath))
-		{
-			if (!_QMLFileWatcher.addPath(filePath))
-				Log::log()  << "Could not watch: " << filePath.toStdString() << std::endl;
-		}
-		connect(&_QMLFileWatcher, &QFileSystemWatcher::fileChanged, [=] () { this->_analysisQMLFileChanged(analysis); });
-	}
+	connect(analysis,	&Analysis::emptyQMLCache,						this, &Analyses::emptyQMLCache						);
 }
 
 void Analyses::clear()
@@ -206,19 +198,18 @@ void Analyses::clear()
 	emit countChanged();
 }
 
-void Analyses::reload(Analysis *analysis, bool logProblem)
+void Analyses::reload(Analysis *analysis, bool qmlFileNameChanged, bool logProblem)
 {
 
 	for (size_t i = 0; i < _orderedIds.size(); i++)
 		if (_analysisMap[_orderedIds[i]] == analysis)
 		{
 			int ind = int(i);
-			// Force the loader to load again the QML file
-			beginRemoveRows(QModelIndex(), ind, ind);
-			endRemoveRows();
+			if(!qmlFileNameChanged)	Log::log() << "Analyses::reload(" << analysis << ") Force a reload of QML file '" << analysis->qmlFormPath() << "' for " << analysis->name() << "("<< analysis->id() << ")." << std::endl;
+			else					Log::log() << "Analyses::reload(" << analysis << ") emit dataChanged for QML file '" << analysis->qmlFormPath() << "' for " << analysis->name() << "("<< analysis->id() << ")." << std::endl;
 
-			beginInsertRows(QModelIndex(), ind, ind);
-			endInsertRows();
+			if(!qmlFileNameChanged)	analysis->reloadForm();
+			else					emit dataChanged(index(ind), index(ind), QList<int>({ Qt::DisplayRole, formPathRole}));
 
 			return;
 		}
@@ -227,6 +218,7 @@ void Analyses::reload(Analysis *analysis, bool logProblem)
 	if(logProblem)
 		Log::log() << "Analysis " << analysis->title() << " not found!" << std::endl;
 }
+
 
 
 bool Analyses::allFresh() const
@@ -238,11 +230,7 @@ bool Analyses::allFresh() const
 	return true;
 }
 
-void Analyses::_analysisQMLFileChanged(Analysis *analysis)
-{
-	emit emptyQMLCache();
-	reload(analysis, false); //Do not log problem because it can cause trouble!
-}
+
 
 Json::Value Analyses::asJson() const
 {
@@ -308,7 +296,7 @@ void Analyses::removeAnalysesOfDynamicModule(Modules::DynamicModule * module)
 
 void Analyses::refreshAnalysesOfDynamicModule(Modules::DynamicModule * module)
 {
-	Log::log() << "void RibbonModel::dynamicModuleChanged(" << module->toString() << ")" << std::endl;
+	//Log::log() << "void Analyses::refreshAnalysesOfDynamicModule(" << module->toString() << ")" << std::endl;
 
 
 	for(auto & keyval : _analysisMap)
@@ -335,8 +323,18 @@ void Analyses::rescanAnalysisEntriesOfDynamicModule(Modules::DynamicModule * mod
 {
 	std::set<int> removeIds;
 	for(auto & keyval : _analysisMap)
-		if(keyval.second->dynamicModule() == module  && !keyval.second->checkAnalysisEntry()) // Check if the analysisEntry this analysis is based still exists
-			removeIds.insert(keyval.first);
+		if(keyval.second->dynamicModule() == module)
+		{
+			if(!keyval.second->checkAnalysisEntry()) // Check if the analysisEntry this analysis is based still exists
+				removeIds.insert(keyval.first);
+			else
+			{
+				Analysis * a = keyval.second;
+
+				if(!a->form() && a->readyToCreateForm())
+					a->createForm();
+			}
+		}
 
 	for(const int & id : removeIds)
 		removeAnalysisById(size_t(id));
@@ -381,9 +379,6 @@ void Analyses::loadAnalysesFromDatasetPackage(bool & errorFound, stringstream & 
 	{
 		int				corruptAnalyses = 0;
 		stringstream	corruptionStrings;
-		Analysis	*	currentAnalysis = nullptr;
-
-		//This really should all be moved to Analyses!
 
 		Json::Value analysesData = DataSetPackage::pkg()->analysesData();
 		if (analysesData.isNull())
@@ -416,7 +411,7 @@ void Analyses::loadAnalysesFromDatasetPackage(bool & errorFound, stringstream & 
 			{
 				try
 				{
-					currentAnalysis = createFromJaspFileEntry(analysisData, ribbonModel);
+					createFromJaspFileEntry(analysisData, ribbonModel);
 				}
 				catch (Modules::ModuleException modProb)
 				{
--- a/Desktop/analysis/analyses.h
+++ b/Desktop/analysis/analyses.h
@@ -27,7 +27,6 @@
 #include <QString>
 #include <QMap>
 #include <QAbstractListModel>
-#include <QFileSystemWatcher>
 #include <sstream>
 
 class RibbonModel;
@@ -60,19 +59,20 @@ public:
 					idRole};
 
 						Analyses();
-						~Analyses() { _singleton = nullptr; }
-	static Analyses *	analyses() { return _singleton; }
+						~Analyses()	{ _singleton = nullptr; }
+	static Analyses *	analyses()	{ return _singleton; }
 
 	Analysis	*	createFromJaspFileEntry(Json::Value analysisData, RibbonModel* ribbonModel);
 
-	Analysis	*	create(const Json::Value & analysisData, Modules::AnalysisEntry * analysisEntry, size_t id, Analysis::Status status = Analysis::Initializing, bool notifyAll = true, std::string title = "", std::string moduleVersion = "", Json::Value *options = nullptr);
+	Analysis	*	create(const Json::Value & analysisData, Modules::AnalysisEntry * analysisEntry, size_t id, Analysis::Status status = Analysis::Empty, bool notifyAll = true, std::string title = "", std::string moduleVersion = "", Json::Value *options = nullptr);
 	Analysis	*	create(Modules::AnalysisEntry * analysisEntry)													{ return create(Json::nullValue, analysisEntry, _nextId++);						}
 
 	Analysis	*	operator[](size_t index)	{ return _analysisMap[_orderedIds[index]]; }
 	Analysis	*	get(size_t id) const		{ return _analysisMap.count(id) > 0 ? _analysisMap.at(id) : nullptr;	}
 
 	void			clear();
-	void			reload(Analysis* analysis, bool logProblem);
+	void			reload(Analysis* analysis, bool qmlFileChanged, bool logProblem);
+	void			destroyAllForms();
 
 	bool			allFresh() const;
 	void			setAnalysesUserData(Json::Value userData);
@@ -150,6 +150,7 @@ signals:
 	void sendRScript(					QString		script, int requestID, bool whiteListedVersion);
 	void analysisSelectedIndexResults(	int			row);
 	void showAnalysisInResults(			int			id);
+	void reloadQmlForm(					int			row);
 	void currentAnalysisIndexChanged(	int			currentAnalysisIndex);
 	void currentFormHeightChanged(		double		currentFormHeight);
 	void visibleChanged(				bool		visible);
@@ -175,7 +176,6 @@ private:
 	void bindAnalysisHandler(Analysis* analysis);
 	void storeAnalysis(Analysis* analysis, size_t id, bool notifyAll);	
 	void _makeBackwardCompatible(RibbonModel* ribbonModel, Version& version, Json::Value& analysisData);
-	void _analysisQMLFileChanged(Analysis* analysis);
 
 
 private:
@@ -187,7 +187,6 @@ private:
 	std::map<size_t, Analysis*>		_analysisMap;
 	std::vector<size_t>				_orderedIds;
 	std::vector<size_t>				_orderedIdsBeforeMoving;
-	QFileSystemWatcher				_QMLFileWatcher;
 
 	size_t							_nextId					= 0;
 	int								_currentAnalysisIndex	= -1;
--- a/Desktop/analysis/analysis.cpp
+++ b/Desktop/analysis/analysis.cpp
@@ -26,18 +26,20 @@
 #include "log.h"
 #include "utils.h"
 #include "utilities/settings.h"
+#include "utilities/qutils.h"
+
 
 Analysis::Analysis(size_t id, Modules::AnalysisEntry * analysisEntry, std::string title, std::string moduleVersion, Json::Value *data) :
-	  QObject(Analyses::analyses()),
-	  _id(id),
-	  _name(analysisEntry->function()),
-	  _qml(analysisEntry->qml().empty() ? _name : analysisEntry->qml()),
-	  _titleDefault(analysisEntry->title()),
-	  _title(title == "" ? _titleDefault : title),
-	  _moduleVersion(moduleVersion),
-	  _version(AppInfo::version),
-	  _moduleData(analysisEntry),
-	  _dynamicModule(_moduleData->dynamicModule())
+	  QObject(			Analyses::analyses()),
+	  _id(				id),
+	  _name(			analysisEntry->function()),
+	  _qml(				analysisEntry->qml().empty() ? _name : analysisEntry->qml()),
+	  _titleDefault(	analysisEntry->title()),
+	  _title(			title == "" ? _titleDefault : title),
+	  _moduleVersion(	moduleVersion),
+	  _version(			AppInfo::version),
+	  _moduleData(		analysisEntry),
+	  _dynamicModule(	_moduleData->dynamicModule())
 {
 	if(_moduleVersion == "" && _dynamicModule)
 		_moduleVersion = _dynamicModule->version();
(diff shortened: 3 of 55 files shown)


Desktop/components/JASP/Controls/Form.qml
```diff
@@ -184,8 +183,8 @@ AnalysisForm
 		running:		false
 		repeat:			false
 		interval:		0
-		onTriggered:	formCompleted();
+		onTriggered:	formCompletedSignal();
 	}
-	
-	Component.onCompleted:	bindingTimer.start()
+
+	Component.onCompleted:	bindingTimer.start() //Do we really need both a timer and a Queued connection?
```
JorisGoosen: @boutinb no right?

JorisGoosen: Maybe I do a cleanup round btw, I see some things I might as well remove entirely.
Today or tomorrow.

boutinb: Yes this Timer seems to be obsolete.

boutinb:
Your Engines overview is really nice!!
First remark: if my max # engines is 4, and I open more than 4 analyses in different modules, then it kills all engines, and starts again with 1 engine. Why not killing only one engine?

Desktop/engine/enginesync.cpp
```diff
@@ -443,44 +614,156 @@ bool EngineSync::processAnalysisRequests()
...
+
+size_t EngineSync::enginesStartableCount() const
+{
+	size_t enginesPossible = maxEngineCount() - _engines.size();
+
+	//But perhaps they have to cool down for a bit.
+
+	for(long engineStopTime : _engineStopTimes)
+		if(engineStopTime != -1 && engineStopTime + ENGINE_COOLDOWN > Utils::currentMillis() && enginesPossible > 0)
+			enginesPossible--;
```
boutinb, line 696: Some extra parenthesis could be nice:
if(engineStopTime != -1 && ((engineStopTime + ENGINE_COOLDOWN) > Utils::currentMillis()) && enginesPossible > 0)

Desktop/analysis/analysis.cpp
```diff
@@ -26,58 +26,80 @@
...
+
+	if(!_isDuplicate && isNewAnalysis)
+		_status = Empty;
+
 }
 
 Analysis::~Analysis()
 {
+	if(form())
+		destroyForm();
```
boutinb, line 101: I'm a bit skeptical here: the form is automatically destroyed when its parent is destroyed, so there are some chances that _analysisForm points to a destroyed object.

Desktop/analysis/analysis.h
```diff
@@ -156,141 +164,159 @@ class Analysis : public QObject
...
 	void					processResultsForDependenciesToBeShown();
 	bool					processResultsForDependenciesToBeShownMetaTraverser(const Json::Value & array);
-	bool					_editOptionsOfPlot(const Json::Value & results, const std::string & uniqueName, Json::Value & editOptions);
-	bool					_setEditOptionsOfPlot(Json::Value & results, const std::string & uniqueName, const Json::Value & editOptions);
+	bool					_editOptionsOfPlot(const	Json::Value & results, const std::string & uniqueName,			Json::Value & editOptions);
+	bool					_setEditOptionsOfPlot(		Json::Value & results, const std::string & uniqueName, const	Json::Value & editOptions);
 	void					storeUserDataEtc();
 	void					fitOldUserDataEtc();
 	bool					updatePlotSize(const std::string & plotName, int width, int height, Json::Value & root);
 	Modules::AnalysisEntry	*moduleData();
```
boutinb: moduleData() wordt niet meer gebruikt.

Desktop/analysis/analysis.cpp
```diff
@@ -92,12 +114,13 @@ void Analysis::clearOptions()
 
 bool Analysis::checkAnalysisEntry()
 {
+
```
boutinb, line 130: I was wondering what exactly is the purpose of this method.
This method is called only by setDynamicModule, which is called only by replaceAnalysesOfDynamicModule, so if I understand well, only when an engine has to process a new module? But if I set a breakpoint in this method, the debugger never stops...

Desktop/analysis/analysisform.h
```diff
@@ -116,82 +118,80 @@ public slots:
...
+	QStringList										_formErrors,
+													_formWarnings;
+	QQmlComponent*									_controlErrorMessageComponent	= nullptr;
+	QList<QQuickItem*>								_controlErrorMessageCache;
+	bool											_runOnChange					= true,
+													_formCompleted					= false,
+													_initialized					= false,
+													_signalChangedEmittedButBlocked = false;
+	QString											_info;
+	int												_signalValueChangedsBlocked		= 0;
```
boutinb: _signalValueChangedsBlocked -> _signalValueChangedBlocked

boutinb (changes requested):
Difficult to review the whole code...
I did not spend much time with the shared memory changes, since that will be removed soon.
I have tested it, and it seems to be quite stable

JorisGoosen:
> Your Engines overview is really nice!! First remark: if my max # engines is 4, and I open more than 4 analyses in different modules, then it kills all engines, and starts again with 1 engine. Why not killing only one engine?

Well, there are some timing issues and the analyses were probably all done. Then the engines are idle, it needs 1 but it takes a bit before it is started. Also because I decided to give the shutdown a bit of a cooldown. This to give the boost shared memory file release by the engine some time to complete and avoiding some conflicts through that (aka to hopefully avoid "file not found" etc).

But I can have a look and see if I can tweak it a bit. 

I already noticed another "issue", not really though, where if you install a module it start all allowed engines and then only uses 1 :s 

But that is what JASP always did till now anyway, so its not really a regression I guess.

Desktop/analysis/analysis.cpp
```diff
@@ -835,6 +864,118 @@ void Analysis::setErrorInResults(const std::string & msg)
...
+	{
+		Log::log(false) << " it has a AnalysisForm " << _analysisForm << " so let's destroy it" << std::endl;
+
+		_optionsDotJASP = _boundValues;  //So that we can later reload the controls to what we currently have
+
+		_analysisForm->setParent(		nullptr);
+		_analysisForm->setParentItem(	nullptr);
+
+		delete _analysisForm;
+		_analysisForm = nullptr;
```
JorisGoosen, line 956: @boutinb `destroyForm()` always clears `_analysisForm`.

Desktop/analysis/analysis.cpp
JorisGoosen: This is necessary because we have multiple datastructures representing the same data :s

The `RibbonModel` has its own copy and `Analysis` links to it, so whenever a module is installed it needs to run this to make sure that it still points to the same place. To be able to do this `Analysis::_codedAnalysisEntry` exists.
If you want the debugger to stop here you should install a developer module. Or change the `Description.qml` of an installed developer module.

Desktop/analysis/analysisform.h
JorisGoosen: No this is because you count how many "value changed"s have occurred. This way it is less misleading.
But Im open for better suggestions.

Desktop/components/JASP/Controls/Form.qml
JorisGoosen: or tomorrow, with the other changes...

Desktop/analysis/analysis.cpp
boutinb: Yes, but if the parent of the form is destroyed before, QML will destroy the form without going to this destroyForm method, making _analysisForm a dangling pointer.

JorisGoosen: As we've just talked about, it will be fine

boutinb:
Apparently you forgot to add your changes for enumutilities.h (because it does not compile without).
I have tested it, and I have got these issues:

- In preferences, if you set a wrong path for the developer module, then it crashes when you try to load it
- If you have loaded a developer module, as soon as you change the path in preferences, the developer module is removed. This is quite quick... maybe beter to remove the old one, when the new one is loading.
- if you load a developer module which is not already in the ribbon bar, then it does not come to the ribbon, you need first to click it in the module list. I would set it directly in the ribbon.
- If all engines are set to a module, and you decrease in Preferences the number of engines, it crashes.
- If all engines are set to a module and try to load a developer module of one of the module already used, then you get an error:
Screenshot 2022-03-17 at 13 19 40

boutinb (approved):

boutinb:
I approve it, but it would be nice to have a loading indicator when a developer module is loaded (and which is not already in the ribbon).
It still crashes when you load a developer module, add and run an analysis from this module, and change the path to the developer module.

JorisGoosen:
> I approve it, but it would be nice to have a loading indicator when a developer module is loaded (and which is not already in the ribbon). 

Well, I agree! So I will fix it.

> It still crashes when you load a developer module, add and run an analysis from this module, and change the path to the developer module.

Oh, that means that the `uninstallJASPModule` function, or whatever its called, is not working perfectly?
Which development module did you load?

boutinb:
> Which development module did you load?

It was crashing with jaspAnova or jaspDistributions module

JorisGoosen:
> > > I approve it, but it would be nice to have a loading indicator when a developer module is loaded (and which is not already in the ribbon).
> > 
> > Well, I agree! So I will fix it.

Well, I still agree but I think I can spend my time more sensibly then on this to be honest. Ill leave it as is.

> > > It still crashes when you load a developer module, add and run an analysis from this module, and change the path to the developer module.
> >
> > Oh, that means that the `uninstallJASPModule` function, or whatever its called, is not working perfectly? Which development module did you load?

> It was crashing with jaspAnova or jaspDistributions module

I cannot reproduce this crash at all. If you really still get this crash I can have a look on my mac on thursday and see whether there is a problem still?

JorisGoosen:
@amirmasoudabdol  did you have a chance to have a look at this PR? I think it would be worthwhile to have another pair of eyes look at this as Ive changed a ton of stuff

amirmasoudabdol:
Not yet, I will, as soon as we are done with the 0.16.2.

Desktop/components/JASP/Widgets/RCommanderWindow.qml
```diff
@@ -305,7 +305,7 @@ Window
 				anchors
 				{
 					top:		parent.verticalCenter
-					right:		parent.right
+					left:		parent.left
```
amirmasoudabdol: This should be `right`!

Desktop/mainwindow.cpp
```diff
@@ -510,8 +513,10 @@ void MainWindow::loadQML()
 		ActiveModules::getActiveCommonModules(),
 		ActiveModules::getActiveExtraModules());
```
amirmasoudabdol, line 518: I guess something went wrong with the rebasing, but there should be closing `}` here.

amirmasoudabdol:
While I'm still waiting for the rebase, release, etc., I'm slowing starting to look into this, and I noticed that the RCommander doesn't seem to be working on macOS that I've tested.

Desktop/components/JASP/Widgets/EnginesWindow.qml
```diff
@@ -0,0 +1,106 @@
...
+						verticalCenter:	parent.verticalCenter
+					}
+			}
+		}
+
+		JASPMouseAreaToolTipped
+		{
+			acceptedButtons:	Qt.MiddleButton
+
+			toolTipText:		qsTr("Middle mouse button to kill an engine.")
```
amirmasoudabdol: I think this is going to be problematic. Modern trackpads don't have middle buttons, and often don't have any buttons at all, and that makes it really hard to simulate the middle click. I think we should put an actual button for killing the engine, and  upon click ask the user to make sure.

JorisGoosen: Yes that is a very good point! I can add a little `x`-mark in the toprightcorner for instance? Or the red-circle-delete icon from the analyses-view?

The question I won't do because really there is nothing that could go wrong except for waiting slighlty longer for your analysis or whatever to finish. So that would just be useless clicking then.

amirmasoudabdol:
I just noticed that this contains the missing signing identity bug that I introduced, and later fixed! 😬

Desktop/components/JASP/Widgets/EnginesWindow.qml
amirmasoudabdol: I think the red circle from the analysis should work, and it's also familiar. 

Are you saying that we shouldn't have the terminate functionality at all?

JorisGoosen: No Im saying that the terminate functionality is very safe and no work will be lost no matter what.
So that leads me to conclude that having a warning (aka "upon click ask the user to make sure.") is unnecessary.

JorisGoosen:
> I just noticed that this contains the missing signing identity bug that I introduced, and later fixed! 😬

I rebased that into this code? Or where is it?

amirmasoudabdol:
It just needs a rebase to the `stable` at this point, but don't worry about it now, let's have the Qt fixes out first.

amirmasoudabdol:
I think the Engine Overview form needs some work,

- Those circles are massive, and can probably be a bit smaller, and minimal
  - A normal font size should do the job too
- The form probably doesn't need the maximize button, and it can be a Dialog
- The form can be called several times. I think this should not be allowed. 

I noticed that after some times, Engines will be killed, and upon interaction they come back to life. I guess this is something we do, but I've never noticed it! 

<p float="left">



</p>

Desktop/components/JASP/Widgets/FileMenu/PrefsAdvanced.qml
```diff
@@ -333,9 +333,18 @@ ScrollView
 				to:					16
 				defaultValue:		4
 				stepSize:			1
+				KeyNavigation.tab:	showEnginesWindow
+				KeyNavigation.down:	showEnginesWindow
+				text:				qsTr("Maximum # of engines: ")
```
amirmasoudabdol: I'd replace the `#` with "number".

Desktop/analysis/analyses.cpp
```diff
@@ -47,6 +47,14 @@ Analyses::Analyses()
 	new KnownIssues(this);
 }
 
+void Analyses::destroyAllForms() 
+{ 
+	Log::log() << "Analyses::destroyAllForms()" << std::endl;
+	
+	//Destroy all existing forms *before* destroying the rest of QML, to avoid a massive slew of errors
+	applyToAll([](Analysis * a){ if(a->form()) a->destroyForm(); });
```
amirmasoudabdol, line 55: I looked a bit around, and I think you might be able to improve the implementation of these `Apply*` methods by using [`std::for_each`](https://en.cppreference.com/w/cpp/algorithm/for_each) or `std::transform`; and if that's doable, then you might be able to get some free parallelism by adding an `ExecutionPolicy` to it. 

I think it might be worth looking into, especially since this is a frequent action.

Desktop/analysis/analyses.h
```diff
@@ -187,7 +187,6 @@ private slots:
 	std::map<size_t, Analysis*>		_analysisMap;
```
amirmasoudabdol, line 187: This looks like a great candidate for `std::map<size_t, std::unique_ptr<Analysis>>`, or `std::shared_ptr<>` if you really need it. Then, you can just deconstruct them, and don't worry about the `delete` operation, and everything is suddenly great! I mean, of course not, but I feel this is a very good place to start thinking about it. You need a bit of refactoring to make sure that deconstructors are working properly, but then they are safer and more pleasant to work with! 🌻

Desktop/analysis/analysisform.h
```diff
@@ -116,82 +118,80 @@ public slots:
...
 	Q_INVOKABLE bool		initialized()	const	{ return _initialized; }
 
-	void		addControlError(JASPControl* control, QString message, bool temporary = false, bool warning = false);
-	void		clearControlError(JASPControl* control);
-	void		cleanUpForm();
-	bool		hasError();
+	void			addControlError(JASPControl* control, QString message, bool temporary = false, bool warning = false);
+	void			clearControlError(JASPControl* control);
+	void			cleanUpForm();
+	bool			hasError();
```
amirmasoudabdol, line 124: These `bool` methods should _preferably_ have an `[[nodiscard]]` attribute.

Desktop/engine/enginesync.h
```diff
@@ -90,69 +104,84 @@ public slots:
...
-	Json::Value							_requestWideCastModuleJson		= Json::nullValue;
-	std::map<int, std::string>			_requestWideCastModuleResults;
+										_engineInfo;
 
 	std::queue<RScriptStore*>			_waitingScripts;
-	std::set<EngineRepresentation*>		_workers,		//Your typical run of the mill engine, can do analyses and Rscript/filter stuff
-										_rCmders,		//For those special occassions where you just want to shout at R in a more personal manner
-										_engines,		//Keeping track of all those engines isn't easy...
+	std::map<std::string,
+		EngineRepresentation * >		_moduleEngines;					///<An engine per module active. Engines will be started and closed as needed.
```
amirmasoudabdol: I think you probably need one space after `///<` but I'm not sure if it's required. I would argue that it's more readable with the space.

Desktop/modules/dynamicmodule.cpp
```diff
@@ -770,26 +641,29 @@ void DynamicModule::setStatus(moduleStatus newStatus)
 	if(_status == moduleStatus::installNeeded && newStatus == moduleStatus::installModPkgNeeded)
 		return;
 
-	_status = newStatus;
+	bool readyForUseInvolved = newStatus == moduleStatus::readyForUse || _status == moduleStatus::readyForUse;
```
amirmasoudabdol: A pair of `()` makes this nicer to read.

Desktop/modules/dynamicmodule.cpp
```diff
@@ -770,26 +641,29 @@ void DynamicModule::setStatus(moduleStatus newStatus)
...
 }
 
 void DynamicModule::reloadDescription()
 {
 	try
 	{
-		loadDescriptionFromFolder(fq(_moduleFolder.absoluteFilePath() + "/" + nameQ() + "/"));
+		loadDescriptionFromFolder(fq(_moduleFolder.absoluteFilePath() + "/" + nameQ() + "/"), false);
 	}
 	catch(std::runtime_error e) { return; } //If it doesnt work then never mind.
```
amirmasoudabdol, line 645: Then, why bother at all! 😅

Desktop/utilities/qutils.cpp
```diff
@@ -241,3 +244,83 @@ QString shortenWinPaths(QString in)
 	return in;
 #endif
 }
+
+
+//Turning QMLENGINE_DOES_ALL_THE_WORK on also works fine, but has slightly less transparent errormsgs so isn't recommended
+//#define QMLENGINE_DOES_ALL_THE_WORK
```
amirmasoudabdol, line 250: I guess we want this as an option, if so, I think it can probably go into the `JASP.cmake`.

Desktop/widgets/listmodelfactorsform.cpp
```diff
@@ -78,7 +78,16 @@ void ListModelFactorsForm::initFactors(const FactorVec &factors)
 	}
 	
 	_setTerms(newTerms);
-	endResetModel();	
+	endResetModel();
+}
+
+int ListModelFactorsForm::countVariables() const
+{
+	int count = 0;
```
amirmasoudabdol: You can probably use an `std::accumulate` instead.

Engine/engine.cpp
```diff
@@ -177,23 +190,25 @@ void Engine::beIdle(bool newlyIdle)
...
 	if (_channel->receive(data, timeout))
 	{
-		if(data =="")
+		if(data == "")
+		{
+			Log::log() << "Received nothing..." << std::endl;
 			return false;
+		}
 
 		// JSONCPP_STRING          err;
```
amirmasoudabdol, line 205: I still need to refactor these into something sensible, but I might not get to them in time. So, I leave this comment in case you want to deal with them in this PR as well! 🙏🏼

amirmasoudabdol (changes requested):
As far as I can tell, this works and I left a few comments/suggestions here and there. If there is anything that needs some attention is the "Engine Overview" form, and the "Engine Kill" behavior that we discussed in one of the other comments before. 

I'm a bit surprised by the fact that we are actively killing the engines, even if there are only a few of them running, but I'm sure there is a good reason for it. However, if we can avoid that, I'd opt for keeping them alive and ready to work, especially if the user has already specified that it wants `X` engines running. 

Also, I'm slightly concerned about the maximum number of engines, ie., 16, but I guess we rarely get to that many anyway!

JorisGoosen:
Thanks for reviewing!


> As far as I can tell, this works and I left a few comments/suggestions here and there. If there is anything that needs some attention is the "Engine Overview" form, and the "Engine Kill" behavior that we discussed in one of the other comments before.

Yeah I know, but it is really only meant for debugging purposes which is why I left it so rough.

> I'm a bit surprised by the fact that we are actively killing the engines, even if there are only a few of them running, but I'm sure there is a good reason for it. However, if we can avoid that, I'd opt for keeping them alive and ready to work, especially if the user has already specified that it wants `X` engines running.

Well, the problem is that R doesn't cleanly unload dlls/dynamic-libs. So the only way to safely reload packages etcetera is by actively killing the engines. In fact, it is the entire purpose of this PR. To make sure the correct stuff is loaded in R by killing it when whatever is loaded isn't needed anymore.

> Also, I'm slightly concerned about the maximum number of engines, ie., 16, but I guess we rarely get to that many anyway!

Well, the default is set at 4 initially so it only grows to 16 if the user changes those settings and a follow-up issue for this is: https://github.com/jasp-stats/jasp-issues/issues/1693

I will start processing your comments from beginning may, but I guess you'll be gone by then.

Desktop/components/JASP/Widgets/RCommanderWindow.qml
JorisGoosen: check

Desktop/analysis/analyses.h
JorisGoosen: Well, I like having some control over when things are constructed and deconstructed.
Especially as there are only a few classes where we do this manually anyway. Most of the time we use the whole qobject tree to clean up.

Desktop/analysis/analyses.cpp
JorisGoosen: Could be nice but really quite outside of the scope of this pr

Desktop/mainwindow.cpp
JorisGoosen: > I think the Engine Overview form needs some work,
>       * A normal font size should do the job too

A bit strange the font is so huge for you though, are you zooming?

>     * The form probably doesn't need the maximize button, and it can be a Dialog

I guess so, but it also doesnt matter really.

>     * The form can be called several times. I think this should not be allowed.

I agree, but I dont find it important. If I have some time I can see if I can at something to avoid that.
 
> 
> I noticed that after some times, Engines will be killed, and upon interaction they come back to life. I guess this is something we do, but I've never noticed it!

Well, this PR changes a lot of things in the behaviour of engines ;)
This is one of the main goals actually :p

>     * Those circles are massive, and can probably be a bit smaller, and minimal
> the screenshot you show
Hmm the circles are much larger relative to the text for you then for me. 

image
Is what it looks like in default width of the form

Desktop/components/JASP/Widgets/FileMenu/PrefsAdvanced.qml
JorisGoosen: yeah you are right, I dont know why I was trying to keep that as short as possible there is quite some space.

Desktop/modules/dynamicmodule.cpp
JorisGoosen: Sure

Desktop/modules/dynamicmodule.cpp
JorisGoosen: Because it might have been triggered while the user was changing it, saved it with a bunch of crap.
Or maybe it just got removed or whatever. You still don't want the entire application to crash and burn then.

Desktop/utilities/qutils.cpp
JorisGoosen: Yeah no not really, that was just something I did for debugging but it turned out the problem was somewhere else entirely.
And the code could turn out to be useful so I hid it under that.

Desktop/widgets/listmodelfactorsform.cpp
JorisGoosen: Yes but that is not my code ;)

JorisGoosen:
I've processed the remarks here insofar as I think is useful.

I did however also just fix the permanent installs of modules and still want to do the following:

- Make sure the permanently installed modules are enabled after install. 
- And now they show up at the beginning of the ribbon.  Maybe they should be behind Common? Or even just in the alphabetical location?
- RCommander is broken as well... So I want to fix that too

JorisGoosen:
@boutinb can you verify that the R commander now works for you as well?

These two points:
- Make sure the permanently installed modules are enabled after install.
- And now they show up at the beginning of the ribbon. Maybe they should be behind Common? Or even just in the alphabetical location?

can be done some other time, I think it is time this goes into `development` at least

PR #4730, opened 2022-03-07
CFA: cannot load or duplicate analyses

### LINKED ISSUE #1640: [Bug]: CFA clearing Factors
### JASP Version

0.16.1

### Commit ID

_No response_

### JASP Module

jaspFactor

CFA

Mac OS 10.15.2

### Bug Description

When saving, closing, and then re-opening JASP with a multifactor CFA, or when using the copy this analysis function for an existing CFA, the factors become blank. i.e., the boxes in the CFA analysis window for each factor you are creating become blank. 

Occasionally this bug will not appear. But then saving the JASP file, closing and reopening will often blank the boxes. 

This has happened to several people running different OS and working on different datasets / factor structures. 

### Expected Behaviour

Copying CFA analyses / opening and closing should not wipe the factor contents

### Steps to Reproduce

1. use the CFA module in the latest JASP build to create a multifactor CFA with items in each
2. copy this analysis - so that you get "copy of" the CFA model you just ran.  
3.  Save the JASP session, close and then re-open. 
...
Screen Shot 2022-02-28 at 4 14 45 pm

### Log (if any)

_No response_

### Final Checklist

boutinb:
Fixes https://github.com/jasp-stats/jasp-issues/issues/1640

This level of this problem is quite high: people can't really work with CFA, and the fix is minimal.

### DIFF
--- a/Desktop/components/JASP/Widgets/FactorsForm.qml
+++ b/Desktop/components/JASP/Widgets/FactorsForm.qml
@@ -77,6 +77,7 @@ FactorsFormBase
                     allowedColumns:     allowAll ? [] : ["scale", "ordinal"]
 					implicitHeight:		factorsForm.factorListHeight // preferredHeight does not work when changing the language: the height is set to the implicitHeight
 					implicitWidth:		listWidth
+					isBound:			false
 
 					onTitleIsChanged:	factorsForm.titleChanged(index, editableTitle)
 				}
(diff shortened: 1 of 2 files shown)


JorisGoosen (approved):

JorisGoosen:
@boutinb does this fix need to be in 0.16.2 btw?
Because then it would've been better to move this to `stable`. But I can still fix it now with a `cherry-pick`

boutinb:
I put it in stable already.

PR #4733, opened 2022-03-10
TextArea in TabView cannot be column encoded

boutinb:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/1745

The RCode Text Area should always generated an option with the 'isRCode' meta.
If a suboption has a meta (isRCode, shouldEncode, ...), this meta should be set also has a submeta.

### DIFF
--- a/Desktop/analysis/analysis.cpp
+++ b/Desktop/analysis/analysis.cpp
@@ -132,11 +132,11 @@ bool Analysis::checkAnalysisEntry()
 			Modules::ModuleException("???", "No coded reference stored or _dynamicModule == nullptr...");
 
 		_moduleData = _dynamicModule->retrieveCorrespondingAnalysisEntry(_codedAnalysisEntry);
-		
+
 		bool updateTitleToDefault = _title == _titleDefault;
-		
+
 		_titleDefault = _moduleData->title();
-		
+
 		if(updateTitleToDefault)
 			setTitle(_titleDefault);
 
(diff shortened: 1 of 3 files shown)


Desktop/analysis/analysis.cpp
```diff
@@ -431,59 +434,66 @@ Json::Value& Analysis::_getParentBoundValue(const QVector<JASPControl::ParentKey
...
+	return result;
 }
 
 bool Analysis::setBoundValue(const std::string &name, const Json::Value &value, const Json::Value &meta, const QVector<JASPControl::ParentKey>& parentKeys)
 {
 	bool found = false;
-	Json::Value &	parentBoundValue	= _getParentBoundValue(parentKeys, found, true),
-					copyPBV				= parentBoundValue; 
+	Json::Value* parentBoundValue	= &_boundValues;
+	QVector<std::string> parents	= _getParentBoundValue(parentKeys, parentBoundValue, found, true);
```
JorisGoosen: Isnt it actually more sensible to give the child-values to the parent and let the parent add them?

Instead of letting the child get a weird ref to a Json::Value in its parent and then setting stuff on there?

Desktop/analysis/analysis.cpp
```diff
@@ -431,59 +434,66 @@ Json::Value& Analysis::_getParentBoundValue(const QVector<JASPControl::ParentKey
...
-	if (found && parentBoundValue.isObject())
+	if (found && parentBoundValue->isObject())
 	{
-		parentBoundValue[name] = value;
+		(*parentBoundValue)[name] = value;
 
 		if ((meta.isObject() || meta.isArray()) && meta.size() > 0)
-			_boundValues[".meta"][name] = meta;
+		{
+			Json::Value* metaBoundValue = &_boundValues[".meta"];
```
JorisGoosen, line 490: This block of code for instance  looks way too complicated. While in fact you just want to pass the `meta` to the parent right?

JorisGoosen (changes requested):
Well, ive requested some changes because I feel like the flow of information is the wrong way round...

I think instead of getting references to Json::Value somewhere in the parent we should be sending stuff to the parent instead and letting it handle putting stuff in it's boundvalues.

Maybe you have really good reasons for this though, in that case we can talk about tomorrow? (also in case you just want to talk about it ;) )

Desktop/analysis/analysis.cpp
boutinb: The main issue here is that I need a reference where the meta has to be set, but this reference depends on the parents vector, and in C++, you cannot reassign a reference to another object. That's why I need a pointer.

JorisGoosen:
Do you want me to review it again?

JorisGoosen:
And this should go into 0.16.3 ? 

Because in that case we need to redirect it to `stable`

boutinb:
@JorisGoosen As https://github.com/jasp-stats/INTERNAL-jasp/issues/1568 is planned only for 0.17, I put this PR to development.
Can you review it again?

Desktop/analysis/analysis.cpp
```diff
@@ -386,38 +386,40 @@ Json::Value Analysis::asJSON(bool withRSource) const
...
+		if (createAnyway && !parentBoundValue->isMember(parent.name))	(*parentBoundValue)[parent.name] = Json::Value(Json::arrayValue);
 
-		if (result->isMember(parent.name))
+		if (parentBoundValue->isMember(parent.name))
 		{
-			Json::Value& parentBoundValues = (*result)[parent.name];
+			Json::Value& parentBoundValues = (*parentBoundValue)[parent.name];
 			if (!parentBoundValues.isNull() && parentBoundValues.isArray())
 			{
 				for (Json::Value & boundValue : parentBoundValues)
```
JorisGoosen, line 425: Should it not be `const Json::Value & boundValue`? 
Just to be sure you aren't changing other objects?

Desktop/analysis/analysis.cpp
```diff
@@ -461,45 +464,53 @@ Json::Value& Analysis::_getParentBoundValue(const QVector<JASPControl::ParentKey
...
 	bool found = false;
-	Json::Value &	parentBoundValue	= _getParentBoundValue(parentKeys, found, true),
-					copyPBV				= parentBoundValue; 
+	QVector<std::string> parents;
+	Json::Value& parentBoundValue = _getParentBoundValue(parentKeys, parents, found, true);
+	Json::Value copyPBV				= parentBoundValue;
 
 	if (found && parentBoundValue.isObject())
 	{
 		parentBoundValue[name] = value;
```
JorisGoosen, line 486: Aaah ok I guess not `const` then otherwise this wouldnt work

JorisGoosen (approved):

PR #4734, opened 2022-03-14
use jaspResults as an R package inside Jasp

vandenman:
Related to https://github.com/jasp-stats/INTERNAL-jasp/issues/1314 and the syntax project in general.

@JorisGoosen this turned out to be a lot easier than expected. Am I overseeing something?

This is only a minimal example so I tried to keep the changes as small as possible.


R-Interface/jasprcpp.cpp
```diff
@@ -71,6 +71,12 @@ static std::string				_R_HOME = "";
 
 bool shouldCrashSoon = false; //Simply here to allow a developer to force a crash
 
+// TODO: this probably won't work on windows
```
JorisGoosen, line 74: Haha no, now you've broken all utf-8 in windows ;)

But the next R has UCRT as runtime and thus utf-8 as native encoding so once we switch to that we can strip all of this nonsense anyway.

R-Interface/jasprcpp.cpp
```diff
@@ -114,7 +120,7 @@ void STDCALL jaspRCPP_init(const char* buildYear, const char* version, RBridgeCa
 	runCallbackCB								= callbacks->runCallbackCB;
 	readDataSetCB								= callbacks->readDataSetCB;
 
-
+	// TODO: none of this should pollute the global environment.
```
JorisGoosen, line 123: Would it be better to manually make a sub-environment?

R-Interface/jasprcpp.cpp
```diff
@@ -158,27 +164,19 @@ void STDCALL jaspRCPP_init(const char* buildYear, const char* version, RBridgeCa
...
-	rInside["jaspResultsModule"]			= givejaspResultsModule();
-
-	//Adding some functions in R to the RefClass (generator) in the module
-	//To do: move this entirely to zzzWrapper if this wasn't done yet.
-	jaspRCPP_parseEvalQNT("jaspResultsModule$jaspTable$methods(addColumnInfo = function(name=NULL, title=NULL, overtitle=NULL, type=NULL, format=NULL, combine=NULL) { addColumnInfoHelper(name, title, type, format, combine, overtitle) })");
-	jaspRCPP_parseEvalQNT("jaspResultsModule$jaspTable$methods(addFootnote =   function(message='', symbol=NULL, col_names=NULL, row_names=NULL) { addFootnoteHelper(message, symbol, col_names, row_names) })");
+	rInside[".sendToDesktopFunction"]	= Rcpp::XPtr<sendFuncDef>(&sendToDesktopFunction);
+	rInside[".pollMessagesFunction"]	= Rcpp::XPtr<pollMessagesFuncDef>(&pollMessagesFunction);
+	rInside[".baseCitation"]			= baseCitation;
+	jaspRCPP_parseEvalQNT("jaspResults:::setSendFunc(.sendToDesktopFunction)");
```
JorisGoosen, line 170: And this actually works?
As in, jaspResults is still sending stuff afterwards?

JorisGoosen:
Looks pretty minimal indeed, nice.

R-Interface/jasprcpp.cpp
vandenman: yes, or we assign it into (a subenvironment of) the jaspBase or jaspResults namespace.

R-Interface/jasprcpp.cpp
vandenman: I tested descriptives + a few other analyses and those worked.

R-Interface/jasprcpp.cpp
vandenman: yeah, I figured, but I didn't want to spend a lot of time on that part before discussing the rest of the approach.

PR #4736, opened 2022-03-18
Cannot open the Column Filter when filter icon is visible

### LINKED ISSUE #1661: Problem with removing filter
### JASP Version

0.16.1

### Commit ID

_No response_

### JASP Module

jaspDescriptives

Data editing mode

Windows 10

### Bug Description

Hi,
When I have a scale variable and that I rename to a nominal variable (id variable), and after try to apply a filter "by clicking off" one observation  I cant returen and deselect the filter. It seem to work with data that is already set as nominal but not when switching.
See pictures below.
Picture1
Picture2
Picture3

### Expected Behaviour

It should be possible to select and derelict filters from converted variables from scale to nominal.

### Steps to Reproduce

1. Have a scale variable
2. Convert it to nominal
3. Filter out one observation
4. Close
5. But after it is not possible to open and deselect.
6. 
...

### Log (if any)

_No response_

### Final Checklist

boutinb:
Fixes https://github.com/jasp-stats/jasp-issues/issues/1661

The problem was that the width of the LoadingIndicator of the column (in DataTableView) was not set to 0 when invisible.
The MouseArea that opens the Labels and Filters should also contain the Filter icon.
Also set the columnType as context variable: seems more logical.

### DIFF
--- a/Desktop/components/JASP/Widgets/DataTableView.qml
+++ b/Desktop/components/JASP/Widgets/DataTableView.qml
@@ -94,9 +94,7 @@ FocusScope
 					anchors.left:			parent.left
 					anchors.margins:		4
 
-					function myColumnType() {return dataSetModel.columnIcon(columnIndex)}
-
-					source: jaspTheme.iconPath + dataSetModel.getColumnTypesWithCorrespondingIcon()[myColumnType()]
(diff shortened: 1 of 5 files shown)


boutinb:
I think this is an easy fix that can be merged in 0.16.4, @JorisGoosen Can you review it?

JorisGoosen (approved):

JorisGoosen:
yeah good idea and fix

PR #4738, opened 2022-03-28
Dont fail configuration at msvc 2019 msm missing

JorisGoosen:
Im not even making an installer...

And I made it more readable, it wasn't as if you were hitting 80 wide anyway.
(Also Im not working on a teletype terminal but on a 4k screen)


amirmasoudabdol:
> I'm not even making an installer...

No, but CMake needs to check things upfront.

For the merge module, are we sure that the MSVC 1930 comes with `Microsoft_VC143_CRT_x64.msm` and MSVC 1920 comes with `Microsoft_VC142_CRT_x64.msm`. I didn't know the answer to this question, so I added a variable for it. If that's the case these changes makes sense; otherwise, the implementation before was assuming that you may want to set it, so passing `-DVC_MERGE_MODULE_NAME=Microsoft_VC142_CRT_x64.msm` would not have resulted in config error. I can default into those, and if we cannot find them then throw as an option.

Regarding the formatting, it _is_ limited to 80 characters, except strings that cannot be written into several lines, because CMake doesn't deal with that nicely. In addition, CMake files formatting comes from [`.cmake-format`](https://github.com/jasp-stats/jasp-desktop/blob/development/.cmake-format.yaml) so that we don't have to adjust it manually, it's consistent. I guess, we can discuss that at some other time.

JorisGoosen:
> > I'm not even making an installer...
> 
> No, but CMake needs to check things upfront.

Alright

> For the merge module, are we sure that the MSVC 1930 comes with `Microsoft_VC143_CRT_x64.msm` and MSVC 1920 comes with `Microsoft_VC142_CRT_x64.msm`. 

Yes we do 

> I didn't know the answer to this question, so I added a variable for it. If that's the case these changes makes sense; otherwise, the implementation before was assuming that you may want to set it, so passing `-DVC_MERGE_MODULE_NAME=Microsoft_VC142_CRT_x64.msm` would not have resulted in config error. I can default into those, and if we cannot find them then throw as an option.

> Regarding the formatting, it _is_ limited to 80 characters, except strings that cannot be written into several lines, because CMake doesn't deal with that nicely. In addition, CMake files formatting comes from [`.cmake-format`](https://github.com/jasp-stats/jasp-desktop/blob/development/.cmake-format.yaml) so that we don't have to adjust it manually, it's consistent. I guess, we can discuss that at some other time.

I think we don't actually enforce that 80 limit anywhere in JASP ever. R code is limited to 120 chars wide.

JorisGoosen:
@amir already fixed this differently

amirmasoudabdol:
I was just going to come here and close this! 😅

PR #4741, opened 2022-03-31
Custom Contrast is not updated when the data is filtered

boutinb:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/1795
2 problems: the Custom Contrast model was not aware of filter changes, and the labels were read not taking into account of the filter.

### DIFF
--- a/Desktop/data/datasetpackage.cpp
+++ b/Desktop/data/datasetpackage.cpp
@@ -1433,8 +1433,38 @@ QStringList DataSetPackage::getColumnLabelsAsStringList(size_t columnIndex)	cons
 	QStringList list;
 	if(columnIndex < 0 || columnIndex >= columnCount()) return list;
 
-	for(const Label & label : _dataSet->column(columnIndex).labels())
-		list.append(tq(label.text()));
+	Column& col = _dataSet->column(columnIndex);
+	QVector<bool> filterRows;
+	bool hasFilter = false;
+	for (int i = 0; i < _dataSet->rowCount(); i++)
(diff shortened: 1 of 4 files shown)


JorisGoosen:
Doesn't this change break the variableswindow?

JorisGoosen:
I think it would be better to let your ListModel take the labels via https://github.com/jasp-stats/jasp-desktop/blob/stable/Desktop/data/datasettablemodel.h because then the filter is already applied anyway right?

JorisGoosen:
Ooh sorry... I suggested work for nothing.

Do you have a testfile? 
Because when I read the issue I don't know how to get to a reproducible path :p

But I think in your latest code it only works if you don't show the filtered rows in the dataview. So I made a small fix:
https://github.com/JorisGoosen/jasp-desktop/commit/64fe7cd570cf77802d8888161260bfd4437503de

(I also changed the `notUsed` to a set, because I figured if you wanted the speedup from `int max=rowCount();` you would also appreciate the hashing powers vs list-searching ;) )

But I can't test whether my code is necessary and/or works

boutinb:
This code adds a new reference between a QML CustomContrast object and the DataSet, and as, for the R Syntax project, I'm busy to separate these kinds of reference in order to make a loose library containing QML objects that can be loaded in a R package, I prefer to base this PR to development, and to merge it only after the PR that implements the loose library is merged.

JorisGoosen (approved):
too bad I cant test it but it looks fine

PR #4743, opened 2022-04-04
Improvements to the README file

amirmasoudabdol:
I noticed that the README file is a bit outdated, and it maybe needs a bit of improvement. 

- I edited it a bit, and remove the HTML menu that we had, but I was not sure, so, I made the PR that we can decide. Those links are/were all available in the text, and I think they don't adhere to general format of GitHub readme files, and therefore visitors might simply overlook them, and continue scrolling.
- I also removed bunch of those `align` attributes, they are deprecated anyway! 
- I think the text needs some attention, some of the sentences don't read very well in my opinion, and it might be a good idea if someone can take a look at them at some point.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,39 +1,24 @@
-<h2 align="center">
-  <img src="https://static.jasp-stats.org/green_logo_dark_text.png" height=150/>
-</h2>
+![](https://static.jasp-stats.org/green_logo_dark_text_for_github.png)
 
-<h5 align="center">
-  <a target="_blank" href="https://jasp-stats.org">Home</a> |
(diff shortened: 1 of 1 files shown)


Kucharssim:
Since you are doing this, could you also change "Rebasing your repository" to "Git guide"? The new document gives much more info now than just rebasing :)

amirmasoudabdol:
Done! 🙂

Kucharssim:
I think this could be merged? @AlexanderLyNL

JorisGoosen (approved):

PR #4747, opened 2022-04-06
Resolved some of the compiler warnings

boutinb:
Fixes jasp-stats/jasp-test-release#1775


amirmasoudabdol:
You don't need to do this on CMake. If the setting is not set, it'll be set to Debug but Qt Creator always sets it when you change the profile. So, you just need to change your profile when building and you get what you have selected.

amirmasoudabdol (changes requested):
I think you should revert the CMake changes, and keep the rest.

boutinb:
OK, I've just kept the warning fixes.

amirmasoudabdol (approved):

PR #4751, opened 2022-04-08
Better CRYPT_KEY

JorisGoosen:
Previously if you set `CRYPT_KEY` in `cmake` options it would just be ignored because `ENVIRONMENT_CRYPTKEY` wasn't set...
And the cryptkey in the environment was actually called `CRYPTKEY`so let's use that.

With the new logic it will use whatever you pass on to cmake first, then from `CRYPTKEY` in environment and lastly the default value as a fallback.
This way it also cannot fail which is nice and friendly.

### DIFF
--- a/Tools/CMake/JASP.cmake
+++ b/Tools/CMake/JASP.cmake
@@ -94,23 +94,25 @@ set(CRYPT_KEY
     CACHE STRING "")
 
 if(CRYPT_KEY STREQUAL "")
-	# lets see if the user set something in the environment
+  # Let's see if the user has set something in the environment
   set(CRYPT_KEY $ENV{CRYPTKEY})
 
   if(CRYPT_KEY STREQUAL "")
-	  message(CHECK_PASS "found default.")
-	  
-    message(WARNING "CRYPT_KEY is not found, using default replace if this is to go out to users!")
-	set(CRYPT_KEY "0x0c2ad4a4acb9f023")
-	
+
+    message(CHECK_PASS "set to default.")
+
+    set(CRYPT_KEY "0x0c2ad4a4acb9f023")
+
   else()
+
     message(CHECK_PASS "found in environment.")
-    message(STATUS "  ${CRYPT_KEY}")
-	
+
   endif()
 
+  message(STATUS "  ${CRYPT_KEY}")
+
 else()
-	message(CHECK_PASS "set by user in cmake config.")
+  message(CHECK_PASS "set by user in cmake config.")
 endif()
 
 # Dealing with Git submodules
--- a/Tools/CMake/JASP.cmake
+++ b/Tools/CMake/JASP.cmake
@@ -92,19 +92,25 @@ message(CHECK_START "Checking for CRYPT_KEY")
 set(CRYPT_KEY
     ""
     CACHE STRING "")
+
 if(CRYPT_KEY STREQUAL "")
-  set(CRYPT_KEY $ENV{ENVIRONMENT_CRYPTKEY})
+	# lets see if the user set something in the environment
+  set(CRYPT_KEY $ENV{CRYPTKEY})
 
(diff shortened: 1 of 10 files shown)


JorisGoosen:
But maybe we could even remove the whole `simplecryptkey.h` ?

Tools/CMake/JASP.cmake
```diff
@@ -92,19 +92,25 @@ message(CHECK_START "Checking for CRYPT_KEY")
...
-  set(CRYPT_KEY $ENV{ENVIRONMENT_CRYPTKEY})
+	# lets see if the user set something in the environment
+  set(CRYPT_KEY $ENV{CRYPTKEY})
 
   if(CRYPT_KEY STREQUAL "")
-    message(CHECK_FAIL "not found.")
-    message(WARNING "CRYPT_KEY is not found.")
+	  message(CHECK_PASS "found default.")
+	  
+    message(WARNING "CRYPT_KEY is not found, using default replace if this is to go out to users!")
```
amirmasoudabdol: I'd remove this warning. This is a very specific variable, and if someone wants to set it, they know what they are doing and this warning is only going to be distraction for most people.

Tools/CMake/JASP.cmake
```diff
@@ -92,19 +92,25 @@ message(CHECK_START "Checking for CRYPT_KEY")
...
+	set(CRYPT_KEY "0x0c2ad4a4acb9f023")
+	
   else()
-    message(CHECK_PASS "found.")
+    message(CHECK_PASS "found in environment.")
     message(STATUS "  ${CRYPT_KEY}")
```
amirmasoudabdol: I'd put this after the `if` that it prints the key either way.

amirmasoudabdol (changes requested):
This certainly looks nicer and simpler! 👍🏼 

I thought I removed the `simplecryptkey.h` but it could be that git didn't want to let it go, and didn't care about it after I put it into the `.gitignore`, but yes, we can remove that and only have the template.

Tools/CMake/JASP.cmake
JorisGoosen: Fair enough

Tools/CMake/JASP.cmake
JorisGoosen: Well, I thought it might be useful to get feedback on where it came from?
You know, in case something is set up wrong?

On the other hand, it should also be obvious from the key itself

JorisGoosen:
> I thought I removed the `simplecryptkey.h` but it could be that git didn't want to let it go, and didn't care about it after I put it into the `.gitignore`, but yes, we can remove that and only have the template.

I meant removing both the template and the resulting header. Because if is now just a simple define right?
We could just add that from `cmake` directly?

amirmasoudabdol:
As far as I know, the safe way of passing a defined value to CMake is by using the template. It's annoying, but that's why I did it like that most of the time. So, I'd keep the template for now, and look around again to see if CMake can actually play nice with passing the defined value.

JorisGoosen:
Sure, there is no hurry in doing this. I just noticed the template was pretty empty right now ;)

But it definitely does work so that is something

amirmasoudabdol:
What about putting this into the `appinfo.cpp.in`? That's already a template and it contains a lot of info, and this is an info as well! 🤔

amirmasoudabdol:
I made the changes that I mentioned, it looks like [this](https://github.com/jasp-stats/jasp-desktop/pull/4751/commits/1f05a2e2085db42d723a8b9809ac12132fca1f82) now. I also fixed a few headers that were imported like `../Common/`. It's not been used consistency and CMake just includes the `Common`, so it's not necessary.

I pass the review to you now! 😬

JorisGoosen:
Works fine ^^

PR #4752, opened 2022-04-08
Centralised Qt set-up, git PAT.

AlexanderLyNL:
Missing:
- GITHUB_PAT rights
- Swap order regarding first PAT,
- Qt Kit configuration

### DIFF
--- a/Docs/development/jasp-build-guide-macos.md
+++ b/Docs/development/jasp-build-guide-macos.md
@@ -5,6 +5,7 @@ If you have not cloned the `jasp-desktop` repository, please head back to the [b
 ## Requirements
 
 - [Xcode](https://apps.apple.com/us/app/xcode/id497799835?mt=12)
+	- Tick macOS + Qt WebEngine
 - [Qt Creator](https://www.qt.io/download) / Qt >= 6.2
     - Qt Creator 7
 - [Homebrew](http://brew.sh)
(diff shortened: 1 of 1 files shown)


Docs/development/jasp-build-guide-macos.md
```diff
@@ -30,28 +31,44 @@ You can download and install Homebrew using the following command. After running
 /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
 ```
 
+Note the instructions in mentioned under "Next Steps" after installing homebrew, where `userName` below should be changed to your username
```
amirmasoudabdol: I have never done this part, and never had any issue with installing Brew. I would not put it in the instruction.

Docs/development/jasp-build-guide-macos.md
```diff
@@ -30,28 +31,44 @@ You can download and install Homebrew using the following command. After running
...
+Install CMake as follows:
+
+```bash
+brew install cmake
+```
+
 > 💡 Optionally, you can install the Qt from the command line, and build JASP using that, e.g., `brew install qt`. 
 
 ### Qt Creator
+<img width="1136" alt="qtInstallComponents" src="https://user-images.githubusercontent.com/10271675/162528342-d7db3c71-9747-4e10-b193-efedb7b52191.png">
```
amirmasoudabdol: This is probably more clear, but I have two issues with it. First, it is not complete, there are more items that needs to go in; second, I don't want to put a figure here because it makes it harder to maintain. I think a developer needs to be able to find an item in a list, and we should just list things that are necessary.

Docs/development/jasp-build-guide-macos.md
AlexanderLyNL: I think it's good to add. People tend not to read stuff, for instance, me. Fortunately, I this time read my terminal output carefully.

Docs/development/jasp-build-guide-macos.md
AlexanderLyNL: We can remove the image. It's still unclear for me which packages should be installed. I only mentioned macOS and webengine, because that's what was needed. So what are the other packages we need to install?

amirmasoudabdol:
Thanks for looking into this! It definitely needs some more improvements and refinements. As you go through them feel free to ping me that we can get it to a good state.

Docs/development/jasp-build-guide-macos.md
```diff
@@ -33,25 +32,43 @@ You can download and install Homebrew using the following command. After running
...
 	- After creating a new PAT, you can set that `GITHUB_PAT` variable in two ways,
 		- If you are using the Qt Creator, you need to either set the `GITHUB_PAT` directly to your CMake variables, under the "Projects->Build->Current Configuration". Search the list for `GITHUB_PAT`, and paste your PAT value there. This will instruct the CMake to use your PAT whenever necessary. 
 	- If you are using the command line, CMake looks for this variable in your environment variable and is able to automatically find and use it, if your `GITHUB_PAT` can be found in your environment variables.
-		- Adding this like to your `.bash_profile`, or `.zshrc` will fasciliate this process, `export GITHUB_PAT=your github pat`
-- `INSTALL_R_MODULES`
-	- If you wish to install JASP modules, you need to toggle this option `ON`. In Qt Creator->Projects->Builds->Current Configuration, you can find this variable, and check the checkbox. Or if you are using a command line, you can include this in your command line by `-DINSTALL_R_MODULES=ON`.
```
AlexanderLyNL, line 54: @amirmasoudabdol did you remove this? Or did something go wrong with me pulling and pushing?

amirmasoudabdol: Oh, sorry my bad, I only wanted to remove the part about `INSTALL_R_MODULES`. I can fix it...

amirmasoudabdol: It's ok apparently. The part about `INSTALL_R_MODULES` isn't necessarily because it's on by default. That's it.

amirmasoudabdol (approved):
It looks good to me, if there is nothing else to add, feel free to merge it.

AlexanderLyNL:
Yeah also to me, but I'm still missing the Qt kit set-up.

Issue #1234, opened 2016-03-11
OSF Login screen still looking suboptimal


EJWagenmakers:
JASP developers' version: Maybe it's because I have a touch-screen on my laptop? They sometimes mess up the format. See screenshot.
![image](https://cloud.githubusercontent.com/assets/8603015/13702401/4a9a1e52-e786-11e5-861d-35dca610532e.png)

boutinb:
@EJWagenmakers: Is this solved?

EJWagenmakers:
Nope, see attach. Could just be my laptop; some other text in the input panel is also "cut off".
E.J.
![osflogin](https://cloud.githubusercontent.com/assets/8603015/23948245/7bb18d94-0982-11e7-8315-3a266a71fb40.png)

FransMeerhoff:
Oldie,
Solved with #2308

PR #4753, opened 2022-04-11
Fixes this, https://github.com/jasp-stats/jasp-test-release/issues/1767

amirmasoudabdol:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/1767

P.S. I'm not 100% sure why you needed those, but let me know if I'm missing something


JorisGoosen:
@amirmasoudabdol next time, could you also describe in a few words what the issue is about.
Because when looking through the log it is never all that helpful to just see random `#1234` numbers...

JorisGoosen (approved):

JorisGoosen:
the `xstr` hack was to get a defined string as an actual string, but that isnt necessary with the templating stuff.

PR #4755, opened 2022-04-12
Fixes to Modules.cmake

JorisGoosen:
we do not need jaspTools at all in JASP.

Also, not setting `.libPaths(...)` means it will check for the normal R-library on your system as well..
So here locally for me building JASP I didnt get `jaspBase` because the check considered it fulfilled, jaspBase was already installed.
But then when running JASP (at least in the Install folder or in the resulting actual install of it) jaspBase was missing and no results showed up...
On a clean system this isnt a problem but I rarely work on a clean system and for people using jaspTools this will be the same.
So this

### DIFF
--- a/Tools/CMake/Modules.cmake
+++ b/Tools/CMake/Modules.cmake
@@ -145,32 +145,50 @@ execute_process(
   COMMAND ${CMAKE_COMMAND} -E copy_if_different R/symlinkTools.R
           ${MODULES_BINARY_PATH}/)
 
-add_custom_target(jaspBase WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/R-Interface)
+add_custom_target(
+  jaspBase
+  WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/R-Interface
+  DEPENDS ${MODULES_BINARY_PATH}/jaspBase-installed-successfully.log
+          ${MODULES_BINARY_PATH}/jaspGraphs-installed-successfully.log)
 
 # This happens during the configuration!
 file(
   WRITE ${MODULES_RENV_ROOT_PATH}/install-jaspBase.R
   "
-    .libPaths('${R_LIBRARY_PATH}') # make sure to only look in our local library
-
-    install.packages(c('ggplot2', 'gridExtra', 'gridGraphics',
-                      'jsonlite', 'modules', 'officer', 'pkgbuild',
-                      'plyr', 'qgraph', 'ragg', 'R6', 'renv',
-                      'rjson', 'rvg', 'svglite', 'systemfonts',
-                      'withr', 'testthat',
-                      'data.table', 'httr', 'lifecycle',
-                      'pkgload', 'remotes', 'stringi', 'stringr',
-                      'vdiffr'), type='${R_PKG_TYPE}', repos='${R_REPOSITORY}' ${USE_LOCAL_R_LIBS_PATH})
-    install.packages('${PROJECT_SOURCE_DIR}/Engine/jaspBase/', type='source', repos=NULL, INSTALL_opts='--no-multiarch --no-docs --no-test-load' ${USE_LOCAL_R_LIBS_PATH})
-
+    # same hash as jaspBase::installModule uses
+    computeHash <- function(modulePkg) {
+      srcFiles <- c(
+        list.files(modulePkg,                         recursive=TRUE, full.names = TRUE, pattern = '(NAMESPACE|DESCRIPTION)$'),
+        #list.files(file.path(modulePkg, 'src'),       recursive=TRUE, full.names = TRUE, pattern = '(\\\\.(cpp|c|hpp|h)|(Makevars|Makevars\\\\.win))$'),
+        list.files(file.path(modulePkg, 'R'),         recursive=TRUE, full.names = TRUE, pattern = '\\\\.R$'),
+        #list.files(file.path(modulePkg, 'inst'),      recursive=TRUE, full.names = TRUE, pattern = '\\\\.(qml|po|svg|png|jpg|md)$'),
+        #list.files(file.path(modulePkg, 'inst'),      recursive=TRUE, full.names = TRUE, pattern = '\\\\qmldir$'),
+        list.files(modulePkg,                         recursive=TRUE, full.names = TRUE, pattern = 'renv\\\\.lock')
+      )
+      tools::md5sum(srcFiles)
+    }
+    hashPath <- file.path('${R_LIBRARY_PATH}', 'jaspBase', 'jaspBaseHash.txt')
+    currentHash <- computeHash('${PROJECT_SOURCE_DIR}/Engine/jaspBase/')
+    if (!(file.exists(hashPath) && 'jaspBase' %in% installed.packages() && identical(currentHash, try(readRDS(hashPath))))) {
+      options(install_opts = '--no-multiarch --no-docs --no-test-load')
+      options('renv.cache.linkable' = TRUE)
+      renv::install('${PROJECT_SOURCE_DIR}/Engine/jaspBase/', repos=NULL, library='${R_LIBRARY_PATH}')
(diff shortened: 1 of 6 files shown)


Tools/CMake/Modules.cmake
```diff
@@ -150,12 +150,14 @@ add_custom_target(
   WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/R-Interface
   DEPENDS ${MODULES_BINARY_PATH}/jaspBase-installed-successfully.log
           ${MODULES_BINARY_PATH}/jaspGraphs-installed-successfully.log
-          ${MODULES_BINARY_PATH}/jaspTools-installed-successfully.log)
```
amirmasoudabdol, line 153: I was just working on this too actually... I pick it up from this PR...

Tools/CMake/Modules.cmake
```diff
@@ -229,22 +234,22 @@ add_custom_command(
     -P ${PROJECT_SOURCE_DIR}/Tools/CMake/Patch.cmake
   COMMENT "------ Installing 'jaspGraphs'")
 
-add_custom_command(
```
amirmasoudabdol, line 232: You can comment the entirety of this, but I see what I can do, especially because it deals with signing, etc.

Tools/CMake/Modules.cmake
```diff
@@ -150,12 +150,14 @@ add_custom_target(
...
   DEPENDS ${MODULES_BINARY_PATH}/jaspBase-installed-successfully.log
           ${MODULES_BINARY_PATH}/jaspGraphs-installed-successfully.log
-          ${MODULES_BINARY_PATH}/jaspTools-installed-successfully.log)
+          )
 
 # This happens during the configuration!
 file(
   WRITE ${MODULES_RENV_ROOT_PATH}/install-jaspBase.R
   "
+    .libPaths('${${R_LIBRARY_PATH}}') # make sure to only look in our local library
```
amirmasoudabdol: Good point, but `${${R_LIBRARY_PATH}}` should be `${R_LIBRARY_PATH}` unless I'm missing something.

Tools/CMake/Modules.cmake
amirmasoudabdol: Nevermind, this is for the `jaspTools`.

amirmasoudabdol:
Is the `.libPaths()` necessary when I already pass the `${USE_LOCAL_R_LIBS_PATH}`, which is set to `set(USE_LOCAL_R_LIBS_PATH ", lib='${R_LIBRARY_PATH}'")`?

```
install.packages(c('ggplot2', 'gridExtra', 'gridGraphics',
                        'jsonlite', 'modules', 'officer', 'pkgbuild',
                        'plyr', 'qgraph', 'ragg', 'R6', 'renv',
                        'rjson', 'rvg', 'svglite', 'systemfonts',
                        'withr', 'testthat',
                        'data.table', 'httr', 'lifecycle',
                        'pkgload', 'remotes', 'stringi', 'stringr',
                        'vdiffr'), type='${R_PKG_TYPE}', repos='${R_REPOSITORY}' ${USE_LOCAL_R_LIBS_PATH})

install.packages('${PROJECT_SOURCE_DIR}/Engine/jaspBase/', type='source', repos=NULL ${USE_LOCAL_R_LIBS_PATH}, INSTALL_opts='--no-multiarch --no-docs --no-test-load')
```

amirmasoudabdol:
I know what you are trying to fix here, but I don't think this fixes it. I think the only way to actually fix this is to remove the checks that I have, and let those to be installed, or let R handles their installation, and either skip installing them if they are there, or installs them every time. This probably improves the incremental builds so that those libraries will be updated, and if submodules get updates, that will be reflected. The problem will be that we will be at R's mercy and intelligence.

Tools/CMake/Modules.cmake
JorisGoosen: Oops

JorisGoosen:
> I know what you are trying to fix here, but I don't think this fixes it. I think the only way to actually fix this is to remove the checks that I have, and let those to be installed, or let R handles their installation, and either skip installing them if they are there, or installs them every time. This probably improves the incremental builds so that those libraries will be updated, and if submodules get updates, that will be reflected. The problem will be that we will be at R's mercy and intelligence.

well, it fixed my problem of having a build of JASP without `jaspBase` in `R/library` so.. you know, there is that.
I dont think you would have this problem on macos, but I certainly did have it here on windows.

And R's mercy and intelligence with regards to dependencies is questionable.

I think it would be better to just always install `jaspBase` to `R/library` and instead of hardcoding the dependencies in cmake use something like `devtools::install_dependencies` or whatever it is called.

But these fixes are all just to make sure that whatever it is doing now it does a little bit better.

JorisGoosen:
> Is the `.libPaths()` necessary when I already pass the `${USE_LOCAL_R_LIBS_PATH}`, which is set to `set(USE_LOCAL_R_LIBS_PATH ", lib='${R_LIBRARY_PATH}'")`?
> 
> ```
> install.packages(c('ggplot2', 'gridExtra', 'gridGraphics',
>                         'jsonlite', 'modules', 'officer', 'pkgbuild',
>                         'plyr', 'qgraph', 'ragg', 'R6', 'renv',
>                         'rjson', 'rvg', 'svglite', 'systemfonts',
>                         'withr', 'testthat',
>                         'data.table', 'httr', 'lifecycle',
>                         'pkgload', 'remotes', 'stringi', 'stringr',
>                         'vdiffr'), type='${R_PKG_TYPE}', repos='${R_REPOSITORY}' ${USE_LOCAL_R_LIBS_PATH})
> 
> install.packages('${PROJECT_SOURCE_DIR}/Engine/jaspBase/', type='source', repos=NULL ${USE_LOCAL_R_LIBS_PATH}, INSTALL_opts='--no-multiarch --no-docs --no-test-load')
> ```

yes of course, because you have the `if ('jaspBase' %in% installed.packages())` , where is it looking you think?

Tools/CMake/Modules.cmake
JorisGoosen: Should I fix it here or are you doing it in your followup branch?

amirmasoudabdol:
Ok, that's a good starting point then for this issue. Let me work on it a bit, and see what we can do. For now, I'm going to remove the dependencies and let them to be installed every time, and pass them to the R regardless of their existence, and then we can see how we can improve it even further.

Tools/CMake/Modules.cmake
JorisGoosen: Weird, but I guess you fixed it and github showed me a bunch of outdated commits

amirmasoudabdol: Yes, I fixed a few things already!

amirmasoudabdol:
I think these changes will guarantee that we don't end up with an outdated `jaspBase` and `jaspGraphs` but I don't want to push them to the stable at the moment. I am going to test them a bit more, and reserve this PR for the development.

amirmasoudabdol:
@vandenman, I bring the discussion here as we already have an option issue on it.

> Ideally, these are installed through renv though. I can set that up if you want.

At the moment, they are not, but if you think that helps, then sure. I'm in favor of leaving everything to Renv. Now, some of the requirements, e.g., `tidyr`, `systemfonts`, etc. are being installed in the `R/library`, as well as `jaspBase` and `jaspGraphs` and the rest are in the Renv through the `jaspBase::install...`

[Here](https://github.com/jasp-stats/jasp-desktop/blob/f38f5263c8060b0bbe1c7f07983bba22b7db46d0/Tools/CMake/Modules.cmake#L156), line 156—247, you can see how the install is handled, and how they are being installed in the main `R_HOME/library`. Feel free to continue working on this PR, and make changes if it affects the CMake, and let me know if you have any questions.

amirmasoudabdol (approved):
Thanks @vandenman! This looks very good! 👍🏼

PR #4757, opened 2022-04-13
The source of a Dropdown should be fetched before its value is set

boutinb:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/1780
This solves 2 problems:
. a dropdown with a source is not set properly when a JASP file is read: if the dropdown was set with a value coming from the source, when reading the JASP file, if the values of the source are not yet read, setting the dropdown generates an error. This is the case in the ANOVA analysis: the 'Reference model for weight ratios' dropdown in the Order Restriction, has for source the titles of the tabs of the Order Restriction models. If the dropdown is set with one of these values, reading a JASP file with this setting will lead to an error.
. if an error is set to a control, removing this error lets the analysis is an error state, and the analysis cannot run anymore.
In AnalysisForm, the terms of an available model (which is the case for a dropdown) should be fetched before the value of the dropdown is set.

Also remove the _getControlErrorMessageOfControl method which is not used anymore.
In ControlErrorMessage, when the message is closed, its association with its control should be removed: in this way the AnalysisForm::hasError becomes false.

### DIFF
--- a/Desktop/analysis/analysisform.cpp
+++ b/Desktop/analysis/analysisform.cpp
@@ -362,6 +362,24 @@ void AnalysisForm::bindTo()
 		BoundControl* boundControl = control->boundControl();
 		JASPListControl* listControl = dynamic_cast<JASPListControl *>(control);
 
+		if (listControl && listControl->hasSource())
+		{
+			ListModelAvailableInterface* availableModel = qobject_cast<ListModelAvailableInterface*>(listControl->model());
+			// The availableList control are not bound with options, but they have to be updated from their sources when the form is initialized.
+			// The availableList cannot signal their assigned models now because they are not yet bound (the controls are ordered by dependency)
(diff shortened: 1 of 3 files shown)


JorisGoosen:
@boutinb I am "on holiday" since today, so I won't have time to review this until 28-04 or so. Maybe not the best choice to pick me, unless you want it to wait a while :p

Also, (not that I always do this) but as I mentioned to Amir quite recently: It would be superhelpful if next to links to an issue you also put a very short description of whatever the issue was. Because otherwise the log in git, you know without this glorious webinterface around it that is called github, it is rather hard to guess what is going on. Of course here you mention the solution but not the problem...

boutinb:
I have added some comments in the commit self, so that in  Git only, you can see what was the problem exactly.
I'm quite sure of my fix, and it is quite important I find. so let's push it.

PR #4765, opened 2022-05-03
uncomment jaspProcessControl

vandenman:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/1827

It's already there in .gitmodules. Also, do we want `(beta)` or not?

### DIFF
--- a/Tools/CMake/Modules.cmake
+++ b/Tools/CMake/Modules.cmake
@@ -39,7 +39,7 @@ set(JASP_EXTRA_MODULES
     "jaspMachineLearning"
     "jaspMetaAnalysis"
     "jaspNetwork"
-    # "jaspProcessControl"
+    "jaspProcessControl"
     "jaspReliability"
     "jaspSem"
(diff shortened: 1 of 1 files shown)


AlexanderLyNL:
Yes, please add the (Beta) tag.

JorisGoosen (approved):

JorisGoosen:
Uhm, I just merged this because the changes looked fine, but it is now in `development` and not in `stable`.
I guess that is where it was supposed to go seeing as how we want it in 0.16.3

EJWagenmakers:
Does this refer to our Quality Control Tools? If so (and I am not sure!), this is not beta, and we should not create that impression  -- SKF is using it instead of Minitab.

EJWagenmakers:
OK, I checked, and it does refer to the Quality Control Tools. These have been developed for well over a year, and have had several SKF feedback rounds. If we call thos beta, we can call everything beta. So we should remove that tag.

PR #4767, opened 2022-05-03
Cannot set custom data editor

### LINKED ISSUE #1696: [Bug]: Select custom editor no longer works in JASP 0.16.2
### JASP Version

0.16.2

### Commit ID

_No response_

### JASP Module

_No response_

_No response_

MacOS Monterey 12.3.1

### Bug Description

With JASP 0.16.2 there is an error when Data Preferences are set to "Select custom editor." When I select an editor application (i.e., Excel, Numbers, SPSS - it doesn't matter which one), JASP produces the following error when I click a cell in the JASP data spreadsheet:

"Unable to start the editor : Microsoft Excel.app. Please check your editor settings in the preference menu."

Screen Shot 2022-04-28 at 5 38 07 PM

If I change Data Preferences to "Use default spreadsheet editor," then clicking a cell in the JASP data spreadsheet correctly opens Excel (i.e. my default editor).  However, this means I can't edit the data in anything but the default editor.

### Expected Behaviour

I reinstalled JASP 0.16.1 alongside the current version.  When I use JASP 0.16.1, "Select custom editor" works perfectly. When I click a cell in the data spreadsheet, whatever I have set as the custom editor opens  The problem only happens with JASP 0.16.2

### Steps to Reproduce

1. I tried four different JASP datasets.  The error happened with all four .jasp files in 0.16.2
2. The error did not happen on these four JASP datasets when using 0.16.1
3. 
...

### Log (if any)

_No response_

### Final Checklist

boutinb:
Fixes https://github.com/jasp-stats/jasp-issues/issues/1696
If in the Data Preferences, you set the spreadsheet editor not to be the default one, but a custom one, then this editor does not start when you double-click the data in the data panel.

### DIFF
--- a/Desktop/mainwindow.cpp
+++ b/Desktop/mainwindow.cpp
@@ -1473,16 +1473,16 @@ void MainWindow::startDataEditor(QString path)
 	if (appname.isEmpty())
 		useDefaultSpreadsheetEditor = true;
 
-	QString startProcess;
 	if (!useDefaultSpreadsheetEditor)
 	{
+		QStringList args;
 #ifdef __APPLE__
-		appname = appname.mid(appname.lastIndexOf('/') + 1);
-		startProcess = "open -a \"" + appname + "\" \"" + path + "\"";
+		args = {"-a", appname, path};
+		appname = "open";
 #else
-		startProcess = "\"" + appname + "\" \"" + path + "\"";
+		args = {path};
 #endif
-		if (!QProcess::startDetached(startProcess))
+		if (!QProcess::startDetached(appname, args))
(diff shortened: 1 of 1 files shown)


JorisGoosen (approved):
The changes look fine, but I wont have access to a macos system until I am back in the netherlands. So I cannot test whether it actually fixes the bug

boutinb:
> The changes look fine, but I wont have access to a macos system until I am back in the netherlands. So I cannot test whether it actually fixes the bug

I did test. With success.

JorisGoosen:
I figured ;)

Something to check as well is if it now still works on windows (unless you checked that as well) and then it can be merged.
But today Ive been fighting with cmake+linux so didnt get that chance

PR #4769, opened 2022-05-04
These fixes stop JASP from crashing on linux

JorisGoosen:
And the behaviour should still be fine

The fix in `Description` is only useful when doubleclicking a jaspModule.tar.gz with the `enginePerModule` branch (Which I will rebase on this).

The fix in the mainwindow destructor is to not make a copy of the list of root-objects because apparently some of them get destroyed through the other ones. So be requesting a new list all the time we avoid a crash.

But maybe Im the only one with this crash?
On Arch linux with Qt 6.3

### DIFF
--- a/Desktop/mainwindow.cpp
+++ b/Desktop/mainwindow.cpp
@@ -205,11 +205,10 @@ MainWindow::~MainWindow()
 	try
 	{
 		//Clean up all QML to get rid of warnings and hopefully fix https://github.com/jasp-stats/jasp-issues/issues/667
-		QList<QObject *> rootObjs = _qml->rootObjects();
-
 		//Going backwards to make sure the theme isnt deleted before everything that depends on it
-		for(int i=rootObjs.size() - 1; i >= 0; i--)
-			delete rootObjs[i];
+		for(int i=_qml->rootObjects().size() - 1; i >= 0; i--)
+			delete _qml->rootObjects().at(i);
+
 
 		delete _qml;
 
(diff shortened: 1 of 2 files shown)


vandenman:
Two remarks.

1. We're still on Qt 6.2.3 right? See also https://github.com/jasp-stats/INTERNAL-jasp/issues/1822. On 6.2.3 I do not see any "crashes in qml-destructor-asserts".
2. Can you post a reproducible path or screenshot of the bug that is supposed to fix? I don't experience any crashes (or errors in the logs) with dynamic modules or elsewhere. If I close JASP I do see this, but that's been there for a long time:
image

JorisGoosen:
> Two remarks.
> 
>     1. We're still on Qt 6.2.3 right? See also [Qt 6.3 INTERNAL-jasp#1822](https://github.com/jasp-stats/INTERNAL-jasp/issues/1822). On 6.2.3 I do not see any "crashes in qml-destructor-asserts".

Thanks for that link, I am on 6.3 and then this should fix that problem. Merci!

>     2. Can you post a reproducible path or screenshot of the bug that is supposed to fix? I don't experience any crashes (or errors in the logs) with dynamic modules or elsewhere. If I close JASP I do see this, but that's been there for a long time:
>        image

Well, the repro-path was closing JASP, for the qml-destructor stuff at well closing.
The other path is simply using "Install Module" and selecting a `jaspModule.tar.gz`.
But I guess that problem would also be only in 6.3

boutinb (approved):

PR #4772, opened 2022-05-11
Add palmer penguins data library file for Machine Learning

koenderks:
Closes https://github.com/jasp-stats/jaspMachineLearning/issues/128


Kucharssim (approved):

Kucharssim:
This should be safe to merge, the only thing is whether we want to include it in the release or not?

koenderks:
I think it would be nice, but maybe @AlexanderLyNL has an opinion about this as well?

PR #4773, opened 2022-05-13
add Turbines example (GLMs) to data library

Kucharssim:


vandenman (approved):
Looks good to me! The only remark I have is that the data are very small (274 B), but the jasp file is quite big (1.1 MiB). That's for another time I guess.

PR #4776, opened 2022-05-17
The code should really always have something set, but if not mention it to the user!

JorisGoosen:
@Kucharssim can you check if this actually shows an error if you do not manually set CMAKE_OSX_ARCHITECTURES?


Kucharssim:
For some reason, this leads to 

```
-- Making sure CMAKE_OSX_ARCHITECTURES is set to something. - Cross compiling for  architecture
```

(and proceeds as if nothing is missing)

JorisGoosen:
Can you try with the new change and then we can see what is going on perhaps?

Kucharssim:
Ah, now it does this:

```
-- Making sure CMAKE_OSX_ARCHITECTURES (now:  while host: ) is set to something.
-- Making sure CMAKE_OSX_ARCHITECTURES (now:  while host: ) is set to something. - Building for  architecture
CMake Warning (dev) at CMakeLists.txt:41 (message):
  Ignored CHECK_FAIL without CHECK_START
This warning is for project developers.  Use -Wno-dev to suppress it.
```

JorisGoosen:
Aah yeah if that is also empty... Well ive changed the `else` and I think this should be better

Kucharssim:
The output is still the same...

JorisGoosen:
Hmm, @amirmasoudabdol what is going on here? You have a clue?

JorisGoosen:
Anyway, we have a workaround for now, which would be just specifying the architecture manually, but it would be nice if this just worked (like it did before I guess?)

PR #4780, opened 2022-06-03
Closing a dataset may crash if it contains an analysis having its own computed columns

boutinb:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/1826

The destruction of a computed column emits an event to change the dataset, but this should not be send if the dataset does not exist anymore.
Moreover, the computed columns of an analysis are destroyed twice, because their destruction did not remove their reference in the analysis.


JorisGoosen:
So ive made some changes that make the fix a bit more selfcontained and clean, if you agree we can merge it ^^

JorisGoosen (approved):
Wij van wceend raden wceend aan ;)

PR #4784, opened 2022-06-08
The dropdown value of old JASP files have sometimes not exactly the right value

boutinb:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/1836

Due to old bugs, the value set in a dropdown might be set with the label and not with its value. This could be solved by adding entries in the upgrade.qml, but this can be quite cumbersome, since the problem can have occurred in many older JASP versions.
We can also argue, that if a value in a JASP file is not found in the possible value list of a DropDown, then before throwing an error message, we can try to find the value in the labels of the DropDown.

### DIFF
--- a/Desktop/widgets/comboboxbase.cpp
+++ b/Desktop/widgets/comboboxbase.cpp
@@ -42,9 +42,12 @@ void ComboBoxBase::bindTo(const Json::Value& value)
 		else
 		{
 			auto itr = std::find(values.begin(), values.end(), selectedValue);
+
 			if (itr == values.end())
 			{
-				// Let's be a bit flexible: if we can find a value which is case-insensitive equal to the selectedValue,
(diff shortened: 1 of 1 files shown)


Resources/Data Sets/Data Library/3. ANOVA/Alcohol Attitudes copy/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,4 @@
+Manifest-Version: 1.0
```
JorisGoosen: Why is this file here? Accidentally committed?

JorisGoosen (changes requested):
Remove the Alcohol Attittudes copy...

boutinb:
> Remove the Alcohol Attittudes copy...

Oeps... weggehaald

JorisGoosen (approved):

PR #4789, opened 2022-06-14
Row controls depending on a source should be disconnected when not used

boutinb:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/1786

The MixedModels model uses a  ComponentsList with VariablesList having a source from another VariablesList. As the ComponentsList caches its controls, it must care that if a control has a source, changes in this source should not affect this control: the signals should be disconnected.

### DIFF
--- a/Desktop/widgets/sourceitem.cpp
+++ b/Desktop/widgets/sourceitem.cpp
@@ -100,15 +100,15 @@ void SourceItem::connectModels()
 	if (_isRSource && form)
 		connect(form,	&AnalysisForm::rSourceChanged,				this, &SourceItem::_rSourceChanged);
 
-	if (!_nativeModel) return;
-
 	ColumnsModel* columnsModel = qobject_cast<ColumnsModel*>(_nativeModel);
-
-	connect(_nativeModel, &QAbstractItemModel::dataChanged,			this, &SourceItem::_dataChangedHandler);
-	connect(_nativeModel, &QAbstractItemModel::rowsInserted,		this, &SourceItem::_resetModel);
-	connect(_nativeModel, &QAbstractItemModel::rowsRemoved,			this, &SourceItem::_resetModel);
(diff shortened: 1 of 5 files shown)


Desktop/widgets/sourceitem.cpp
```diff
@@ -127,8 +127,35 @@ void SourceItem::_connectModels()
...
+{
+	if (!_connected) return;
+
+	ListModel *controlModel = _listControl->model();
+	AnalysisForm* form		= _listControl->form();
+
+	if (_isRSource && form)
+		disconnect(form,	&AnalysisForm::rSourceChanged,				this, &SourceItem::_rSourceChanged);
+
+	if (!_nativeModel) return;
```
JorisGoosen: shouldn't `_connected` still be set to `false` even if there isnt a nativeModel ?

JorisGoosen:
If I use the mixed.models.jasp file from [the-issue](https://github.com/jasp-stats/jasp-test-release/issues/1786) I get:

image

At least, after I remove all "fixed effects variables" and re-add them and do the same in "random effects grouping factors".

So while the original error disappeared (and thus this issue might be considered solved) there is still some problem.
But at least this seems output from the actual error-checking in the analysis.

JorisGoosen (changes requested):
I would guess that also for models without a native model there should be `_connected = false;` but that is just a feeling.

Either Im missing something about that and this can be merged (because it looks fine otherwise) or I do in fact request a change.

boutinb:
> I would guess that also for models without a native model there should be `_connected = false;` but that is just a feeling.

_connected must be set also without native model. Good point...

> So while the original error disappeared (and thus this issue might be considered solved) there is still some problem.
But at least this seems output from the actual error-checking in the analysis.

The error is now correct (and informative). That was the goal of this fix.

JorisGoosen (approved):

PR #4798, opened 2022-06-21
Importing database from SQL

### LINKED ISSUE #29: Enhancement request: Import data from database.
As mentioned in the meeting:
An import from database(s) might be useful to some (commercial) users.
First thoughts on construction:
1. Use an ODBC connector to access the databases.
2. Have the user enter an SQL statement ("JOIN") somewhere (An "Import options" dialogue?).
A full featured DB browser would be overkill.
Points to think about:
- DB type to JASP type mapping.
- A RDMS could easily produce more data than the memory available to JASP can hold.
- If people are using large datasets (from a DB or elsewhere), might we get performance issues?

JorisGoosen:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/1693
Fixes https://github.com/jasp-stats/jasp-issues/issues/29


JorisGoosen:
@boutinb suggested adding a "remember me" checkmark like at OSF which could work like:
-   check remember me -> stores password in settings and in JASP-file if saved.
 -  dont check remember me -> uses password + user while loaded. but it isnt saved and everytime the file is loaded the user is prompted with a user+password dialog?

Desktop/components/JASP/Widgets/ErrorMessage.qml
```diff
@@ -0,0 +1,100 @@
...
+		enabled: preferencesModel.animationsOn;
+
+		NumberAnimation
+		{
+			duration			: 300
+			easing.type			: Easing.InOutQuad
+		}
+	}
+
+	CrossButton { onCrossClicked: {	errorMsg.opacity = 0 } }
```
boutinb, line 62: By clicking the cross button, this will just make the message disappear, but it is not removed: the place it takes will be still used, making a big empty place.

Desktop/components/JASP/Widgets/FileMenu/Database.qml
```diff
@@ -0,0 +1,379 @@
...
+                    onTextChanged:			fileMenuModel.database.password = text;
+
+                    textInput.echoMode:		TextInput.Password
+
+                    x:						dbHostnameLabel.width
+                    width:					dbHostnameInput.width
+                }
+            }
+
+            RoundedButton
```
boutinb: This button should appear only when the database is not yet connected. When connected it should show 'Close database', and the connection settings should be disabled.

Desktop/components/JASP/Widgets/FileMenu/Database.qml
```diff
@@ -0,0 +1,379 @@
...
+                {
+                    id:						loadResults
+                    text:					qsTr("Load into JASP")
+                    onClicked:				fileMenuModel.database.importResults();
+                    KeyNavigation.tab:		intervalSpinner
+                    anchors.left:			parent.left
+                }
+
+                SpinBox
+                {
```
boutinb, line 351: If I change back the synchronization to 0, then it still tries to synchronize.

Desktop/components/JASP/Widgets/FileMenu/Database.qml
```diff
@@ -0,0 +1,379 @@
...
+                    id:					intervalSpinner
+                    value:				fileMenuModel.database.interval
+                    onValueChanged:		if(value != "") fileMenuModel.database.interval = value
+                    from:				0
+                    to:					24 * 60 * 7 // do we want to allow for an interval bigger than a week?
```
boutinb: Why to put a max limitation? It's not necessary, you never know how JASP will be used in te future.

Desktop/components/JASP/Widgets/FileMenu/Database.qml
```diff
@@ -0,0 +1,379 @@
...
+                height:			childrenRect.height
+                anchors
+                {
+                    left:		parent.left
+                    right:		parent.right
+                    margins:	jaspTheme.generalAnchorMargin
+                }
+
+                RoundedButton
+                {
```
boutinb, line 342: If I change the query, execute it, and load it again, then it does not change the data in JASP.

Desktop/components/JASP/Widgets/FileMenu/Database.qml
```diff
@@ -0,0 +1,379 @@
...
+                right:		parent.right
+                margins:	jaspTheme.generalAnchorMargin
+            }
+        }
+
+        PrefsGroupRect
+        {
+            id:		databaseGroup
+
+            title:	qsTr("Database")
```
boutinb, line 83: Are all fields needed for all kinds of database? I don't think a port is needed for SQLLite for example.

boutinb (changes requested):
It works really easily, but the interface can be fine-tuned. Especially when having loaded the data, and you come back to the Open/Database, then it is not clear what you can change or not (in fact any change does not seem to work..).

boutinb:
Also, if I modify the SQL statement, save it as JASP file, and open the JASP file again, it still uses the old settings.
Another point the Sync Data menu, should be disabled when a Database connection is used, or the Synchronization setting should be set in this menu.

Desktop/components/JASP/Widgets/FileMenu/Database.qml
JorisGoosen: In fact you could leave empty what you want and port==0 is empty there.
sqlite for sure doesnt need that indeed, I also mention this in the guide: https://github.com/JorisGoosen/jasp-desktop/blob/importDB/Resources/Help/filemenu/Database.md#sqlite

Desktop/components/JASP/Widgets/FileMenu/Database.qml
JorisGoosen: I guess, I will remove it.

JorisGoosen:
Well, using the open-interface would result in a new dataset being loaded right?
There is no provision made for changing the database connection in JASP later... 

So changing the query would also have no effect on the currently loaded JASP. But if you press "Load into JASP" it *should* open a new dataset (but I dont think Ive tested that).

Desktop/components/JASP/Widgets/FileMenu/Database.qml
JorisGoosen: Ok but I think this is a fundamental thing here, the "Open" menu is not for changing the currently loaded dataset at all.
So a database is never "connected" once you enter that screen right?

What would the behaviour even be?
You press "Load into JASP" and then you lose all changes you made? 
Why would you change the connection once it works?

Desktop/components/JASP/Widgets/FileMenu/Database.qml
JorisGoosen: and it also doesnt open a new JASP?
because that is what I would expect

Desktop/components/JASP/Widgets/FileMenu/Database.qml
JorisGoosen: so you mean after you loaded it? Again the same open != change-current

Desktop/components/JASP/Widgets/ErrorMessage.qml
JorisGoosen: oops

JorisGoosen:
Well, @boutinb is on vacation and I think all the prvevious issues have been resolved so Im just gonna merge it.

PR #4799, opened 2022-06-22
add power and logit transforms

Kucharssim:
closes https://github.com/jasp-stats/INTERNAL-jasp/issues/1728
closes https://github.com/jasp-stats/INTERNAL-jasp/issues/1866


Desktop/components/JASP/Widgets/ComputeColumnWindow.qml
```diff
@@ -214,13 +214,19 @@ FocusScope
...
-					ListElement	{ type: "function";	functionName: "fishZ";      functionParameters: "y";		functionParamTypes: "number";			toolTip: qsTr("Fisher's Z-transform (i.e., the inverse hyperbolic tangent) to transform correlations, numbers between -1 and 1 to the real line") }
-					ListElement	{ type: "function";	functionName: "invFishZ";	functionParameters: "y";		functionParamTypes: "number";			toolTip: qsTr("Inverse Fisher's Z-transform (i.e., the hyperbolic tangent) to transform real numbers to numbers between -1 and 1") }
+					ListElement	{ type: "function";	functionName: "log";			functionParameters: "y";				functionParamTypes: "number";						toolTip: qsTr("natural logarithm") }
+					ListElement	{ type: "function";	functionName: "log2";			functionParameters: "y";				functionParamTypes: "number";						toolTip: qsTr("base 2 logarithm") }
+					ListElement	{ type: "function";	functionName: "log10";			functionParameters: "y";				functionParamTypes: "number";						toolTip: qsTr("base 10 logarithm") }
+					ListElement	{ type: "function";	functionName: "logb";			functionParameters: "y,base";			functionParamTypes: "number";						toolTip: qsTr("logarithm of y in 'base'") }
```
vandenman: This doesn't seem to work for me:
image

vandenman (changes requested):
So it looks all fine to me, except that `logb` doesn't seem to work (maybe the functionName should just be `log`?). It also shows up like this in the list:

image

shouldn't it be `logb(y, b)`? Although some other functions, like `normalDist`, also accept multiple parameters but are shown without them.

Also, I was wondering if we should typeset inverse functions like `invLogit` like `logit⁻¹(y)`. On the one hand, `logit⁻¹(y)` is more readable (at least to me), while on the other hand, it's convenient that `invLogit` can also be used in the R code. Then again, there are a few other symbols that cannot be typed in R either: 
image

any thoughts?

Desktop/components/JASP/Widgets/ComputeColumnWindow.qml
Kucharssim: it seems like you found an existing bug :) I will fix this

Kucharssim:
> Also, I was wondering if we should typeset inverse functions like invLogit like logit⁻¹(y). On the one hand, logit⁻¹(y) is more readable (at least to me), while on the other hand, it's convenient that invLogit can also be used in the R code. Then again, there are a few other symbols that cannot be typed in R either

Good point. I think the whole point of the point and click interface is to be human friendly (and not programmer friendly), so it does make sense to me to typeset it like logit⁻¹(y).

Just a question about implementation. I thought I could just change it to `logit\u207B\u00B9` (assuming unicode would work). However, it seems like we did not implement the option to replace the name of the function with another string? I could only replace it with an image like for the symbols here?

https://github.com/jasp-stats/jasp-desktop/blob/7c8c2a91fc914bc3a1c3b998611517d2fe0bdaf6/Desktop/components/JASP/Widgets/FilterConstructor/Function.qml#L17

 (If @vandenman does not know perhaps @JorisGoosen would). I must admit I am not happy about the prospect of making these images and in two themes 😅

JorisGoosen:
I could easily add a "pretty functionname" property and then you could enter the unicode directly there?

Kucharssim:
If you can easily add that, that would be nice!

JorisGoosen:
> If you can easily add that, that would be nice!

I added it to `stable` just now, see `log2` and `log10` in `ComputeColumnWindow.qml`

Kucharssim:
For some reason I cannot build right now to test it, but the needed changes looked easy enough to make.

vandenman:
so this probably works, but because we no longer do `library(jaspBase)` by default I'm now seeing 

image

for all our custom functions...

JorisGoosen:
So... I can move https://github.com/jasp-stats/jasp-desktop/commit/e9e8011c7d663cfefc75eb3e23b05b390064b072 to `stable`

vandenman:
> So... I can move https://github.com/jasp-stats/jasp-desktop/commit/e9e8011c7d663cfefc75eb3e23b05b390064b072 to stable

I don't like it, but maybe it's best to do that for now. The downside is that every module needs to be aware that they should not clash with anything in `https://github.com/jasp-stats/jaspBase/blob/master/NAMESPACE`, but perhaps they should do that anyway because otherwise, we need to provide syntax with `jaspBase::` and `jaspModule::` rather than `library`.

JorisGoosen:
Ok, I moved the commit to `stable` so here it is: https://github.com/jasp-stats/jasp-desktop/commit/e4b61bc23767e1bb1c9d64b6ef6683880aed9a13

JorisGoosen:
so those custom functions should work again

Kucharssim:
Now it works!

There is just some funky business going on with the pretty labels:
Screenshot 2022-08-25 at 15 47 08

You can see on the right that the pretty labels are very pretty indeed, and they also look pretty if I drag them to the computation window alone, but if I drag it inside of another function, then the R function name is displayed instead of the pretty label (i.e., `logit(invLogit(contNormal))` should look like `logit(logit⁻¹((contNormal))`)

JorisGoosen:
Aah hey could you make a bug for it in 0.16.4? SHould be easy to fix but also to forget im afraid :p

Kucharssim:
Ok so this should be fine now, I guess @vandenman can give it a go once more but otherwise it should be ready

vandenman:
Looks good to me! The only nitpick I have is that the log2 and log10 should use subscript base rather than superscript base:

image

but Wikipedia (and others) do this:

image

Kucharssim:
Thanks for spotting that! Should be fixed now

vandenman (approved):

PR #4801, opened 2022-06-30
R 4.2.1

JorisGoosen:
Makes JASP use R 4.2.1 

fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/1792 
fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/1886

I also removed nowide entirely and set the codepage to `.65001` on windows which *should* make everything work with utf-8. This might break the "local codepage reading when no bom present in a csv on windows" so: https://github.com/jasp-stats/INTERNAL-jasp/issues/1894


koenderks:
I can build this PR on Windows but only without the jaspProphet module (which fails to install)

JorisGoosen:
I already disable jaspNetwork as well so Ill turn off the prophet as well and make some issues.

JorisGoosen:
Right before this PR is merged we should first merge: https://github.com/jasp-stats/jaspResults/pull/17
Then when it is merged we can just update the commit in the PR to that one and merge it like that.

JorisGoosen:
@vandenman suggested we might want to remove: https://github.com/jasp-stats/jaspBase/blob/5d79025af56f6c26788302af0d38cb2b780e001f/R/moduleInstall.R#L316

JorisGoosen:
> @vandenman suggested we might want to remove: [jasp-stats/jaspBase@`5d79025`/R/moduleInstall.R#L316](https://github.com/jasp-stats/jaspBase/blob/5d79025af56f6c26788302af0d38cb2b780e001f/R/moduleInstall.R#L316)

And does he want it only for linux or for everything?

vandenman:
For everything.

My idea was to move the installation files out of jaspBase. That way, we don't need to explicitly install jaspBase anymore because it becomes installed automatically when we install the modules. This allows different modules to use different versions of jaspBase (which allows us to make breaking changes to jaspBase, should we ever want to). This is by no means necessary though, and even if we'd want to go this way it doesn't make sense until we start on an actual module store.

JorisGoosen:
So we would just use "pure" `renv` then?

vandenman:
> So we would just use "pure" renv then?

Exactly, base R and "pure" renv.

JorisGoosen:
but how would the fixing of the otool-paths then work on mac?

vandenman:
The same way it works now? These lines would be unchanged:

https://github.com/jasp-stats/jaspBase/blob/5d79025af56f6c26788302af0d38cb2b780e001f/R/moduleInstall.R#L1-L24

The main difference is that this file (and the file that generates checkSums) would no longer be a part of jaspBase, but are sourced and called separately. This could even be an R package to facilitate testing and all that but just separated from jaspBase.

JorisGoosen:
Ja ok, so you mean a separate "installer-pkg/r-file" then, doesnt sound to bad

JorisGoosen:
But maybe we should make a proper issue for it?
I think this PR is fine like this, after it is merged we can upgrade to qt 6.3.1 and R 4.2.1 and then see about other changes.

vandenman:
I made https://github.com/jasp-stats/INTERNAL-jasp/issues/1899 to continue the discussion

Tools/CMake/Modules.cmake
```diff
@@ -162,21 +162,32 @@ configure_file("${PROJECT_SOURCE_DIR}/Modules/install-jaspBase.R.in"
...
-    ${CMAKE_COMMAND} -D
-    NAME_TOOL_PREFIX_PATCHER=${PROJECT_SOURCE_DIR}/Tools/macOS/install_name_prefix_tool.sh
-    -D PATH=${R_HOME_PATH}/library -D R_HOME_PATH=${R_HOME_PATH} -D
-    R_DIR_NAME=${R_DIR_NAME} -D SIGNING_IDENTITY=${APPLE_CODESIGN_IDENTITY} -D
-    SIGNING=${IS_SIGNING} -D CODESIGN_TIMESTAMP_FLAG=${CODESIGN_TIMESTAMP_FLAG}
-    -P ${PROJECT_SOURCE_DIR}/Tools/CMake/Patch.cmake
-  COMMENT "------ Installing 'jaspBase'")
-
+if(APPLE)
+	add_custom_command(
```
vandenman, line 166: spaces?

Tools/CMake/R.cmake
```diff
@@ -9,47 +9,58 @@
...
 set(R_BINARY_HASHES
     "61d3909bc070f7fb86c5a2bd67209fda9408faaa"
     "69e8845ffa134c822d4bdcf458220e841a9eeaa5"
+    "c72e68bc50e84bea68a2379073c9fedbdfaeda0c"
     "45121f2c830b0cd7d180aee3fc4cd80d0de1e582"
     "dad405d4f58349403c4976ba50e944502070b209"
-    "c72e68bc50e84bea68a2379073c9fedbdfaeda0c"
-    "d4068fdc75334c850d5948a0dc8356d34d3512e1")
+    "d4068fdc75334c850d5948a0dc8356d34d3512e1"
+	"f83a6c96cedd19193255f94cb01381a273073a3a"
```
vandenman, line 54: spaces

vandenman (approved):
looks good to me! Also, boost is no longer used by jaspResults, correct? I'll remove `BH` as a dependency in a later PR.

JorisGoosen:
> looks good to me! Also, boost is no longer used by jaspResults, correct? I'll remove `BH` as a dependency in a later PR.

Indeed I removed the two places where we called `boost` :p

Desktop/main.cpp
```diff
@@ -314,6 +314,8 @@ int main(int argc, char *argv[])
 	// This does slow down QML quite a bit and disables gradients
 	// qputenv("QT_QUICK_BACKEND", "software");
 	//Turned this off because it causes some weird resizing problems QCoreApplication::setAttribute(Qt::AA_UseOpenGLES); //might fix weirdlooking QML on Windows when using not-so-goo drivers? ( https://github.com/jasp-stats/jasp-desktop/issues/2669 )
+	
+	setlocale(LC_ALL, ".65001"); //use utf8
```
boutinb, line 318: Why not use ".UTF8" as locale?

boutinb (approved):
I suppose this is OK...
When prophet and network can be added?
And why rtools42 replaces mingw?

Desktop/main.cpp
JorisGoosen: Because this is windows and that is what it is called right?
Where do you see ".UTF8" as a possible codepage? (as in, if you can show me some docs that state it is acceptable on windows and will gladly change it cause it looks better but I didnt encounter that anywhere)

JorisGoosen:
> I suppose this is OK... 

You don't sound too enthusiastic :p

> When prophet and network can be added? 

See https://github.com/jasp-stats/INTERNAL-jasp/issues/1897
See https://github.com/jasp-stats/INTERNAL-jasp/issues/1896
I think prophet can already be enabled but t

> And why rtools42 replaces mingw?

https://developer.r-project.org/Blog/public/2022/06/16/upcoming-changes-in-r-4.2.1-on-windows/
Because rtools42 is basically mingw anyway and it was weird we were using both rtools40 *and* mingw before.
Now we just use rtools42 which people need to install to do R-dev work on windows anyway.
Also it contains the UCRT stuff we need to be able to properly use that `.65001` codepage ^^

Desktop/main.cpp
shun2wang: I think docs [here](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/setlocale-wsetlocale?view=msvc-170) is helpful😀

shun2wang:
Could you make a nightly build of this as I think this might fix some character encoding issues, I can't wait to test it...thx😊

boutinb:
> Could you make a nightly build of this as I think this might fix some character encoding issues, I can't wait to test it...thx😊

The buildbot has just started. It takes a bit more than 1 hour

Desktop/main.cpp
boutinb: This doc just says that ".UTF-8" is the codepage...

JorisGoosen:
And if you run into any problems @shun2wang you can add them to: https://github.com/jasp-stats/INTERNAL-jasp/issues/1894

shun2wang:
I can't visit internal issues...

JorisGoosen:
Well, I will see whether we can give the translators access.

the new nightly is up btw ^^

JorisGoosen:
@shun2wang you should be able to see it now I think

PR #4804, opened 2022-07-05
Add combineTerms property for sources

boutinb:
With the combineTerms source, it is now possible to combine the variables in different ways (cross, 2-way, 3-way, etc)
Also add the maxTermsWidth property in all list QML Widget, so that we can know what is the maximum length (depending of the metrics) of the terms of the list.
This is needed for @JohnnyDoorn 's new feature.

You can use this new feature with this QML for example:

```
	VariablesForm
	{
		AvailableVariablesList	{ name: "allVariablesList"								}
		AssignedVariablesList	{ name: "variables";		title: qsTr("Variables");	id: variables 	}
	}

	ComponentsList
	{
		id:combinations
		title: "Combinations"
		name: "Combinations"
		source: [ {name: "variables", combineTerms: JASP.Combination2Way}]
		property int textWidth: variables.maxTermsWidth

		rowComponent: RowLayout
		{
			property var pair : rowValue.split(INTERACTION_SEPARATOR)
			Text { Layout.preferredWidth: combinations.textWidth; text: pair[0] }
			Text { text: "-" }
			Text { Layout.preferredWidth: combinations.textWidth; text: pair[1] }
			DropDown { name: "mediation"; addEmptyValue: true; placeholderText: qsTr("No Mediation"); source: [{name: "variables", discard: {values: pair} }] }
		}

	}
```

Then you should get this:
Screenshot 2022-07-05 at 14 18 28


JorisGoosen:
Can you also add some documentation for this to https://github.com/jasp-stats/jasp-desktop/blob/stable/Docs/development/jasp-qml-guide.md#variable-specification ?

boutinb:
> Can you also add some documentation for this to https://github.com/jasp-stats/jasp-desktop/blob/stable/Docs/development/jasp-qml-guide.md#variable-specification ?

Yes I'll do

Desktop/analysis/terms.cpp
```diff
@@ -391,6 +391,42 @@ Terms Terms::ffCombinations(const Terms &terms)
...
+{
+	Terms combinedTerms;
+
+	size_t nbTerms = size();
+
+	switch (type)
+	{
+	case JASPControl::CombinationType::CombinationCross:
+		combinedTerms = crossCombinations();
+		break;
```
JorisGoosen, line 404: newlines after each `break` would have been nice but I won't block the PR on that :p

Desktop/widgets/jasplistcontrol.cpp
```diff
@@ -232,6 +236,21 @@ int JASPListControl::count()
...
 
+double JASPListControl::maxTermsWidth()
+{
+	if (!model()) return 0;
+	double maxWidth = 0;
+
+	QFontMetricsF& metrics = JaspTheme::fontMetrics();
+	for (const Term& term : model()->terms())
+	{
+		double width = metrics.horizontalAdvance(term.asQString());
```
JorisGoosen:
```
for (const Term& term : model()->terms())
	maxWidth = std::max(maxWidth, metrics.horizontalAdvance(term.asQString()));
``` 
Is a bit cleaner don't you think? ;)

JorisGoosen (approved):

Desktop/widgets/jasplistcontrol.cpp
JorisGoosen: But if you don't want that you can also just merge it as is ofc

PR #4806, opened 2022-07-08
Data library updates (GLMMS: Larks & Owls)

FBartos:
So, I have a few JASP analyses that could be added to the JASP library.

First, I'm adding this one to the mixed models, if the PR looks good, I would add the remaining ones.
(I changed the numbering as Mixed Models should not be last.)

fixes: https://github.com/jasp-stats/INTERNAL-jasp/issues/1877


Kucharssim (changes requested):
Looks good to me. Just two little comments for the .jasp file

1. Can you make the title inside of the results panel "Larks and Owls: Classical Generalized Linear Models"?
2. Can you provide description for all variables in the data set, not only the variables that you use in the analysis?

Do you want to add more files into this PR, or are you planning to make additional PRs?

FBartos:
Thanks, I will add the suggestions.

What do you prefer? Adding the other library files here or in a new PR?

Kucharssim:
Here is fine :)

FBartos:
I changed the title, added a description to some of the variables, and I removed other variables that I did not want to spend time figuring out and describing :)

Kucharssim (approved):

PR #4809, opened 2022-07-15
Empty R Source gives one empty row instead of no row

boutinb:
Issue can be easily reproduced by just opening the Classical Continuous Cochrane:
Screenshot 2022-07-15 at 11 21 48

This is because in c++, a vector of vector of string initialized by {{}} is a vector of one empty element. It should be initialized by {} which is an empty vector.


JorisGoosen:
ooh haha, my bad...

PR #4814, opened 2022-08-01
adds msgs to upgrades.qml

JorisGoosen:
Implements https://github.com/jasp-stats/INTERNAL-jasp/issues/1914

However, I couldnt get the warnings/errors from qml properly out because it would not be very useful compared to the effort.


JorisGoosen:
@vandenman could you at least give this a compile and see if my added instructions and code make sense/are clear?
Especially in the light of https://github.com/jasp-stats/INTERNAL-jasp/issues/1912

I did test of course and I dont expect you to review all the code but a sanity check for the PR would be appreciated ^^

Desktop/modules/upgrader/changecopy.cpp
```diff
@@ -24,6 +24,9 @@ void ChangeCopy::applyUpgrade(Json::Value & options, UpgradeMsgs & msgs) const
 	msgs[newName]		= msgs[oldName];
 
 	msgs[logId].push_back(prefixLog + "Copied option '" + oldName + "' to '" + newName + "'");
+
+	if(msg() != "")
```
vandenman, line 28: Is there a reason why you're not doing `msg().isEmpty()`?

Desktop/modules/upgrader/upgrade.cpp
```diff
@@ -66,7 +69,7 @@ Version Upgrade::fromVersion() const
 	}
 	catch(Version::encodingError & e)
 	{
-		throw upgradeError("Incorrect from version: '" + fq(_fromVersion) + "' supplied to Upgrade...");
+		throw upgradeError("Incorrect from-version: '" + fq(_fromVersion) + "' supplied to Upgrade...");
```
vandenman:
```suggestion
		throw upgradeError("Incorrect fromVersion: '" + fq(_fromVersion) + "' supplied to Upgrade...");
```
If this should match the properties in `Upgrader.qml`, I would use their exact names.

vandenman (approved):
The explanation makes sense to me and it also works in jasp:

image

Desktop/modules/upgrader/changecopy.cpp
JorisGoosen: save like 4 characters at least and I like the look of it ^^

PR #4820, opened 2022-08-08
Remove japBase after module install adds it

JorisGoosen:
This way bundled modules always use bundled jaspBase, to ensure this we add the main library first to .libPaths()

### DIFF
--- a/Desktop/modules/dynamicmodule.cpp
+++ b/Desktop/modules/dynamicmodule.cpp
@@ -470,12 +470,12 @@ void DynamicModule::unpackage()
 
 std::string DynamicModule::getLibPathsToUse()
 {
-	std::string libPathsToUse = "c('" + shortenWinPaths(moduleRLibrary()).toStdString()	+ "')";
+	std::string libPathsToUse = "c('" + AppDirs::rHome().toStdString() + "/library', '" + shortenWinPaths(moduleRLibrary()).toStdString()	+ "')";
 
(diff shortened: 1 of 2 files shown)


Desktop/modules/dynamicmodule.cpp
```diff
@@ -470,12 +470,12 @@ void DynamicModule::unpackage()
...
-	std::string libPathsToUse = "c('" + shortenWinPaths(moduleRLibrary()).toStdString()	+ "')";
+	std::string libPathsToUse = "c('" + AppDirs::rHome().toStdString() + "/library', '" + shortenWinPaths(moduleRLibrary()).toStdString()	+ "')";
 
-	std::vector<std::string> requiredLibPaths = fq(DynamicModules::dynMods()->requiredModulesLibPaths(tq(_name)));
+	//std::vector<std::string> requiredLibPaths = fq(DynamicModules::dynMods()->requiredModulesLibPaths(tq(_name)));
 
-	for(const std::string & path : requiredLibPaths)
-		libPathsToUse += ", '" + path + "'";
+	//for(const std::string & path : requiredLibPaths)
+	//	libPathsToUse += ", '" + path + "'";
```
vandenman, line 478: does this remove the dependency among modules? So the libPaths of e.g., jaspAnova become

Modules/jaspAnova
R/library

?

vandenman (approved):

Desktop/modules/dynamicmodule.cpp
JorisGoosen: No that is only solved in `development` (the old deps getting loaded and thus messing up new dynamic modules installed later)

PR #4826, opened 2022-08-11
Almost fixes  weird font in windows results

### LINKED ISSUE #1747: [Bug]: 0.16.2 & 0.16.3 Font not displaying correctly on Windows 10
### JASP Version

0.16.2

### Commit ID

_No response_

### JASP Module

jaspRegression

Linear Regression, but I think this is not limited to the regression

Windows 10; 10.0.19044 Build 19044

### Bug Description

Hi,
 
I would like to report a bug with the font display on 0.16.2. I upgraded from 16.1 to 16.2, and immediately noticed that numbers are not displaying correctly. Please see the attached ppt file (I used screen capture). So I uninstalled 16.2 immediately, and reinstalled 16.1. And it fixed the problem. I also attached the screen capture of 16.1. As you can see from the clock (right bottom), it happened in 14 minutes, from the same laptop. 

[JASP 0.16.2 bug report.pptx](https://github.com/jasp-stats/jasp-issues/files/8913301/JASP.0.16.2.bug.report.pptx)

Below is my system info. 

OS Name	Microsoft Windows 10 Home
Version	10.0.19044 Build 19044
Other OS Description 	Not Available
OS Manufacturer	Microsoft Corporation
System Name	*****
System Manufacturer	Dell Inc.
System Model	XPS 13 9310 2-in-1
System Type	x64-based PC
System SKU	*****
Processor	11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz, 2803 Mhz, 4 Core(s), 8 Logical Processor(s)
BIOS Version/Date	Dell Inc. 2.9.0, 3/17/2022
SMBIOS Version	3.2
Embedded Controller Version	255.255
BIOS Mode	UEFI
BaseBoard Manufacturer	Dell Inc.
BaseBoard Product	*****
BaseBoard Version	*****
Platform Role	Mobile
Secure Boot State	On

Thank you.

### Expected Behaviour

The numbers should not be mushed. I should be able to read them with clarity.

### Steps to Reproduce

1. Run any data analysis on Windows 10
2. 
3. 
...

### Log (if any)

_No response_

### Final Checklist

JorisGoosen:
Fixes https://github.com/jasp-stats/jasp-issues/issues/1747 mostly

You can still get it to look slightly weird if you try but it seems to usually get it right now

### DIFF
--- a/Desktop/components/JASP/Widgets/MainPage.qml
+++ b/Desktop/components/JASP/Widgets/MainPage.qml
@@ -178,15 +178,16 @@ Item
 			WebEngineView
 			{
 				id:						resultsView
+				clip:                   true
 
 				anchors
 				{
 					top:				parent.top
-					left:				parent.left
 					bottom:				parent.bottom
 				}
 
-				width:					giveResultsSomeSpace.width - panelSplit.hackySplitHandlerHideWidth
+				x:						1 + (Math.floor(parent.x) - parent.x)
+				width:					Math.floor(giveResultsSomeSpace.width - panelSplit.hackySplitHandlerHideWidth)
 
 				url:					resultsJsInterface.resultsPageUrl
 				onContextMenuRequested: (request)=>{ request.accepted = true; }


JorisGoosen:
This is done by simply trying to compensate for any fractional `x` value

Desktop/components/JASP/Widgets/MainPage.qml
```diff
@@ -178,15 +178,16 @@ Item
 			WebEngineView
 			{
 				id:						resultsView
+				clip:                   true
```
boutinb, line 181: I suppose you have tested this fix on the Windows laptop, but why do you need to add this `clip: true`?

JorisGoosen: Well we don't "need" this clip, but because clipping forces the to-be-clipped texture to be rendered separately from the rest of the qml-scenegraph I figured it couldnt hurt to force that.

If it doesn't do anything it would only slow down the whole mess marginally so I think we can just leave it in.

JorisGoosen: Also, yes I tested it ;)

boutinb (approved):

PR #4833, opened 2022-08-18
Userdefined label for missing values in data

JorisGoosen:
Implements https://github.com/jasp-stats/jasp-issues/issues/1708

image

From JASP with love:
image

image

Haha, I like this feature + unicode support, such a colorful data view ^^

### DIFF
--- a/Common/column.cpp
+++ b/Common/column.cpp
@@ -1011,31 +1011,31 @@ bool Column::isValueEqual(int row, const string &value)
 	return false;
 }
 
-string Column::_getScaleValue(int row)
+string Column::_getScaleValue(int row, bool forDisplay)
 {
 	double v = AsDoubles[row];
 
-	if (v > std::numeric_limits<double>::max())					return string({ (char)0xE2, (char)0x88, (char)0x9E, 0 });
-	else if (v < std::numeric_limits<double>::lowest())			return string({ (char)0x2D, (char)0xE2, (char)0x88, (char)0x9E, 0 });
-	else if (Utils::isEmptyValue(v))							return Utils::emptyValue;
+	if (v > std::numeric_limits<double>::max())					return "∞";
+	else if (v < std::numeric_limits<double>::lowest())			return "-∞";
+	else if (Utils::isEmptyValue(v))							return forDisplay ? Utils::emptyValue : "";
 	else														return Utils::doubleToString(v);
 }
 
 string Column::getOriginalValue(int row)
 {
-	string result = Utils::emptyValue;
+	string result = "";
 
 	if (row < _rowCount)
 	{
 		if (_columnType == columnType::scale)
 		{
-			result = _getScaleValue(row);
+			result = _getScaleValue(row, false);
 		}
 		else
 		{
 			int key = AsInts[row];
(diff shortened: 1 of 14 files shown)


Desktop/utilities/settings.h
```diff
@@ -69,7 +69,7 @@ class Settings {
 		DB_IMPORT_INTERVAL,
 		DB_SHOW_WARNING,
 		DB_REMEMBER_ME,
-		REPORT_SHOW
```
boutinb, line 73: What was this REPORT_SHOW?

boutinb (changes requested):
Quite tricky...
The Utils::emptyValue was expected to be always the empty string, but now it can be something else. You have changed the use of Utils::emptyValue in several cases so that in these cases it always use the empty string in place of the user-defined empty string. But you did not change it in Column::getOriginalValue: this method is used to display the value in a table, but is also used in DataSetPackage::writeDataSetToOStream which writes the data export file. So the data export will use the User defined empty string, which is not what is expected I think.

Desktop/utilities/settings.h
JorisGoosen: accidentally committed it to `stable` but this is for `jaspReport` (for using jasp as a reportingtool)

Common/column.cpp
```diff
@@ -1011,31 +1011,31 @@ bool Column::isValueEqual(int row, const string &value)
...
 
-string Column::_getScaleValue(int row)
+string Column::_getScaleValue(int row, bool forDisplay)
 {
 	double v = AsDoubles[row];
 
-	if (v > std::numeric_limits<double>::max())					return string({ (char)0xE2, (char)0x88, (char)0x9E, 0 });
-	else if (v < std::numeric_limits<double>::lowest())			return string({ (char)0x2D, (char)0xE2, (char)0x88, (char)0x9E, 0 });
-	else if (Utils::isEmptyValue(v))							return Utils::emptyValue;
+	if (v > std::numeric_limits<double>::max())					return "∞";
```
JorisGoosen, line 1018: much nicer like this right?

And its not like anyone wouldnt be using utf-8 by now.

Desktop/data/datasetpackage.cpp
```diff
@@ -1183,7 +1183,7 @@ void DataSetPackage::writeDataSetToOStream(std::ostream & out, bool includeCompu
 			Column *column = cols[i];
 
 			std::string value = column->getOriginalValue(r);
-			if (value != ".")
```
JorisGoosen, line 1186: this was apparently wrong for a while already :p

JorisGoosen:
I think ive fixed it all, would appreciate another look though @boutinb ^^

Desktop/gui/preferencesmodel.h
```diff
@@ -227,8 +230,16 @@ public slots:
 	void maxEnginesChanged(				int			maxEngines);
 	void windowsNoBomNativeChanged(		bool		windowsNoBomNative);
 	void dbShowWarningChanged(			bool		dbShowWarning);
```
boutinb, line 232: Oeps: forgot to solve this conflict!

boutinb (approved):

PR #4835, opened 2022-08-26
Integer Field shows wrong error

boutinb:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2076

When an IntegerField (or a DoubleFiled) has a constraint (min or max for example), the control may show sometimes a red border as if it has wrong value, even if the value is OK.
This is due to a wrong fix made some time ago that activates this red border for controls that have wrong default value. This was done by assigning a width and a color to the focus indicator when the control was still initializing. This is wrong since the control gets then wrong default values for its width and color border.
The check whether the control has wrong default value must be done just after it is initialized.

### DIFF
--- a/Desktop/components/JASP/Controls/TextField.qml
+++ b/Desktop/components/JASP/Controls/TextField.qml
@@ -55,8 +55,8 @@ TextInputBase
 
 	signal editingFinished()
 	signal textEdited()
-	signal pressed()
-	signal released()
+	signal pressed(var event)
+	signal released(var event)
 
 	function doEditingFinished()
 	{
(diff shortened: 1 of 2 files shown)


Desktop/components/JASP/Controls/TextField.qml
```diff
@@ -128,8 +128,7 @@ TextInputBase
 			height:				parent.height + jaspTheme.jaspControlHighlightWidth
 			width:				parent.width  + jaspTheme.jaspControlHighlightWidth
 			color:				"transparent"
-			border.width:		control.acceptableInput ? 0 : 3 //See comment below
-			border.color:		jaspTheme.red // Needed when the QML file has wrong default value
+			border.width:		0
```
JorisGoosen: Might be useful to mention here that it is set programmatically because that is not quite obvious from qml...

JorisGoosen (approved):

PR #4842, opened 2022-09-01
Use qt rendering

JorisGoosen:
Implements https://github.com/jasp-stats/INTERNAL-jasp/issues/2091


JorisGoosen:
Alrighty I have added the setting.

However the native output today looks less bad on my win11 system, I still think we should have the qt-textrendering set by default though?

Or maybe not... I know what Tim wouldve wanted but it can really look terrible with Intel on windows right @vandenman  ?

vandenman:
> it can really look terrible with Intel on windows right @vandenman ?

I can only build on amd and have no recollection of how (bad) it looked before... If there is a nightly I can try it out on intel though. So not sure :shrug:

boutinb (changes requested):
If you resolve the conflict, it's perfect!

PR #4844, opened 2022-09-06
Make sure to not move analyses out of screen when adding a second one

JorisGoosen:
fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2090
should also keep https://github.com/jasp-stats/INTERNAL-jasp/issues/1752 fine

### DIFF
--- a/Desktop/components/JASP/Controls/Form.qml
+++ b/Desktop/components/JASP/Controls/Form.qml
@@ -100,7 +100,7 @@ AnalysisForm
 			id:					errorMessagesBox
 			visible:			form.errors !== ""
 			color:				jaspTheme.controlErrorBackgroundColor
-			width:				parent.width
+			width:				form.implicitWidth
 			height:				visible ? errorMessagesText.height : 0
 			anchors.top:		oldFileMessagesBox.bottom
 			anchors.margins:	visible ? jaspTheme.generalAnchorMargin : 0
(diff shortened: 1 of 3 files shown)


Desktop/components/JASP/Controls/Form.qml
```diff
@@ -100,7 +100,7 @@ AnalysisForm
 			id:					errorMessagesBox
 			visible:			form.errors !== ""
 			color:				jaspTheme.controlErrorBackgroundColor
-			width:				parent.width
+			width:				form.implicitWidth
```
boutinb, line 103: Why do you change it to implicitWidth ?

boutinb (approved):
I'm not sure to understand it completely. I have the impression it could be simpler, but I have tested it, it works: that's for now the main goal.

Desktop/components/JASP/Controls/Form.qml
JorisGoosen: Because the width of the parent changes when the scrollbar appears and dissappears. I figured this would give it a bit more stable sizing (and avoid some property-update-loops).

Not 100% it is necessary but yeah...

JorisGoosen:
> I'm not sure to understand it completely. I have the impression it could be simpler, but I have tested it, it works: that's for now the main goal.

Well, I wrote it originally (most of it anyway) and I also don't always understand whats going on.
Its why ive started adding the comments.

The problem is that when I tried to simplify this whole qml in the end it started getting more complicated because we want to do something not really supported by SplitView.

Such as having a resizable handler that can move out of screen left and right. And to work around that and to make sure stuff is draggable and all that... Well I think this is as simple as it is going to get.
But if we just put more comments every time we work on it maybe at some point it will become more organized.

PR #4847, opened 2022-09-08
Zoom on auxiliary windows and build fixes

RensDofferhoff:
Fixes a build error due to missing include
Fixes a linker error on Linux.

Fixes zoom on R-command and Help-windows.
Scales Help-window when zooming to keep text readable.
Makes CTRL-Q close the application on these windows.

### DIFF
--- a/Common/ipcchannel.h
+++ b/Common/ipcchannel.h
@@ -66,7 +66,7 @@ public:
 
 private:
 	bool tryWait(int timeout = 0);
-    void catchAndRepeat(const std::string & taskDescription, std::function<void()> doThis);
+	void catchAndRepeat(const std::string & taskDescription, std::function<void()> doThis);
 
 	void doubleMemoryOut();
 	void rebindMemoryInIfSizeChanged();
--- a/Common/ipcchannel.h
+++ b/Common/ipcchannel.h
@@ -36,6 +36,7 @@
 
 #include <boost/interprocess/managed_shared_memory.hpp>
 #include <boost/container/string.hpp>
+#include <functional>
 
 typedef boost::interprocess::allocator<char,	boost::interprocess::managed_shared_memory::segment_manager	> CharAllocator;
 typedef boost::container::basic_string<char,	std::char_traits<char>, CharAllocator						> String;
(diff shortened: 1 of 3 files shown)


Common/ipcchannel.h
```diff
@@ -36,6 +36,7 @@
 
 #include <boost/interprocess/managed_shared_memory.hpp>
 #include <boost/container/string.hpp>
+#include <functional>
```
JorisGoosen, line 39: Why add this?

Desktop/components/JASP/Widgets/HelpWindow.qml
```diff
@@ -7,7 +7,7 @@ import JASP.Controls	1.0
 Window
 {
 	id:					helpWindowRoot
-	width:				400
+	width:				400 * preferencesModel.uiScale
 	height:				700
```
JorisGoosen: Maybe also scale the height?

QMLComponents/CMakeLists.txt
```diff
@@ -29,6 +29,7 @@ target_include_directories(
 target_link_libraries(
   QMLComponents
   PUBLIC
+         Common
```
JorisGoosen, line 32: Aaah that really should have been there before yeah, I guess we werent using anything from it yet though. As it was compiling fine.

But good idea to avoid some future issues ^^

JorisGoosen:
If you make sure the aboutwindow height also scales I can approve and merge this

Common/ipcchannel.h
RensDofferhoff: It will not compile otherwise since std::function<void()> is used.
Not sure why it compiles for you without it honestly.

QMLComponents/CMakeLists.txt
RensDofferhoff: This already generates linking issues for me currently

JorisGoosen: I figured ^^

JorisGoosen:
I periodically rebase `development` on `stable` so this is what leads to the big mess of conflicts above here.
When rebasing your branch make sure you dont duplicate any commits (probably not though, git is pretty smart despite its name)

JorisGoosen (approved):

PR #4850, opened 2022-09-11
Left Censored data

EdsonRuiMontoro:
Normally the first statistics applied to environmental data is simply to calculate the average and standard deviation, of course after a probability distribution model. When we have  the left censored data (this is very common in environmental data), the average and standard deviation are too much affected. So, the simple substitution (1/2  Detection Limit) is not the recommended solution. The most technical solution is an estimation using MLE or ROS (Regression in Ordered Statistics).
There is no commercial statistics package to handle this situation. Minitab has only the right censored option, we have to flip the original data and then transform back to the original values. I believe this is an opportunity, because the application of Statistics in the environmental area is growing very fast.


JorisGoosen:
@EdsonRuiMontoro are you trying to request a feature because the best place for that would be our issuetracker: https://github.com/jasp-stats/jasp-issues/issues/new/choose

PR #4855, opened 2022-09-16
Make sure we use UTF-8 support from R 4.2.1 on windows

### LINKED ISSUE #1758: [Bug]: special character in user name results in engine crash with plots
### JASP Version

0.16.3

### Commit ID

_No response_

### JASP Module

jaspDescriptives, jaspRegression, jaspTTests

Any plot in any module

Windows 11

### Bug Description

Hi everybody,

all of my calculations in JASP work fine as long as I don't use any plots. As soon as I chosse to display descriptive plots or raincloud plot in T-Tests,  distribution plot in descriptives and so on, it will display a white window with heading and as soon as I try to scale it, the error message is dropped "The engine crashed while trying to run this analysis". 
Even more interesting: When I add a second type of plot, the first one will eventually display, sometimes for all of the items, sometimes for 1/4, however, again as soon as I try doing anything with it, the error message drops again. Tried my own data file and also the famous "kitchen rolls" testfile, same error.
I've read similar problems here already, but the solutions did not work for me (Region settings for instance).

I'd be glad about any help!

PS: First occurred to me in 0.16.2, now I was hoping to solve it with 0.16.3, which is not the case.

### Expected Behaviour

In versions prior to 0.16.2 plots would always display correctly and don't drop any error message when e.g. trying to scale them

### Steps to Reproduce

1. Load any variable in any test,
2. Check any analysis that is a plot,
3. Scale the window or try "edit image" and hit "ok"
...

### Log (if any)

[JASP 2022-06-22 12_27_15 Desktop.log](https://github.com/jasp-stats/jasp-issues/files/8956759/JASP.2022-06-22.12_27_15.Desktop.log)
[JASP 2022-06-22 12_27_15 Engine 0.log](https://github.com/jasp-stats/jasp-issues/files/8956760/JASP.2022-06-22.12_27_15.Engine.0.log)
Screenshot 2022-06-22 122627

### Final Checklist

JorisGoosen:
Fixes https://github.com/jasp-stats/jasp-issues/issues/1758 (engine crash on nonlatin characters in username on windows)

ps. All the nativeToUtf8 workaround functions can also be removed, but there is no hurry

### DIFF
--- a/Engine/main.cpp
+++ b/Engine/main.cpp
@@ -51,16 +51,12 @@ void openConsoleOutput(unsigned long slaveNo, unsigned parentPID)
 #ifdef _WIN32
 int wmain( int argc, wchar_t *argv[ ], wchar_t *envp[ ] )
 {
-   // setlocale(LC_ALL, ".UTF8"); //use utf8
-
-   // Utils::sleep(3000);
-
 	if(argc > 4)
 	{
 		unsigned long	slaveNo			= wcstoul(argv[1], NULL, 10),
 						parentPID		= wcstoul(argv[2], NULL, 10);
-        std::string		logFileBase		= Utils::wstringToString(argv[3]),
-                        logFileWhere	= Utils::wstringToString(argv[4]);
+		std::string		logFileBase		= Utils::wstringToString(argv[3]),
+						logFileWhere	= Utils::wstringToString(argv[4]);
 			
 #else
 int main(int argc, char *argv[])
@@ -114,9 +110,9 @@ int main(int argc, char *argv[])
 		Engine e(0, 0); //It needs to start to make sure rbridge functions work
 
 #ifdef _WIN32
-        _moduleLibraryFixer(Utils::wstringToString(argv[1]), true, true);
+	_m	oduleLibraryFixer(Utils::wstringToString(argv[1]), true, true);
 #else
-        _moduleLibraryFixer(argv[1], true, true);
+		_moduleLibraryFixer(argv[1], true, true);
 #endif
 
 		exit(0);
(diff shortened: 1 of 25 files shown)


JorisGoosen:
And also, some of the changes mightve become obsolete due to the active codepage.
But on the other hand, Id like to keep JASP.exe running in local codepage so that we can still support loading chinese-locale csvs in JASP and such shenanigans

Engine/main.cpp
```diff
@@ -109,14 +109,18 @@ int main(int argc, char *argv[])
 
 		Engine e(0, 0); //It needs to start to make sure rbridge functions work
 
+#ifdef _WIN32
+	_m	oduleLibraryFixer(Utils::wstringToString(argv[1]), true, true);
```
boutinb: oeps

R-Interface/jasprcpp.cpp
```diff
@@ -98,7 +101,10 @@ void STDCALL jaspRCPP_init(const char* buildYear, const char* version, RBridgeCa
 	_stringNativeToUtf8		= nativeToUtf8;
 
 	jaspRCPP_logString("Creating RInside.\n");
+
 	rinside = new RInside();
+
+	R_TempDir = (char*)tempDir;
```
boutinb, line 107: Don't understand: this variable does not seem to be used.

boutinb (approved):

R-Interface/jasprcpp.cpp
JorisGoosen: If you check where it is defined you'll see it is `extern` and it is in fact in R.
There were some issues on windows where the environment variables were converted to local codepage... Which messed up stuff if it couldnt fit. So I now just worked around it like this.

But actually, I think setting the active codepage to utf-8 probably already solved that. 

However, it works now and we can always clean stuff up later.

JorisGoosen: Previously this value was shared through `TMPDIR`

PR #4859, opened 2022-09-19
Stale QML cache issue workaround (linux only)

RensDofferhoff:
Tested on linux local build and flatpak.
Extends standard qmlcache path with commit num.
Deletes old ones to save space.

If the issue ever appears on other platforms extend the ifdefs


Desktop/utilities/qmlcacheconfiguration.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include "qmlcacheconfiguration.h"
+#include <QtGlobal>
+#include <QStandardPaths>
+#include "appinfo.h"
+#include "utilities/qutils.h"
+#include "log.h"
+
+
+void QMLCacheConfiguration::configureQMLCacheDir() {
+    //set cache environment variable
```
JorisGoosen: small nitpick, we use 4-wide tabs as indentation in C++ and QML.
(And 2 spaces in R)

Desktop/utilities/qmlcacheconfiguration.h
```diff
@@ -0,0 +1,18 @@
+#ifndef QMLCACHECONFIGURATION_H
+#define QMLCACHECONFIGURATION_H
+
+#ifdef linux
+#include <QDir>
+
+class QMLCacheConfiguration
+{
+public:
+    static void configureQMLCacheDir();
```
JorisGoosen: Also I think it might be better to add these functions to `Desktop/utilities/qmlutils.h/cpp`

JorisGoosen (changes requested):
Code looks fine!

However, I would like to see the indentation changed to tabs and the functions to be added to `qmlutils.h` instead of adding another class ^^

Desktop/utilities/qmlcacheconfiguration.h
RensDofferhoff: I believe the changes are as specified.

vandenman (approved):
Looks good to me, except there are still some spaces instead of tabs :wink:

Desktop/utilities/qmlutils.cpp
```diff
@@ -31,3 +31,51 @@ QJSValue	QmlUtils::decodeJson(const QJSValue	& val, QQuickItem * caller)
 	ColumnEncoder::decodeJson(v);
 	return tqj(v, caller);
 }
+
+
+#ifdef linux
+
+#include <QtGlobal>
```
JorisGoosen: I would prefer it if the includes were at the top to be honest.
That way it is a bit simpler to figure out what is being included when you open a file.

Desktop/utilities/qmlutils.cpp
```diff
@@ -31,3 +31,51 @@ QJSValue	QmlUtils::decodeJson(const QJSValue	& val, QQuickItem * caller)
...
+	}
+	Log::log() << "QML cache directory: " + cacheDir.absolutePath() << std::endl;
+}
+
+QDir QmlUtils::generateQMLCacheDir() {
+	QString commit = tq(AppInfo::gitCommit);
+	QString basePath = qEnvironmentVariable("QML_DISK_CACHE_PATH", QStandardPaths::writableLocation(QStandardPaths::CacheLocation));
+	QString path = basePath + "/qmlcache_" + commit;
+	QDir cacheDir(path);
+	if(!cacheDir.exists()) {
```
JorisGoosen:
```
if(!cacheDir.exists() && !cacheDir.mkpath("."))
			throw std::runtime_error("Could not create qml cache directory: " + fq(cacheDir.absolutePath()));
```
Is maybe a bit shorter and doesnt use a superfluous boolean

JorisGoosen (changes requested):
Moving the includes up in qmlutils.cpp I would request as a change and the other comment is more of a suggestion (but one I would like a lot).

JorisGoosen (approved):

PR #4864, opened 2022-09-20
Fixing computed columns and more

JorisGoosen:
Set environment variable JASP_R_INTERFACE_LIBRARY to something during jaspBase and jaspModule installs. jaspBase checks for it and sets the corresponding define if not ""

Passes pointers to some jaspRcpp functions to jaspBase so we do not need the header there (or JASP)

Also couldnt resist removing all the workaround code for utf-8 on windows...


JorisGoosen:
This PR totally needs: https://github.com/jasp-stats/jaspBase/pull/80

So if this is all to be merged, first merge jaspBase, then make a new commit in my branch (you both should be able to push) referencing that new commit in jaspBase@master en then pushing it here.
After that this can be squash-merged.

Assuming I haven't made any mistakes...

vandenman (approved):

boutinb (approved):

PR #4879, opened 2022-09-30
make jaspBase cacheAble in renv to avoid double installs

vandenman:
@JorisGoosen we made this not linkable because there were issues on macOS with signing (if I remember correctly). This change makes jaspBase linkable but only on Linux. 

This fixes a bug when building on flatpak. Because jaspBase is not linkable, we now install it twice. Once when installing jaspBase, and a second time when installing the first module (because it never ended up in the cache). The second time we're installing it, we haven't set the include dir so we end up with a compilation failure because jaspBase tries to clone jaspColumnEncoder which obviously fails on the flatpak buildbot.

Alternatively, we can set the include dir also in installModule.R.in, but I prefer this solution.

### DIFF
--- a/Modules/install-jaspBase.R.in
+++ b/Modules/install-jaspBase.R.in
@@ -28,7 +28,7 @@ prompt        <- FALSE
 
 if (md5SumsChanged(modulePkg, moduleLibrary)) {
   options(
-    renv.cache.linkable = getOS() == "linux",
+    renv.cache.linkable = FALSE,
     configure.vars = c(jaspBase = "INCLUDE_DIR='@PROJECT_SOURCE_DIR@/Common/jaspColumnEncoder'")
   )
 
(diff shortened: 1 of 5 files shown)


JorisGoosen:
Correct me if I'm wrong, but this still means we are wastefully and confusingly (if we make local changes to `jaspBase`) install second copies of jaspBase on windows and macos still?

Because it would be much better to treat everyone equally ^^

Although I think there is also only 1 reason to set the install on non-linkable and that is more for the modules.
I ran into problems where installing a development module (for instance `jaspDescriptives`) had a lasting effect on my buildfolder. So I build JASP in qt-creator, run it and install a development module.
I restart JASP and run descriptives again expecting the analysis to be back to normal, but in fact the development module is still loaded. Crazy thing is that the actual install-library in userdata for the development module is then wiped. But that is because "LINKABLE" also means (apparently anyway) that your links might be replaced...

It then also got into the code for jaspBase by mimicry and this probably means we can simplify your PR to simply setting jaspBase to be always linkable. Ill give it a try on macos and put it in the flatpak to see.

JorisGoosen:
It was the wrong thing to set anyway, we wanted to turn on "cacheAble" I think.
If I do this on macos the modules then all link to `jaspBase` so it looks like its going alright.
Ill backport the commit to a separate branch for flatpak because the new module stuff is also in `stable` right now.

JorisGoosen (approved):

JorisGoosen (approved):

vandenman:
> Correct me if I'm wrong, but this still means we are wastefully and confusingly (if we make local changes to jaspBase) install second copies of jaspBase on windows and macos still?

Yes. The problem is that the first time we install jaspBase we intentionally avoid creating a symlink (because macOS I think). That's why the second time we need to reinstall jaspBase. And this fails on flatpak.

Your second issue is avoidable with the new build approach I was working on, I'll make a PR to development with that soon-ish.

If we can just always make jaspBase linkable that would be ideal of course :)

PR #4882, opened 2022-10-04
Fixes the offset issue for scrollbars in drop-downs.

RensDofferhoff:
Looked really ugly in dark Theme.
Looks good now.


boutinb (approved):

boutinb:
The `addScrollBar` property should be removed: the scrollbar should appear only when necessary, and it is its default behaviour. Setting the visibility to `addScrollBar` removes this default behaviour. I don't know why this `addScrollBar` was added. It is used in the PrefsUI.qml (to list the fonts), and in Database.qml (to list the drivers), but it's completely unnecessary.
@RensDofferhoff Can you do that also?

RensDofferhoff:
@boutinb Done. The default behavior does indeed seem fine.

PR #4883, opened 2022-10-05
Update `jasp-human-guide.md`

shun2wang:
Part of https://github.com/jasp-stats/INTERNAL-jasp/issues/2125

- added Writing for i18n guide
- also placeholder and unicode

### DIFF
--- a/Docs/development/jasp-human-guide.md
+++ b/Docs/development/jasp-human-guide.md
@@ -29,6 +29,34 @@ Hence, in developing an analysis, it is good to think through each and every sor
 
 Error messages can either be placed over the top of the results table, or put in the footnotes. In general, an error which makes the whole analysis meaningless, should be placed over the top of the results table. A t-test where the user has specified an independent variable with three levels would be one such example. In contrast, if the error only affects one or a handful of values in the table, then an NaN should be placed in that cell, and a footnote marker added.
 
+### Writing for internationalization
+
+Internationalization(i18n) is the adapting of module to different languages. An inspiring news is that JASP is moving towards i18n, which means that the analysis and modules you developed will be translated into different languages. Therefore, it is necessary to pay attention to the readability and flexibility of the message during writing.
+Passing your message with [gettext()](https://en.wikipedia.org/wiki/Gettext), JASP will generate the string  to be translated automatically in a `.po` file.
+
+**Placeholders**
+
+If you have multiple placeholders, you can provide them inside `gettext()` as `$1`, `$2`, and `$3` and so on. 
+
+Some examples:
+
+_Not good:_
+
+  ```r
+  gettextf("Number of factor levels is %s in %s", "{{factorLevels.amount}}", "{{variables}}")
+  ```
+ 
+
+_Good:_
+	
+  ```r
+  gettextf("Number of factor levels is %1$s in %2$s", "{{factorLevels.amount}}", "{{variables}}")
+  ```
+
+**Use Unicode everywhere**
+
+Internationalization includes support for multiple character sets (usually we recommend Unicode), e.g, use `\u03B2` is better than a `β` symbol in messages. This allows your writing characters available on multiple language environments.
+
 ### Thorough testing
 
 It is important that software is thoroughly tested, and that a proactive approach (proactively thinking through what sorts of problems may occur) is taken rather than reactive (simply waiting until people complain about things not working).
--- a/Docs/development/jasp-human-guide.md
+++ b/Docs/development/jasp-human-guide.md
@@ -31,16 +31,24 @@ Error messages can either be placed over the top of the results table, or put in
 
 ### Writing for internationalization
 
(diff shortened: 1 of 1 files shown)


Docs/development/jasp-human-guide.md
```diff
@@ -29,6 +29,34 @@ Hence, in developing an analysis, it is good to think through each and every sor
 
 Error messages can either be placed over the top of the results table, or put in the footnotes. In general, an error which makes the whole analysis meaningless, should be placed over the top of the results table. A t-test where the user has specified an independent variable with three levels would be one such example. In contrast, if the error only affects one or a handful of values in the table, then an NaN should be placed in that cell, and a footnote marker added.
 
+### Writing for internationalization
+
+Internationalization(i18n) is the adapting of module to different languages. An inspiring news is that JASP is moving towards i18n, which means that the analysis and modules you developed will be translated into different languages. Therefore, it is necessary to pay attention to the readability and flexibility of the message during writing.
```
boutinb: ... which means that the analysis and modules you develop will be translated

Docs/development/jasp-human-guide.md
```diff
@@ -29,6 +29,34 @@ Hence, in developing an analysis, it is good to think through each and every sor
...
+Passing your message with [gettext()](https://en.wikipedia.org/wiki/Gettext), JASP will generate the string  to be translated automatically in a `.po` file.
```
boutinb: Passing your messages with [gettext() or gettextf()](https://en.wikipedia.org/wiki/Gettext), JASP will add automatically the strings in a `.po` file, that will be exposed in [WebLate](https://jasp-stats.org/translation-guidelines) so that people can translate them in their own language.
gettextf should be used in place of gettext when arguments are needed, or when special characters are used in the string (non latin characters like 'β', or the '%' character which is used for placeholder).

boutinb: It is important to know that the translator will just see the string (without the value of the arguments) inside gettext. So for example instead of writing:
`gettextf("File %s is %s protected", filename, rw ? "write" : "read");`
It's better to write:
`gettextf (rw ? "File %s is write protected" : "File %s is read protected", filename);`
This is more understandable for the translator.

Docs/development/jasp-human-guide.md
```diff
@@ -29,6 +29,34 @@ Hence, in developing an analysis, it is good to think through each and every sor
...
+
+**Placeholders**
+
+If you have multiple placeholders, you can provide them inside `gettext()` as `$1`, `$2`, and `$3` and so on.
```
boutinb: You may have multiple placeholders for one message: 
`gettextf("Number of factor levels is %s in %s", amount, variables)`.
The problem here is that the order of the arguments depends on the language, and that for some languages the `variables` argument will be used before the `amount` argument.
Therefore it is important to specify the argument order:
`gettextf("Number of factor levels is %1$s in %2$s", amount, variables)`
So that the translation in some language can place the `variables` argument first:
`Dans %2$s, le nombre de niveaux des facteurs est %1$s`

Docs/development/jasp-human-guide.md
```diff
@@ -29,6 +29,34 @@ Hence, in developing an analysis, it is good to think through each and every sor
...
+_Good:_
+	
+  ```r
+  gettextf("Number of factor levels is %1$s in %2$s", "{{factorLevels.amount}}", "{{variables}}")
+  ```
+
+**Use Unicode everywhere**
+
+Internationalization includes support for multiple character sets (usually we recommend Unicode), e.g, use `\u03B2` is better than a `β` symbol in messages. This allows your writing characters available on multiple language environments.
+
```
boutinb, line 88: You have then to use gettextf with a placeholder %s and argument `\u03B2`.

Another tricky point is that the `%` character is a special character in gettext: it means that it expects a placeholder. So if you just want to print a `%` character, you need to double it, you ned also to use gettextf in place of gettext:
`gettextf("%s%% CI for Mean Difference")`

boutinb (changes requested):
Hi @shun2wang, I have added some changes to make it clearer I think. What do you think?

Docs/development/jasp-human-guide.md
shun2wang: Hi Bruno,I understand your idea. But I think the readers of `jasp-human-guide.md`  are developers but not translators, do you agree? So all this documentation is for developers who develop new analysis or modules for JASP:)

EDIT: I'm sorry Bruno I misunderstood you, now I understand

Docs/development/jasp-human-guide.md
shun2wang: Yes, you are right, the different cases should indeed be differentiated for the use of gettext, however I think the documentation [here](https://github.com/jasp-stats/jasp-desktop/blob/stable/Docs/development/translate.md#4-utilities-needed-for-translations) seems to be more useful for developers (and outdated for translators), it seems we should merge the content of this documentation into Developer Guidelines...?

shun2wang: Guidelines for translators https://jasp-stats.org/translation-guidelines/,  needs to be updated directly on the website after all developer documentation and code changes done

boutinb (approved):

PR #4889, opened 2022-10-12
Add R syntax inside JASP

boutinb:
This is the first (alpha) version.
It adds a 'Generate Wrapper' button to generate the R analysis wrapper, that must be copied in the R file so that the R Syntax can work. This generate wrapper function must be automatise in a GitHub action.
This adds also a 'Add Analysis' button in the R Commander: this works only if the R code in this format with <ModuleName>::<AnalysisNameWrapper>(options)


EJWagenmakers:
:-))))))))

JorisGoosen:
@boutinb Im changing "AnalysisForm::getError()" to:
```
QString AnalysisForm::getError()
{
	QString message;

	for (QQuickItem* item : _controlErrorMessageCache)
		if (item->property("control").value<JASPControl*>() != nullptr)
			message += (message != "" ? ", " : "") + item->property("message").toString();

	return message;
}
```

If we know for sure there'd be only 1 error Id do:
```
QString AnalysisForm::getError()
{
	for (QQuickItem* item : _controlErrorMessageCache)
		if (item->property("control").value<JASPControl*>() != nullptr)
			return item->property("message").toString();

	return "";
}
``

File: QMLComponents/controls/radiobuttonsgroupbase.cpp [hunk 1]
```diff
@@ -33,7 +33,6 @@ RadioButtonsGroupBase::RadioButtonsGroupBase(QQuickItem* item)
 
 void RadioButtonsGroupBase::setUp()
 {
-	JASPControl::setUp();
```
JorisGoosen, line 36: This isnt necessary anymore?

JorisGoosen (approved):

PR #4895, opened 2022-10-18
Catch installation errors when installing dynamic module

vandenman:
Without this change, the error is not caught. There might be a better place to set this flag, but I couldn't find it.

Prerequisite for https://github.com/jasp-stats/jaspBase/pull/90

### DIFF
--- a/Desktop/modules/dynamicmodule.cpp
+++ b/Desktop/modules/dynamicmodule.cpp
@@ -436,7 +436,7 @@ std::string DynamicModule::generateModuleInstallingR(bool onlyModPkg)
 	}
 	setInstallLog("Installing module " + _name + ".\n");
 	
-	return "jaspBase::installJaspModule(modulePkg='" + _modulePackage + "', libPathsToUse=" + getLibPathsToUse() + ", moduleLibrary='" + moduleRLibrary().toStdString() + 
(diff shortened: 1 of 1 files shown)


Desktop/modules/dynamicmodule.cpp
```diff
@@ -436,7 +436,7 @@ std::string DynamicModule::generateModuleInstallingR(bool onlyModPkg)
 	}
 	setInstallLog("Installing module " + _name + ".\n");
 	
-	return "jaspBase::installJaspModule(modulePkg='" + _modulePackage + "', libPathsToUse=" + getLibPathsToUse() + ", moduleLibrary='" + moduleRLibrary().toStdString() + 
+	return "options(\"renv.config.install.verbose\" = FALSE);jaspBase::installJaspModule(modulePkg='" + _modulePackage + "', libPathsToUse=" + getLibPathsToUse() + ", moduleLibrary='" + moduleRLibrary().toStdString() +
```
JorisGoosen, line 439: Why not just inside `installJaspModule` with an argument "verboseStdout" or something?
Not that it matters too much. This would be fine too

vandenman: also works, I was just annoyed why it didn't work and it turns out this flag was set somewhere :)

JorisGoosen (approved):

PR #4899, opened 2022-10-26
Update R.cmake

FBartos:
This is the default include path on Ubuntu 22.04

### DIFF
--- a/Tools/CMake/R.cmake
+++ b/Tools/CMake/R.cmake
@@ -766,6 +766,7 @@ elseif(LINUX)
   endif()
 
   set(R_EXECUTABLE "${R_HOME_PATH}/bin/R")
+  set(Rscript_EXECUTABLE "${R_HOME_PATH}/bin/Rscript")
   set(RCPP_PATH "${R_LIBRARY_PATH}/Rcpp")
   set(RINSIDE_PATH "${R_LIBRARY_PATH}/RInside")
   set(RENV_PATH "${R_LIBRARY_PATH}/renv")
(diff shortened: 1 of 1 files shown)


vandenman:
So I modified this a bit more. Now

1. Cmake asks R where it thinks its include directory is.
2. If R returns a nonexisting directory, we try exactly what we did before (with the path from Frantisek's default location added).

This only affects Linux.

@JorisGoosen any comments?

JorisGoosen:
This looks pretty good but I am afraid it might break a linux build where you use a local "R" (aka in buildfolder) but also have an R installed to your system. I guess that could lead to a different R.h being loaded then is actually in the buildfolder?

But the code looks benign enough to be merged if it works on Dons system. (my linuxpc is rather busy running windows atm)

vandenman:
> This looks pretty good but I am afraid it might break a linux build where you use a local "R" (aka in buildfolder) but also have an R installed to your system.

So my idea was that exactly that shouldn't be possible (or at least, become more difficult). Because we ask R itself where its includes are, and that's either the R somebody specified manually or the system R (or whatever is on the path first).

JorisGoosen:
Yeah but we ask R through a `R_HOME_PATH` we already got from somewhere and it could be different from `R_HOME` envvar in theory right? If I read R.cmake corectly.
As in, if `R_HOME` points to a different place then `CUSTOM_R_PATH`. See [line 735](https://github.com/FBartos/jasp-desktop/blob/7aaada71084192792e1071978d71ce94e1091469/Tools/CMake/R.cmake#L735)

vandenman:
I'm not really understanding the problem.

Basically, CMake does this:

If `${CUSTOM_R_PATH}` is not set, then
- `set(R_HOME_PATH $ENV{R_HOME})`

Else
- `set(R_HOME_PATH ${CUSTOM_R_PATH})`

End

Next, using `R_HOME_PATH` we determine the R executable, Rscript, etc. that are used throughout the remainder of Cmake steps.
This means it should be fine to ask R (through Rscript) where it thinks its include directories are because that must be the same R that is used in other steps (e.g., to install Rcpp and RInside).

I'm not entirely sure how/ what `$ENV{R_HOME}` contains (for me it's empty) but I assume this would be equal to the R_HOME of the system R.

JorisGoosen:
Well the problem only occurs if:
`$ENV{R_HOME} == "LocationA"' and `${CUSTOM_R_PATH} == "LocationB"'

Because then we set `R_HOME_PATH==CUSTOM_R_PATH` but when R starts `R_HOME!=CUSTOM_R_PATH` right?

Aaaaaah, but we of course set in the R scripts again the `JASP_R_HOME=R_HOME_PATH` and that sets `R_HOME`.

Yeah ok so there is no problem. Nvm!

JorisGoosen:
So should we merge this or?

vandenman:
Preferably yes. Since this only affects building Linux anyway, we'll find out pretty quick if this breaks something but also do not risk much.

JorisGoosen (approved):

PR #4900, opened 2022-10-31
Scroll along whith focus changes due to tab/backtab

RensDofferhoff:
Uses a filter on _innercontrol to catch the focus reason so that we can discern between mouse click and tabs.
If you really dislike it I can revert to the simpler version.

adds focus reason property to JASPControls
adds activeJASPControl property to analysis forms.
Implements a little logic for scrolling in to view.

### DIFF
--- a/Desktop/components/JASP/Widgets/AnalysisFormExpander.qml
+++ b/Desktop/components/JASP/Widgets/AnalysisFormExpander.qml
@@ -569,20 +569,25 @@ DropArea
 					Connections
 					{
 						target: myForm
-						function onActiveItemChanged()
+						function onActiveJASPControlChanged()
 						{
-							if (!myForm || !myForm.activeItem)
+							if (!myForm || !myForm.activeJASPControl)
 								return;
 
-							const control = myForm.activeItem;
-							const coordinates = myForm.activeItem.mapToItem(scrollAnalyses, 0, 0);
-							const diffYBottom = coordinates.y + myForm.activeItem.height - scrollAnalyses.height;
-							const diffYTop = coordinates.y;
-							//check if the object is visisble in the scrollAnalyses and scroll to it if not
-							if (diffYBottom > 0 && !contentYBehaviour.animation.running)
-								backgroundFlickable.contentY = backgroundFlickable.contentY + diffYBottom;
-							if (diffYTop < 0 && !contentYBehaviour.animation.running)
-								backgroundFlickable.contentY = Math.max(0, backgroundFlickable.contentY + diffYTop);
+							const control = myForm.activeJASPControl;
+							if (control.focusReason === Qt.BacktabFocusReason || control.focusReason === Qt.TabFocusReason)
+							{
+								const coordinates = control.mapToItem(scrollAnalyses, 0, 0);
+								const diffYBottom = coordinates.y + control.height - scrollAnalyses.height;
+								const diffYTop = coordinates.y;
+								const margin = 50 * jaspTheme.uiScale;
+
+								//check if the object is visisble in the scrollAnalyses and scroll to it if not
+								if (diffYBottom > -margin && !contentYBehaviour.animation.running)
+									backgroundFlickable.contentY = backgroundFlickable.contentY + Math.max(0, diffYBottom + margin);
+								if (diffYTop < margin && !contentYBehaviour.animation.running)
+									backgroundFlickable.contentY = Math.max(0, backgroundFlickable.contentY + Math.min(0, diffYTop - margin));
+							}
 						}
 					}
 				}
--- a/Desktop/components/JASP/Widgets/AnalysisFormExpander.qml
+++ b/Desktop/components/JASP/Widgets/AnalysisFormExpander.qml
@@ -578,14 +578,14 @@ DropArea
 							if (control.focusReason === Qt.BacktabFocusReason || control.focusReason === Qt.TabFocusReason)
 							{
 								const coordinates = control.mapToItem(scrollAnalyses, 0, 0);
-								const diffYBottom = coordinates.y + control.height - scrollAnalyses.height;
-								const diffYTop = coordinates.y;
+								const diffYBottom = coordinates.y + Math.min(control.height, scrollAnalyses.height) - scrollAnalyses.height; //positive if not visible
+								const diffYTop = coordinates.y; //negative if not visible
 								const margin = 50 * jaspTheme.uiScale;
 
-								//check if the object is visisble in the scrollAnalyses and scroll to it if not
-								if (diffYBottom > -margin && !contentYBehaviour.animation.running)
+								//check if the object is visisble in the scrollAnalyses (with margin) and scroll to it if not
+								if (diffYBottom > -margin && !contentYBehaviour.animation.running) // scroll up
 									backgroundFlickable.contentY = backgroundFlickable.contentY + Math.max(0, diffYBottom + margin);
-								if (diffYTop < margin && !contentYBehaviour.animation.running)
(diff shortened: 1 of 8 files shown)


Desktop/components/JASP/Widgets/AnalysisFormExpander.qml
```diff
@@ -530,6 +531,31 @@ DropArea
...
+								const coordinates = control.mapToItem(scrollAnalyses, 0, 0);
+								const diffYBottom = coordinates.y + Math.min(control.height, scrollAnalyses.height) - scrollAnalyses.height; //positive if not visible
+								const diffYTop = coordinates.y; //negative if not visible
+								const margin = 50 * jaspTheme.uiScale;
+
+								//check if the object is visisble in the scrollAnalyses (with margin) and scroll to it if not
+								if (diffYBottom > -margin && !contentYBehaviour.animation.running) // scroll up
+									backgroundFlickable.contentY = backgroundFlickable.contentY + Math.max(0, diffYBottom + margin);
+								else if (diffYTop < margin && !contentYBehaviour.animation.running) //scroll down
+									backgroundFlickable.contentY = Math.max(0, backgroundFlickable.contentY + Math.min(0, diffYTop - margin));
```
JorisGoosen:
```suggestion
								//check if the object is visisble in the scrollAnalyses (with margin) and scroll to it if not
								if(!contentYBehaviour.animation.running)
									return;
								
								if (diffYBottom > -margin) // scroll up
									backgroundFlickable.contentY = backgroundFlickable.contentY + Math.max(0, diffYBottom + margin);
								else if (diffYTop < margin) //scroll down
									backgroundFlickable.contentY = Math.max(0, backgroundFlickable.contentY + Math.min(0, diffYTop - margin));
```

JorisGoosen (changes requested):

JorisGoosen:
If I build this code tabbing only works within the variableslists-component. Not in the form...
This is on macos m1

RensDofferhoff:
> If I build this code tabbing only works within the variableslists-component. Not in the form... This is on macos m1

Yes, this is a known (unrelated) issue. It has not worked properly for some time on mac.
Accessibly and tab behavior is weird on @boutinb  Mac as well.
I will set my mac up on Wednesday and dive in.

JorisGoosen:
Aah yes on windows this all works a lot better indeed.

But there is still a bug though, in for instance descriptives:
- tabbing works fine, also out-of-variablelists
- back-tabbing works fine in most places but gets stuck in variableslist. In descriptives in '[variable, density-plots, heatmaps]`

RensDofferhoff:
> But there is still a bug though, in for instance descriptives:

ah yes I see what is happening.
I can put the fix in this PR if you want or make another.

JorisGoosen:
Putting the fix in this pr would be best

RensDofferhoff:
The requested fix has been implemented.

QMLComponents/components/JASP/Controls/VariablesForm.qml
```diff
@@ -33,7 +33,15 @@ VariablesFormBase
...
 	Layout.preferredHeight		: preferredHeight // Cannot set Layout attached property in c++
-	focusOnTab					: false
+
+	onActiveFocusChanged		:
+	{
+		if (activeFocus) {
+			availableVariablesList.forceActiveFocus();
+			//the first button knows the way out...
+			availableVariablesList.KeyNavigation.backtab = assignButtonRepeater.itemAt(0).nextItemInFocusChain(false);
+		}
```
boutinb, line 43: I'm not sure to understand this code... How can the first assign button know which control should get the focus after a backtab on the Available VariablesList?

QMLComponents/controls/jaspcontrol.cpp
```diff
@@ -248,6 +255,13 @@ void JASPControl::componentComplete()
 
 	if (_form)
 		connect(this, &JASPControl::boundValueChanged, _form, &AnalysisForm::boundValueChangedHandler);
+
+	//capture focus reason
+	for (QQuickItem* childItem : childItems())
+	{
+		if (!qobject_cast<JASPControl*>(childItem))
+			childItem->installEventFilter(this);
+	}
```
boutinb: Why all/only child items that are not JASPControl should get this event filter. Should it be only the inner control? Maybe the right control that receives the focus is not a direct child?

QMLComponents/controls/jaspcontrol.cpp
```diff
@@ -633,9 +674,22 @@ void JASPControl::rScriptDoneHandler(const QString &)
...
+	{
+		if (!hasActiveFocus())
+		{
+			_form->setActiveJASPControl(nullptr);
+			_activeJASPControl = false;
+		}
+		if(_activeJASPControl)
+			_form->setActiveJASPControl(this);
+	}
+}
```
boutinb, line 683: I see a concurrency problem here: If the new active JASP control runs first this _notifyFormOfActiveFocus method, and then the JASP control that loses the active focus, then the form will lose the active JASP control.
I see that you have solved this problem by adding the line:
`if(!control && _activeJASPControl && _activeJASPControl->hasActiveFocus())` in setActiveJASPControl...
What I'd rather do is:
```
setActiveJASPControl(JASPControl* control, hasActiveFocus)
{
  bool emitSignal = false;
  if (hasActiveFocus)
  {
     if (_activeJASPControl != control) emitSignal = true;
     _activeJASPControl = control
  }
  else if (control == _activeJASPControl)
  {
     if (_activeJASPControl) emitSignal = true;
     _activeJASPControl = nullptr;
  }

  if (emitSignal)
     emit activeJASPControlChanged();
}
```
Also I would rename the JASPControl `_activeJASPControl` to `_hasActiveFocus`  because it is quite confusing with the form `_activeJASPControl`

boutinb (changes requested):

QMLComponents/components/JASP/Controls/VariablesForm.qml
RensDofferhoff: made some changes to clarify

boutinb (approved):

JorisGoosen (changes requested):
I just rebased this on `stable` and I cannot at all navigate through analyses by tab-key...

The scrolling seems to work alright but for the rest it is kaput.
Not sure why though...

RensDofferhoff:
> I just rebased this on `stable` and I cannot at all navigate through analyses by tab-key...
> 
> The scrolling seems to work alright but for the rest it is kaput. Not sure why though...

I rebased it myself and cannot reproduce this issue on Mac or Linux

JorisGoosen:
Ill build on windows and will see

JorisGoosen:
well this is what it looks like on my macbook: 


https://user-images.githubusercontent.com/29062713/204289079-ea79b9ec-78ad-4b54-a4e6-0c642797417b.mov


I can bring it along wednesday so you can have a look in debugmode what is up

JorisGoosen:
Or maybe its just because of the RM level/factors thing?

JorisGoosen:
Turns out it is more general:

https://user-images.githubusercontent.com/29062713/204299640-c658b8f5-2c4a-4629-b504-e346ae6636d5.mov

RensDofferhoff:
> Turns out it is more general:
> Screen.Recording.2022-11-28.at.15.08.21.mov

Bizar! I cannot reproduce this on any of my 3 systems.
Just for reference which Qt version is that ?

JorisGoosen:
6.3.1

Ive apparently already installed 6.4.1 so ill build with that too

JorisGoosen (approved):

PR #4902, opened 2022-10-31
Accesiblity: Focus analysis expanders

RensDofferhoff:
Analysis expanders now show focus and the internal buttons are reachable by keyboard.
Expanders now show&get focus on click and drag

### DIFF
--- a/Desktop/components/JASP/Widgets/AnalysisFormExpander.qml
+++ b/Desktop/components/JASP/Widgets/AnalysisFormExpander.qml
@@ -59,6 +59,8 @@ DropArea
 		Drag.hotSpot.x:		width/2
 		Drag.hotSpot.y:		height/2
 
+		Component.onCompleted:	{ forceActiveFocus(); }
+
 		states:
 		[
 			State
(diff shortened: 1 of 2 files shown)


Desktop/components/JASP/Widgets/AnalysisFormExpander.qml
```diff
@@ -77,6 +79,15 @@ DropArea
...
 				}
+
+				PropertyChanges
+				{
+					restoreEntryValues: false
+					draggableItem
+					{
+						focus:			true
+					}
+				}
```
boutinb, line 92: I don't think you need to specify draggableItem here since the PropertyChanges is for a state of the draggableItem.
```
				PropertyChanges
				{
					restoreEntryValues: false
				    focus:			true
				}
```

boutinb (changes requested):

Desktop/components/JASP/Widgets/AnalysisFormExpander.qml
RensDofferhoff: It seems necessary otherwise the error:
"QML PropertyChanges: Cannot assign to non-existent property "focus""
Is generated.
It doesn't hurt for the sake of clarity either.

boutinb (approved):

JorisGoosen (approved):

PR #4909, opened 2022-11-07
A validation error must be removed when its constraints change dynamically

boutinb:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2158
Bug mentioned also in https://github.com/jasp-stats/jaspLearnBayes/pull/119

When a min or max depends on other controls, the (default) value of Double/Integer field might be temporary wrong as long as these other controls are not set. When these min or max are set, if the value becomes correct, the error must disappear.


JorisGoosen (approved):

FBartos (approved):
The error is no longer there

PR #4910, opened 2022-11-07
Allow Allowed & Suggested columns to be changed dynamically

boutinb:
This is a request from @fqixiang 

The _setAllowedVariables function must be connected to the allowedColumnsChanged & suggestedColumnsChanged signals.
Moreover _setAllowedVariables should not mix up the _allowColumns & _suggestedColumns property: these properties should not be set in this function, only _variableTypesAllowed & _suggestedColumnsIcons should be set: _variableTypesAllowed is used to check which column is really allowed, and _suggestedColumnsIcons is used to set the icons in the VariablesList.

### DIFF
--- a/QMLComponents/controls/variableslistbase.cpp
+++ b/QMLComponents/controls/variableslistbase.cpp
@@ -85,6 +85,8 @@ void VariablesListBase::setUp()
 	//We use macros here because the signals come from QML
 	QQuickItem::connect(this, SIGNAL(itemDoubleClicked(int)),						this, SLOT(itemDoubleClickedHandler(int)));
 	QQuickItem::connect(this, SIGNAL(itemsDropped(QVariant, QVariant, int)),		this, SLOT(itemsDroppedHandler(QVariant, QVariant, int)));
+	connect(this,	&VariablesListBase::allowedColumnsChanged,						this, &VariablesListBase::_setAllowedVariables);
+	connect(this,	&VariablesListBase::suggestedColumnsChanged,					this, &VariablesListBase::_setAllowedVariables);
 }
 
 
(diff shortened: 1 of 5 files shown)


QMLComponents/controls/variableslistbase.cpp
```diff
@@ -338,59 +340,63 @@ void VariablesListBase::termsChangedHandler()
...
-		allowedColumnsTypes = 0;
-		for (const QString& allowedColumn: allowedColumns())
+		columnsTypes = 0;
+		for (const QString& typeStr: types)
 		{
-			columnType allowedType = columnTypeFromString(fq(allowedColumn), columnType::unknown);
-			if (allowedType != columnType::unknown)
-				allowedColumnsTypes |= int(allowedType);
+			columnType type = columnTypeFromString(fq(typeStr), columnType::unknown);
+			if (type != columnType::unknown)
```
JorisGoosen: This check is kind of useless as `unknown` == 0 anyway. Although I guess it does make sense for the error/warning underneath.

JorisGoosen:
Did you fix this?

boutinb:
> Did you fix this?

It was working... I think you made a mistake when you tested it.

JorisGoosen:
If you set only suggested on a assigned form and no allowed (or just ordinal anyway) you get something that looks like:
image

But then you can't actually drop anything there unless it is allowed as well...
Of course it is kind of silly to set both but in this case I would want to see only the join of `allowed & suggested ` instead of `allowed | suggested' as this looks confusing.

JorisGoosen:
Also weird: if nothing suggested and nothing allowed you cannot drop anything in an assigned field.
If only scale as suggested (for instance) you can then drop everything....

Edit:
In fact this only happens if you first select as "suggested"  "sacle + ordinal" and then remove first ordinal and then scale.
The variables disappear and you cant add anything anymore.

If I then turn scale back on I can add everything....

JorisGoosen:
also nominal text icon is wrong in darkmode:
image

JorisGoosen:
With a single "allowed type" that gets removed the assigned form remembers it. Not really a problem but ugly nonetheless.

JorisGoosen (changes requested):
While it *does* indeed update dynamically now there are several issues Ive described and would like to see fixed or at least handled.

boutinb:
> Of course it is kind of silly to set both but in this case I would want to see only the join of `allowed & suggested ` instead of `allowed | suggested' as this looks confusing.

I don't want to change the current behaviour.
allowedColumns and suggestedColumns should normally not be used at the same time (maybe only for really specific goal, when you want for example to suggest 'nominal', and to have only this icon displayed, but allowing implicitly only 'ordinal', and not 'nominalText'. This can be achieved by using suggestedColumns set to 'nominal' and allowedColumns to 'nominal' and 'ordinal'. I don't know if this case will be ever used). 
I have updated the doc to describe more precisely how allowedColumns and suggestedColumns work and how they should be used.

boutinb:
> Edit:
> In fact this only happens if you first select as "suggested" "sacle + ordinal" and then remove first ordinal and then scale.
> The variables disappear and you cant add anything anymore.

This is now solved.

boutinb:
> With a single "allowed type" that gets removed the assigned form remembers it. Not really a problem but ugly nonetheless.

This is also solved.

boutinb:
> also nominal text icon is wrong in darkmode:
> image
I get this:
Screenshot 2022-11-24 at 16 17 48

JorisGoosen (approved):

PR #4911, opened 2022-11-07
Add support for GITHUB_PAT_DEF

JorisGoosen:
This was originally a feature of JASP, using a default GITHUB_PAT but the cmake switch lost control of that.

Here it is returned!

Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2074 (Default GITHUB_PAT totally broken)

### DIFF
--- a/Tools/CMake/Config.cmake
+++ b/Tools/CMake/Config.cmake
@@ -237,28 +237,37 @@ endif()
 # I will consider turning this off, and letting Qt does it
 # when everything else worked properly
 
-set(GITHUB_PAT
-    ""
-    CACHE STRING "GitHub Personal Access Token")
+set(GITHUB_PAT     "" CACHE STRING "GitHub Personal Access Token to use during building")
+set(GITHUB_PAT_DEF "" CACHE STRING "GitHub Personal Access Token to use in released version as default")
 
-if(INSTALL_R_MODULES)
 
-  message(CHECK_START "Looking if its set as an environment variable.")
-  set(GITHUB_PAT $ENV{GITHUB_PAT})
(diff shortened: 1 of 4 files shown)


Tools/CMake/Config.cmake
```diff
@@ -237,28 +237,37 @@ endif()
...
+message(CHECK_START "Looking if GITHUB_PAT_DEF is set as an environment variable.")
+set(GITHUB_PAT_DEF      $ENV{GITHUB_PAT_DEF})
+
+if(GITHUB_PAT_DEF STREQUAL "")
+  message(CHECK_FAIL "not found")
+  set(GITHUB_PAT_DEF ${GITHUB_PAT})
+  message(
+    WARNING
+      "Your GITHUB_PAT is used as the default PAT for any JASP build with this config, if this is inteded as released software you will want to set GITHUB_PAT_DEF to something else than your personal PAT!"
+  )
```
boutinb, line 265: I'm afraid this warning won't be read...
Maybe in Release mode, this warning can become an error, to be sure that a persona GITHUB_PAT won't be used in a released version. What do you think?

boutinb (changes requested):

Tools/CMake/Config.cmake
JorisGoosen: That could be a good idea but I didnt want to annoy Koen, Jonas etc who are building on windows (and probably in release as it goes faster). (They could of course set their own GITHUB_PAT in it then, but the code already does that now)

I figured you, me and Rens could be counted on remembering to set this up?
And also, the systems that will come back online to build JASP at some point would also have these settings.
I want to get away from this manual building sooner rather than later.

boutinb (approved):

PR #4914, opened 2022-11-10
Remove unneeded dependencies in QMLComponents and Common libraries

boutinb:
For the R Syntax project, we need a QMLComponents library that depends to a minimal set of extra libraries, so that we can easily link it to the jaspBase.
As QMLComponents depends on Common, and Common has dependencies with Boost and libArchive libraries, the Common library is split into 2 libraries: CommonData & Common. In CommonData there are all classes that need Boosts & LibArchive library: these classes are essentially classes that deals with the DataSet (with the Boost Shared Memory), and the class that reads and write JASP file (archivereader). That's why it's called CommonData (but any idea for a better name is welcome).
CommonData is needed only in Desktop. Common is then needed in Desktop, Engine & QMLComponents, so the Common library should not depend on Boosts or archive libraries (for QMLComponents) and also not on Qt libraries (Engine is not linked with Qt library).
Common and QMLComponents will be later in 2 apart submodules, so that QMLComponents can be compiled and generate a static library outside of JASP.
The jaspColumnEncoder does not need anymore to be an apart submodule. The classes can be added to Common, jaspBase can use then Common instead of jaspColumnEncoder. The jaspColumnEncoder should not be updated anymore. The files come from this commit (https://github.com/jasp-stats/jaspColumnEncoder/commit/c63c0203d728ecc1bb88165c4003e29aa17faf5f) and no other commit should be added.


Common/log.cpp
```diff
@@ -9,12 +9,12 @@
...
 #include "utils.h"
-#include <boost/iostreams/stream.hpp>
-#include <boost/iostreams/device/null.hpp>
 #include <codecvt>
 #include <fstream>
 
-static std::ofstream _logFile;// = bofstream();
+std::ofstream Log::_logFile;// = bofstream();
+
+std::ostream* Log::_nullStream = &std::cout;
```
JorisGoosen, line 17: the point of the nullstream is to not have buffers constantly be in use...

If you then point everything to stdout... What is the difference between release and debug?

JorisGoosen (approved):

Common/log.cpp
boutinb: It's just the default. Desktop and Engine set up the Log class by setting the _nullstream with the boost null_sink.
This is done to avoid Common to include the boost library.

JorisGoosen: okidoki

PR #4920, opened 2022-11-15
update R version to 4.2.2

vandenman:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2146

merge https://github.com/jasp-stats/jasp-actions/pull/17 at the same time

the new versions are uploaded at https://static.jasp-stats.org/development/{version}


JorisGoosen:
Hmm, the PR seems right but the R_VERSION is cached and therefore qt keeps downloading R.4.2.1 instead of the newer one.
After changing the cmake-var manually, throwing away the Frameworks folder and retrying I get the new one.
But Ill see if I can just turn off the caching

JorisGoosen (approved):
now it doesnt cache the version anymore, although to use the new version of R the framework ought to be thrown away anyway

PR #4921, opened 2022-11-15
Jasp report

JorisGoosen:
JASP reporting adds a special jaspObject in https://github.com/jasp-stats/jaspBase/pull/100 that can be used to give feedback to a user either through the interface or JASP or by making a webpage of the results available in a directory if the user so chooses. This can then be displayed anywhere by [spinning up a simple webserver](https://www.redhat.com/sysadmin/simple-http-server) and opening the required ports etc.

JASP is also able to collect these reports from the entire resultswindow and write them to files that can then be easily served through a webserver. All of this functionality is hidden/disabled if reporting-mode is off. And reportingmode can only be enabled/disabled when running in development mode. (for now anyway, as this is really meant for users atm)


Desktop/html/index-jasp.html
```diff
@@ -0,0 +1,57 @@
+<!DOCTYPE html>
```
JorisGoosen, line 1: this one is now the one used *inside* jasp

Desktop/html/index.html
```diff
@@ -8,48 +8,57 @@
 
     <link rel="stylesheet" href="css/jquery-ui-1.10.1.custom.css"			type="text/css" />
```
JorisGoosen, line 9: and this *outside*

Desktop/mainwindow.cpp
```diff
@@ -1086,15 +1113,12 @@ void MainWindow::dataSetIOCompleted(FileEvent *event)
...
-				_engineSync->resumeEngines();
-
-				//Also give it like 3secs to have the ribbon load
+			{				
+				//Give it like 3secs to have the ribbon load and the engines to load the data
 				QTimer::singleShot(3000, this, &MainWindow::startComparingResults);
 			}
-
+			else if(_reporter && !_reporter->isJaspFileNotDabaseOrSynching())
+					exit(12);
```
boutinb, line 1121: What is this 12 number?

boutinb (approved):

PR #4922, opened 2022-11-16
set TableView dimensions from QML

vandenman:


QMLComponents/models/listmodeltableviewbase.cpp
```diff
@@ -180,6 +180,66 @@ void ListModelTableViewBase::removeRow(size_t row, bool emitStuff)
...
+				_tableTerms.values.push_back(values);
+			}
+		}
+	}
+
+	if (emitStuff)
+	{
+		endResetModel();
+
+		emit columnCountChanged();
```
JorisGoosen: maybe only emit it when changed?

vandenman: done

shun2wang:
is this related to [tableview](https://github.com/jasp-stats/jasp-desktop/blob/stable/Docs/development/jasp-qml-guide.md#tableview)?

@vandenman Could you please also add these componets or arguments into QML [developer guide](https://github.com/jasp-stats/jasp-desktop/blob/stable/Docs/development/jasp-qml-guide.md)?

Looks like this is what I was looking for, thanks.

JorisGoosen:
@vandenman how did you test this?
Something in the testmodule perhaps?

vandenman:
I tested it with the skf module, but that isn't online anywhere. I can push that, or should I copy-ish the tableview I used to the testmodule?

vandenman:
@shun2wang I just added some documentation, but the source code is unfortunately the best documentation we have right now. There are a bunch of other things possible with table views that I didn't go into, I only discussed the things I changed in this PR.

boutinb (approved):

JorisGoosen (approved):

PR #4923, opened 2022-11-16
Use C++20 and remove warnings

boutinb:


boutinb:
I have now no warning on Mac.

Desktop/data/importers/readstat/readstatimportcolumn.cpp
```diff
@@ -63,6 +63,8 @@ bool ReadStatImportColumn::canConvertToType(columnType newType)
 	case columnType::scale:
 		switch(newType)
 		{
+		case columnType::scale:			[[fallthrough]];
```
JorisGoosen, line 66: Before you added this "unknown" would have not been a 'return true' but a 'return false' at line 130

JorisGoosen: and same for scale but that wont ever be called.

I dont think it matters too much but changing the behaviour of the code like this can have unforeseen consequences...

JorisGoosen (approved):

PR #4924, opened 2022-11-17
Make sure Rcode scripts are run on engines that load the required module first

JorisGoosen:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2164 (crash/bug after engine unloads module)

### DIFF
--- a/Desktop/engine/enginesync.cpp
+++ b/Desktop/engine/enginesync.cpp
@@ -519,6 +519,14 @@ void EngineSync::processReloadData()
 }
 
 
+/**
+Checks if the top scriptstruct of the _waitingScripts queue is an rcode script that needs to run. Each of these belongs to a specific module, so it needs to find that engine or start it.
+
+    If there is no engine that is already coupled with the module, then it tries to find an idle engine that is not coupled with any module. If such an engine exists, it loads the module: it cannot run the script yet, because the loading of the module can take time. As this method is called every 50 milliseconds, the script will be run automatically when the module is loaded.
+    If an engine exists with the right module and is idle, it runs the script.
+
+If an engine is found that can run the script (immediately, or later when the module is loaded), then this method returns an empty set, else it returns a set with only the module name. This is done in order to facilitate the work of the process() function so that it can aggregate the modules that could not be handled, and starts maybe new engines.
+**/
 stringset EngineSync::processRScriptQueue()
 {
 	bool	foundEngine		= false, 
--- a/Desktop/engine/enginesync.cpp
+++ b/Desktop/engine/enginesync.cpp
@@ -380,14 +380,19 @@ void EngineSync::process()
 	if(_engines.size() == 0)
 		startExtraEngines();
 	
-	bool		notEnoughIdlesForScript		=	processScriptQueue();
+	stringset	notEnoughIdlesForScript		=	processRScriptQueue();
+	bool		notEnoughIdlesForCompCol	=	processComputedColumnQueue();
 	stringset	notEnoughIdlesForModule		=	processDynamicModules();
 	auto		notEnoughIdlesForAnalysis	=	processAnalysisRequests();
-	bool		notEnoughIdles				=	notEnoughIdlesForScript || notEnoughIdlesForModule.size() || notEnoughIdlesForAnalysis.size();
-	int			wantThisManyEngines			=	notEnoughIdlesForModule.size();
+	bool		notEnoughIdles				=	notEnoughIdlesForCompCol || notEnoughIdlesForScript.size() || notEnoughIdlesForModule.size() || notEnoughIdlesForAnalysis.size();
+	
+	stringset notEnoughIdlesSet(notEnoughIdlesForModule);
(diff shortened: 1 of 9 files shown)


Desktop/engine/enginesync.cpp
```diff
@@ -514,26 +519,89 @@ void EngineSync::processReloadData()
...
+					foundEngine = true;
+					if(_moduleEngines[mod]->idle())		_moduleEngines[mod]->runScriptOnProcess(waiting);
+					else								engineNotIdle = true;
+				}
+			
+				
+				if(!foundEngine)
+					return { mod };
+				
+				else if(!engineNotIdle)
```
boutinb, line 569: It would be better to call `engineNotIdle` `engineBusy`. This not not might be confusing.

Desktop/engine/enginesync.cpp
```diff
@@ -514,26 +519,89 @@ void EngineSync::processReloadData()
...
+				{
+					for(auto & engine : _engines)
+						if(engine->idle() && engine->module() == "")
+						{
+							registerEngineForModule(engine, mod);	
+							foundEngine		= true;
+							engineNotIdle	= true; //it still needs to load the module
+							
+							if(!engine->moduleLoaded() && !engine->moduleLoading())
+								engine->moduleLoad();
```
boutinb, line 555: Why the registerEngineForModule does not care that the moduleLoad function is called?

JorisGoosen:
You can test whether this works by using the engines-window to kill engines and make sure they have nothing loaded. (That is, if this fix isnt in there ;) )

boutinb (changes requested):

Desktop/engine/enginesync.cpp
```diff
@@ -514,26 +519,89 @@ void EngineSync::processReloadData()
 }
```
boutinb, line 521: Would be nice to have a comment for this method:
This methods tries to run the first R Script of the _waitingScripts queue (if it is not a computed column script).An R Script is associated with a module.

- If there is no engine that is already coupled with the module, then it tries to find an idle engine that is not coupled with any module. If such an engine exists, it loads the module: it cannot run the script yet, because the loading of the module can take time. As this method is called every 50 milliseconds, the script will be run automatically when the module is loaded.
- If an engine exists with the right module and is idle, it runs the script.

If an engine is found that can run the script (immediately, or later when the module is loaded), then this method returns an empty set, else it returns a set with only the module name. This is done in order to facilitate the work of the process() function so that it can aggregate the modules that could not be handled, and starts maybe new engines.

Desktop/engine/enginesync.cpp
```diff
@@ -514,26 +519,89 @@ void EngineSync::processReloadData()
...
 
 
-bool EngineSync::processScriptQueue()
+stringset EngineSync::processRScriptQueue()
```
boutinb, line 530: Would be nice to have a comment for this method:
This methods tries to run the first R Script of the _waitingScripts queue (if it is not a computed column script).An R Script is associated with a module.

- If there is no engine that is already coupled with the module, then it tries to find an idle engine that is not coupled with any module. If such an engine exists, it loads the module: it cannot run the script yet, because the loading of the module can take time. As this method is called every 50 milliseconds, the script will be run automatically when the module is loaded.
- If an engine exists with the right module and is idle, it runs the script.

If an engine is found that can run the script (immediately, or later when the module is loaded), then this method returns an empty set, else it returns a set with only the module name. This is done in order to facilitate the work of the process() function so that it can aggregate the modules that could not be handled, and starts maybe new engines.

Desktop/engine/enginesync.cpp
JorisGoosen: because it just "registers" that this engine and module belong together, but that could also be used for installing a module. well, how would you load a module you've yet to install? ;)

Desktop/engine/enginesync.cpp
JorisGoosen, line 530:
```suggestion
/**
Checks if the top scriptstruct of the _waitingScripts queue is an rcode script that needs to run. Each of these belongs to a specific module, so it needs to find that engine or start it.

    If there is no engine that is already coupled with the module, then it tries to find an idle engine that is not coupled with any module. If such an engine exists, it loads the module: it cannot run the script yet, because the loading of the module can take time. As this method is called every 50 milliseconds, the script will be run automatically when the module is loaded.
    If an engine exists with the right module and is idle, it runs the script.

If an engine is found that can run the script (immediately, or later when the module is loaded), then this method returns an empty set, else it returns a set with only the module name. This is done in order to facilitate the work of the process() function so that it can aggregate the modules that could not be handled, and starts maybe new engines.
**/
stringset EngineSync::processRScriptQueue()
```

Of course this is all quite similar to what happens in the other `EngineSync::process*` functions but seeing as how the behaviour of these functions could still radically change (like in this PR) Id hold off on writing too many large comments for now...

JorisGoosen: the above suggestion could now be applied easily by choosing "Commit suggestion" if you'd like

PR #4925, opened 2022-11-22
Rscript on macos fixes

JorisGoosen:
Rscript can be called during installing of pkgs, such as jfa. On macos however the Rscript executable has a builtin R_HOME and ignores whatever was set in the environment To be sure we use *our* Rscript we put the build-R first in PATH And because Rscript looks at RHOME envvar instead of R_HOME for some reason we have to set that as well...


JorisGoosen:
You can test this locally by:
- Removing `/Library/Frameworks/R.framework` (or renaming whatever)
- Cleaning up buildfolder, throw away `Modules`, `_cache` at least
- try to build `stable`, jaspAudit will fail because jfa is not starting
- checkout this branch, build again and be amazed how well it works now ;)

PR #4928, opened 2022-11-28
Fixed overflow error & made left/right available while custom menu open

RensDofferhoff:
Fixed overflow error on Ribbon button enter & made left/right available while custom menu open.

### DIFF
--- a/Desktop/components/JASP/Widgets/CustomMenu.qml
+++ b/Desktop/components/JASP/Widgets/CustomMenu.qml
@@ -49,39 +49,41 @@ FocusScope
 
 	Keys.onPressed: (event)=>
 	{
-		if (event.key === Qt.Key_Up || event.key === Qt.Key_Backtab)
+		switch(event.key)
 		{
+		case Qt.Key_Up:
+		case Qt.Key_Backtab:
 			navigate(-1);
 			event.accepted = true;
-		}
-		else if (event.key === Qt.Key_Down || event.key === Qt.Key_Tab)
-		{
+			break;
+		case Qt.Key_Down:
+		case Qt.Key_Tab:
 			navigate(1);
 			event.accepted = true;
-		}
-		else if (event.key === Qt.Key_Left)
-		{
+			break;
+		case Qt.Key_Left:
 			parentNavigate(-1);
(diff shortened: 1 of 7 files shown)


Desktop/components/JASP/Widgets/CustomMenu.qml
```diff
@@ -45,22 +45,28 @@ FocusScope
 
 	property int	currentIndex: -1
 
+	onSourceItemChanged: { menu.currentIndex = -1; }
+
 	Keys.onPressed: (event)=>
 	{
 		if (event.key === Qt.Key_Up || event.key === Qt.Key_Backtab)
```
JorisGoosen: Can you make this a switch?

RensDofferhoff: Sure

JorisGoosen (approved):
Nice, feels very natural!

Too bad the filemenu/modulesmenu dont open when going there from an opened custommenu+ribbonbutton. But that really is more of a nice to have anyway.

RensDofferhoff:
> Nice, feels very natural!
> 
> Too bad the filemenu/modulesmenu dont open when going there from an opened custommenu+ribbonbutton. But that really is more of a nice to have anyway.

I can make it so they always open when they get activeFocus. But tab walks you into the custom menu, so only we would need to turn that off in that case.  What is best is subjective I suppose.

JorisGoosen:
> I make it so they always open when they get activeFocus. But tab walks you into the custom menu, so only we would need to turn that off in that case. What is best is subjective I suppose.

Hmm, Ill have to see again to understand exactly what you mean. Ill see what happens once ive got a jasp with qt 6.4.1 up and running

RensDofferhoff:
> > I make it so they always open when they get activeFocus. But tab walks you into the custom menu, so only we would need to turn that off in that case. What is best is subjective I suppose.
> 
> Hmm, Ill have to see again to understand exactly what you mean. Ill see what happens once ive got a jasp with qt 6.4.1 up and running

We can discuss it Wednesday. As you said its nice to haves.

boutinb (approved):

PR #4933, opened 2022-12-02
Add version to R Syntax

boutinb:
Also 'data' and 'version' should be forbidden as option names.
Better handling of the option name, so that they don't conflict with each other, a formula name, or a reserved word.

### DIFF
--- a/Desktop/analysis/analysis.cpp
+++ b/Desktop/analysis/analysis.cpp
@@ -39,7 +39,7 @@ Analysis::Analysis(size_t id, Modules::AnalysisEntry * analysisEntry, std::strin
 	  _moduleData(		analysisEntry),
 	  _dynamicModule(	_moduleData->dynamicModule())
 {
-	if(moduleVersion == "" && _dynamicModule)
+	if(_moduleVersion.isEmpty() && _dynamicModule)
 		_moduleVersion = _dynamicModule->version();
 
 	if (data)
--- a/Desktop/analysis/analysis.cpp
+++ b/Desktop/analysis/analysis.cpp
@@ -30,17 +30,16 @@
 #include "utilities/reporter.h"
 
 Analysis::Analysis(size_t id, Modules::AnalysisEntry * analysisEntry, std::string title, std::string moduleVersion, Json::Value *data) :
-	  AnalysisBase(Analyses::analyses()),
+	  AnalysisBase(Analyses::analyses(), moduleVersion),
 	  _id(				id),
 	  _name(			analysisEntry->function()),
 	  _qml(				analysisEntry->qml().empty() ? _name : analysisEntry->qml()),
 	  _titleDefault(	analysisEntry->title()),
 	  _title(			title == "" ? _titleDefault : title),
-	  _moduleVersion(	moduleVersion),
 	  _moduleData(		analysisEntry),
 	  _dynamicModule(	_moduleData->dynamicModule())
 {
-	if(_moduleVersion == "" && _dynamicModule)
+	if(moduleVersion == "" && _dynamicModule)
 		_moduleVersion = _dynamicModule->version();
 
 	if (data)
@@ -391,7 +390,6 @@ Json::Value Analysis::asJSON(bool withRSource) const
 	analysisAsJson["rfile"]			= _rfile;
 	analysisAsJson["hasReport"]		= _hasReport;
 	analysisAsJson["progress"]		= _progress;
-	analysisAsJson["version"]		= _version.asString();
 	analysisAsJson["results"]		= _results;
 	analysisAsJson["status"]		= statusToString(_status);
 	analysisAsJson["options"]		= boundValues();
@@ -415,7 +413,7 @@ void Analysis::checkDefaultTitleFromJASPFile(const Json::Value & analysisData)
 	if(_title == oldTitleDefault && _titleDefault != oldTitleDefault)
 		_title = _titleDefault;
 
-	_oldVersion		= analysisData.get("preUpgradeVersion", _results.get("version", AppInfo::version.asString())).asString();
+	_preUpgraderVersion	= analysisData.get("preUpgradeVersion", _results.get("version", AppInfo::version.asString())).asString();
 }
 
 void Analysis::loadResultsUserdataAndRSourcesFromJASPFile(const Json::Value & analysisData, Status status)
@@ -450,9 +448,9 @@ void Analysis::setStatus(Analysis::Status status)
 		bool neededRefresh = needsRefresh();
 
 		TempFiles::deleteList(TempFiles::retrieveList(_id));
-		setVersion(AppInfo::version, true);
+		_wasUpgraded = false;
 
-		_moduleVersion = _dynamicModule->version();
+		_moduleVersion = _dynamicModule ?  _dynamicModule->version() : AppInfo::version;
 
 		if(neededRefresh != needsRefresh())
 			emit needsRefreshChanged();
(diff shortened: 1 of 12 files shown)


QMLComponents/controls/jaspcontrol.cpp
```diff
@@ -203,10 +206,22 @@ void JASPControl::componentComplete()
...
+		setUp();
+		setInitialized();
+	}
+	else if (!isDynamic)
+		// For statically build controls in a form, the form self will setup the controls when the form is completely loaded
+		// (by calling the AnalysisForm::setAnalysisUp function).
 		_form->addControl(this);
 	else
 	{
+		// The control is created dynamically, this is the case for row components.
```
JorisGoosen, line 222: 🙇

QMLComponents/rsyntax/rsyntax.cpp
```diff
@@ -155,7 +158,8 @@ QString RSyntax::generateWrapper() const
 \n\
 ";
 	result += _form->name() + "Wrapper <- function(\n";
-	result += FunctionOptionIndent + "data = NULL";
+	result += FunctionOptionIndent + "data = NULL\n";
+	result += FunctionOptionIndent + "version = \"" + tq(AppInfo::version.asString()) + "\"";
```
JorisGoosen: This is wrong, you need to use the version of the module

JorisGoosen (changes requested):

QMLComponents/rsyntax/rsyntax.cpp
```diff
@@ -85,7 +86,9 @@ QString RSyntax::generateSyntax() const
 		formulaSources.append(formula->modelSources());
 
 	result = _analysisFullName(true) + "(\n";
-	result += FunctionOptionIndent + "data = NULL";
+	result += FunctionOptionIndent + "data = NULL\n";
+	result += FunctionOptionIndent + "version = \"" + tq(AppInfo::version.asString()) + "\"";
```
JorisGoosen: as you should here

QMLComponents/analysisbase.h
```diff
@@ -27,6 +28,7 @@ class AnalysisBase : public QObject
 	virtual const std::string & module()									const	{ return emptyString;		}
 	virtual const std::string & name()										const	{ return emptyString;		}
 	virtual const std::string & title()										const	{ return emptyString;		}
+	virtual const Version & version()										const	{ return AppInfo::version;	}
```
JorisGoosen: Doesnt this mean that the versionnumbers in R-only mode will be totally wrong?

JorisGoosen: As in, with the current code all the stuff that has only access to AnalysisBase will always return whatever was the version of JASP when jaspBase gets compiled right?

Why not just add `_version` in AnalysisBase and not have an override in Analysis and then just make sure a value gets set into it.

JorisGoosen (changes requested):

JorisGoosen (changes requested):
This is better... Except for a crucial problem.

The moduleversion is stored in `_moduleVersion` and the JASP version is `_version` .
_version isnt really used in JASP anymore I think.

JorisGoosen:
In fact, "version" on Analysis is there basically only to load jaspfiles from pre 0.15 jasps

JorisGoosen:
Maybe add that as a comment to the definition of `_version` in Analysis (as there is no reason to have the jasp-version specified when running in pure R anyway)

Desktop/analysis/analysis.cpp
```diff
@@ -453,9 +448,9 @@ void Analysis::setStatus(Analysis::Status status)
 		bool neededRefresh = needsRefresh();
 
 		TempFiles::deleteList(TempFiles::retrieveList(_id));
-		setVersion(AppInfo::version, true);
+		_wasUpgraded = false;
 
-		_moduleVersion = _dynamicModule->version();
+		_moduleVersion = _dynamicModule ?  _dynamicModule->version() : AppInfo::version;
```
JorisGoosen, line 453: Why are you changing the behaviour?

JorisGoosen: this will all break loading old old files won't it?

JorisGoosen:
```suggestion
	_moduleVersion = _dynamicModule->version();
```

Desktop/analysis/analysis.cpp
```diff
@@ -394,7 +390,6 @@ Json::Value Analysis::asJSON(bool withRSource) const
 	analysisAsJson["rfile"]			= _rfile;
 	analysisAsJson["hasReport"]		= _hasReport;
 	analysisAsJson["progress"]		= _progress;
-	analysisAsJson["version"]		= _version.asString();
```
JorisGoosen, line 397: Its still useful to know which version of JASP was used to make an analysis right?

JorisGoosen: Especially as the dynamic module might be, you know, dynamic

Desktop/analysis/analysis.cpp
```diff
@@ -394,7 +390,6 @@ Json::Value Analysis::asJSON(bool withRSource) const
 	analysisAsJson["rfile"]			= _rfile;
 	analysisAsJson["hasReport"]		= _hasReport;
 	analysisAsJson["progress"]		= _progress;
```
JorisGoosen, line 392:
```suggestion
	analysisAsJson["progress"]		= _progress;
	analysisAsJson["version"]		= _version.asString();
```

JorisGoosen:
I think some of those changes you made with regards to needsRefresh might have caused this bug:
image

Left is with your code and right is 0.16.4, file below:

[oudJaspFile-bLinReg.zip](https://github.com/jasp-stats/jasp-desktop/files/10168279/oudJaspFile-bLinReg.zip)

Desktop/analysis/analysis.cpp
boutinb: The _version is an obsolete property. Only _moduleVersion is used, and it is stored with the "dyamicModule" part of the json value (see Analyses::createFromJaspFileEntry)

Desktop/analysis/analysis.cpp
boutinb: Apparently for old JASP file, the _dynamicModule can be null, so it is here only to make it not crash, maybe I can then better set the  moduleVersion to 0.

JorisGoosen: image

the right diff is from a log from your branch, it just loaded the attached file. 
[oudJaspFile-johnnyRfClassification.zip](https://github.com/jasp-stats/jasp-desktop/files/10168327/oudJaspFile-johnnyRfClassification.zip)

It should not have changed the version yet, as it hasn't been refreshed. I think this might cause the errormessage to dissappear

JorisGoosen (changes requested):

JorisGoosen:
So if you could change this PR to *not change the behaviour* that would be lovely.

~So keeping "version" in there so we can track the original version of JASP used.~
And keep the `setVersion` function on Analysis, or was there a reason you removed it?

Cause I think it is implicated in getting the error message to show, but am not sure

Desktop/analysis/analysis.cpp
JorisGoosen: First of all the upgrader handles `dynamicModule` being null, see: 
```
void DynamicModule::applyUpgrade(const std::string	& function,	const Version & version, Json::Value & analysesJson, UpgradeMsgs & msgs, StepsTaken & stepsTaken)
{
	if(!analysesJson.isMember("dynamicModule"))
		analysesJson["dynamicModule"] = asJsonForJaspFile(function);
	
	_upgrades->applyUpgrade(function, version, analysesJson, msgs, stepsTaken);
}
```

But it needs to have the information available to do this as seen at:
https://github.com/jasp-stats/jasp-desktop/blob/4bed733e759f11b14b19d0aa1f992221f21e07b9/Desktop/modules/upgrader/upgrader.cpp#L122

Desktop/analysis/analysis.cpp
JorisGoosen: I guess we could rely on the `version` in `Analysis` for the JASP version.

JorisGoosen:
Hmm, it seems the problem was already there in `stable`:
image

Ill check your code again later.

JorisGoosen (approved):

PR #4936, opened 2022-12-06
Remove Wrapper in wrapper function

boutinb:
Add in Descriptions.qml boolean hasWrappers, and for each analysis, the boolean hasWrapper and functionInternal properties. 
If hasWrappers is true for the module, then all analyses have hasWrapper set to true. 
If functionInternal is not set, then it takes the name of function with as suffix 'Internal' if hasWrapper is true. 
When an analysis is called, the functionInternal is called: so this work for analysis having a wrapper or not. No upgrade is needed, since the name of the analysis stays the same. 
In R Syntax the R code generated does not use the 'Wrapper' suffix anymore

### DIFF
--- a/Desktop/modules/description/entrybase.cpp
+++ b/Desktop/modules/description/entrybase.cpp
@@ -20,12 +20,14 @@ EntryBase::EntryBase(EntryType entryType) : DescriptionChildBase(), _entryType(e
 	connect(this, &EntryBase::menuChanged,			this, &EntryBase::somethingChanged);
 	connect(this, &EntryBase::titleChanged,			this, &EntryBase::somethingChanged);
 	connect(this, &EntryBase::functionChanged,		this, &EntryBase::somethingChanged);
+	connect(this, &EntryBase::functionInternalChanged, this, &EntryBase::somethingChanged);
 	connect(this, &EntryBase::iconChanged,			this, &EntryBase::somethingChanged);
 	connect(this, &EntryBase::entryTypeChanged,		this, &EntryBase::somethingChanged);
 	connect(this, &EntryBase::requiresDataChanged,	this, &EntryBase::somethingChanged);
 	connect(this, &EntryBase::enabledChanged,		this, &EntryBase::somethingChanged);
 	connect(this, &EntryBase::qmlChanged,			this, &EntryBase::somethingChanged);
 	connect(this, &EntryBase::debugChanged,			this, &EntryBase::somethingChanged);
+	connect(this, &EntryBase::hasWrapperChanged,	this, &EntryBase::somethingChanged);
 }
 
 void EntryBase::devModeChanged(bool)
@@ -43,7 +45,9 @@ QString EntryBase::toString() const
 		return	"-- analysis '"		+ title()
 				+ ( menu() != title() ? " menu: '" + menu() + "'" : "" )
 				+ ", function: '"	+ function() + "'"
+				+ ", functionInternal: '"	+ functionInternal() + "'"
 				+ ", qml: '"		+ qml() + "'"
+				+ ", hasWrapper: "	+ (hasWrapper() ? "TRUE" : "FALSE")
 				+ ", reqData: "		+ (requiresData() ? "yes":"no")
 				+ ", icon: '"		+ icon() + "' --";
 	default: 
@@ -92,6 +96,18 @@ void EntryBase::setFunction(QString function)
 
 	_function = function;
 	emit functionChanged();
+}
+
+void EntryBase::setFunctionInternal(QString functionInternal)
+{
+	if(_entryType != EntryType::analysis)
+		throw EntryError("You cannot set 'func(tion)' of an entry that is not of type 'analysis'.");
+
+	if (_functionInternal == functionInternal)
+		return;
+
+	_functionInternal = functionInternal;
+	emit functionInternalChanged();
 }
 
 void EntryBase::setIcon(QString icon)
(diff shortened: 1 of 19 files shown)


Desktop/modules/description/entrybase.cpp
```diff
@@ -20,12 +20,14 @@ EntryBase::EntryBase(EntryType entryType) : DescriptionChildBase(), _entryType(e
 	connect(this, &EntryBase::menuChanged,			this, &EntryBase::somethingChanged);
 	connect(this, &EntryBase::titleChanged,			this, &EntryBase::somethingChanged);
 	connect(this, &EntryBase::functionChanged,		this, &EntryBase::somethingChanged);
+	connect(this, &EntryBase::functionInternalChanged, this, &EntryBase::somethingChanged);
```
JorisGoosen: ouch

Desktop/modules/description/entrybase.h
```diff
@@ -23,72 +23,82 @@ class EntryBase : public DescriptionChildBase
...
-	bool		requiresData()	const { return _requiresData;	}
-	bool		enabled()		const { return _enabled;		}
-	QString		qml()			const { return _qml;			}
-	bool		debug()			const { return _debug;			}
-	QString		toString()		const;
-	bool		shouldBeAdded()	const;
+	QString		menu()				const { return _menu;				}
+	QString		title()				const { return _title;				}
+	QString		function()			const { return _function;			}
+	QString		functionInternal()	const { return _functionInternal;	}
```
JorisGoosen: I think it is better if you remove `_functionInternal` and just make it dynamic here in this function. 
That way we dont have to keep doing administration if things change.
And anyway, there really is no need to store it, which is the main reason.

If you have both `_hasWrappers` and `_function` you do not need `_functionInternal` as well

Desktop/modules/description/entrybase.h
```diff
@@ -23,72 +23,82 @@ class EntryBase : public DescriptionChildBase
...
+	bool		enabled()			const { return _enabled;			}
+	QString		qml()				const { return _qml;				}
+	bool		debug()				const { return _debug;				}
+	bool		hasWrapper()		const { return _hasWrapper;			}
+	QString		toString()			const;
+	bool		shouldBeAdded()		const;
 
 	///This function is a stopgap and these two classes must be merged together later
-	AnalysisEntry * convertToAnalysisEntry(bool requiresDataDefault) const;
+	AnalysisEntry * convertToAnalysisEntry(bool requiresDataDefault, bool hasWrappers) const;
```
JorisGoosen: Why does it have this flag?

entryBase already knows whether it has a wrapper...

Desktop/qquick/rcommander.cpp
```diff
@@ -63,15 +63,13 @@ bool RCommander::addAnalysis(const QString &code)
 
 	setRunning(true);
 
-	// Temporary solution: check whether the code starts with '<moduleName>::<analysisNameWrapper>(...'
+	// Temporary solution: check whether the code starts with '<moduleName>::<analysisName>(...'
 	QString codeTrimmed = code.trimmed();
 	QStringList analysisParts = codeTrimmed.mid(0, codeTrimmed.indexOf("(")).split("::");
 
 	if (analysisParts.length() == 2)
```
JorisGoosen: Instead of doing all this you could also simply make a `std::map<std::string....` that maps all the known "wrappers" or "syntax mode calls" to `Analysis` but that isnt quite necessary in this PR

QMLComponents/analysisform.cpp
```diff
@@ -943,6 +943,26 @@ void AnalysisForm::setShowRSyntax(bool showRSyntax)
...
+		return;
+
+	_showRButton = showRButton;
+
+	emit showRButtonChanged();
+}
+
+void AnalysisForm::setDeveloperMode(bool developerMode)
+{
+	if (_developerMode == developerMode)
```
JorisGoosen, line 952: what is this?

JorisGoosen (changes requested):
some changes and questions

Desktop/modules/description/entrybase.h
boutinb: There `hasWrappers` for the whole module, and `hasWrapper` for the analysis. I have added `hasWrappers`so that you don't have to specify it for all analyses, if all have wrappers.

Desktop/qquick/rcommander.cpp
boutinb: I have made the parsing a bit more clever so that it checked that the moduleName & analysisName exist

Desktop/modules/description/entrybase.h
boutinb: functionInternal is per default <function name> + "Internal", but why not to make it settable? It costs so little...

QMLComponents/analysisform.cpp
boutinb: developerMode is an AnalysisForm property, and must be set by the Analysis object when created

Desktop/modules/description/entrybase.cpp
boutinb: ??

JorisGoosen: the spacing looked messed up haha

QMLComponents/analysisform.cpp
JorisGoosen: yes I see that but why?

boutinb: So that the "Generate Wrappers" button appears only in Developer Mode.

Desktop/qquick/rcommander.cpp
```diff
@@ -56,41 +57,76 @@ bool RCommander::runCode(const QString & code)
...
+	if (analysisParts.length() != 2) return false;
+
+	moduleName = analysisParts[0];
+	analysisName = analysisParts[1];
+
+	Modules::DynamicModule* module = Modules::DynamicModules::dynMods()->dynamicModule(moduleName);
+	if (!module) return false;
+
+	for (const Modules::AnalysisEntry* entry : module->menu())
+	{
```
JorisGoosen:
```suggestion
```

Desktop/qquick/rcommander.cpp
```diff
@@ -56,41 +57,76 @@ bool RCommander::runCode(const QString & code)
...
+		if (entry->isAnalysis() && entry->isEnabled() && entry->hasWrapper() && entry->function() == fq(analysisName))
+			return true;
+	}
```
JorisGoosen:
```suggestion
```

JorisGoosen (approved):
Much nicer wrappers this way ^^

PR #4937, opened 2022-12-06
Make decimal settings available in R

JorisGoosen:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2172 Also adds some small default values in jaspBase for when not in JASP

This makes the following global values available in R:
- .fixedDecimals        (BOOL -> should decimals be fixed to a certain number?)
- .numDecimals          (INT  -> if above is TRUE then how many decimals)
- .normalizedNotation   (BOOl -> 10^n notation vs exponent notation)
- .exactPValues         (BOOL -> whether to show all decimals for p-values ( or at least 4 ) and not APA style (if Im reading the code correctly)

See also: https://github.com/jasp-stats/jaspBase/pull/111 but that isnt necessary for running this in normal JASP


boutinb (approved):

boutinb:
If you remove jaspBase from your commit, then the conflict will be solved... then you can merge it.

PR #4938, opened 2022-12-06
extra space underneath last analysis

JorisGoosen:
it isnt always perfect but I think Ive got the bases mostly covered:
- on add analysis, remove and move
- on change analysis (results updated)
- on reRender

Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2089

### DIFF
--- a/Desktop/html/js/analyses.js
+++ b/Desktop/html/js/analyses.js
@@ -100,6 +100,8 @@ JASPWidgets.Analyses = JASPWidgets.View.extend({
 
 	removeAnalysis: function (analysis) {
 
+		var analysesVar = this;
+
 		analysis.$el.animate({ opacity: 0 }, 400, "easeOutCubic", function () {
 			analysis.$el.slideUp(400, function () {
 				// TODO: See why analysis is NULL here.
@@ -107,7 +109,7 @@ JASPWidgets.Analyses = JASPWidgets.View.extend({
 				this.analyses = _.without(this.analyses, analysis);
 				this.views = _.without(this.analyses, analysis);
 
-				this.setBottomSpacerHeight();
+				analysesVar.setBottomSpacerHeight();
 			});
 		});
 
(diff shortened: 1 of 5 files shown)


Desktop/html/js/analyses.js
```diff
@@ -122,6 +130,23 @@ JASPWidgets.Analyses = JASPWidgets.View.extend({
...
+	},
+
+	setBottomSpacerHeight: function() {
+		var bottomSpacer = $("#bottomSpacer");
+		var height		 = window.innerHeight;
+
+		if(this.analyses.length > 0)
+		{
+			var lastAnalysis = this.analyses[this.analyses.length - 1];
+			height = height - lastAnalysis.$el.outerHeight();
```
JorisGoosen:
```suggestion
			height = Math.max(0, height - lastAnalysis.$el.outerHeight());
```

Desktop/html/js/analyses.js
```diff
@@ -122,6 +130,23 @@ JASPWidgets.Analyses = JASPWidgets.View.extend({
 
 	reRender: function() {
 		this.analyses.forEach(function(analysis) {analysis.render();});
+		this.setBottomSpacerHeight();
+	},
+
+	setBottomSpacerHeight: function() {
+		var bottomSpacer = $("#bottomSpacer");
+		var height		 = window.innerHeight;
```
boutinb: Use better $(windows).innerHeight(): jQuery takes better into account the scrollbars.
I have noticed that when the horizontal scrollbar appears, then the bottomSpacer becomes a bit too high.

boutinb (changes requested):

Desktop/html/js/analyses.js
JorisGoosen:
```suggestion
		var height		 = $(windows).innerHeight();
```

like this?

PR #4939, opened 2022-12-12
Improvements to "info" field

JorisGoosen:
This should make at least all components show up as markdown when Info is filled.
To test this you can use the `jaspTestModule` and analysis "info testing"

Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/1554 at least partly.

### DIFF
--- a/QMLComponents/analysisbase.cpp
+++ b/QMLComponents/analysisbase.cpp
@@ -48,13 +48,21 @@ void AnalysisBase::createForm(QQuickItem* parentItem)
 	Log::log() << "Analysis(" << this << ")::createForm() called with parentItem " << parentItem << std::endl;
 
 	setQmlError("");
+	
+	if(parentItem)
+		_parentItem = parentItem;
 
 	if(_analysisForm)
 	{
 		Log::log() << "It already has a form, so we destroy it." << std::endl;
-		if (!parentItem) parentItem = _analysisForm->parentItem();
+		
+		if (!parentItem) 
+			parentItem = _analysisForm->parentItem();
+		
 		destroyForm();
 	}
+	else if(!parentItem)
+		parentItem = _parentItem;
 
 	try
 	{
@@ -75,7 +83,12 @@ void AnalysisBase::createForm(QQuickItem* parentItem)
 
 		emit formItemChanged();
 	}
-	catch(std::exception e)
+	catch(qmlLoadError & e)
+	{
+		setQmlError(e.what());
+		_analysisForm = nullptr;
+	}
+	catch(std::exception & e)
 	{
 		setQmlError(e.what());
 		_analysisForm = nullptr;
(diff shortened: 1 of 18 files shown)


JorisGoosen:
As @Kucharssim said in the issue there were some problems, Ive implemented most of that except for:

> 6. Support basic math. Ideally used as \eqn{} and \deqn{} with LaTeX expressions inside for inline and block math, respectivelly. (This would mean we do not need to do any shannenigans for the R help files as roxygen understands \eqn{} and \deqn{})

Im not sure how we would integrate this into the html output. We are using `mrkdwn.js` to render the MD to html, I guess we could see if we could add a hook to process latex in there? But that honestly seems a bit hard to do? We would have to ship some kind of (la)tex with JASP then right?

> 7. Should we support gifs in help files? Could be relatively difficult for the R help files to include that. It's possible for html help files, but not really for pdf documentation (see below).

I think we do support gifs within the jasp-help right now. I think in something for Rstudio it would be ok too, pdf... yeah, well...

Kucharssim:
> We would have to ship some kind of (la)tex with JASP then right?

Well if I got to finish https://github.com/jasp-stats/INTERNAL-jasp/issues/395 we could use that to render the math in JASP in helpfiles since we could just let `katex` do its thing on the generated html with LaTeX chunks.

An alternative would be to stick to unicode to render math symbols in help files, but then things like `\u03BC` would have to be replaced with `` `r "\u03BC"` `` in roxygen (see https://mpn.metworx.com/packages/roxygen2/7.2.0/articles/rd-formatting.html#inline-code) and we would have to forbid things like `&mu;`, `&sigma;`, etc., since R does not know how to deal with it.

> I think we do support gifs within the jasp-help right now. I think in something for Rstudio it would be ok too, pdf... yeah, well...

Yeah we did talk about it already right? Chances are that if a developer wants to include a gif in the help, the gif probably shows something in the user interface, so not immediately helpful for people working in R. So it's not a big deal I think. But, I think at worst we can always just include a link to the gif so that users can open it in whatever program they want (or whatever is the default program set up for opening gif on their computer), right?

QMLComponents/analysisbase.cpp
```diff
@@ -48,13 +48,21 @@ void AnalysisBase::createForm(QQuickItem* parentItem)
...
 		Log::log() << "It already has a form, so we destroy it." << std::endl;
-		if (!parentItem) parentItem = _analysisForm->parentItem();
+		
+		if (!parentItem) 
+			parentItem = _analysisForm->parentItem();
+		
 		destroyForm();
 	}
+	else if(!parentItem)
+		parentItem = _parentItem;
```
boutinb, line 65: If I understand well _parentItem is used to cache the parent item in case a form is re-created (can be nice to be added in a comment).
But if the parentItem is not given as argument in the `createForm` function, and is found in the old form, then it won't be set to _parentItem.
I would then move here the line 52/53:
```
	if(parentItem)
		_parentItem = parentItem;
```

boutinb (changes requested):
Seems to work quite well!
Just a little comment, and this can be merged

QMLComponents/analysisbase.cpp
JorisGoosen: the problem is more when you load a qml form with an error it fails. Then the next reload it doesnt have access to the parentItem and due to the way it is triggered it doesnt get it passed.

By moving the thing you suggest it wouldnt store the parentItem at the moment it gets it and thus it gets lost.

PR #4941, opened 2022-12-13
Upgrader should not throw error if not a developer module

boutinb:
- The errors were not send to the form: this was due to a wrong overriding of the upgradeMsgsForOption method. 
- An error message box should pop up only if the module is a Developer Module.
- Make a difference between error and warning: If the issue comes from the fact that an option is not found in the JASP file, this is just a warning and should not be displayed in the form, only log it.

### DIFF
--- a/Desktop/modules/upgrader/upgrade.cpp
+++ b/Desktop/modules/upgrader/upgrade.cpp
@@ -51,12 +51,18 @@ void Upgrade::applyUpgrade(const std::string & function, const Version & version
 		{
 			Log::log() << "Change had a" << ( error.isWarning ? " warning" : "n error") <<  ": " << error.what() << std::endl;
 
-			if(module && module->isDevMod())
+			if(isModuleDev())
 				throw error;
+			
 			if (!error.isWarning)
 				msgs[""].push_back(error.what());
 		}
 
+}
+
+bool Upgrade::isModuleDev() const
+{
+	return DynamicModules::developmentModuleName() == fq(module());
 }
 
 QString Upgrade::toString() 
--- a/Desktop/modules/upgrader/upgrade.cpp
+++ b/Desktop/modules/upgrader/upgrade.cpp
@@ -4,6 +4,7 @@
 #include "upgrades.h"
 #include "upgrade.h"
 #include "log.h"
+#include "modules/dynamicmodules.h"
 
 namespace Modules
 {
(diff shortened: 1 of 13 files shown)


Desktop/modules/upgrader/upgrade.cpp
```diff
@@ -48,12 +49,14 @@ void Upgrade::applyUpgrade(const std::string & function, const Version & version
 		}
 		catch(upgradeError & error)
 		{
-			//If we are not in developermode it would be better to keep going instead of crashing the whole upgradeprocess
-			if(PreferencesModel::prefs()->developerMode())
+			Log::log() << "Change had a problem: " << error.what() << std::endl;
```
JorisGoosen:
```suggestion
			Log::log() << "Change had a" << ( error.isWarning ? " warning" : "n error") <<  ": " << error.what() << std::endl;
```

Desktop/modules/upgrader/upgradeDefinitions.h
```diff
@@ -18,10 +18,12 @@ extern const char			*	logId;
 
 struct upgradeError  : public std::runtime_error
 {
-	upgradeError(std::string msg) : std::runtime_error(msg) {}
+	bool isWarning = false;
+	upgradeError(std::string msg, bool _isWarning = false) : std::runtime_error(msg) { isWarning = _isWarning; }
```
JorisGoosen:
```suggestion
	upgradeError(std::string msg, bool isWarning = false) : std::runtime_error(msg), isWarning(isWarning) {}
```

QMLComponents/analysisbase.cpp
```diff
@@ -4,6 +4,7 @@
 #include "utilities/qmlutils.h"
 
 const std::string AnalysisBase::emptyString;
+const stringvec AnalysisBase::emptyStringVec;
```
JorisGoosen, line 7: `stringvec upgradeMsgsForOption(std::string name)` used to return `{""}` not `{}`...

```suggestion
const stringvec AnalysisBase::emptyStringVec = {""};
```

JorisGoosen: Aah ok, but that seems to be caused by the move to AnalysisBase and wasnt originally like that at all

JorisGoosen (commented):
mostly looks fine, I mde some small suggestions, ill commit and em and test the result. If all goes well Ill merge

JorisGoosen:
this should work better, but I cant test it cause something seems terribly wrong with my build.

Perhaps you @boutinb could test whether it now properly behaves and if so merge it?

Might be nice to test with this file and jaspRegression dev module
[bLinReg.jasp.zip](https://github.com/jasp-stats/jasp-desktop/files/10238924/bLinReg.jasp.zip)

boutinb:
It works for me.

PR #4942, opened 2022-12-13
passing ppi, imageBackground and resultFont via settings

JorisGoosen:
instead of in each function again and agin

this cleans up the api a bit and makes transitioning to some kind of jaspOptions R object much easier in the future
changing the settings in jasp nows triggers `registerFonts()` after setting .resultsFont

depends on https://github.com/jasp-stats/jaspBase/pull/113 (Although I also pushed it to https://github.com/jasp-stats/jaspBase/tree/fixSaveAsPdf for easy `git submodule update` pleasure.

These changes let `pdf` work on macos and I assume they dont mess up Windows but cnt really test that at the moment.

Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2136

### DIFF
--- a/R-Interface/jasprcpp.cpp
+++ b/R-Interface/jasprcpp.cpp
@@ -158,7 +158,6 @@ void STDCALL jaspRCPP_init(const char* buildYear, const char* version, RBridgeCa
 	char baseCitation[200];
 	sprintf(baseCitation, baseCitationFormat, buildYear, version);
 	rInside[".baseCitation"]		= baseCitation;
-	rInside[".resultsFont"]			= resultsFont;
 
 	//XPtr doesnt like it if it can't run a finalizer so here are some dummy variables:
 	static logFuncDef			_logFuncDef					= jaspRCPP_logString;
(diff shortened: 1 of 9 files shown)


R-Interface/jasprcpp.cpp
```diff
@@ -259,23 +258,31 @@ void STDCALL jaspRCPP_setDecimalSettings(int numDecimals, bool fixedDecimals, bo
...
 
+void STDCALL jaspRCPP_setFontAndPlotSettings(const char * resultFont, const int ppi, const char* imageBackground)
+{
+	RInside &rInside			= rinside->instance();
+
+	rInside[".resultFont"]			= resultFont;
+	rInside[".imageBackground"]		= imageBackground;
+	rInside[".ppi"]					= ppi;
+
+	jaspRCPP_parseEvalQNT("jaspBase::registerFonts()");
```
vandenman:
```suggestion
	jaspRCPP_parseEvalQNT("jaspBase:::registerFonts()");
```
so we wouldn't need to export it in jaspBase?

Also, I'm not 100% sure about which part is responsibile for setting the fonts now. 

It looks like sometimes, jaspBase calls `registerFonts` and sometimes the engine does. Shouldn't it be the case that jaspBase calls this whenever needed (runAnalysis, saveImage, and reWriteImages) and that jaspEngine makes sure .resultFont is available/ updated?

JorisGoosen: > Also, I'm not 100% sure about which part is responsibile for setting the fonts now.
> It looks like sometimes, jaspBase calls registerFonts and sometimes the engine does. Shouldn't it be the case that jaspBase calls this whenever needed (runAnalysis, saveImage, and reWriteImages) and that jaspEngine makes sure .resultFont is available/ updated?

I think the main thing registerFonts seems to do is set `.resultFont` in `graphOptions` in jspGraphs.

As far as it being called in the right places:

Well right now `initEnvironment` calls it, which I assume we'd also run in syntaxmode or so. And the user cant easily switch fonts after that. 

But inside JASP it gets called every time the resultsFont *could have* changed.

boutinb (approved):

PR #4946, opened 2022-12-14
Slightly darker blue and thinner border on arrow on analysisformexpander

JorisGoosen:
As requested by @EJWagenmakers 

image

image


JorisGoosen:
@EJWagenmakers what do you think?

EJWagenmakers:
Yes, I think this is more crisp

PR #4958, opened 2022-12-21
Dont auto scroll upon mouseclick

RensDofferhoff:
Focus event chains are a little weird so had to be more specific

### DIFF
--- a/Desktop/components/JASP/Widgets/AnalysisFormExpander.qml
+++ b/Desktop/components/JASP/Widgets/AnalysisFormExpander.qml
@@ -225,7 +225,7 @@ DropArea
 
 					backgroundFlickable.scrollToElement(expanderButton);
 					if(firstExpansion) //only focus first item on analysis creation
-						formParent.nextItemInFocusChain().forceActiveFocus();
+						formParent.nextItemInFocusChain().forceActiveFocus(Qt.OtherFocusReason);
 					else
 						draggableItem.forceActiveFocus();
 			}
@@ -595,7 +595,8 @@ DropArea
 								return;
 
 							const control = myForm.activeJASPControl;
-							if (control.focusReason !== Qt.MouseFocusReason)
+							var x =  control.focusReason;
+							if (control.focusReason === Qt.TabFocusReason || control.focusReason === Qt.BacktabFocusReason || control.focusReason === Qt.OtherFocusReason)
 								backgroundFlickable.scrollToElement(control, 50 * jaspTheme.uiScale);
 						}
 					}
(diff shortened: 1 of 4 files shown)


Desktop/components/JASP/Widgets/AnalysisFormExpander.qml
```diff
@@ -225,7 +225,7 @@ DropArea
 
 					backgroundFlickable.scrollToElement(expanderButton);
 					if(firstExpansion) //only focus first item on analysis creation
-						formParent.nextItemInFocusChain().forceActiveFocus();
+						formParent.nextItemInFocusChain().forceActiveFocus(Qt.OtherFocusReason);
```
boutinb: Why this change, since OtherFocusReason is the default?

Desktop/components/JASP/Widgets/AnalysisFormExpander.qml
```diff
@@ -595,7 +595,8 @@ DropArea
 								return;
 
 							const control = myForm.activeJASPControl;
-							if (control.focusReason !== Qt.MouseFocusReason)
+							var x =  control.focusReason;
+							if (control.focusReason === Qt.TabFocusReason || control.focusReason === Qt.BacktabFocusReason || control.focusReason === Qt.OtherFocusReason)
```
boutinb: Why this change again? Which focus reasons you don't want here? With this change you exclude now MouseFocusReason, but also ActiveWindowFocusReason, PopupFocusReason, ShortcutFocusReason and MenuBarFocusReason. Is the ShortcutFocusReason not a good reason for scrolling?

boutinb (changes requested):

RensDofferhoff:
Fixed the fix @boutinb.
Good suggestions

QMLComponents/controls/jaspcontrol.cpp
```diff
@@ -793,6 +793,9 @@ void JASPControl::_notifyFormOfActiveFocus()
 	if (!hasActiveFocus())
 		_hasActiveFocus = false;
 
-	if (_form)
+	//All JASP controls are focusscopes when they receive focus due to a child being clicked the reason is 8 for some reason (undocumented as of 3-1-2023)
+	//Objects like jasp groupboxes could unrightfully be marked as the active jasp control when a child jasp control gets focus
+	//For this reason we check if the focus reason is the result of user input or the focus scope system.
+	if (_form && _focusReason >= Qt::MouseFocusReason && _focusReason <= Qt::OtherFocusReason)
```
boutinb, line 799: Sorry, I don't understand this extra check : the only focus reason that cannot be between MouseFocusReason and OtherFocusReason is NoFocusReason (also, as this function is called only when the focus changes, that would be really strange).

boutinb (changes requested):

QMLComponents/controls/jaspcontrol.cpp
RensDofferhoff: There are hidden values for the FocusReason Enum that do not concern user input.
Such as when a FocusScope gets focus due to a focus event of  one of its children (value 8) .
You can check it under the debugger. It is weird but true.

boutinb (approved):

PR #4968, opened 2023-01-05
Add missing translation strings

shun2wang:


JorisGoosen (approved):
Hey thanks!

I do see I shouldve probably formatted those strings differently but this should suffice for now

PR #4970, opened 2023-01-08
Dynamic Radiobuttons

RensDofferhoff:
Test With: https://github.com/jasp-stats/jaspTestModule/pull/47

Implements Dynamic Radiobuttons.
Not really happy with the implementation.
The buttons search upwards for a RadioButtonGroup upon component complete & parent change.
Name changes are propagated to the group.

Some internal issues get exposed upon name change.
I think they are related to the engine but the debugger is not very descriptive.
They occur upon very quick name changes.
A non descriptive uncaught std error is thrown.

Another issue relates to the set width of RadioButtonGroup when it has no direct Child to encapsulated.
The Group does not scale for Component Lists for example.

When there is no RadioButton with checked set, the item selected py the group depends on the order of onComplete & parentItem change signals of the buttons. This might not always be the same as the child order which defines the current behavior.  If anyone actually depends on the current behavior (by not setting a single Button as checked), I do not know. It would be unwise either way.

Unless really needed by @FBartos  I recommend we delay these changes until the next release.


QMLComponents/controls/radiobuttonsgroupbase.cpp
```diff
@@ -33,50 +32,49 @@ RadioButtonsGroupBase::RadioButtonsGroupBase(QQuickItem* item)
...
-			if (controlType == ControlType::RadioButton)
-				buttons.append(qobject_cast<RadioButtonBase*>(jaspControl));
-			else if (controlType != ControlType::RadioButtonGroup)
-				_getRadioButtons(child, buttons);
-		}
-		else
-			_getRadioButtons(child, buttons);
-	}	
+		QString oldValue = _buttonValueMap[button];
+		_valueButtonMap.remove(oldValue);
```
JorisGoosen: Why store the value in a map?

When you need the value you can just get it from the RadioButton through its pointer, no?

QMLComponents/controls/radiobuttonsgroupbase.h
```diff
@@ -38,23 +40,30 @@ class RadioButtonsGroupBase : public JASPControl, public BoundControlBase
...
 
+	RadioButtonBase* checkedButton() { auto it = _valueButtonMap.find(_value); return it == _valueButtonMap.end() ? nullptr : it.value(); }
+	QList<RadioButtonBase*> buttons() { return _valueButtonMap.values(); }
+
 	void _setCheckedButton(RadioButtonBase* button);
-	void _getRadioButtons(QQuickItem* item, QList<RadioButtonBase* >& buttons);
 
-	QMap<QString, RadioButtonBase *>	_buttons;
+
+	QMap<QString, RadioButtonBase *>	_valueButtonMap;
```
JorisGoosen: Im not really sure why you opted to use two maps instead of a single set, but I could very well be missing something

QMLComponents/controls/radiobuttonsgroupbase.cpp
RensDofferhoff: The value -> button map is for the bindTo function. I suppose I could have done a linear scan over a set instead but a map is most natural.
The second map is a bit silly but when the name changes I need to have to old name() value to check if it is the current group  value.
There is probably a nicer way but I didnt want to alter the changeName logic in jaspControl so I took the easy route.

QMLComponents/controls/radiobuttonsgroupbase.h
RensDofferhoff: Need to have access to previous values.
But now that I think of it should just store a value reference/pointer instead.
Or simply store the selected button instead of the string value.
Bit stupid will change

QMLComponents/components/JASP/Controls/RadioButton.qml
```diff
@@ -43,26 +43,18 @@ RadioButtonBase
 	property alias	label:					control.text
 	property alias	checked:				control.checked
 	property alias	value:					radioButton.name
-	property var	buttonGroup:			null
 	property bool	childrenOnSameRow:		false
 	property alias	columns:				childControlsArea.columns
 	property bool	enableChildrenOnChecked: true
 	property bool	indentChildren:			true
 
-	function click()
```
boutinb, line 52: click is a function that is called by some analyses. So you can't remove it.
I thought click() is the action that clicks the button.
clicked is the signal.
onClicked is the slot.

QMLComponents/controls/radiobuttonbase.h
```diff
@@ -30,6 +30,16 @@ class RadioButtonBase : public JASPControl
 public:
 	RadioButtonBase(QQuickItem* parent = nullptr);
 
+public slots:
+	Q_INVOKABLE void registerWithParent();
+	Q_INVOKABLE void clicked();
```
boutinb: clicked should be a signal!

QMLComponents/components/JASP/Controls/RadioButton.qml
```diff
@@ -43,26 +43,18 @@ RadioButtonBase
...
-
-
 	RadioButton
 	{
 		id:					control
-		ButtonGroup.group:	buttonGroup
 		padding:			jaspTheme.jaspControlPadding
 		focus:				true
 
-		onCheckedChanged:	if (buttonGroup && checked) buttonGroup.clicked(control)
```
boutinb, line 65: In some analyses the checked property is directly used: if the checked property is changed, an action should be called so the analysis options are changed and the analysis is run.

QMLComponents/components/JASP/Controls/RadioButtonGroup.qml
```diff
@@ -31,9 +31,6 @@ RadioButtonsGroupBase
 	shouldStealHover:	false
 
 	default property alias	content:				contentArea.children
-			property alias	buttons:				buttonGroup.buttons
-			property alias	checkedButton:			buttonGroup.checkedButton
-			property alias	buttonGroup:			buttonGroup
```
boutinb, line 36: the buttonGroup alias should be still be present: all the controls give access to their internal Qt control

QMLComponents/controls/radiobuttonsgroupbase.cpp
```diff
@@ -33,50 +32,46 @@ RadioButtonsGroupBase::RadioButtonsGroupBase(QQuickItem* item)
...
-		else
-		{
-			_buttons[controlName] = button;
-			if (button->property("checked").toBool())	setValue(controlName);
-			button->setProperty("buttonGroup", buttonGroup);
-		}
+		if (_buttons.size() == 0)
+			_setCheckedButton(button);
+		if (button->property("checked").toBool())
+			_setCheckedButton(button);
```
boutinb: To avoid checking radio buttons multiple times during the initialization of the RadioButtonGroup: if it is not yet initialized, don't set the checked button, and connect the initialized property to set the right checked button.

boutinb (changes requested):

QMLComponents/components/JASP/Controls/RadioButtonGroup.qml
RensDofferhoff: There is no internal control all logic is handled in the component now.

boutinb (approved):

PR #4972, opened 2023-01-09
Double Validation can be too intrusive

boutinb:
When 2 DoubleFields determine the min and max of each other (to simulate a range), if the user changes the value of one field, the min can temporary exceed the max of one DoubleField, though the user did not finish editing the value.. As long a user did not finish editing a value, no error messages should be displayed.

Fixes https://github.com/jasp-stats/jasp-test-release/issues/2170

### DIFF
--- a/QMLComponents/components/JASP/Controls/TextField.qml
+++ b/QMLComponents/components/JASP/Controls/TextField.qml
@@ -33,7 +33,7 @@ TextInputBase
 	
 	property alias	control:			control
 	property alias	text:				textField.label
-	property alias	value:				control.text
+	property alias	transientValue:		control.text
 	property string lastValidValue:		defaultValue
 	property int	fieldWidth:			jaspTheme.textFieldWidth
 	property int	fieldHeight:		0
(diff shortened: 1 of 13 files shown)


Kucharssim (commented):
It is improved; I don't get stuck in an infinite loop, but it is still a bit intrusive (both validation errors fire up at the same time and I need to fix the value problem in one of the fields and then manually close both messages)

JorisGoosen:
mybe @boutinb can add a `editing` property to textfield so you can set the min and mx to infinity while they are edited?

boutinb:
I have defined the `displayValue` property which is the value the user is typing.
The `value` property is then the value when the user has finished editing.
`displayValue` property should be used only in the onEditingFinished slots in QML (which should not be used in the analyses).
So when in a QML, the `value` of a DoubleField or IntegerField is used in a property of another control,  this value won't be updated each time the user is typing, but only when the user has finished editing.
Also this makes the handling of this min - max constraints easier: when modifying the min or max, the constraint is checked only when `value` is changed, so only when the user has finished typing.
I have also add a Min Max test in jaspTestModule

QMLComponents/components/JASP/Controls/TextField.qml
```diff
@@ -33,8 +33,8 @@ TextInputBase
 	
 	property alias	control:			control
 	property alias	text:				textField.label
-	property alias	value:				control.text
-	property string lastValidValue:		defaultValue
+	property alias	displayValue:		control.text
```
JorisGoosen:
```suggestion
	property alias	displayValue:		control.text	///< Only use in onEditingFinished!
```

JorisGoosen (approved):

JorisGoosen (approved):

PR #4979, opened 2023-01-12
Fix error in R Syntax for Repeated Measure ANOVA

boutinb:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/2177

2 problems are solved here:

- a formula should be used only if column variables are used: if a factor or level name is used (like in  Repeated Measures ANOVA) then an error is thrown by the formula checker is R. 
- when an option value of a Variables List control uses a list instead of a formula, then the R syntax tries to simplify the structure by removing if possible extra depth in the list of list of list... As the json value that is generated by such simplified value does not fit the expected json value by the control, this json value must be first adjusted

### DIFF
--- a/QMLComponents/rsyntax/rsyntax.cpp
+++ b/QMLComponents/rsyntax/rsyntax.cpp
@@ -415,15 +415,14 @@ QString RSyntax::transformJsonToR(const Json::Value &json, const Json::Value& co
 
 bool RSyntax::_areTermsVariables(ListModel* model, const Terms& terms) const
 {
-	bool result = true;
-
 	QStringList variables = model->requestInfo(VariableInfo::VariableNames).toStringList();
 
 	for (const Term& term : terms)
 		for (const QString& comp : term.components())
-			result = result && variables.contains(comp);
+			if(!variables.contains(comp))
+				return false;
 
-	return result;
+	return true;
 }
 
 QString RSyntax::_transformInteractionTerms(ListModel* model) const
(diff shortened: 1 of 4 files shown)


QMLComponents/rsyntax/rsyntax.cpp
```diff
@@ -413,10 +413,28 @@ QString RSyntax::transformJsonToR(const Json::Value &json, const Json::Value& co
...
 {
+	bool result = true;
+
+	QStringList variables = model->requestInfo(VariableInfo::VariableNames).toStringList();
+
+	for (const Term& term : terms)
+	{
+		for (const QString& comp : term.components())
+			result = result && variables.contains(comp);
+	}
```
JorisGoosen:
```suggestion
		for (const QString& comp : term.components())
			result = result && variables.contains(comp);
```

QMLComponents/rsyntax/rsyntax.cpp
```diff
@@ -413,10 +413,26 @@ QString RSyntax::transformJsonToR(const Json::Value &json, const Json::Value& co
...
 {
+	bool result = true;
+
+	QStringList variables = model->requestInfo(VariableInfo::VariableNames).toStringList();
+
+	for (const Term& term : terms)
+		for (const QString& comp : term.components())
+			result = result && variables.contains(comp);
+
+	return result;
```
JorisGoosen:
```suggestion
	QStringList variables = model->requestInfo(VariableInfo::VariableNames).toStringList();

	for (const Term& term : terms)
		for (const QString& comp : term.components())
			if(!variables.contains(comp))
				return false;

	return true;
```

JorisGoosen (approved):
Code looks fine although im not too sure about all the possible interactions the interactions can have

PR #5001, opened 2023-01-19
Properly clean up mem-mapped files at shutdown and fix long pause crashes due to their accidental removal

RensDofferhoff:
For a less involved dirty fix see PR: https://github.com/jasp-stats/jasp-desktop/pull/5002
But with that patch the trash pile will mount in the temp folder for 24 hours which could be GBs if the user uses large files.
We are currently not properly cleaning up at shutdown  on MacOS. This PR fixes this.

On MacOS crucial memory mapped files of other instances are deleted upon start up. 
This PR should fix the unnecessary removal of the temp files which can lead to crashing engine and entering eternal resume-crash loop.

This PR builds in safety margin by extending time limit to 24 hours and fastening the heartbeat to guard against possible hidden bugs, likely related to post sleep and app-nap interference with timers on MacOS. 

This PR needs to be tested heavily on all platforms

Fixes: https://github.com/jasp-stats/jasp-test-release/issues/2241

### DIFF
--- a/Common/tempfiles.cpp
+++ b/Common/tempfiles.cpp
@@ -28,6 +28,7 @@
 #include "log.h"
 using namespace std;
 
+const long 				outOfDateDelta = 24 * 3600;
 long					TempFiles::_sessionId		= 0;
 std::string				TempFiles::_sessionDirName	= "";
 std::string				TempFiles::_statusFileName	= "";
@@ -122,7 +123,6 @@ void TempFiles::deleteOrphans()
 
 	try
 	{
-		long outOfDateDelta = 24 * 3600;
 		std::filesystem::path tempPath		= Utils::osPath(Dirs::tempDir());
 		std::filesystem::path sessionPath	= Utils::osPath(_sessionDirName); 
 		stringvec aliveIDs;
--- a/Common/tempfiles.cpp
+++ b/Common/tempfiles.cpp
@@ -122,9 +122,10 @@ void TempFiles::deleteOrphans()
 
 	try
 	{
-
+		long outOfDateDelta = 24 * 3600;
 		std::filesystem::path tempPath		= Utils::osPath(Dirs::tempDir());
-		std::filesystem::path sessionPath	= Utils::osPath(_sessionDirName);
+		std::filesystem::path sessionPath	= Utils::osPath(_sessionDirName); 
+		stringvec aliveIDs;
 
 		std::filesystem::directory_iterator itr(tempPath, error);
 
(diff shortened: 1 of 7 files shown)


boutinb:
Too tricky for this release. @JorisGoosen should also have a look.

boutinb:
This PR should go to next release. @RensDofferhoff could you solve the conflicts?

RensDofferhoff:
Certainly. I will rebase.

RensDofferhoff:
Okay Resolved @boutinb

Common/tempfiles.cpp
```diff
@@ -122,9 +122,10 @@ void TempFiles::deleteOrphans()
 
 	try
 	{
-
+		long outOfDateDelta = 24 * 3600;
```
JorisGoosen: This is never changed right?
Maybe make it a const member of tempfiles?

Desktop/mainwindow.cpp
```diff
@@ -239,8 +239,11 @@ MainWindow::~MainWindow()
 
 		delete _resultsJsInterface;
 
-		if (_package->hasDataSet())
-			_package->reset();
+//probably to late for signals
```
JorisGoosen:
```suggestion
//probably too late for signals
```

Engine/engine.cpp
```diff
@@ -133,12 +133,8 @@ void Engine::initialize()
 
 Engine::~Engine()
 {
-	TempFiles::deleteAll();
```
JorisGoosen, line 136: ooh yeah that is a good remove :p

Desktop/engine/enginesync.cpp
```diff
@@ -94,6 +94,9 @@ EngineSync::~EngineSync()
 	_moduleEngines.clear();
 	_engines.clear();
 
+	for(auto* channel : _channels)
```
JorisGoosen, line 97: I guess it is a bit cleaner to actually clean them up...

JorisGoosen (approved):
So i havent tested this extensively, but the code changes look really sound and I think we will hear soon enough if this breaks JASP after its in the nightlies

JorisGoosen:
And I think this is a good time to merge it anyway.

RensDofferhoff:
This is from quite a while ago. It might conflict with some of your changes in development @JorisGoosen.

JorisGoosen:
I noticed, but Ill handle that when I rebase development on stable again in a while

JorisGoosen (approved):

PR #5002, opened 2023-01-19
Fix long pause crashes.

RensDofferhoff:
For an actual Fix of the underlying problem see PR: https://github.com/jasp-stats/jasp-desktop/pull/5001
There are many issues with cleanup.
Trash may mount in the GBs within the tmp folder when large datasets are used on MacOS.
But this gets rid of the crashes within a 24 hour window and possibly beyond.

Fixes: https://github.com/jasp-stats/jasp-test-release/issues/2241

### DIFF
--- a/Desktop/engine/enginesync.cpp
+++ b/Desktop/engine/enginesync.cpp
@@ -306,7 +306,7 @@ void EngineSync::start(int )
 	connect(timerBeat,		&QTimer::timeout, this, &EngineSync::heartbeatTempFiles,	Qt::QueuedConnection);
 
 	timerProcess->start(50);
-	timerBeat->start(30000);
+	timerBeat->start(50);
 }
 
 void EngineSync::restartEngines()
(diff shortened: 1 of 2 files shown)


Desktop/engine/enginesync.cpp
```diff
@@ -306,7 +306,7 @@ void EngineSync::start(int )
 	connect(timerBeat,		&QTimer::timeout, this, &EngineSync::heartbeatTempFiles,	Qt::QueuedConnection);
 
 	timerProcess->start(50);
```
boutinb, line 308: Why the timerBeat should be so often?

RensDofferhoff: It helps after wake up so files don't get deleted. But since the limit is 24 hours now it's not necessarily needed.
But on the other hand 50 milliseconds is still a lifetime so it doesn't matter for performance.

boutinb (approved):

PR #5010, opened 2023-01-31
Repeated Measures ANOVA fix

### LINKED ISSUE #1987: [Bug]: Repeated measures ANOVA does not run
### JASP Version

0.17.0

### Commit ID

_No response_

### JASP Module

ANOVA

repeated measures ANOVA

macOS Intel, macOS Silicon

### Bug Description

After selecting the variables involved in a RM ANOVA, the tables pop up but no results are shown.  This happens on both the Intel and Apple Silicon version of the application.  No such issues on 16.4.  I only noticed it on 17.  The issue is replicable and can be demonstrated on different Macs even after an uninstall and reinstall.

### Expected Behaviour

Data should appear on the tables.

### Steps to Reproduce

1. ANOVA
2. Repeated Measures ANOVA
3. Select variables for analysis

Note bottom of screenshot shows a non-RM ANOVA, which runs properly.
...
Screen Shot 2023-01-30 at 8 44 44 AMGMT-6

### Log (if any)

_No response_

### Final Checklist

boutinb:
Fixes https://github.com/jasp-stats/jasp-issues/issues/1987

There were 2 problems:
. when the Repeated Measures ANOVA is started, it starts with one default factor. This factor should be set per default in the Model terms. This does not work anymore.
. when changing the name of a factor, this factor and its interactions with other variables should stay in the model terms (if they already there). The factor (and its interactions) are now removed when changing its name: this must be also fixed

### DIFF
--- a/QMLComponents/analysisform.cpp
+++ b/QMLComponents/analysisform.cpp
@@ -366,7 +366,7 @@ void AnalysisForm::bindTo(const Json::Value & defaultOptions)
 			// As their assigned models are not yet bound, resetTermsFromSourceModels (with updateAssigned argument set to true) must be called afterwards.
 			if (availableModel)
 			{
-				if (defaultOptions != Json::nullValue || _analysis->isDuplicate())
+				if (defaultOptions.size() != 0 || _analysis->isDuplicate())
 					availableModel->resetTermsFromSources(false);
 				else
 					availableModelsToBeReset.push_back(availableModel);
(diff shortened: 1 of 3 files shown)


vandenman (approved):
I like how you added more comments than code :stuck_out_tongue: 

The ANOVA works again as intended, but I cannot judge whether the changes make sense.

RensDofferhoff (approved):

PR #5011, opened 2023-01-31
Catch exception on version parse failure

RensDofferhoff:
Should Mitigate: https://github.com/jasp-stats/jasp-issues/issues/1986

I could not reproduce the issue on any platform with all kinds of network shenanigans.
So I just catch the exception and leave it at that.

### DIFF
--- a/Desktop/gui/jaspversionchecker.cpp
+++ b/Desktop/gui/jaspversionchecker.cpp
@@ -50,14 +50,20 @@ void JASPVersionChecker::downloadVersionFinished()
 
 	if(version != "")
 	{
+		try
+		{
+			Version cv		= AppInfo::version,
+					lv		= version.toStdString();
+			long	cur		= cv.major()*1000000 + cv.minor()*100000 + cv.release()*1000 + cv.fourth(),
+					latest	= lv.major()*1000000 + lv.minor()*100000 + lv.release()*1000 + lv.fourth();
 
-		Version cv		= AppInfo::version,
-				lv		= version.toStdString();
-		long	cur		= cv.major()*1000000 + cv.minor()*100000 + cv.release()*1000 + cv.fourth(),
-				latest	= lv.major()*1000000 + lv.minor()*100000 + lv.release()*1000 + lv.fourth();
-
-		if (latest > cur)
-			emit showDownloadButton(downloadfile);
+			if (latest > cur)
+				emit showDownloadButton(downloadfile);
+		}
+		catch(std::runtime_error& e)
+		{
+			Log::log() << "Unable to parse version number:\n " << e.what() << std::endl;
+		}
 	}
 
 	if(KnownIssues::issues()->downloadNeededOrLoad())	downloadKnownIssues();
--- a/Desktop/gui/jaspversionchecker.cpp
+++ b/Desktop/gui/jaspversionchecker.cpp
@@ -73,7 +73,14 @@ void JASPVersionChecker::downloadVersionFinished()
 void JASPVersionChecker::downloadIssuesFinished()
 {
 	Log::log() << "New version of knownIssues downloaded!" << std::endl;
-	KnownIssues::issues()->loadJson(_networkReply->readAll().trimmed(), true);
+	try
+	{
+		KnownIssues::issues()->loadJson(_networkReply->readAll().trimmed(), true);
+	}
+	catch (...)
+	{
+		Log::log() << "Something went wrong parsing the known issues" << std::endl;
+	}
 
 	deleteLater(); //Remove yourself!
 }


boutinb (approved):

boutinb:
I think more exception handling is needed: the version is also read from JASP file, from Description.qml file... any malformed version string will crash JASP. Also any malformed description.qml file will crash JASP...
I have made another issue for this: https://github.com/jasp-stats/INTERNAL-jasp/issues/2235

JorisGoosen:
But now nothing is done with a "refresh" reply like the user got in the issue no?

RensDofferhoff:
> But now nothing is done with a "refresh" reply like the user got in the issue no?

That's true but I supposed it's a rather rare event and not worth adding handling for. But if you disagree I can take a look

boutinb:
> > But now nothing is done with a "refresh" reply like the user got in the issue no?
> 
> That's true but I supposed it's a rather rare event and not worth adding handling for. But if you disagree I can take a look

Maybe we could catch the redirected signal: https://doc.qt.io/qt-5/qnetworkreply.html#redirected
These people may never hear about a new JASP version. That can be a pity.

JorisGoosen:
> Maybe we could catch the redirected signal: [doc.qt.io/qt-5/qnetworkreply.html#redirected](https://doc.qt.io/qt-5/qnetworkreply.html#redirected)
These people may never hear about a new JASP version. That can be a pity.

Exactly! And this looks like an easy thing to add

RensDofferhoff:
> > Maybe we could catch the redirected signal: [doc.qt.io/qt-5/qnetworkreply.html#redirected](https://doc.qt.io/qt-5/qnetworkreply.html#redirected)
> > These people may never hear about a new JASP version. That can be a pity.
> 
> Exactly! And this looks like an easy thing to add

Will do!

boutinb:
I had a look further, and I see https://doc.qt.io/qt-6/qnetworkrequest.html#RedirectPolicy-enum, so a redirect http -> http should be automatic (NoLessSafeRedirectPolicy is the default). So we can still try to connect this redirect signal, but it does not seem to be the solution...

RensDofferhoff:
> I had a look further, and I see https://doc.qt.io/qt-6/qnetworkrequest.html#RedirectPolicy-enum, so a redirect http -> http should be automatic (NoLessSafeRedirectPolicy is the default). So we can still try to connect this redirect signal, but it does not seem to be the solution...

Its probably because this is a html meta pragma that browsers use, not a proper redirect by http code 301/304. It's not part of http at all (status is 200)
See 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections
Do we still want to deal with it @JorisGoosen @boutinb?  I'm not sure this occurs all that much

JorisGoosen:
Hmm, well if it isnt a proper redirect at all... Then lets drop it for now.

Although I do hope we don't end up with the situation where lots of people dont know about new versions :p

JorisGoosen:
One problem I see with whats in this PR is that the logging at:
`Log::log() << "Unable to parse version number:\n " << e.what() << std::endl;` doesnt actually include the "version number" aka the html output. So even if we get a log from someone with this problem we couldnt even see whether there was such a weird response or not right?

As in the exception is a standard one no? So only the one number that fails is in it I guess

shun2wang:
Ok, seems good and I cannot reproduce crash with nightly build. logs now like this:

```
......
00:00:02.431: Unable to parse version number:
 Module version not understood: <html>
<head>
<meta http-equiv='refresh' content='1; url=http://static.jasp-stats.org/JASP-Version.txt?cmd=redirect&arubalp=12345'>
</head>
</html>
00:00:03.323: MainWindow::resultsPageLoaded()
00:00:03.324: ResultsJsInterface::dequeueJsQueue has 1 waiting Js-scripts.
00:00:03.324: Running delayed script: 'window.setSelection(false);'.
00:00:06.369: EngineRepresentation::processEngineResumedReply() for engine #0
00:00:06.369: EngineRepresentation::sendSettings()
00:00:42.358: AsyncLoader::loadPackage(asyncloader)
00:00:42.359: SharedMemory::retrieveDataSet(17528)
00:00:42.457: Created shared mem with name JASP-DATA-17528
00:00:42.457: (Re)created dataset in shared mem with name JASP-DATA-17528
00:00:42.718: DataSetPackage::setDatabaseJson got:null
00:00:42.814: Loading analyses from jasp-file, entering loop.
00:00:42.814: Analyses::createFromJaspFileEntry
00:00:42.814: Checking if there are upgrade options for module 'jaspAnova' with function 'AnovaRepeatedMeasures' and version '0.17'!
00:00:42.814: Nope, no upgrades to be done.
......

```

PR #5014, opened 2023-02-03
Add translation strings

shun2wang:
Not sure if it's reasonable using `tr( )` or `QObject::tr( )` here, but I just wanted to point out that it is possible to become translatable here:)

### DIFF
--- a/Desktop/modules/analysisentry.cpp
+++ b/Desktop/modules/analysisentry.cpp
@@ -54,7 +54,7 @@ Json::Value AnalysisEntry::getDefaultResults() const
 	res["title"]			= title();
 	res[".meta"]			= Json::arrayValue;
 	res["notice"]			= Json::objectValue;
-	res["notice"]["title"]	= "Waiting for initialization of module: " + dynamicModule()->title();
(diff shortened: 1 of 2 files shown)


Desktop/modules/analysisentry.cpp
```diff
@@ -54,7 +54,7 @@ Json::Value AnalysisEntry::getDefaultResults() const
 	res["title"]			= title();
 	res[".meta"]			= Json::arrayValue;
 	res["notice"]			= Json::objectValue;
-	res["notice"]["title"]	= "Waiting for initialization of module: " + dynamicModule()->title();
+	res["notice"]["title"]	= tr("Waiting for initialization of module: ") + dynamicModule()->title();
```
boutinb: Shouldn't it be better with an argument '%s'?

boutinb (approved):

Desktop/modules/analysisentry.cpp
shun2wang: yeah you are right it's should be better...

PR #5015, opened 2023-02-06
Move preprocess of %HELP_FOLDER% in markdown to HelpModel::setMarkdown

JorisGoosen:
Otherwise its only applied to the helpfile-mds.

This should be a fairly harmless change that could be useful for having people switch over to using `info:`

### DIFF
--- a/Desktop/utilities/helpmodel.cpp
+++ b/Desktop/utilities/helpmodel.cpp
@@ -66,6 +66,9 @@ void HelpModel::setMarkdown(QString markdown)
 		_pagePath = "";
 	
 	_markdown = markdown;
+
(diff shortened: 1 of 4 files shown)


Desktop/utilities/helpmodel.cpp
```diff
@@ -157,13 +161,11 @@ void HelpModel::showOrToggleParticularPageForAnalysis(Analysis * analysis, QStri
 	else 
 	{
 		_analysis = analysis;
```
boutinb, line 163: `setAnalysis(analysis) ` could be added as first line of the function.

boutinb (approved):
Just a small suggestion

PR #5020, opened 2023-02-09
Db data set

JorisGoosen:
Implements https://github.com/jasp-stats/INTERNAL-jasp/issues/1194


JorisGoosen:
So reviews would be nice ^^

JorisGoosen:
@RensDofferhoff if I load debugdataset and check whether the missing values are correct:
image

So im not sure why it isnt working with your compile?
Did you manually make the data as opposed to loading it maybe?

JorisGoosen:
Ah yes 
image

JorisGoosen:
Fixed it
afbeelding

Problem was that SQLITE couldnt properly store nan/inf

Tools/CMake/Modules.cmake
boutinb, line 1: Set right alphabetical order

QMLComponents/controls/jaspcontrol.h
```diff
@@ -10,7 +10,7 @@
 class AnalysisForm;
 class JASPListControl;
 class BoundControl;
-class ComputedColumn;
+class Column;
```
boutinb: Forward def not needed anymore apparently

QMLComponents/components/JASP/Theme/Theme.qml
```diff
@@ -61,7 +61,7 @@ JaspTheme
 	fontRCode.weight:									Font.Normal
 
 	fontCode.family:									preferencesModel.codeFont
-	fontCode.pixelSize:									Math.round(10 * uiScale)
+	fontCode.pixelSize:									Math.round(11 * uiScale)
```
boutinb, line 64: Why?

JorisGoosen: Because it was supertiny

Tools/CMake/Modules.cmake
```diff
@@ -19,38 +19,39 @@
...
+  "jaspLearnBayes"
+  "jaspLearnStats"
+  "jaspMachineLearning"
+  "jaspMetaAnalysis"
+  "jaspNetwork"
+  "jaspQualityControl"
+  "jaspPower"
+  "jaspPredictiveAnalytics"
+  "jaspProphet"
+  "jaspReliability"
```
JorisGoosen:
```suggestion
  "jaspPower"
  "jaspPredictiveAnalytics"
  "jaspProphet"
  "jaspQualityControl"
  "jaspReliability"
```

boutinb:
Add row on the last row, or add column after the last column or before the first column crash

boutinb:
I can open a file of 250 MB without problem (it is quicker than in 0.17.1), but closing the dataset is really long. I think the dbDelete function can be more efficient.

JorisGoosen:
> I can open a file of 250 MB without problem (it is quicker than in 0.17.1), but closing the dataset is really long. I think the dbDelete function can be more efficient.

Good point, I didnt really look into this (just killed JASP every time so didnt really notice :p)

boutinb:
This branch could not be compiled anymore. I've just fixed it ([Suppress _info redefinition in columnsModel.cpp](https://github.com/jasp-stats/jasp-desktop/pull/5020/commits/16da84b813a208e82760423f677374a1622f7512), but now JASP crashes when it starts...

JorisGoosen:
Ah you should not have removed the definition in the cpp.
Instead you should have made it static again in ColumnsModel.h

boutinb:
> Ah you should not have removed the definition in the cpp.
> Instead you should have made it static again in ColumnsModel.h

I set it as it is in stable.

JorisGoosen:
Ah exporting to csv also seems to have a small problem.
And the crash on insert is only for loaded datasets

boutinb:
I can now add columns on an existing datafile.
But adding a row still deletes values of some columns

JorisGoosen:
New commit should fix that ^^

Ill have a look at dbDelete

JorisGoosen:
Reminder to myself: be sure to check loading from an external database

boutinb:
Work better, but deleting a row remove data in some columns.

JorisGoosen:
Haha, well then I know what I will start with tomorrow :p

JorisGoosen:
Hmm, you cant actually click a cell twice for editing, you first need to click something else...

Desktop/qquick/datasetview.cpp
```diff
@@ -884,39 +903,59 @@ void DataSetView::updateExtraColumnItem()
...
+	
+	if(createItem && !(_prevEditRow == -1 || _prevEditCol == -1))
+	{
+		Log::log() << "Restoring text item for old edit item at " << _prevEditRow << ", " << _prevEditCol << std::endl;
+		QQuickItem * item = createTextItem(_prevEditRow, _prevEditCol);
+
+		size_t col=_prevEditCol, row=_prevEditRow;
+
+		//A delay might help the focus problem? No it doesnt...
+		QTimer::singleShot(10, _cellTextItems[col][row]->item, [col, row, this](){ _cellTextItems[col][row]->item->forceActiveFocus(); });
```
boutinb:
```suggestion
		QTimer::singleShot(10, _cellTextItems[col][row]->item, [col, row, this](){ if (_cellTextItems.contains(col) && _cellTextItems[col].contains(row) && _cellTextItems[col][row]->item) _cellTextItems[col][row]->item->forceActiveFocus(); });
```

boutinb:
Let's merge it!

boutinb (approved):

PR #5025, opened 2023-02-10
ALT Navigation should be in QMLComponents

boutinb:


RensDofferhoff (approved):
Its a bit awkward to move it here but I suppose we don't have a choice when using the components outside of jasp.

PR #5039, opened 2023-03-01
Add tooltip for checkbox and radiobutton

shun2wang:
This may fix tooltip text was not showing in preference buttons,such as [here](https://github.com/jasp-stats/jasp-desktop/blob/03ed0bfb070731694e7f4ef8ab650e8d3c5000aa/Desktop/components/JASP/Widgets/FileMenu/PrefsUI.qml#L140)

### DIFF
--- a/QMLComponents/components/JASP/Controls/CheckBox.qml
+++ b/QMLComponents/components/JASP/Controls/CheckBox.qml
@@ -49,6 +49,7 @@ CheckBoxBase
 			property alias	columns:				childControlsArea.columns
 			property bool	enableChildrenOnChecked: true
 			property bool	forwardKeys:			false
+			property string toolTip: 				""
 
 	function click()	{ control.toggle(); }
 	function toggle()	{ control.toggle(); }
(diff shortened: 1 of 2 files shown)


shun2wang:
i just want to make sure can this into 0.17.2? Not urgent but minor fixes for gui things (*^_^*)

RensDofferhoff:
We have a different way to generate tooltips for JASP controls which is seemingly broken. I will discuss with Joris.

JorisGoosen:
Aah yes `ToolTip` is already integrated into JASP another way. If it is broken we should really fix another way...

This ought to be the globally used single and themed ToolTip for JASP:
https://github.com/jasp-stats/jasp-desktop/blob/f8809601c8c0410ef9e03dc2e72bc344a24c9525/Desktop/components/JASP/Widgets/MainWindow.qml#L259

shun2wang:
OK，thanks. I will close this for now.

shun2wang:
@RensDofferhoff @JorisGoosen But if I look at other JASP control (MenuArrowButton, SpinBox...) seems also using the method what I used. what's the difference or if I missing something here😀

JorisGoosen:
Well, those two arent JASPControl based...
In any case, even your hacky way doesnt seem to fix the problem

PR #5044, opened 2023-03-02
Fix for export issue on darktheme

### LINKED ISSUE #2031: [Bug]: Exporting results with dark theme doesn't show correctly: either dark background or light text
### JASP Version

0.17.1

### Commit ID

_No response_

### JASP Module

Unrelated

_No response_

Windows 10

### Bug Description

When the interface is set to dark theme, exporting results is very difficult: more often than not, results are exported with incorrect appearance. I've seen either: dark background with dark text (I think this is mostly when exporting PDF), and light background with light text (I think this is mostly when exporting html). See screenshot here (HTML export):

image

Only graphics are shown correctly, for example:

image

### Expected Behaviour

The export should be the same as when theme is light (i.e. light background, dark text). For now, I've resorted to switch back to light theme when exporting, and set it back to dark after exporting is complete.

### Steps to Reproduce

1. set dark theme in preference > interface
2. export results (either html or pdf)

### Log (if any)

_No response_

### Final Checklist

RensDofferhoff:
Fixes: https://github.com/jasp-stats/jasp-issues/issues/2031

Its an ugly fix since waiting is an anti pattern.
I wait for theme change and unset for results JS.
It doesn't seem to always work for unset somehow, so I add a small wait on the export thread.
Overall not a clean fix but it works.

### DIFF
--- a/Desktop/components/JASP/Widgets/MainPage.qml
+++ b/Desktop/components/JASP/Widgets/MainPage.qml
@@ -282,11 +282,13 @@ Item
 
 				function onExportToPDF(pdfPath)
 				{
-					if(preferencesModel.currentThemeName !== "lightTheme")
-						resultsJsInterface.setThemeCss("lightTheme");
-
-					resultsJsInterface.unselect(); //Otherwise we get the selected analysis highlighted in the pdf...
 					resultsView.printToPdf(pdfPath);
+				}
+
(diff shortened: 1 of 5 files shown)


boutinb (approved):
Not beautiful, but better that than a not working pdf.

PR #5046, opened 2023-03-08
Add hide dependencies to result menu

RensDofferhoff:
Fixes: https://github.com/jasp-stats/INTERNAL-jasp/issues/2254

A lot of changes for something so small.

### DIFF
--- a/Desktop/results/resultmenumodel.cpp
+++ b/Desktop/results/resultmenumodel.cpp
@@ -49,7 +49,6 @@ void ResultMenuModel::_generateCorrectlyTranslatedResultEntries()
 		{	"hasRefreshAllAnalyses",	ResultMenuEntry(tr("Refresh All"),			"hasRefreshAllAnalyses",	"",							"")									},
 		{	"hasShowDeps",				ResultMenuEntry(tr("Show Dependencies"),	"hasShowDeps",				"",							"window.showDependenciesClicked()")	},
 		{	"hasExportResults",			ResultMenuEntry(tr("Export Results"),		"hasExportResults",			"",							"")									},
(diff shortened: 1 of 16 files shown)


Desktop/results/resultmenumodel.cpp
```diff
@@ -126,11 +127,19 @@ void ResultMenuModel::setOptions(QString options, QStringList selected)
 		}
 		else if (key == "hasShowDeps")
 		{
-			if(Settings::value(Settings::DEVELOPER_MODE).toBool())
+			if (Settings::value(Settings::DEVELOPER_MODE).toBool())
 			{
 				//It's developerMode time!
-				entries.push_back(separator);
-				entries.push_back(entry);
+				if (!_dependenciesShown)
```
boutinb: I would have done this like for the notesMenu option: extend the showDependenciesClicked function with a boolean argument, and build the jsFunction string 'showDependenciesClicked(%1)'.arg(_dependenciesShown ? 'true' : 'false'). Them you just need to maintain 1 js function.

boutinb (changes requested):

boutinb:
Your fix does not work correctly. If I open a Descriptive analysis, I click on 'Show dependencies' for the Descriptives Statistics table, then Hide dependencies. Afterwards I select one variable, then the Variables & Split get the green broder:

Screenshot 2023-04-18 at 12 26 33

JorisGoosen (changes requested):
"Hide Dependencies" should only be shown for the one that is actually showing dependencies...

See screenshot with two analyses or two jaspObjects:
image

image

The shown dependencies are for "Descriptive Statistics" table in both cases.

And perhaps I want to see more dependencies for more objects?
Now I have to click twice...

PR #5047, opened 2023-03-10
Fix for options clear upon translation

RensDofferhoff:
Fixes: https://github.com/jasp-stats/INTERNAL-jasp/issues/2262

Its not a nice solution but its pragmatic.
A nice fix would be quite involved due to a delayed signal in description.cpp that other parts seem to require.

### DIFF
--- a/Desktop/analysis/analyses.cpp
+++ b/Desktop/analysis/analyses.cpp
@@ -340,10 +340,10 @@ void Analyses::rescanAnalysisEntriesOfDynamicModule(Modules::DynamicModule * mod
 				Analysis * a = keyval.second;
 
 				//This function is called once after 300 ms upon translation due to delayedUpdate in description
-				//and causes the set analysis options to be cleared without this additional boolean
-				if(a->readyToCreateForm() && !a->justTranslated())
+				//and causes the set analysis options to be cleared without this additional flag check set at the start of translations
+				if(a->readyToCreateForm() && !a->beingTranslated())
 					a->createForm();
-				a->aboutToBeTranslated(false);
+				a->setBeingTranslated(false);
 			}
 		}
 
@@ -767,7 +767,7 @@ void Analyses::prepareForLanguageChange()
 {
 	applyToAll([&](Analysis * a)
 	{ 
-		a->aboutToBeTranslated(true);
+		a->setBeingTranslated(true);
 		a->setRefreshBlocked(true); 
 		a->abort();
 	});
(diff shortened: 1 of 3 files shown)


Desktop/analysis/analyses.cpp
```diff
@@ -339,8 +339,9 @@ void Analyses::rescanAnalysisEntriesOfDynamicModule(Modules::DynamicModule * mod
 			{
 				Analysis * a = keyval.second;
 
-				if(a->readyToCreateForm())
+				if(a->readyToCreateForm() && !a->justTranslated())
```
boutinb: I'm not sure to understand this fix... I see a risk that the form is not created at all...

boutinb (commented):
You have to put more explanation here: is the `rescanAnalysisEntriesOfDynamicModule` function called several times? If there is a race issue, which one is it?

RensDofferhoff:
> You have to put more explanation here: is the `rescanAnalysisEntriesOfDynamicModule` function called several times? If there is a race issue, which one is it?

When _qml->retranslate(); is called eventually rescanAnalysisEntriesOfDynamicModule is called exactly once after a delay of 300ms and a bunch of signals around the alteration of the description.
A new form is created and options are cleared.

There is no nicer solution due to the 300 ms delay. You cant store and load the option state or clear a boolean upon the currentLanguageChanged signal because this signal is emitted way before rescanAnalysisEntriesOfDynamicModule is called.

The only way to get a clean solution is to get rid of the 300 ms delay but other parts seem to depend on it.
Therefore we set a boolean upon prepareForLanguageChange signal and clear it in rescanAnalysisEntriesOfDynamicModule.

Desktop/analysis/analysis.h
```diff
@@ -241,7 +242,8 @@ public slots:
 	bool						_isDuplicate					= false,
 								_wasUpgraded					= false,
 								_tryToFixNotes					= false,
-								_hasReport						= false;
+								_hasReport						= false,
+	_aboutToBeTranslated					= false;
```
JorisGoosen:
```suggestion
								_aboutToBeTranslated					= false;
```

Desktop/analysis/analyses.cpp
JorisGoosen: I dont really understand why the form would not be there?
Is `checkAnalysisEntry()` returning `false`?

JorisGoosen:
> A new form is created and options are cleared.

Where did the old one go?

Desktop/analysis/analyses.cpp
RensDofferhoff: Its returning true.  therefore we reach the else statement

RensDofferhoff:
> > A new form is created and options are cleared.
> 
> Where did the old one go?

its destroyed in createForm()

boutinb (approved):

PR #5048, opened 2023-03-13
Make sure install to ProgramFiles64Folder

shun2wang:
Related to https://github.com/jasp-stats/jasp-issues/issues/1974

I've had this problem before, maybe more efficient to do like this. although I think it's not a big deal to install in `C:\Program Files (x86)\ ` path.

While testing, also need to pay attention to whether msi installer can find the existing old version installation when uninstalling and upgrading:-)

### DIFF
--- a/Tools/wix/JASP.wxs
+++ b/Tools/wix/JASP.wxs
@@ -9,7 +9,7 @@
 	<?define ApplicationName="JASP $(var.VersionNumber)" ?>
 
 	<Product Id="*" Name="$(var.ApplicationName)" Language="1033" Version="$(var.VersionNumber)" Manufacturer="Universiteit van Amsterdam" UpgradeCode="$(var.UpgradeCode)">
-		<Package InstallerVersion="500" Compressed="yes" InstallScope="perMachine"/>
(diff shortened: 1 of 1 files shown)


boutinb (approved):

boutinb:
Unfortunately, this does not work... it does not harm either.

shun2wang:
I can confirm that this setup is available. the old [build](https://github.com/jasp-stats/jasp-desktop/commit/70c4d961819b4a8deaec9348cc75d65bf8eb46b6) will be (snapshot from a msi tool like advanced installer):
advinst_PBQmXBiV3O
sample [latest build](https://github.com/jasp-stats/jasp-desktop/commit/37d1113eced66f1721ce8fb5773a6d8b613959da)  was:
advinst_K45s5LkqWZ

But we have to more setup because it will install to `C:\Program Files (x86)\` and pack from this path, so the `JASP-*.msi `will not install to `C:\Program Files\` on user's PC.

shun2wang:
Or can I get some buildbot build logs?

boutinb:
Here is the log:
[builbotlog.txt](https://github.com/jasp-stats/jasp-desktop/files/11123035/builbotlog.txt)

PR #5049, opened 2023-03-14
Fix helpbutton in RM ANOVA

RensDofferhoff:
Fixes: https://github.com/jasp-stats/INTERNAL-jasp/issues/2265

ANOVA declares a property "analysis", due to this the form property "analysis" can not be reached by internal components (it also defines an id "form" :( ).

We should discourage the use of names in Modules of form properties we reference in QML components or we should alias them as jaspPropertyName in form.qml as I propose in this PR.

### DIFF
--- a/QMLComponents/components/JASP/Controls/Form.qml
+++ b/QMLComponents/components/JASP/Controls/Form.qml
@@ -31,6 +31,8 @@ AnalysisForm
 	
 	default property alias	content		: contentArea.children
 	property alias	form				: form
+	property alias	jaspForm			: form
+	property alias	jaspAnalysis		: form.analysis
 	property alias	_contentArea		: contentArea
 	property real	minimumYMsgs		: warningMessagesBox.height + warningMessagesBox.y
 	property int	majorVersion		: 1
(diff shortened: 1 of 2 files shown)


shun2wang:
>We should discourage the use of names in Modules of form properties we reference in QML components

I think so, duplicated root id may cause other unexpected problems in analysis control, should we add this sample into Dev Docs?

RensDofferhoff:
> > We should discourage the use of names in Modules of form properties we reference in QML components
> 
> I think so, duplicated root id may cause other unexpected problems in analysis control, should we add this sample into Dev Docs?

I will discuss it with @boutinb . Sounds like a good idea but on the one hand just an alias used for components is probably better since "form" and "analysis"  are arguably natural names to use by  the analysis creators.

boutinb (approved):
`analysis` is used only here in QML analyses. This is quite unfortunate that `analysis` is used as another property for the ANOVA analyses.
I find quite a good idea to set some general properties at the beginning of an analysis, so we should not disapprove this.
I guess to change the `analysis` property to `jaspAnalysis` is the best thing to do.

For the `form` property I have made another issue. Can be done later.

PR #5055, opened 2023-03-31
Update supported win OS for installer

shun2wang:
Sould avoid https://github.com/jasp-stats/jasp-issues/issues/2061 and https://github.com/jasp-stats/jasp-issues/issues/1822 during install step but not a crash or something.

Could you trigger a build for this before merge? I will have a test on VM for OS support, I not have a arm CPU...


shun2wang:
not sure why it closed when I push force from Github Codespace.../(ㄒoㄒ)/~~
If anyone knows why please let me know...

boutinb:
Hi @shun2wang your commit contains 1501 files!

shun2wang:
I pushed this from Github_codespace which will sync all project files when push to remote force...so it was an accidents...

PR #5060, opened 2023-04-14
R syntax in results

### LINKED ISSUE #2008: [Feature Request]: Print R code in output
### Description

Print the R code as used in Syntax Mode in the output screen

### Purpose

code review; transparancy without the need for data sharing (i.e., sharing the JASP-file)

### Use-case

Medical settings, where data sharing isn't always viable

no

Unrelated

Print the R code as used in Syntax Mode in the output screen

copypasting the "syntax" in a note above the analysis

### Additional context

First of all, the Syntax Mode that was added in version 0.17 is fantastic. I think it was one of the most longed-for functionalities among the users and has been called the top priority by JASP developers for a while. It's nice that you can not only see the R code, but edit it as well. 

I was wondering if it's possible to show the used R code in the output as well. Usually, you would say that sharing the .jasp file is sufficient for transparancy and reproducibility, however sometimes you can't share your data (e.g., when you have medical data and/or case anonimity cannot be achieved because of the nature/contents of the dataset). If you then still want to maintain transparancy, not only sharing your output, but also sharing the underlying code might be a good thing. Also, this might be demanded by a journal editor or a teacher who is grading your statistics assignment - they often want to see whether your reasoning was ok without the need to repeat the analyses themselves.

Of course, what you can do, is add a note above the analysis output, and paste the used syntax there. But with many analyses, and checking and unchecking boxes repeatedly in the test pane which obviously changes the code at every change, this might become quite a hassle. So I was wondering if this could be done automatically, perhaps even in a nice command-line-like font (comparable to the syntax in SPSS output, and RMarkdown/Quarto code chunks).

boutinb:
Fixes https://github.com/jasp-stats/jasp-issues/issues/2008

R Syntax can be now displayed in the results, either via the popup menu of the title of the result, or via the preferences/Results settings. This displays (or hides) the R syntax for all analyses.

This works in light and dark theme, and the syntax is also exported in HTML or PDF.

### DIFF
--- a/Desktop/html/js/jaspwidgets.js
+++ b/Desktop/html/js/jaspwidgets.js
@@ -972,6 +972,7 @@ JASPWidgets.Toolbar = JASPWidgets.View.extend({
 			hasRemoveAllAnalyses:	parent.menuName			=== 'All',
 			hasRefreshAllAnalyses:	parent.menuName			=== 'All',
 			hasExportResults:		parent.menuName			=== 'All',
+			hasShowRSyntax:			parent.menuName			=== 'All',
 
 			objectName:				parent.menuName
 		};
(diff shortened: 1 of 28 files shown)


RensDofferhoff:
Looks good! But it does mess with the auto scroll in the results screen.

Desktop/html/js/jaspwidgets.js
RensDofferhoff, line 1: If you remove `window.scrollIntoView(self.$el,` function () {});` the behaviour is correct again as far as I can see

Desktop/html/js/jaspwidgets.js
```diff
@@ -1009,6 +1010,97 @@ JASPWidgets.Toolbar = JASPWidgets.View.extend({
...
+	},
+	setVisibility: function (value) {
+		var self = this;
+		self.$el.css("opacity", value ? 0 : 1);
+
+		if (value === true) {
+			self.$el.slideDown(200, function () {
+				self._setVisibility(value);
+				self.$el.animate({ "opacity": 1 }, 200, "easeOutCubic", function () {
+					window.scrollIntoView(self.$el, function () {});
```
RensDofferhoff, line 1039: If you remove window.scrollIntoView(self.$el, function () {});` the behaviour is correct again as far as I can see

RensDofferhoff (approved):

PR #5063, opened 2023-04-17
Fix JASP note editor for link input

shun2wang:
This was discovered when I introduced the Latex Formula editor (almost complete and this will also be effective for it), and I think I can move this fix to the next release?

Before: not work
After:
image

### DIFF
--- a/Desktop/html/js/jaspwidgets.js
+++ b/Desktop/html/js/jaspwidgets.js
@@ -537,7 +537,7 @@ JASPWidgets.NoteBox = JASPWidgets.View.extend({
 
         	// Change example link from quilljs.com to a sample link
         	var linkInput = quillTooltipTheme.root.querySelector('input[data-link]');
-        	linkInput.dataset.link = 'https://example.com';
+        	linkInput.dataset.link = 'https://jasp-stats.org';
 
 		// Add tooltips to the toolbar buttons
 		// Quilljs website mentions changing the toolbar html element (https://quilljs.com/playground/#snow-toolbar-tooltips),
--- a/Desktop/html/js/jaspwidgets.js
+++ b/Desktop/html/js/jaspwidgets.js
@@ -10,6 +10,12 @@ if(insideJASP)
 
 
 	Quill.register('formats/linebreak', LineBreakClass);
+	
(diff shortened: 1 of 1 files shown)


RensDofferhoff (approved):
Great find!

Desktop/html/js/jaspwidgets.js
```diff
@@ -525,6 +531,13 @@ JASPWidgets.NoteBox = JASPWidgets.View.extend({
 
 		this.$quillToolbar     = this.$el.find(".ql-toolbar").get(0);
 		let quillEditorElement = this.$el.find(".ql-editor").get(0);
+		
+		this.$quillTooltip     = this.$el.find(".ql-tooltip");
+        	var quillTooltipTheme  = this.$quill.theme.tooltip;
+
+        	// Change example link from quilljs.com to a sample link
+        	var linkInput = quillTooltipTheme.root.querySelector('input[data-link]');
+        	linkInput.dataset.link = 'https://example.com';
```
JorisGoosen:
```suggestion
        	linkInput.dataset.link = 'https://jasp-stats.org';
```

^^

JorisGoosen:
Code looks good, I just need to test it

JorisGoosen:
Well, if one forgets to dd https: you get:
image

But that is hardly a problem

JorisGoosen:
I dont see the sample link anywhere anyway though but the links now work so thats good.

shun2wang:
>But that is hardly a problem

This is indeed difficult because we do not know whether the user's link is HTTPS, HTTP, or even a 'mailto'url. or we can set conditions...

But what I find another question when I edit a link and add a new analysis then the new analysis will not work, I can see this error:
image

@RensDofferhoff Do you know why?
 If I set` this.$quillTooltip.is(':visible')` here and mess up something? if so I apologize for introducing a problem, but I am currently not sure why it happened/(ㄒoㄒ)

Maybe another better way may be to modify the focu out to obtain more accurate mouse click positions (such as considering the case of the Quill Tooltip editor)

JorisGoosen:
Well if we dont know what the user want but they entered a link we can just check for lack of protocol and if so add `https://`.

@shun2wang those warnings you see are no problem

PR #5079, opened 2023-05-03
Makes multibridge work on macos Arm!

JorisGoosen:
To use this commit and actually have bug jasp-stats/jasp-test-release#2325 fixed
you need to remove `gmp` and `mpfr` as installed by `brew`.

Then download a snapshot of `gmp`: https://gmplib.org/download/snapshot/gmp-next/gmp-6.2.99-20221117121717.tar.zst
Build with:
./configure --enable-static --disable-shared --with-pic --enable-cxx
make -j
make check #These NEED to pass
sudo make install

Then build `mpfr` and use:
./configure --enable-static --disable-shared --with-pic --with-gmp=/usr/local
make -j
sudo make install

Then to make sure Qt creator can actually find pkg-config (sigh) you need to start it from a terminal

Make sure to `git submodule update --init` too

Then use a fresh buildfolder or remove at least `_cache`, `_deps`, `_scripts`, `Modules`, and `Frameworks`
Run cmake and verify `multibridge` is installed at the end, it will probably give some loading namespace errors during help generation but that can be ignored

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -122,3 +122,7 @@
 	path = Modules/jaspPredictiveAnalytics
 	url = https://github.com/jasp-stats/jaspPredictiveAnalytics
 	branch = main
+[submodule "RPkgs/multibridge"]
+	path = RPkgs/multibridge
+	url = https://github.com/Asarafoglou/multibridge
+	branch = master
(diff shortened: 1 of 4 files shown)


JorisGoosen:
This rather involved and I think it might be better to *not* merge it at all. Or to make a 0.17.2 release branch where we merge this.

This all will not be necessary by the time `gmp` has a new release and this trickles down into CRAN. Dont know how long it takes but at least we can have people use jaspFrequencies without crashing the engine :p

JorisGoosen:
Ill upload my local build

JorisGoosen:
https://static.jasp-stats.org/Nightlies/macOS-arm64/JASP-nightly-multibridgeMacArmWorkaround-2d964ce7adc96bf195c9e4162d0b434b53254eed.dmg

PR #5088, opened 2023-05-11
Quality Control data library

JTPetter:
A data library for the Quality Control module. I have moved "Miscellaneous" to position 19., I think it makes sense to keep it at the bottom.


JTPetter:
JASP files are not annotated yet. This will take a lot of work and I think this is already very useful as it is, so it's probably better to get this out already and then add annotation later.

JorisGoosen:
@JTPetter after you finish making the upgrader and set the QC version to 0.17.2.1 you can perhaps save these again so they have the right version number in there?

JorisGoosen:
and then we could merge em in with the hotfix

JTPetter:
@JorisGoosen Done

JorisGoosen (approved):

PR #5096, opened 2023-05-24
[Do Not Merge] update to R-4.3.1

vandenman:
the installers are on static already

Note that I didn't test this, since this only works for macOs / windows.

### DIFF
--- a/Docs/development/jasp-build-guide-windows.md
+++ b/Docs/development/jasp-build-guide-windows.md
@@ -7,7 +7,7 @@ If you have not cloned the `jasp-desktop` repository, please head back to the [b
 - [Microsoft Visual Studio](https://visualstudio.microsoft.com/downloads/)
 - [Qt Creator](https://www.qt.io/download) / Qt >= 6.2
     - Qt Creator 7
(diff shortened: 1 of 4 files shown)


shun2wang:
I think should also bump version of Rtools42 to Rtools43 because now R4.3.0 only build with this tools version. such [here ](https://github.com/jasp-stats/jasp-desktop/blob/9d22ede735743081fef178922ed02f8d5a908952/R-Interface/CMakeLists.txt#L62-L63)and in docs:-)

EDIT: A suggest is we can update to R-4.3.1 after [2023-6-16](https://www.r-project.org/)

JorisGoosen:
Should we still wait for 4.3.1 ?

JorisGoosen (approved):

vandenman:
I've updated this PR to R-4.3.1, updated the Rtools stuff everywhere, and updated the building guide for Windows. @JorisGoosen it's not entirely clear what needs to be done inside `R-Interface/CmakeLists.txt`, so I've improvised that part. On the other hand, since we tell people in the build guide that they must add Rtools to the path, we could also make them set `RTOOLS4X_HOME` and make our lives easier. Or merge the R-Interface and Desktop projects on windows.

PR #5100, opened 2023-05-25
Add special source for E-Science project

boutinb:
- Add the discardIndex attribute for a source to add the possibility to remove some values from a source (for the E-Science project, the last value of the source must be always removed).
- Add the titles property in ComponentsList: there was no (easy) way to set titles as header of the columns of the ComponentsList


RensDofferhoff (approved):
I see no issue

PR #5102, opened 2023-05-30
Implement R code highlight

shun2wang:
Close: https://github.com/jasp-stats/INTERNAL-jasp/issues/2328

Now looks like:
image

TODO:

- [ ] Highlighting in exported HTML (Not now)
- [x] Highlighting in Help Docs
- [x] Style improvement
- [x] Performance Testing (EDIT: about 16ms for a r code syntax of one  analysis)


EJWagenmakers:
awesome

Desktop/html/index-jasp.html
```diff
@@ -13,6 +13,10 @@
 	<link rel="stylesheet" href="css/lightTheme-jasp.css"					type="text/css" id="style"/>
 	<link rel="stylesheet" href="css/printing.css"							type="text/css" />
 
+	<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/11.7.0/styles/default.min.css">
+	<script src="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/11.7.0/highlight.min.js"></script>
+	<script src="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/11.7.0/languages/r.min.js"></script>
```
boutinb: This is not a good idea to load the libraries from the net: JASP should work without internet, so  it is better to add these libraries in our own resources.

boutinb (changes requested):
Really nice!!!
Just one remark.

Desktop/html/index-jasp.html
shun2wang: yes I know this, thanks for point that. 
I just left it for now to live debugging, once I'm done debugging I'll import from local:-) Because reloading resources from qtwebengine devtool by `qrc://` during live debug phase requires compiling every time.

JorisGoosen:
Could we also use this to highlight the rcode in textareas in the gui?

Ive done some reading in the https://doc.qt.io/qt-6/qsyntaxhighlighter.html and some related pages.
It seems that we cant reuse the js highlighter.

But if we have a regex that can highlight R we could use that.

And looking in the PR I guess the highlighter itself already has some definition inside. Ill have a look at their docs

Edit: https://github.com/highlightjs/highlight.js/blob/main/src/languages/r.js seems quite useful for implementing this later. 
But reading https://doc.qt.io/qt-6/qsyntaxhighlighter.html#highlightBlock indicates it might require a bunch more work.
Better postpone that.

Having R code look awesome in the results-notes will already be very nice.

shun2wang:
> Could we also use this to highlight the rcode in textareas in the gui?

Is there a way to import js directly in QML? However that seems to require using qmake build system? if so sounds terrible🤪

>But if we have a regex that can highlight R we could use that.


I recommend here https://prismjs.com/download.html#themes=prism&languages=r

JorisGoosen:
>  Is there a way to import js directly in QML? However that seems to require using qmake build system? if so sounds terrible🤪

Well it *is* possible to include `js` directly for use in qml it wont then be available to the `QSyntaxHighlighter` without a bunch of C++ trickery. So it wouldnt make things much easier :p

> prism.js

https://github.com/PrismJS/prism/blob/master/components/prism-r.js
It does look slightly cleaner

shun2wang:
Now in help docs:
```
\`\`\`r
if (i == 0) {
  y = TRUE
} else {
  y = FALSE
}
\`\`\`
```
image

So I think it ready to review.
**Legacy :**
Implemented code highlighting for exporting HTML, in fact only needed to come with css styles, but I don't want to "add more css styles inline" right now so it's left to next time I clean up all styles. some of them in https://github.com/jasp-stats/jasp-desktop/pull/5073

Resources/Help/marked.js
```diff
@@ -1,1286 +1,2838 @@
 /**
- * marked - a markdown parser
- * Copyright (c) 2011-2014, Christopher Jeffrey. (MIT Licensed)
- * https://github.com/chjj/marked
+ * marked v5.0.1 - a markdown parser
```
JorisGoosen, line 2: Good idea!

JorisGoosen (changes requested):
Changes look fine and it works splendidly!

But...Dark theme support is still lacking for the rsyntax box:
image

Resources/Help/marked.js
shun2wang: Ah, I did forget the dark mode!! will do.

shun2wang: I made a few simple tweaks to fit dark mode, mostly sticking to the default theme, maybe I'll add a full dark theme later.

JorisGoosen (approved):

JorisGoosen:
Looks good to me ^^

JorisGoosen:
@boutinb you think you can also review this?

boutinb (approved):

PR #5107, opened 2023-06-07
Data library files for range method (long) & annotation type 1 study files

JTPetter:
Added test-retest data example in long format, only merge after https://github.com/jasp-stats/jaspQualityControl/pull/254, because this PR fixes the long format for this analysis, else the analysis is broken.


JTPetter:
@Kucharssim I have also added annotations for the type 1 measurement analysis data example

Kucharssim (approved):

PR #5109, opened 2023-06-12
initial implement LaTeX render using mathJax

shun2wang:
closes: https://github.com/jasp-stats/INTERNAL-jasp/issues/395 
closes: https://github.com/jasp-stats/INTERNAL-jasp/issues/2331

This is attempt to solve the copy-paste consistency after I discussing with @JorisGoosen ,  mathJax provides output rendered as svg, in this way we can use `canvas` stuff, which is useful for online editors (like Google docs). If it does what we want, it will replace the https://github.com/jasp-stats/jasp-desktop/pull/5073
but as I said, mathJax is not lightweight, although its current performance is very good.

we will get a copy as menu for latex...like this:
image


JorisGoosen:
image
 ^^
I made a small tweak to the PR and now we get this

JorisGoosen:
image

JorisGoosen (approved):

JorisGoosen:
@shun2wang I added some code to remove the mml-assistive tags because they showed up in the copy.
Now I can easily copy the results including the pngs to google-docs/libreoffice and it looks awesome

@Kucharssim I gave this a try and it seems to be working splendidly, maybe you want to have a look and see if it does all you expect?

shun2wang:
Hi joris this is what I left on purpose, as I said when I talked to you，I am now not sure if the quality of the png is ok, we can remove it in the wrapHTML() function such:
``` JavaScript
var wrapHTML = function (html, exportParams, doctype = false) {
    var completehtml = ""
    if(doctype)
        completehtml = "<!DOCTYPE HTML>\n"
	completehtml += "<html>\n"
	completehtml += "	<head>\n"
	completehtml += "		<meta http-equiv='Content-Type' content='text/html; charset=utf-8' />\n"
	completehtml += "		<title>JASP</title>"
	completehtml += "		<style>"
	completehtml += "			p {margin-top:1em; margin-bottom:1em;}"
	if (exportParams.isFormatted())
		completehtml += "			body {font-family: sans-serif;}"
	completehtml += "		</style>"
	completehtml += "	</head>\n"

	var styles = JASPWidgets.Exporter.getStyles($("body"), ["display", "padding", "margin"]);

	completehtml += "	<body " + styles + ">\n";

    const htmlObj = $.parseHTML(html);
        $(htmlObj).find('mjx-assistive-mml math').remove();
    const result = $(htmlObj).prop('outerHTML');
    completehtml += result

	completehtml += "	</body>\n"
	completehtml += "</html>";
	return completehtml;
};
```
😉so that someday we can re-enable mathml in exported html to reduce HTML file size

JorisGoosen (changes requested):
Ah, html export isnt good yet.

JorisGoosen:
image

shun2wang:
You can try what I did above, it will work for all cases(*^_^*)

shun2wang:
Ideally, I only want to include png images in the copy-paste. For exported HTML files and PDFs, including svg or mathml is the optimal solution. Unfortunately I haven't found a solution how to do this only in copy-paste.

JorisGoosen:
Awesome, now exporting to pdf, html goes right. Copy also goes right on macos to libreoffice or google docs in chrome.

I also tried copying the latex test table from `jaspTestModule$"test latex"` and that looks pretty good already:
image

It worked fine after I added `\usepackage{amsfonts} ` to the latex manually.
Perhaps we should add that (or another one that contains things like `\mathbb` ) to the "copy to latex" function?

JorisGoosen (approved):

shun2wang:
sorry i just added render for help docs. test cases for it in .md:

```
\\\\(e=mc^2\\\\)

\\\\[a^x\\\\]
```

$$e=mc^2$$

$$a^x$$

Kucharssim (commented):
Looks great! Just two things:

1. The mathJax menu can go outside of the JASP window, which makes it not accessible in some cases. Could we fix that?

Screenshot 2023-06-19 at 08 24 25

2. Now adding math inside of a note does not work: Make a new note, click on formula, type in a formula + `Save`, I get only a new line without the typed in formula.

shun2wang:
>Could we fix that?

Sure,but right now screenreader can't read the whole JASP window content (at least on my Windows)...although it's easy to fix this independent for MathJax.

>Now adding math inside of a note does not work: Make a new note, click on formula, type in a formula + Save, I get only a new line without the typed in formula.

ah, I cannot reproduce this, add formula for note is works on my build. 
image

can @JorisGoosen reproduce that?

Kucharssim:
> Sure,but right now screenreader can't read the whole JASP window content (at least on my Windows)...although it's easy to fix this independent for MathJax.

I was not necessarily asking to implement screen reader - just that the dialog menu can go outside of the visible JASP window, so you cannot access it. It would be also fine to hide options from this menu that are not relevant currently.

shun2wang:
>just that the dialog menu can go outside of the visible JASP window

@Kucharssim OK i see your problem.just a quick question, are you using the build from my branch, because your menus look a little different😛, I didn't enable those menu content for MathJax...

Kucharssim:
> ah, I cannot reproduce this, add formula for note is works on my build.

I tried it again, and it works! Not sure what was the problem before...

shun2wang:
> I tried it again, and it works! Not sure what was the problem before...

Great! maybe a resource cache problems in html/js compiling...

OK, I checked what you said about the menu out of bounds problem, this can be fixed but it may be a bit hard, mathjax's menu is hardcoded in JavaScript (no separate css style file), if I create a separate rule to detection, there is more work to be done here.

In fact I think people can simply to stretching the window of JASP to access it and fortunately the third level is completely within the JASP window, so given the usage of multilevel menus, I probably won't fix it in this stage.

JorisGoosen:
Well, the menu being slightly out of screen is no reason to block this PR in my opinion.

Kucharssim:
I agree, I think we can merge this

PR #5112, opened 2023-06-15
commit edit on editFinished + navigation + name change

RensDofferhoff:
Commit edit on editFinished instead of textEdit, this solves a myriad of issues for scalar columns. 
Alter navigation so cursor can actually be  moved over text content in cells .
Add shift tab and tab

+ small name change.

@boutinb prefers this navigation/editing behavior to the ones seen in excel, libre.
I dont mind it.


JorisGoosen:
Well, that is what I originally had but because the model gets reset in the meantime the commit edit can wreak havock on the data... 
I like the navigation through the textfield, however now JASP crashes pretty bad if you add a column of some kind and then click anything...

Ill try to fix it

JorisGoosen:
the new name is way better btw haha

JorisGoosen (approved):

PR #5116, opened 2023-06-15
Expand data set automatically

boutinb:
Add some virtual rows and columns. As soon a value is set on these virtual cells, the dataset is automatically expanded to contain these new cells.


Desktop/qquick/datasetview.h
```diff
@@ -55,20 +54,23 @@ class DataSetView : public QQuickItem
 	Q_PROPERTY( double					headerHeight			READ headerHeight											NOTIFY headerHeightChanged			)
 	Q_PROPERTY( double					rowNumberWidth			READ rowNumberWidth			WRITE setRowNumberWidth			NOTIFY rowNumberWidthChanged		)
 	Q_PROPERTY( bool					cacheItems				READ cacheItems				WRITE setCacheItems				NOTIFY cacheItemsChanged			)
+	Q_PROPERTY( bool					expandDataSet			READ expandDataSet			WRITE setExpandDataSet			NOTIFY expandDataSetChanged			)
 	Q_PROPERTY( QQuickItem			*	tableViewItem			READ tableViewItem			WRITE setTableViewItem												)
 	Q_PROPERTY( QItemSelectionModel *	selection				READ selectionModel											NOTIFY selectionModelChanged		)
-	Q_PROPERTY(	QModelIndex				selectionStart			READ selectionStart			WRITE setSelectionStart			NOTIFY selectionStartChanged		)
-	Q_PROPERTY(	QModelIndex				selectionEnd			READ selectionEnd			WRITE setSelectionEnd			NOTIFY selectionEndChanged			)
+	Q_PROPERTY(	QPoint					selectionStart			READ selectionStart			WRITE setSelectionStart			NOTIFY selectionStartChanged		)
```
JorisGoosen, line 60: good idea

Desktop/qquick/datasetview.h
```diff
@@ -55,20 +54,23 @@ class DataSetView : public QQuickItem
...
+	friend ExpandDataProxyModel;
+
 							DataSetView(QQuickItem *parent = nullptr);
 
 	static DataSetView *	lastInstancedDataSetView()					{ return _lastInstancedDataSetView; }
 
 	void					setModel(QAbstractItemModel * model);
 	
-	QAbstractItemModel	*	model()								const	{ return _model;					}
+	QAbstractItemModel	*	model()								const	{ return _model->sourceModel();		}
```
JorisGoosen, line 73: ????

JorisGoosen: Why is `model` secretly `sourceModel`?

Desktop/data/expanddataproxymodel.h
```diff
@@ -0,0 +1,54 @@
+#ifndef EXPANDDATAPROXYMODEL_H
+#define EXPANDDATAPROXYMODEL_H
+
+#include <QAbstractItemModel>
+
+class DataSetView;
+
+class ExpandDataProxyModel : public QObject
```
JorisGoosen, line 8: Shouldnt this be a `QAbstractItemModel` or something?
Or why not?

Desktop/qquick/datasetview.h
JorisGoosen: I think Id prefer having both models stored as private variables, `_model` for the actual real model and `_expandModel` for your extension (which works great btw ^^)

Desktop/data/expanddataproxymodel.h
JorisGoosen: If it was we could reuse this for other models, although I guess that isnt to necessary.
But I think this class would also work fine if it *did* use QModelIndex

JorisGoosen (changes requested):
Works pretty reasonably but I have some questions about the architectural choices and also filtering is now broken...

image

Desktop/qquick/datasetview.h
boutinb: It is what I did first (_model & _expandModel as private), but I found it not satisfying: sometimes _model was used, sometimes _expandModel, I found it tricky. So I thought that all the call should go through the expanded Model, that wrapped completely the dataSetTableModel. 
Only the model property uses the dataSetTableModel: this is because the user of the DataSetView must set the dataSetTableModel as property. Maybe we could change the name of this property: innerModel or so....

Desktop/data/expanddataproxymodel.h
boutinb: The problem is that all the calls should not use an index, but a row and column indices.
An index returns an index of the source model: this is still used for the selectionModel, and it is used only here. I could not see how to manage to generate my own indexes that wrapped the source model indexes.

JorisGoosen:
See: https://github.com/jasp-stats/jasp-desktop/tree/extendDataSet
It is based on https://github.com/jasp-stats/jasp-desktop/pull/5120, so after that gets merged I can open a new clean PR.

PR #5118, opened 2023-06-19
Upgrade jQuery library

shun2wang:
The jQuery library is out of date (even 10 years😛), which has security (fortunately we're not a web hosted software) and performance issues for modern browsers (Qt webengine).
All deprecated functions have been refactored and also bump jQuery ui version.


JorisGoosen (approved):
Looks good and works fine!
Thanks for getting this up to date ^^

we can try again in 10 years haha

JorisGoosen:
Ah, but I only now saw this in my application output:
```
js: One of your code blocks includes unescaped HTML. This is a potentially serious security risk.
js: https://github.com/highlightjs/highlight.js/wiki/security
js: The element with unescaped HTML:
js: [object HTMLElement]
```

You know if we can fix/silence this warning?

shun2wang:
this waring is no related to this upgrade but a highlight.js waring from code highlighting.The reason is that we did not escape some html tags.

Yes, we can suppress this warning.

JorisGoosen:
> this waring is no related to this upgrade but a highlight.js waring from code highlighting.The reason is that we did not escape some html tags.

ah ok!

> Yes, we can suppress this warning.

That would be lovely

PR #5119, opened 2023-06-20
Suppress unescaped  html warnings from hljs

shun2wang:
This also avoids the safe risk of unescaped characters.

### DIFF
--- a/Desktop/html/js/jaspwidgets.js
+++ b/Desktop/html/js/jaspwidgets.js
@@ -1101,6 +1101,7 @@ JASPWidgets.RSyntaxView = JASPWidgets.View.extend({
 					.html("<pre><code class='language-r'>" + rScript + "</code></pre>");
 		setTimeout(() => {
 			this.$el.find(".jasp-rsyntax")[0].querySelectorAll('pre code').forEach((el) => {
+				el.innerHTML = el.textContent; //Suppress warnings of html code tags
 				hljs.highlightElement(el);
 			});
 		}, 200);


JorisGoosen:
So this works and the warnings dissappear, but I dont fully understand how this fixes it?
Is the `el.textContent` the parsed version of the HTML or how does this help?

JorisGoosen (approved):

shun2wang:
`el.innerHTML` is such `<pre><code class='language-r'>" + rScript + "</code></pre>` Html element, the `textContent` is a Plain Text of our r code, so that avoid `<>` characters unescaped.

So, This is a trick. The better way is that the r code output from the cpp side should not contain any html tags.(Because there is a newline tag`</br> `in the current r script but a `\n` to replace, I have escaped it above）

JorisGoosen:
Ah we literally replace that to make sure things like newline normally show up.

https://github.com/jasp-stats/jaspBase/blob/b2cdc13c86a0233d0531b8232b921c7aa754dc26/src/jaspHtml.cpp#L29

JorisGoosen:
but this solution is fine by me

shun2wang:
https://github.com/jasp-stats/jaspBase/blob/b2cdc13c86a0233d0531b8232b921c7aa754dc26/src/jaspHtml.cpp#L29

This also fine because without that we must do it in js so both are ok.

shun2wang:
Because that's really the best way to go before there's code highlighting.

PR #5120, opened 2023-06-20
Some fixes and improvements for data editing

JorisGoosen:
- show warnings and stop the proxymodel from crashing all the time
- fix some "wrong model" errors
- added column title and description suppot
- variableswindow improvements so that title and description are editable and also usable for scalar columns
- general fixes

Also one of the main things I tried to do is make sure that data flows through all the proxies down to DataSetPackage so that filtering etc will keep working. Oh yeah, and this PR fixes a bug where filtering stopped working at all


Desktop/components/JASP/Widgets/VariablesWindow.qml
```diff
@@ -80,11 +92,63 @@ FocusScope
 				enabled:			ribbonModel.dataMode
 				selectByMouse:		true
```
boutinb:
```suggestion
padding:			jaspTheme.jaspControlPadding
```

Desktop/components/JASP/Widgets/VariablesWindow.qml
```diff
@@ -80,11 +92,63 @@ FocusScope
...
+			TextInput
+			{
+				id:					columnTitleVariablesWindow
+				text:				labelModel.columnTitle
+				onTextChanged:		if(labelModel.columnTitle != text) labelModel.columnTitle = text
+				color:				jaspTheme.textEnabled
+				font:				jaspTheme.fontGroupTitle
+				enabled:			ribbonModel.dataMode
+				selectByMouse:		true
+
```
boutinb, line 147:
```suggestion
padding:			jaspTheme.jaspControlPadding
```

boutinb (changes requested):
I have to test further, but at least 2 bugs found:

- The Variables Window shows the label/values 2 times
- The Variables types are not well passed to the VariablesList component.

JorisGoosen:
>     * The Variables Window shows the label/values 2 times

Already fixed in https://github.com/JorisGoosen/jasp-desktop/tree/extendDataSet

>     * The Variables types are not well passed to the VariablesList component.

Ill check it out

boutinb:
Qua usability, this is quite strange to be able to change the labels in the Variables Window, but you can't change the column name, title or description as long as you are not in Edit Data mode. I think it's better to allow always this change.

It's also a pity to use the Qt TextInput and not the JASP TextField control, they don't have the same look and feel. But I can change this afterwards.

JorisGoosen:
> Qua usability, this is quite strange to be able to change the labels in the Variables Window, but you can't change the column name, title or description as long as you are not in Edit Data mode. I think it's better to allow always this change.

I agree!
I keep trying to click em :p
Ill change it

> It's also a pity to use the Qt TextInput and not the JASP TextField control, they don't have the same look and feel. But I can change this afterwards.

Yeah sure, good idea!

boutinb:
Hmm I have still 2 times the labels if I scroll to the right.
Screenshot 2023-06-21 at 12 11 55

Also I would set the description field just under the Name/Title TextFields

boutinb:
Also there are no lines above the header cells in the Variables Window

boutinb:
If I go to joris/furtherModelImprovements, it is much better.... so apperently you haven't updated  well upstream/furtherModelImprovements

JorisGoosen:
As i said, I fixed more things in the extendDataSet. Also everything from origin/further.. is in the upstream one

JorisGoosen:
image
is in extendDataSet

PR #5121, opened 2023-06-21
Add Automatic Extended Data

JorisGoosen:
Do not use the view in the model

Remove comment

Delete ResizeDataDialog.qml


Desktop/data/labelmodel.cpp
```diff
@@ -372,6 +372,8 @@ void LabelModel::unselectAll()
 
 bool LabelModel::setChecked(int rowIndex, bool checked)
 {
+	Log::log() << "setChecked(" << rowIndex << ", " << (checked ? "checked" : "unchecked") << ")" << std::endl;
```
JorisGoosen:
```suggestion
	//Log::log() << "setChecked(" << rowIndex << ", " << (checked ? "checked" : "unchecked") << ")" << std::endl;
```

boutinb (approved):

PR #5124, opened 2023-06-22
Better datafile synchronisation handling

JorisGoosen:
- Now when you open a non-jasp file datafile synchronisation is on by default
- an older jaspfile with a datafile thats is present will synch with it unless edited/etc
- a new jaspfile with sqlite will remember the setting
- if you go back from editing to synchronising it gives the option to reload or generate a new datafile


boutinb (changes requested):
In Preferences/Data. there is still the Data Synchronization setting. It should be removed.

It is much better so!!! It makes the synchronization feature more visible also.

JorisGoosen:
Oops!
Ill remove it ^^

PR #5127, opened 2023-06-23
Add Undo/Redo functionality

boutinb:
This adds Undo/Redo functionality for the Data Edit mode.
The Variables & Filter Windows have no undo/redo functions. This might be added later (the need to have undo/redo for these windows were not high since we did not receive any feature request for this). This would add also more complexity since the Variables Windows is accessible without Data Edit mode.
But for the Data Edit mode the need of a Undo/Redo is more compelling; the user may remove accidentally data.

### DIFF
--- a/CommonData/column.h
+++ b/CommonData/column.h
@@ -162,6 +162,8 @@ public:
 
 			void					checkForLoopInDependencies(std::string code);
 			const	stringset	 &	dependsOnColumns(bool refresh = true);
+			Json::Value				serialize()																const;
+			void					deserialize(const Json::Value& info);
 
 protected:
 			void					_checkForDependencyLoop(stringset foundNames, std::list<std::string> loopList);
--- a/Desktop/data/datasetpackage.cpp
+++ b/Desktop/data/datasetpackage.cpp
@@ -59,6 +59,8 @@ DataSetPackage::DataSetPackage(QObject * parent) : QAbstractItemModel(parent)
 	_labelsSubModel = new SubNodeModel("labels");
 	
 	connect(&_databaseIntervalSyncher, &QTimer::timeout, this, &DataSetPackage::synchingIntervalPassed);
+
+	_undoStack = new UndoStack(this);
 }
 
 DataSetPackage::~DataSetPackage() 
@@ -1272,6 +1274,7 @@ void DataSetPackage::deleteDataSet()
 	
 	delete _dataSet;
 	_dataSet = nullptr;
+	_undoStack->clear();
 }
 
 bool DataSetPackage::initColumnAsScale(size_t colNo, const std::string & newName, const doublevec & values, const std::string & title)
@@ -1701,12 +1704,16 @@ void DataSetPackage::setColumnName(size_t columnIndex, const std::string & newNa
 	if(!_dataSet)
 		return;
 
+	Column* column = _dataSet->column(columnIndex);
+	if (!column)
+		return;
+
 	std::string oldName = getColumnName(columnIndex);
 
 	if(resetModel)
 		beginResetModel();
 
-	_dataSet->column(columnIndex)->setName(newName);
+	column->setName(newName);
 
 	if(resetModel)
 		endResetModel();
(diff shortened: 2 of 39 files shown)


CommonData/column.h
```diff
@@ -162,6 +162,8 @@ class Column : public DataSetBaseNode
 
 			void					checkForLoopInDependencies(std::string code);
 			const	stringset	 &	dependsOnColumns(bool refresh = true);
+			Json::Value				serialize()																const;
```
JorisGoosen, line 165: Would it make sense to make `serialize()` and `deserialize()` virtual functions on `DataSetBaseNode`?

Desktop/data/datasetpackage.cpp
```diff
@@ -1959,6 +1965,21 @@ void DataSetPackage::unicifyColumnNames()
 	}
 }
 
+Json::Value DataSetPackage::getColumn(const std::string& columnName) const
+{
+	Column		*	column	= _dataSet->column(columnName);
+	if (column)
+		return column->serialize();
+	else
+		return Json::nullValue;
```
```suggestion
	return column ? column->serialize() : Json::nullValue;
```

Desktop/data/datasetpackage.h
```diff
@@ -250,6 +250,8 @@ class DataSetPackage : public QAbstractItemModel //Not QAbstractTableModel becau
 				QStringList					getColumnValuesAsStringList(size_t columnIndex)				const;
 				QList<QVariant>				getColumnValuesAsDoubleList(size_t columnIndex)				const;
 				void						unicifyColumnNames();
+				Json::Value					getColumn(const std::string& columnName)					const;
```
JorisGoosen: these function names are bad, I suggest:
```
serializeColumn(
deserializeColumn(
```

Desktop/data/expanddataproxymodel.h
```diff
@@ -3,18 +3,134 @@
 
 #include <QAbstractItemModel>
 #include "utils.h"
+#include <QUndoCommand>
+#include <json/json.h>
 
-class DataSetView;
+class ExpandDataProxyModel;
+
+class UndoModelCommand : public QUndoCommand
```
JorisGoosen: Can these undo model classes be moved to their own header + implementation file? Not necessarely for each one a file but I think they need to be separated from the expanded proxy model nonetheless

JorisGoosen:
Also, if you do:
- open a file, edit a bunch of stuff
- checkout undo redo, it works great
- close file and open another file
- undo stuff...
- you'll see your fresh data getting clobbered

JorisGoosen:
Also, undo redo does not take into account switching to synching with the datafile again.
So you should probably add a copy/diff of the data when switching to the linked datafile and there are changes

JorisGoosen:
I also think ctrl-z and ctrl-shift-z should work in analyses mode when datapanel/varwindow/filter is selected.
But also switch to datamode when doing so to make it clear to the user whats going on

Edit: unless others disagree of course

Desktop/data/expanddataproxymodel.h
```diff
@@ -32,19 +148,33 @@ class ExpandDataProxyModel : public QObject
 	void				removeRow(int row);
 	void				removeColumn(int col);
 	void				insertRow(int row);
-	void				insertColumn(int col);
-	QString				insertColumnSpecial(int col, bool computed, bool R);
+	void				insertColumn(int col, bool computed, bool R);
```
JorisGoosen, line 36: I called it "special" cause it doesnt take into account the proxymodels possible columnfiltering.
This wont bite us now but it is good to keep it in mind.

JorisGoosen:
Another bug is the fact that you do not use proxy rows but datsetpackages rows, this leads to problems, try:
- open debug.csv
- remove, insert and generally mess with rows
- make a filter "V1 > 3"
- make filtered out row invisible by clicking eye
- `(undo/redo)+`
- look at weird result...

JorisGoosen (changes requested):

PR #5130, opened 2023-06-27
Make js/html interface translatable

### LINKED ISSUE #1474: [Bug]: Untranslatable string "Click here to add text"
### JASP Version

0.15

### Commit ID

_No response_

### JASP Module

_No response_

Any analysis, the bug is in jaspDesktop

Ubuntu

### Bug Description

When adding a Note to analysis results, there is a hard-coded English string that cannot yet be translated.

In current development branch, the string is at https://github.com/jasp-stats/jasp-desktop/blob/development/Desktop/html/js/jaspwidgets.js#L375:

```
this.ghostTextDefault = 'Click here to add text';
```

### Expected Behaviour

The string should be translatable.

### Steps to Reproduce

1. Switch to any language other than English.
2. In any analysis result, click on *Add Note*.

### Log (if any)

_No response_

### Final Checklist

shun2wang:
Fix https://github.com/jasp-stats/jasp-issues/issues/1474

Inspired by joris' comment [here](https://github.com/jasp-stats/jasp-issues/issues/1474#issuecomment-957304957), reused the `qsTr() ` to translate the string, I don't know if it's messy, but it works.😜

a test case [here ](https://github.com/shun2wang/jasp-desktop/commit/4b74ebe75ecc44b3b122e8fa7c76f461738df2d6) you can patch the `.qm` files and have a look in jaspNote "Bold" option.

### DIFF
--- a/Desktop/html/js/jaspwidgets.js
+++ b/Desktop/html/js/jaspwidgets.js
@@ -1,5 +1,26 @@
 var JASPWidgets = {};
 
+var i18nStrs = null;
+// The global translation function needs to be loaded before all other .js files.
+window.setI18nStrings = function (i18nObject) {
+        i18nStrs = i18nObject
+        return i18nStrs
+}
+
+// Use i18n(...) to call strings to be translated then update them in MainPage.qml
+function i18n(str) {
+	if (typeof str === "string" && str instanceof String) {
+		console.log("i18n can only be used for literal strings")
+	} else {
+		let findI18nStr = i18nStrs[str] !== null && i18nStrs[str] !== undefined;
+		if (findI18nStr) {
+			return i18nStrs[str];
+		} else {
+			return str;
+		}
+	}
+}
+
 if(insideJASP)
 {
 	var Parchment = Quill.import('parchment');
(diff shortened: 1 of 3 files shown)


JorisGoosen (approved):

JorisGoosen:
Awesome

Desktop/html/js/jaspwidgets.js
```diff
@@ -1,5 +1,26 @@
...
+        i18nStrs = i18nObject
+        return i18nStrs
+}
+
+// Use i18n(...) to call strings to be translated then update them in MainPage.qml
+function i18n(str) {
+	if (typeof str === "string" && str instanceof String) {
+		console.log("i18n can only be used for literal strings")
+	} else {
+		let findI18nStr = i18nStrs[str] !== null && i18nStrs[str] !== undefined;
```
boutinb: Why don't you use the hasOwnProperty method? 
I'm not sure, but I'm afraid by testing ` i18nStrs[str] !== null` this will insert automatically an entry in the object if the key does not exist.

boutinb (changes requested):
Really good initiative. Just a small comment.

Desktop/html/js/jaspwidgets.js
shun2wang: yes you are right, that should be more robust.

shun2wang: But this still cannot avoid returning `null `and `undefined`, because this is still valid for undefined.

We can be optimistic though, since the strings to be translated are created by the programmer and not the user, so I intend to return a value in any error case and just report the error in the console, anyway future reviewers of the PR should notice

boutinb (approved):

PR #5131, opened 2023-06-27
Synchronise the Variable Window with the selected column

boutinb:
When the Variable Window is open, it would be user friendly to synchronise it with the selected column in the the Data view


CommonData/column.cpp
```diff
@@ -1409,7 +1409,7 @@ strintmap Column::labelsSyncStrings(const stringvec &new_values, const strstrmap
 
 		maxLabelKey = std::max(labelValue, maxLabelKey); //to be used in labelsResetValues
 
-		if(mapValuesToAdd.count(labelText))	//Found it!
+		if(mapValuesToAdd.contains(labelText))
```
JorisGoosen, line 1412: Its identical in behaviour but sure ^^

JorisGoosen (approved):

PR #5132, opened 2023-06-27
Adds Ejs requests

JorisGoosen:
As discussed this morning in the meeting this PR adds:
- selecting (multiple) rows or columns through the headers
- add a separator to the ribbonbar
- adds clear cells to right menu and ribbon, but Im not sure which key to use for macos backspace+fn or something? (Using backspace itself would mess up editing cells quite a bit) (As does delete perhaps?)
- fix a problem with converting ordinal / nominal -> scale would mess up values from labels vs db-id
- cancel a data-synch request now also disables synching bool + icon


boutinb (changes requested):
I would set the Clear menu at the same level as the Cut/Copy/Paste level. It seems to me.a bit awkward alone at the end of the menu.
The 'Del' shortcut does not work: it is used already in the TextField to remove one character.

boutinb:
Deleting a Computed column is still an issue; it deletes random columns and sometimes it crahes.
... But it is not mentionned in this PR. I'll make another issue then

JorisGoosen:
Well maybe we shouldnt have the delete short cut then? @EJWagenmakers

RensDofferhoff:
Nice improvements! 
But not being able to use delete while editing a cell is maybe a bit awkward.
Drawback of automatically starting editing upon selection

boutinb:
By adding the possibility to select a row or a column by clicking the header row/column, then it makes quite frustrating not to have a pop-up menu (like for the cells, by right-clicking), and to be able to add/remove rows and columns.
I think to add this pop-up menu is quite easy.

JorisGoosen:
Actually, fn+backspace works for me on my mac

JorisGoosen:
But there was some room for improvement ^^

boutinb:
The pop-up menu is now available on the Row and Column header, but it does not work well:
- the pop-up shows menus that has nothing to do with the context: on a row header, there should be a menu for the column and vice-versa.
- on the column header, the right-click makes already toggling the Variable Window, which is quite annoying.

boutinb (approved):

PR #5133, opened 2023-06-29
Unify column edit windows

RensDofferhoff:
I will clean this up. let me know what you think.

As requested by: https://github.com/jasp-stats/INTERNAL-jasp/issues/2348
Suggested test file:
[a.zip](https://github.com/jasp-stats/jasp-desktop/files/11903979/a.zip)

Opens unified column editor containing label filtering and computed column editor (when clicking on column header for computed columns)
Fix name not changing for computed column window upon column name change.

For computed columns of ordinal, nominal types there is not enough screen real estate to show both the computed columns editor and the label filter window.  Therefore I added a button to switch between them:
image

Note:
(1) Can be cleaned up
(2) label filtering has broken in the last few days


boutinb:
When you create a computed column, the computed should be automatically opened

boutinb:
Also the close button is on a strange place: at the bottom-right. The close button should be at the top-right. We could use the same icon as the one in the analysis form.

boutinb:
If I open the Variable WindowComputed column with a computed column, then it gets the right height. If I click then on an ordinal column, then it is quite bad:
Screenshot 2023-06-29 at 15 13 53

PR #5134, opened 2023-06-29
Beter handling of Selection. mouse click and right-click behaviour on the Data view

boutinb:
In the DataView, different issues have been solved: 

- Select a cell, right click on another cell: the selected cell should be unselected, and the pop-up menu should appear on the new cell.
- Select multiple cells (by using shift):
  - right-click on one of these cells: the selected cells should stay selected, and the pop-uo menu should appear.
  - right-click on an unselected cell: the selected cells are unselected, the new cell is selected, and the pop-up menu should appear.
  - A simple click (on a selected cell or not): the other cells are unselected , the current cell is selected, and this cell gets the focus..
- Click on a row header: the whole row should be selected.
  - Click on one cell (selected or not): the row is unselected, only the current cell is selected and has the focus.
  - Right-click on a selected cell: the row stays selected, and a pop-up menu appears on the clicked cell.
  - Right-click on a unselected cell, the row is unselected and a pop-up menu appears on the clicked cell. 
- Click on a row header and shift-click on another row header: all the rows between the 2 rows should be selected. Same behaviour as with only one row selected. If delete row is selected from the pop-up menu or from the ribbon menu: all the selected rows should be removed. 
- Same behaviour with column header

### DIFF
--- a/Desktop/components/JASP/Widgets/DataTableView.qml
+++ b/Desktop/components/JASP/Widgets/DataTableView.qml
@@ -414,7 +414,7 @@ FocusScope
 						onClicked: 			(mouseEvent)=>
 											{
 												if(mouseEvent.button === Qt.LeftButton || mouseEvent.button === Qt.RightButton)
-													dataTableView.view.rowSelect(rowIndex, mouseEvent.modifiers & Qt.ShiftModifier );
+													dataTableView.view.rowSelect(rowIndex, mouseEvent.modifiers & Qt.ShiftModifier, mouseEvent.button === Qt.RightButton);
 												if(mouseEvent.button === Qt.RightButton)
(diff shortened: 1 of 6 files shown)


RensDofferhoff:
Shift selected cells get dropped to moment the cursor moves over virtual rows

RensDofferhoff:
> Click on one cell (selected or not): the row is unselected, only the current cell is selected and has the focus.

This doesnt work if the selected cell is in the first column of the selected row

RensDofferhoff:
> This doesnt work if the selected cell is in the first column of the selected row

Same for column select. if you click on first row of column the entire column stays selected

RensDofferhoff:
Great improvements!

RensDofferhoff (approved):

PR #5144, opened 2023-07-05
Add more JS i18n strings

shun2wang:
This shoud **not** added into 0.17.3 because we are during testing release.

### DIFF
--- a/Desktop/html/js/analysis.js
+++ b/Desktop/html/js/analysis.js
@@ -126,8 +126,8 @@ JASPWidgets.AnalysisView = JASPWidgets.View.extend({
 
 	_setTitle: function (title, format) {
 
-		this.viewNotes.firstNoteNoteBox.ghostText = title + ` - ${i18n("Introduction")}: ` + this.viewNotes.firstNoteNoteBox.ghostTextDefault;
-		this.viewNotes.lastNoteNoteBox.ghostText  = title + ` - ${i18n("Conclusion")}: `   + this.viewNotes.lastNoteNoteBox.ghostTextDefault;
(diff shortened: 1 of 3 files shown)


Desktop/html/js/analysis.js
```diff
@@ -126,8 +126,8 @@ JASPWidgets.AnalysisView = JASPWidgets.View.extend({
 
 	_setTitle: function (title, format) {
 
-		this.viewNotes.firstNoteNoteBox.ghostText = title + ' - Introduction: ' + this.viewNotes.firstNoteNoteBox.ghostTextDefault;
-		this.viewNotes.lastNoteNoteBox.ghostText  = title + ' - Conclusion: '   + this.viewNotes.lastNoteNoteBox.ghostTextDefault;
+		this.viewNotes.firstNoteNoteBox.ghostText = title + ` - ${i18n("Introduction")}: ` + this.viewNotes.firstNoteNoteBox.ghostTextDefault;
+		this.viewNotes.lastNoteNoteBox.ghostText  = title + ` - ${i18n("Conclusion")}: `   + this.viewNotes.lastNoteNoteBox.ghostTextDefault;
```
boutinb: Only `Introduction` and `Conclusion` is translated, but you should add the semicolon: in French you should normally add a space before the semicolon, so it should be translated by `Introduction :`

boutinb (changes requested):

Desktop/html/js/analysis.js
shun2wang: >Only Introduction and Conclusion is translated

that's OK ,other variables (such `title`) are translated some where by `i18n(...)` call or from QML/C++ level, because we just translate strings Literal.

> add the semicolon

I just let `:`including translation string, so that translator can change the style for they locale?

boutinb (approved):

PR #5189, opened 2023-08-13
Bugfix: Fix cell show blank after remove label

shun2wang:
Fix https://github.com/jasp-stats/jasp-issues/issues/1655

Showing empty content is not a good idea especially for table and plot results. so let's just set value as label while label is empty.

### DIFF
--- a/Desktop/components/JASP/Widgets/VariablesWindow.qml
+++ b/Desktop/components/JASP/Widgets/VariablesWindow.qml
@@ -554,6 +554,7 @@ FocusScope
 
 											Text
 											{
+												id	 :					cellValue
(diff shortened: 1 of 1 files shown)


JorisGoosen:
Your fix works but has a little beauty problem.

If you remove the label everything works as expected, but if the previous label was already exactly equal to the value it will stay empty in the variableswindow...

shun2wang:
> If you remove the label everything works as expected, but if the previous label was already exactly equal to the value it will stay empty in the variableswindow...

that's may because the variableswindow does not update immediately, if we click the data table then it updated and not empty. maybe needs a lose focus event to update label cell.or something change in cpp level is more beautiful.😃

PR #5190, opened 2023-08-14
originalValueAsLabelIfEmpty

JorisGoosen:
Better in the model than in QML

Fixes https://github.com/jasp-stats/jasp-issues/issues/1655

### DIFF
--- a/CommonData/label.cpp
+++ b/CommonData/label.cpp
@@ -205,7 +205,7 @@ bool Label::setLabel(const std::string & label)
 {
 	if(_label != label)
 	{
-		_label = label;
+		_label = label.empty() ? originalValueAsString() : label;
 
 		dbUpdate();
 		return true;


shun2wang (commented):
Suppose there are three corresponding value and label pairs: 1-A, 2-B, 3-C. If we remove the value of 3, but the legacy value and label are still leave in the variable window, it seems that `VariablesWindow.qml` is not getting the signal?

But this fix does make it clearer:-)

JorisGoosen:
You mean the label of 3?
Ill try it though and see what you mean. But I cant repro

image
like this?

If I remove label C it just turns into 3

shun2wang:
i mean remove all the 3 value from data cell.

JorisGoosen:
> i mean remove all the 3 value from data cell.

And the problem would be what?
image
(I removed D from data nd after that the label D from 4)

shun2wang:
图片

JorisGoosen:
Ah you mean that 3/C is still in the vriables window?
That is not a problem is it?

shun2wang:
yes, I do mean this.

JorisGoosen:
Seems like proper behaviour to me, this way you would keep the original value if you add the label back again later during editing

shun2wang:
yes, you are right at least it doesn't enter the analysis results and affect the data analysis.

shun2wang (approved):

PR #5191, opened 2023-08-15
Fix flatpak mime type export on Linux

shun2wang:
JASP's mime type is not exported by flatpak and registed correctly on Debian/Ubuntu series with gnome desktop (at least on my Ubuntu 22.04), so the JASP cannot opened by double click on .jasp file which registed as zip file so will open by zip app default.

* Add MIME type and file extension for Linux flatpak
* Fix to mime type icon
* Update org.jaspstats.JASP.desktop


RensDofferhoff (approved):
Very nice!

PR #5201, opened 2023-08-17
Double click for variableswindow/column

JorisGoosen:
As suggested in https://github.com/jasp-stats/jasp-test-release/issues/2343 and for https://github.com/jasp-stats/jasp-test-release/issues/2356

Also makes clicking around in the data waaaay less twitchy, also solves the clearing columns problem

### DIFF
--- a/Desktop/components/JASP/Widgets/DataTableView.qml
+++ b/Desktop/components/JASP/Widgets/DataTableView.qml
@@ -404,7 +404,7 @@ FocusScope
 									dataTableView.view.clearEdit()
 									dataTableView.showPopupMenu(itemHighlight, mapToGlobal(mouse.x, mouse.y), rowIndex, columnIndex);
 								}
-								else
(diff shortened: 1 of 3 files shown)


RensDofferhoff:
Delete still doesn't always work in selection.
When clicking on a cell while selecting instead of just hovering over it, the delete key is caught by the text edit element that is activated.
Is this desired behavior?

RensDofferhoff:
Perhaps we shouldn't enter the text edit while shift is hold?
Otherwise we should probably cancel the selection on the click

JorisGoosen:
Aah yeah I only looked at clicking the columnheaders for this issue, but I can have a look at the shift-clicking a cell.

> Perhaps we shouldn't enter the text edit while shift is hold?

Yeah that probably makes sense.

JorisGoosen:
easy peasy

RensDofferhoff (approved):
Great!

PR #5205, opened 2023-08-21
make filter responsive by switching to dataMode on changes

JorisGoosen:
fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2145 and https://github.com/jasp-stats/jasp-issues/issues/1644 Also adds a few profiling timers

Im fairly sure this shouldnt add any problems as the only difference between this and the current nightly is that filters wont ever be ran in "Analyses" mode but will just switch to data mode. This alredy worked fine and apparently obviates the gui becoming unresponsive by not hving to pause/resume the engines repeatedly.

### DIFF
--- a/CommonData/filter.cpp
+++ b/CommonData/filter.cpp
@@ -14,6 +14,8 @@ void Filter::dbCreate()
 
 void Filter::dbUpdate()
 {
+	JASPTIMER_SCOPE(Filter::dbUpdate);
+
 	assert(_id != -1);
 
 	db().transactionWriteBegin();
--- a/CommonData/label.cpp
+++ b/CommonData/label.cpp
@@ -45,6 +45,8 @@ void Label::dbDelete()
 
 void Label::dbCreate()
 {
+	JASPTIMER_SCOPE(Label::dbCreate);
+
 	if(_column->batchedLabel())
 		return;
 		
@@ -74,6 +76,8 @@ void Label::dbLoad(int labelId)
 
 void Label::dbUpdate()
 {
+	JASPTIMER_SCOPE(Label::dbUpdate);
+
 	if(_column->batchedLabel())
 		return;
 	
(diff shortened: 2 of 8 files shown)


RensDofferhoff (approved):
Looks & works fine for me

boutinb (changes requested):
I don't like this solution. First that the engineSync class changes JASP in data mode does not seem to me a logic place to do this action.
But more important, the switch to data mode makes JASP less responsive: before you could switch some labels on and off, and see immediately the consequences in the analyses. Now you have to switch back to analyses mode each time before seeing the results.
Switching to data mode to block the analyses to run before the filters are processed, seems to me a workaround than a real solution.

shun2wang:
Just a noise, a buffer might help, we use a timer and timeout to decide whether to initiate engine synchronization:
This is like the “Debounce” function: For example, if the mouse clicks on the filter and does not click it again within 2s, the synchronization is initiated, otherwise the timeout is reset to zero let the engine continues wait.
Or a "Throttle" function: No matter how many times the user clicks the filter, engine sync will be performed within 1 second for only once.

In this way, we can keep the filter working even if the user clicks frequently, because maybe in the future, the filter will be used independently of the analysis module (I saw some feature requests).

JorisGoosen:
Well, the problem lies mostly in the process of stopping the engines, they run in separate processes. JASP desktop needs to communicate with those processes in a tight loop, well that is how it is setup now anyway, which causes the GUI to become unresponsive if one of those processes doesnt respond nice and quick. 

The problem is however that the analyses will get triggered on data changes, load data, load module and run analysis.

We should really have a better way of communicating with the engines but that isnt trivial. A (priority)message queue would do wonders already.

For now though this solution has a low chance of breaking JASP while changing the flow of filter -> datachange -> analysis-refresh has a lot of potential for problems...

JorisGoosen:
@boutinb shall I merge this?

boutinb (approved):

PR #5206, opened 2023-08-21
Some changes in the GUI

boutinb:
- After selecting a menu in the custom menu (popup menu), the source item of the popup should get back the focus (behalve if the action of the menu sets the focus). 
- Undo/Redo ribbon menu should be set enabled/disabled in place of visible/invisible 
- Change tooltip: when variable settings window is open, only a click (not a doubleclick) on another variable header changes the selected variable.

### DIFF
--- a/Desktop/components/JASP/Widgets/CustomMenu.qml
+++ b/Desktop/components/JASP/Widgets/CustomMenu.qml
@@ -73,7 +73,7 @@ FocusScope
 		case Qt.Key_Space:
 			if (currentIndex > -1)
 			{
-				menu.props['functionCall'](currentIndex)
+				callMenuAction(currentIndex)
 				menu.currentIndex = -1;
 			}
 			closeMenu();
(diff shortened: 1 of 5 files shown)


RensDofferhoff (approved):

JorisGoosen:
I cnt click "New Data" anymore: 
image

JorisGoosen:
> After selecting a menu in the custom menu (popup menu), the source item of the popup should get back the focus (behalve if the action of the menu sets the focus).

I dont understand what this means?

JorisGoosen (changes requested):
So I would like to have new data again and I dont understand that one sentence so Im not sure what to look for :p

boutinb:
> So I would like to have new data again and I dont understand that one sentence so Im not sure what to look for :p

New data is back.
The sentence says that when a popup menu shows up,  and a menu item is selected, this menu item gets the focus. But afterwards the popup menu disappears, so no item has the focus anymore. The fix is to set the focus back to the item from which the popup menu was generated.
For example, if from a data cell, you right-click and select copy from the popup menu, the data cell gets back the focus.

JorisGoosen (approved):

PR #5208, opened 2023-08-22
Open variables window upon computed column creation

RensDofferhoff:


Desktop/data/columnModel.cpp
```diff
@@ -342,6 +349,12 @@ void ColumnModel::checkRemovedColumns(int columnIndex, int count)
 	}
 }
 
+void ColumnModel::openComputedColumn(QString name)
```
JorisGoosen:
```suggestion
void ColumnModel::openComputedColumn(const QString & name)
```

Desktop/data/columnModel.h
```diff
@@ -78,6 +78,7 @@ public slots:
 	void filteredOutChangedHandler(int col);
 	void setVisible(bool visible);
 	void setChosenColumn(int chosenColumn);
+	void setChosenColumn(QString& chosenName);
```
JorisGoosen:
```suggestion
	void setChosenColumn(const QString & chosenName);
```

Desktop/data/columnModel.h
```diff
@@ -87,6 +88,7 @@ public slots:
 	void refresh();
 	void changeSelectedColumn(QPoint selectionStart);
 	void checkRemovedColumns(int columnIndex, int count);
+	void openComputedColumn(QString name);
```
JorisGoosen:
```suggestion
	void openComputedColumn(const QString & name);
```

Desktop/data/columnModel.cpp
```diff
@@ -260,6 +260,13 @@ void ColumnModel::setChosenColumn(int chosenColumn)
 	emit showComputedColumnChanged();
 }
 
+void ColumnModel::setChosenColumn(QString &chosenName)
```
JorisGoosen:
```suggestion
void ColumnModel::setChosenColumn(const QString & chosenName)
```

JorisGoosen (changes requested):
Doesnt work when inserting computed column from ribbonmenu... 
Only from clicking virtual headers and getting the creation menu

Desktop/qquick/datasetview.cpp
```diff
@@ -1295,7 +1295,7 @@ QString DataSetView::columnInsertBefore(int col, bool computed, bool R)
 
 	_model->insertColumn(col, computed, R);
 
-	return "";
+	return tq(DataSetPackage::pkg()->dataSet()->column(col)->name());
```
JorisGoosen: This will go wrong as soon as we start sorting or filtering columns, but its probably not the only place

JorisGoosen:
```suggestion
	return tq(DataSetPackage::pkg()->dataSet()->column(col)->name()); // This will break as soon as we start filtering/sorting columns
```

JorisGoosen (approved):

PR #5211, opened 2023-08-23
Dont switch to data editing mode on filterrun, as this gives silly behaviour

JorisGoosen:
now a timer is started after a filter is ran, 2s delay before it allows analyses etc to run this should also fix https://github.com/jasp-stats/INTERNAL-jasp/issues/2145 and reverts some of https://github.com/jasp-stats/jasp-desktop/commit/80c5077268c90524a403b1b33101c7bf8ab0ad32

### DIFF
--- a/Desktop/engine/enginesync.cpp
+++ b/Desktop/engine/enginesync.cpp
@@ -52,7 +52,7 @@ EngineSync::EngineSync(QObject *parent)
 	_singleton = this;
 	
 	_filterRunningResetTimer = new QTimer(this);
-	_filterRunningResetTimer->setInterval(2000);
+	_filterRunningResetTimer->setInterval(1000);
 	_filterRunningResetTimer->setSingleShot(true);
 	
 	connect(_filterRunningResetTimer, &QTimer::timeout, this, [&](){ _filterRunning = false; });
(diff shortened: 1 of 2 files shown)


Desktop/engine/enginesync.cpp
```diff
@@ -52,6 +51,12 @@ EngineSync::EngineSync(QObject *parent)
 	assert(!_singleton);
 	_singleton = this;
 	
+	_filterRunningResetTimer = new QTimer(this);
+	_filterRunningResetTimer->setInterval(2000);
```
JorisGoosen:
```suggestion
	_filterRunningResetTimer->setInterval(1000);
```

shun2wang: emmm, both tested on Win and Linux: 2s is most silly for me when I 'crazy' click filter haha, for 1s  I can feel slow but yes it's also enough. I recommended 2s before since most cases it can be separated from the delay of mouse click/doubleclick events. For the user, after he clicks the filter and returns to the analysis panel, 1s has passed😂

shun2wang (approved):

boutinb (approved):

Desktop/engine/enginesync.cpp
JorisGoosen: I first started with 4s, but that was absolutely ridiculous :p

PR #5218, opened 2023-08-24
Edit/New data Ribbon should be enabled at the right moment

boutinb:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/2361

If there is no data, the New Data icon should be enabled, but if an analysis that does not require data (like a distribution) is added, then the New Data should be disabled. If you remove the analysis, the New Data should be enabled. If you open a JASP file that has no data, the New Data icon should be disabled.

### DIFF
--- a/Desktop/components/JASP/Widgets/DataTableView.qml
+++ b/Desktop/components/JASP/Widgets/DataTableView.qml
@@ -691,14 +691,11 @@ FocusScope
 								createComputeDialog.open()
 							else
 							{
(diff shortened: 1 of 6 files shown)


RensDofferhoff (approved):
Nice improvements!

JorisGoosen:
Very nice!

One nitpicky thing: It looks real weird to have a disabled button "New Data" when you cant have data anyway.
I think it might be prettier to have the button still there, still disabled but then with "No Data" on it.

PR #5221, opened 2023-08-25
Tweaks for parse exported html

shun2wang:
However I still think this is unlikely to solve the inconsistency of comparing .jasp files, since the difference is subtle and the original difference was in the old jasp file,unless we refreshing old .jasp file and use the new file as a benchmark or just using JSON comparing.

This PR contains two equivalent commits
1. Parsing strings into html by jquery will lose end end tag slash?
2. The browser natively parses it as html, or it can be set to xml/xhtml (but it currently has parsing errors: some tags need to be escaped).

### DIFF
--- a/Desktop/html/js/main.js
+++ b/Desktop/html/js/main.js
@@ -548,38 +548,25 @@ $(document).ready(function () {
 	$("body").on("click", window.unselectByClickingBody)
 })
 
-var wrapHTML = function (html, exportParams, doctype = false) {
-	var completehtml = ""
-	if(doctype)
-		completehtml = "<!DOCTYPE HTML>\n"
-	completehtml += "<html>\n"
-	completehtml += "	<head>\n"
-	completehtml += "		<meta http-equiv='Content-Type' content='text/html; charset=utf-8' />\n"
-	completehtml += "		<title>JASP</title>"
-	completehtml += "		<style>"
-	completehtml += "			p {margin-top:1em; margin-bottom:1em;}"
-	if (exportParams.isFormatted())
-		completehtml += "			body {font-family: sans-serif;}"
-	completehtml += "		</style>"
-	completehtml += "	</head>\n"
+var wrapHTML = function (htmlStrings, exportParams, doctype = false) {
 
-	var styles = JASPWidgets.Exporter.getStyles($("body"), ["display", "padding", "margin"]);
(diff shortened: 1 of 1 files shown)


JorisGoosen:
Starting jasp with this code and the argument `--unitTest ../Resources/Data\ Sets/Data\ Library/3.\ ANOVA/Bugs.jasp` (From `build/Desktop`) leads to the result that none of the tables are found anymore...

Also the resulting index.html seems rather different, lots of extra properties on each element. It does look fine in firefox though.

shun2wang:
> Also the resulting index.html seems rather different, lots of extra properties on each element. It does look fine in firefox though.

@JorisGoosen Could you give me some files or screenshots for that because I am on a `release` type local build but not switch to a `debug` build  for now? if `extra properties` is class name like `class='blabala'` that is expected, I added it very early...

JorisGoosen:
image
Left is old

shun2wang:
Build on this PR? why I cannot get this file...but only some `'` `"` diff....

JorisGoosen:
> Build on this PR?

Yes I did 

> why I cannot get this file...but only some `'` `"` diff....

Well, all I can see is that the unit test breaks completely now. 
And there are so many changes Im not sure whether we break something else.
I also just got the regex replace working which is only a single line change.

Im trying to upload it but my internet/github is not cooperating right now...

shun2wang:
I guess what you see may related to my change in https://github.com/jasp-stats/jasp-desktop/pull/5109 like this:

```javascript

		let attrObj = $elObj.prop("attributes");
		$.each(attrObj, function() {
			attrs += this.name + '="' + this.value + '" ';
			});

		if (style)
			attrs = attrs + style
```
but I cannot reproduce that anymore...let me make a debug build to see unittest things. sorry for the inconvenience what I introduced.

JorisGoosen:
> sorry for the inconvenience what I introduced.

Dont worry about that!

Having the formulas in there is real nice and the unittest is hardly the first thing you think of when implementing some js in the resultswindow ;)

JorisGoosen:
https://github.com/jasp-stats/jasp-desktop/pull/5224

We can still see after 0.18 release whether this PR might have additional advantages (if it also makes the unittests function again, probably by switching back from `'` to `"`

shun2wang:
Let's close it for now? No reason to modify it because it works fine.

PR #5223, opened 2023-08-25
When synchronization is on, add menu to be able to start external spreadsheet editor

boutinb:
When you open a datafile or a JASP file that is synchronized with a data file, in Data Edit mode, it is not possible to start directly the external spreadsheet editor: you have first to switch it off, and then switch it on again. This is not quite intuitive. So add a menu in the Synchronzation ribbon button, so to let the user have the possibility to start directly the spreadsheet editor.


JorisGoosen (approved):

JorisGoosen:
Great branchname...

PR #5224, opened 2023-08-28
this regex removes the superfluous mjx-assistive-mml element without dom-parsing

JorisGoosen:
Fixes --unitTest and --unitTestRecursive

would be good to resave the data library after this btw

### DIFF
--- a/Desktop/html/js/table.js
+++ b/Desktop/html/js/table.js
@@ -801,10 +801,10 @@ JASPWidgets.tablePrimitive = JASPWidgets.View.extend({
 
 			style = JASPWidgets.Exporter.getTableContentStyles($elObj, exportParams);
 
-			if ($elObj.prop("rowspan") && $elObj.prop("rowspan") != 1)
+			/*if ($elObj.prop("rowspan") && $elObj.prop("rowspan") != 1)
 				attrs += 'rowspan="' + $elObj.prop("rowspan") + '" '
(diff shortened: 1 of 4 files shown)


Desktop/html/js/table.js
```diff
@@ -801,10 +801,10 @@ JASPWidgets.tablePrimitive = JASPWidgets.View.extend({
 
 			style = JASPWidgets.Exporter.getTableContentStyles($elObj, exportParams);
 
-			if ($elObj.prop("rowspan") && $elObj.prop("rowspan") != 1)
+			/*if ($elObj.prop("rowspan") && $elObj.prop("rowspan") != 1)
 				attrs += 'rowspan="' + $elObj.prop("rowspan") + '" '
 			if ($elObj.prop("colspan") && $elObj.prop("colspan") != 1)
-				attrs += 'colspan="' + $elObj.prop("colspan") + '" '
+				attrs += 'colspan="' + $elObj.prop("colspan") + '" '*/
```
boutinb: I don't understand what it has to do with the fix... maybe could you add a comment to the uncommented stuff (or just remove it completely).

boutinb (changes requested):

Desktop/html/js/table.js
JorisGoosen: there were double colspan entries after I removed the domparsing, but ok ill just remove it then.

boutinb (approved):

PR #5226, opened 2023-08-28
actually use datarowcount and dont go for infinite loops please

JorisGoosen:
fixes https://github.com/jasp-stats/jasp-test-release/issues/2364 

also renamed some things in variableInfo to make it clearer that some things are from the dataset and not from a variable aka a column and thus transposed.

### DIFF
--- a/Desktop/data/columnsmodel.cpp
+++ b/Desktop/data/columnsmodel.cpp
@@ -90,7 +90,7 @@ QVariant ColumnsModel::provideInfo(VariableInfo::InfoType info, const QString& c
 		case VariableInfo::Labels:						return	_getLabels(colIndex);
 		case VariableInfo::DoubleValues:				return	QTransposeProxyModel::data(qColIndex, int(DataSetPackage::specialRoles::valuesDblList));
 		case VariableInfo::NameRole:					return	data(qColIndex, ColumnsModel::NameRole);
-		case VariableInfo::DataSetRowCount:				return	DataSetPackage::pkg()->dataRowCount();
+		case VariableInfo::DataSetRowCount:				return  QTransposeProxyModel::columnCount();
 		case VariableInfo::DataSetValue:				return	QTransposeProxyModel::data(qValIndex, int(DataSetPackage::specialRoles::value));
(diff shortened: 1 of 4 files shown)


Desktop/data/columnsmodel.cpp
```diff
@@ -87,10 +88,10 @@ QVariant ColumnsModel::provideInfo(VariableInfo::InfoType info, const QString& c
 		case VariableInfo::VariableTypeDisabledIcon:	return	VariableInfo::getIconFile(colTypeHere, VariableInfo::DisabledIconType);
 		case VariableInfo::VariableTypeInactiveIcon:	return	VariableInfo::getIconFile(colTypeHere, VariableInfo::InactiveIconType);
 		case VariableInfo::Labels:						return	_getLabels(colIndex);
-		case VariableInfo::DoubleValues:				return	QTransposeProxyModel::data(qIndex, int(DataSetPackage::specialRoles::valuesDblList));
-		case VariableInfo::NameRole:					return	data(qIndex, ColumnsModel::NameRole);
-		case VariableInfo::RowCount:					return	rowCount();
-		case VariableInfo::Value:						return	QTransposeProxyModel::data(qIndex, int(DataSetPackage::specialRoles::value));
+		case VariableInfo::DoubleValues:				return	QTransposeProxyModel::data(qColIndex, int(DataSetPackage::specialRoles::valuesDblList));
+		case VariableInfo::NameRole:					return	data(qColIndex, ColumnsModel::NameRole);
+		case VariableInfo::DataSetRowCount:				return	DataSetPackage::pkg()->dataRowCount();
```
boutinb: But this won't work with filtered data? Toch?

boutinb (changes requested):

Desktop/data/columnsmodel.cpp
JorisGoosen: No it wont work with filtered data at all as it is

JorisGoosen: did it do so before?
well, Ill have another look I guess

boutinb (approved):

PR #5228, opened 2023-08-28
Column type switch can lead to crash

boutinb:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/2362

The crash was due to the fact that _labelByValueMap was not updated when a label was changed. Also this PR adds the automatic switch to from nominalText to nominal if all values becomes integer. This is needed because if a user adds accidentally a string value to a nominal (or ordinal) column, the column is automatically changed to nominalText. If the user removes the wrong value, the column stays to nominalText, and the user has no way to change it back to nominal. JASP should detect this scenario and change the column to the tight type.

### DIFF
--- a/CommonData/column.cpp
+++ b/CommonData/column.cpp
@@ -799,8 +799,15 @@ columnTypeChangeResult Column::changeType(columnType colType)
 	{
 		if (colType == _type)				
 			return columnTypeChangeResult::changed;
-		else								
-			return colType == columnType::scale ? _changeColumnToScale() : _changeColumnToNominalOrOrdinal(colType);
+		else
+		{
+			columnTypeChangeResult changeResult = colType == columnType::scale ? _changeColumnToScale() : _changeColumnToNominalOrOrdinal(colType);
+
+			if(changeResult == columnTypeChangeResult::changed && _type != columnType::nominalText && _preEditType != _type)
+				_preEditType = _type;
+
+			return changeResult;
+		}
 	}
 	else
 	{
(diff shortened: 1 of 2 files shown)


CommonData/column.cpp
```diff
@@ -1661,12 +1666,29 @@ bool Column::setStringValueToRowIfItFits(size_t row, const std::string & value,
                                 ColumnUtils::convertValueToDoubleForImport(value,				dbl))
 							typeChanged = changeType(_preEditType) == columnTypeChangeResult::changed; //Just try it
 					}
+					else if(_type == columnType::nominalText)
```
JorisGoosen: Ah I meant to comment here...
```suggestion
					else if(_type == columnType::nominalText && (_prevEditType == columnType::nominal || _prevEditType == columnType::ordinal))
```
Like for scale?

CommonData/column.cpp
```diff
@@ -1661,12 +1666,29 @@ bool Column::setStringValueToRowIfItFits(size_t row, const std::string & value,
...
+						{
+							if (!_labels[i]->originalValue().isInt())
+							{
+								onlyIntegers = false;
+								break;
+							}
+						}
+
+						if (onlyIntegers)
+							_changeColumnToNominalOrOrdinal(columnType::nominal);
```
JorisGoosen:
```suggestion
							_changeColumnToNominalOrOrdinal(_prevEditType);
```
And this?

Btw, I didnt test this but it seems like better behaviour

JorisGoosen:
Im changing this PR

boutinb:
> Im changing this PR

I've just tested it, and it's working after your last changes. Or do you mean that you are still changing it?

JorisGoosen:
Im done now

JorisGoosen (approved):

JorisGoosen:
It still crashes if you:
- open debug.csv
- change V1 to ordinal
- change label in variablewindow
- edit a non-labelled int in the data editor

JorisGoosen:
This breaks analyses.
Repro:
- load debug
- V1 to Ordinal
- add descriptives with V1 it works
- switch to data mode
- edit a value in V1 to be a string, switches to nominalText
- unedit manually so it switches back to ordinal
- switch to analyses mode and admire:
image

JorisGoosen:
You probably shouldnt mess with the `maxLabelKey` code as it took me a lot of effort to get that to *not* create "malformed factor"

PR #5241, opened 2023-09-04
Bugfix: remove legacy analysis result in clipboard

shun2wang:
Fixes https://github.com/jasp-stats/jasp-issues/issues/1301 and https://github.com/jasp-stats/jasp-issues/issues/2194

It's because we just set 'removed'  analysis view invisible but not really removed. then it was got in exporter.

### DIFF
--- a/Desktop/html/js/analyses.js
+++ b/Desktop/html/js/analyses.js
@@ -110,10 +110,11 @@ JASPWidgets.Analyses = JASPWidgets.View.extend({
 				this.views = _.without(this.analyses, analysis);
 
 				analysesVar.setBottomSpacerHeight();
+
+				analysis.$el.addClass("removed"); 
+				analysis.$el.remove(); 
 			});
 		});
-
-		
 	},
 
 	removeAnalysisId: function (analysisId) {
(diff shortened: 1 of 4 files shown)


Desktop/html/js/analyses.js
```diff
@@ -110,10 +110,11 @@ JASPWidgets.Analyses = JASPWidgets.View.extend({
 				this.views = _.without(this.analyses, analysis);
 
 				analysesVar.setBottomSpacerHeight();
+
+				analysis.$el.addClass("removed"); 
+				analysis.$el.remove();
```
boutinb, line 116: I'm not sure to understand here: if `$el.remove()` is called, then it is removed from the DOM, but apparently it is not enough, the element is kept somewhere else?

shun2wang: yes, it's just removed from DOM but not in backbone.js table view, i call it here to reduce memory usage. for results in clipboard is generated and get in jaspwidgets.

boutinb: If I remove the line 113 (that adds the removed class), II see that it does not work. So this handling is apparently necessary. So just add a comment that explains why, then we can merge it

shun2wang: your understand right, in fact our export is not as what a "visible" page inside jasp, but we using tags name and classname... to collected a view to export. In fact, the removal here is just reduce browser's memory usage, but it is necessary to add the class name.

shun2wang: ill add comments.

boutinb (approved):

PR #5242, opened 2023-09-04
Make remove columns/rows quicker

boutinb:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2374
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2376

Thanks to the expanded model, it is not necessary anymore to add a unique cell when there is no column or row anymore.
When removing all rows, the columns should not be removed anyway
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2373

### DIFF
--- a/CommonData/column.cpp
+++ b/CommonData/column.cpp
@@ -1265,6 +1265,8 @@ int Column::labelsAdd(int value, const std::string & display, bool filterAllows,
 
 void Column::labelsRemoveValues(std::set<int> valuesToRemove)
 {
+	if (valuesToRemove.empty()) return;
+
 	JASPTIMER_SCOPE(Column::labelsRemoveValues);
 
 	_labels.erase(
--- a/Desktop/data/computedcolumnsmodel.cpp
+++ b/Desktop/data/computedcolumnsmodel.cpp
@@ -304,7 +304,7 @@ void ComputedColumnsModel::removeColumn()
 		return;
 
 	// TODO pass RemoveColumnCommand aab
-	_undoStack->pushCommand(new RemoveColumnCommand(DataSetPackage::pkg(), _selectedColumn->id()));
+	_undoStack->pushCommand(new RemoveColumnsCommand(DataSetPackage::pkg(), _selectedColumn->id(), 1));
 
 	DataSetPackage::pkg()->requestComputedColumnDestruction(_selectedColumn->name());
 	setComputeColumnNameSelected("");
--- a/Desktop/data/datasetpackage.cpp
+++ b/Desktop/data/datasetpackage.cpp
@@ -1412,7 +1412,7 @@ bool DataSetPackage::initColumnAsNominalOrOrdinal(	QVariant colID, const std::st
 }
 
 
-void DataSetPackage::initColumnWithStrings(QVariant colId, const std::string & newName, const stringvec &values, const std::string & title)
+void DataSetPackage::initColumnWithStrings(QVariant colId, const std::string & newName, const stringvec &values, const std::string & title, columnType desiredType)
 {
 	JASPTIMER_SCOPE(DataSetPackage::initColumnWithStrings);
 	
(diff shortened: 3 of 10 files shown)


JorisGoosen:
Nice!

Much much faster, also the expanded model does indeed make the resetToOneCell thing obsolete.

However, undo has some problems:
- Open debug.csv
- change contBinom to scalar
- delete some rows from contBinom
- undo
- watch how contBinom turned back into nominal

This is caused by using `pasteSpreadsheet`...

I think this isnt the best behaviour.
We could fix this either for this PR or by fixing pasteSpreadSheet to make it respect the columnTypes already there a bit better?

Probably the easiest way would be to store columnType as well and set that after the undo via pasteSpreadsheet.

Only problem then would be if the column would turn into scalar you lose any order in the labels, but it could be a good temporary solution.
To see what I mean:
- open debug.csv
- change V1 to Ordinal
- reverse label order
- delete some rows
- undo
- see V1 turn into scalar
- turn V1 into Ordinal
- labels are ordered unreversed (of course)

JorisGoosen (changes requested):
It would be much better if the columnType didnt get reset after undo

boutinb:
@JorisGoosen I have solved the type issue.

boutinb:
For https://github.com/jasp-stats/INTERNAL-jasp/issues/2375, as I myself did not understand the purpose of these actions, we need a solution that should be more intuitive. Also I would like to avoid to give to the user a too long menu.

What I thought is that there are in fact 2 cases when we copy/cut from JASP: we paste it in JASP or in another spreadsheet editor. If we paste it in JASP, we can have more information about the column we want to copy (labels, computed column, empty values...), so we could copy all that. In an external spreadsheet, we have to use the clipboard, so we have to choose here only if we insert the header or not.
So for the clipboard self, I would insert the header only if the copy is done from the column header.
When we copy, we should also have an 'internal clipboard' so that we can have a reference of the column which is copied.
I would then add a 'Paste Column(s)' in the menu, which should be enabled only if the copy comes from a copy in the header column in JASP. This paste would overwrite the current focused column: if the user wants to insert a new column with the copied column, he should do an 'Insert Column' action. If this paste is done in a virtual column, it inserts of course directly a new column.

Another point, I would like to change the behaviour when you click the header of a virtual column: now it creates automatically a computed column, which is quite confusing. It would be better to show the Variable Editor instead: this variable editor should have a checbox to switch from 'normal' column to computed column. The focus would go first to the name of this column, and the column would stay virtual as long as there is no name. The default type would be 'scale'.
Right-click on a virtual column header will show this 'Paste Column(s)'  menu, so that the column can be directly created.

JorisGoosen:
> For [jasp-stats/INTERNAL-jasp#2375](https://github.com/jasp-stats/INTERNAL-jasp/issues/2375), as I myself did not understand the purpose of these actions, we need a solution that should be more intuitive. Also I would like to avoid to give to the user a too long menu.
> 
> What I thought is that there are in fact 2 cases when we copy/cut from JASP: we paste it in JASP or in another spreadsheet editor. If we paste it in JASP, we can have more information about the column we want to copy (labels, computed column, empty values...), so we could copy all that. In an external spreadsheet, we have to use the clipboard, so we have to choose here only if we insert the header or not. So for the clipboard self, I would insert the header only if the copy is done from the column header. When we copy, we should also have an 'internal clipboard' so that we can have a reference of the column which is copied. I would then add a 'Paste Column(s)' in the menu, which should be enabled only if the copy comes from a copy in the header column in JASP. This paste would overwrite the current focused column: if the user wants to insert a new column with the copied column, he should do an 'Insert Column' action. If this paste is done in a virtual column, it inserts of course directly a new column.
> 
> Another point, I would like to change the behaviour when you click the header of a virtual column: now it creates automatically a computed column, which is quite confusing. It would be better to show the Variable Editor instead: this variable editor should have a checbox to switch from 'normal' column to computed column. The focus would go first to the name of this column, and the column would stay virtual as long as there is no name. The default type would be 'scale'. Right-click on a virtual column header will show this 'Paste Column(s)' menu, so that the column can be directly created.

Maybe make an issue out of this instead of a comment on a vaugely related PR? ;)

JorisGoosen:
Awesome, much better!
It can be merged like this but a small improvement would be:

`RemoveRowCommand` is not necessary in undo is it?
I'd remove that and the `RemoveColumnCommand` completely but I dont really care, its just slightly more maintenance than necessary.

So remove those or merge it, what you like ;)

JorisGoosen (approved):

PR #5244, opened 2023-09-06
Fix MatrixModels build fail due to conflicting Matrix versions for local linux builds

RensDofferhoff:
R 4.3.1 still has the ancient Matrix version.


RensDofferhoff:
This is also necesarry to make builds work smoothly on the mac buildbot again.

One of my jaspBase updates seems to have been reverted while the submodules were updated in the PR

PR #5246, opened 2023-09-12
Fixes some jasp save and load issues

### LINKED ISSUE #2289: [Bug]: After editing the existing file,  it cannot be archived regardless of whether "save" or "save as" is used.
### JASP Version

0.18.0

### Commit ID

_No response_

### JASP Module

Unrelated

save file

Windows 10

### Bug Description

1. When a JASP file is placed in a folder with a Chinese name, after opening the file and editing it, it cannot be archived regardless of whether "save" or "save as" is used.
2. When the JASP file is placed in a folder with an English name, after opening the file and editing it, using the "save" function, you cannot save it with the original file name, but you can use the "save as" function and save it with a new file name.
3. I have videotaped the operation process, please see the video below:
    https://youtu.be/f6_j8CMQ0ng

### Expected Behaviour

please see the video below:     https://youtu.be/f6_j8CMQ0ng

### Steps to Reproduce

1. the bug is reported in following video：    https://youtu.be/f6_j8CMQ0ng
2.
3.
...

### Log (if any)

_No response_

### Final Checklist

JorisGoosen:
libarchive seemed to mess up the strings from utf8, the code still seemed to assume an oldschool codepage however this is easy to avoid, I just added back the usage of the archive_..._w interface

I also noticed removing an analysis didnt trigger a "modified" flag on datasetpackage so I added that too.

Saving the modified file however also didnt work because the check for the importer somehow blew out the constructor of ArchiveReader thereby avoiding the close() in the destructor we just never noticed because none of us work on windows a lot I think

Fixes https://github.com/jasp-stats/jasp-issues/issues/2289

### DIFF
--- a/Common/utils.cpp
+++ b/Common/utils.cpp
@@ -256,7 +256,10 @@ std::wstring Utils::getShortPathWin(const std::wstring & longPath)
 
 	length = GetShortPathName(longPath.c_str(), buffer, length);
 	if (length == 0)
+	{
+		delete[] buffer;
 		return longPath;
+	}
 
 	std::wstring shortPath(buffer, length);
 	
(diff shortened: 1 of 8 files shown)


JorisGoosen:
Maybe @shun2wang can try it out on a chinese pc, maybe I missed something

shun2wang (approved):
It's really works and I can save and load CJK filename now!

I just add note that  `The current "_w" methods will disappear in libarchive 4` so we can still use it for now from libarchive https://github.com/libarchive/libarchive/wiki/Filenames

and the main issue is `ArchiveReader `  because even ASCII filename was cannot save either. so this should also fix: https://github.com/jasp-stats/jasp-issues/issues/2281#issuecomment-1715960494

boutinb (approved):

boutinb:
This should be merged in stable I think.

JorisGoosen:
Yes and that is what I intended but I forgot we changed the default branch...

shun2wang:
@JorisGoosen maybe trigger a build and let https://github.com/jasp-stats/jasp-issues/issues/2281 make sure if this also fixed?

PR #5247, opened 2023-09-13
Save should not always call Save As.

boutinb:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2383

A JASP file from the Data Library should be read-only, but once it is saved with another name on the PC (Save As), the file should not be read-only anymore

Add CTRL-SHIFT-S shortcut to call Save As.

### DIFF
--- a/Desktop/mainwindow.cpp
+++ b/Desktop/mainwindow.cpp
@@ -838,7 +838,7 @@ void MainWindow::saveKeyPressed()
 
 void MainWindow::saveAsKeyPressed()
 {
-	if (_package->hasDataSet()) _fileMenu->saveAs();
+	if (_package->isLoaded()) _fileMenu->saveAs();
 }
 
 void MainWindow::openKeyPressed()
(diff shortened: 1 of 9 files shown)


RensDofferhoff:
I fill check the shortcut on windows. Works fine on Linux

Desktop/mainwindow.cpp
```diff
@@ -836,6 +836,11 @@ void MainWindow::saveKeyPressed()
 	if (_package->isModified()) _fileMenu->save();
 }
 
+void MainWindow::saveAsKeyPressed()
+{
+	if (_package->hasDataSet()) _fileMenu->saveAs();
```
JorisGoosen: what about jasp-files where there is no dataSet? Like with sumstats?
Why not use `if(_package->isModified())` like in `saveKeyPressed`?

Desktop/widgets/filemenu/filemenu.cpp
```diff
@@ -108,13 +108,24 @@ FileEvent *FileMenu::open(const Json::Value & dbJson)
 	return event;
 }
 
+FileEvent *FileMenu::saveAs()
+{
+	FileEvent *event = _computer->browseSave();
+
+	if (!event->isCompleted())
+		// If the JASP file is saved on the computer, it should not be read only anymore (if it was)
```
JorisGoosen: What if it is still a readonly datafile?

You know the csv itself...

I can in fact easily demonstrate this problem with your PR:
- open debug.csv from data library
- add an analysis
- save it as jaspfile
- go to edit data mode
- turn on synchronization
- see how you load debug.csv from Resources in your build/install-folder (obviously it shouldnt)

Why does it do that?

Because you disable the property called "dataFileReadOnly"

Instead of doing that you shouldnot keep reusing `dataFileReadOnly` for multiple things.
Now it does:
- indicating whether the datafile is readonly
- indication whether we loaded something from data library (but halfbaked...)

Instead it should *only* indicate that the dataFile is in fact *read only*
And another property should be added indicating whether whether it is a readonly file from the data library (like a jasp file or something)

On the other hand you could also rename the property to something that describes its actual behaviour (so "fileFromDataLibrary" or something) and then make sure that the data-file-path is simply annulled on saving a jasp-file so that it won't start synching with that...

Desktop/mainwindow.cpp
JorisGoosen: Ah ok I see why not the isModified...

However!

There is *always* a dataSet in a package, except when exiting.
So now you can press Ctrl+Shift+S on the welcome screen and save that...

I'd suggest `if(_package->isModified() && _package->isLoaded())` or something like that

JorisGoosen (changes requested):
Please implement the changes I requested

Desktop/mainwindow.cpp
boutinb: _dataSet is null when JAPS is opened without file, and the _dataSet is set to nullptr when the workspace is closed, so on the welcome the Ctrol-Shift-S does not work...
But the isLoaded is indeed cleared. (isModified should not be used, since we should be able to do saveAs even is the file is not modified).

Desktop/widgets/filemenu/filemenu.cpp
boutinb: Good catch! I will make 2 properties.

Desktop/mainwindow.cpp
JorisGoosen: > _dataSet is null when JAPS is opened without file, and the _dataSet is set to nullptr when the workspace is closed, so on the welcome the Ctrol-Shift-S does not work.

Well I could get a save-as menu without having any analyses or data.
Maybe I just added and removed a distributions analysis without data or something.
But in any case, it is not a logical function to use for this. 

In fact Id also suggest making a function/property in datasetpackage `workspaceActive` or something.
It could now just be something like `workspaceActive() { return isModified() && isLoaded() }

JorisGoosen:
I moved the base of this PR to `stable` btw

boutinb:
Done!

JorisGoosen (approved):

PR #5248, opened 2023-09-14
Bugfix: handle wide table and Rsyntax printing

shun2wang:
The default output of Qt PDF printing conforms to the standard size of A4 paper. Anything outside this range will be cut off.

- Not a perfect solution, but for now it can handle with wide tables of about <30 columns.Wider tables will overlap rather than be cut off, I believe overlaps are more easily visible to users as incomplete, so this works for users processing tables manually .
- Wide Rsyntax is also can wrap in PDF

Fix https://github.com/jasp-stats/INTERNAL-jasp/issues/2341
Related https://github.com/jasp-stats/INTERNAL-jasp/issues/1715

### DIFF
--- a/Desktop/html/css/printing.css
+++ b/Desktop/html/css/printing.css
@@ -1,24 +1,27 @@
 @media only print
 {
-    #results {  display: block;  }
+    #results    {  display: block;  }
 
     .jasp-toolbar h1, .jasp-toolbar h2, .jasp-toolbar h3, .jasp-toolbar h4, .jasp-toolbar h5, .jasp-toolbar h6, .jasp-toolbar div , .jasp-toolbar span
-	                                { display:				block !important;	}
-	.toolbar-button					{ display:				none  !important;	visibility: hidden; }
-	.jasp-menu						{ display:				none  !important;	visibility: hidden; }
-	.jasp-menu-selected				{ display:				none  !important;	visibility: hidden; }
-	.jaspReportTop					{ display:				none  !important;	visibility: hidden; }
-	body							{ float:				none  !important;	}
+	                                    { display:	block !important;  }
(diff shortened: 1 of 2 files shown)


Desktop/html/css/printing.css
```diff
@@ -1,24 +1,30 @@
...
+	.object-body			    		{ page-break-inside:	auto;	}
+	.object-body .hidden-collection	    { page-break-inside:	auto;	}
+	.jasp-image			    			{ page-break-inside:	avoid;	}
+	.jasp-notes			    			{ page-break-inside:	avoid;	}
+
+	.jasp-table-primitive		   		{ page-break-inside:	avoid;	}
+    .jasp-table table               	{ table-layout:fixed !important;  width: 90em !important; }
+    .jasp-table table td            	{ white-space: pre-wrap !important; word-wrap: break-all !important; }
+
+	.jasp-rsyntax-container 			{ max-width: 90em !important; }
```
boutinb: What is this 90em coming from?

shun2wang: Haha, good point! This is what I calculated based on the A4 paper size - 1cm and the font size. But I rethought it and realized that this is unreasonable and should be set directly to the paper size.

Any suggestions?

boutinb: The max-width must be set with absolute unit. A4 is 21 cm width so let's use cm.
font-size is per default 12px, that is 0.3175 cm. The margin + padding = 1,7 * 2 em = 3,4 * 12px = 1cm. This could be of course bigger if the font-size is set bigger. So the max width could be 20 cm, but with some extra marges, I would say 18 cm.
What do you think?

boutinb (approved):

PR #5249, opened 2023-09-14
Add copy/cut/paste with header

boutinb:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2375

If a copy is done from the column header, the column name is also copied in the clipboard, so that if the user paste it in Excel, then the whole column with its name is copied. 
If a paste is done in a cell in JASP, then the clipboard is copied. 
If a paste is done in a column header, then if the copy was done form a column header, then the whole column (with name, labels, description...) is copied. 
If a cut is done in a column header, the whole column is removed.

Moreover it ensures now that the name of a column is unique.

### DIFF
--- a/CommonData/column.cpp
+++ b/CommonData/column.cpp
@@ -89,10 +89,11 @@ void Column::setName(const std::string &name)
 	if(_name == name)
 		return;
 
+	std::string orgName = _name;
 	_name = getUniqueName(name);
 
-	if(_title.empty() || _title == _name)
-		setTitle(name);
+	if(_title.empty() || _title == orgName)
+		setTitle(_name);
 
 	db().columnSetName(_id, _name);
 	incRevision();
--- a/CommonData/column.cpp
+++ b/CommonData/column.cpp
@@ -89,10 +89,11 @@ void Column::setName(const std::string &name)
 	if(_name == name)
 		return;
 
+	_name = getUniqueName(name);
+
 	if(_title.empty() || _title == _name)
 		setTitle(name);
 
-	_name = name;
 	db().columnSetName(_id, _name);
 	incRevision();
 }
@@ -2085,7 +2086,10 @@ Json::Value Column::serialize() const
 
 void Column::deserialize(const Json::Value &json)
 {
-	_name				= json["name"].asString();
+	if (json.isNull())
+		return;
+
+	_name				= getUniqueName(json["name"].asString());
 	db().columnSetName(_id, _name);
 
 	_title				= json["title"].asString();
@@ -2136,6 +2140,35 @@ void Column::deserialize(const Json::Value &json)
 	}
 
 	incRevision();
+}
+
+std::string Column::getUniqueName(const std::string &name) const
+{
+	std::string result	= name;
+	int	suffix			= 1;
+	bool foundSameName	= false;
+
+	do
+	{
+		foundSameName	= false;
+		for(Column * col : _data->columns())
+		{
+			if (col != this && col->name() == result)
+			{
+				foundSameName = true;
+				break;
+			}
+		}
+
+		if (foundSameName)
+		{
+			suffix++;
+			result = name + " " + std::to_string(suffix);
+		}
+
+	} while (foundSameName);
+
+	return result;
 }
 
 bool Column::dependsOn(const std::string & columnName, bool refresh)
--- a/CommonData/column.cpp
+++ b/CommonData/column.cpp
@@ -2090,10 +2090,14 @@ void Column::deserialize(const Json::Value &json)
 	if (json.isNull())
 		return;
 
-	_name				= getUniqueName(json["name"].asString());
+	std::string name = json["name"].asString(),
+				title = json["title"].asString();
+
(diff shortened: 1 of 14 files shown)


JorisGoosen:
> Moreover it ensures now that the name of a column is unique.

it already did that in paste with header

```
void DataSetPackage::unicifyColumnNames()
{
	for(int c=0; c<dataColumnCount(); c++)
	{
		std::string name = getColumnName(c);

		for(int c1=c+1; c1<dataColumnCount(); c1++)
			if(getColumnName(c1) == name)
			{
				std::string newName = name + "'";
				while(!isColumnNameFree(newName))
					newName = newName + "'";
				setColumnName(c1, newName);
			}
	}
}
```

but now we apparently have two functions for that haha

CommonData/column.cpp
```diff
@@ -89,10 +89,11 @@ void Column::setName(const std::string &name)
 	if(_name == name)
 		return;
 
+	_name = getUniqueName(name);
```
JorisGoosen, line 93: This is quite dangerous btw...

You now can have the situation where you set the columnname to **X** from somewhere else and get confirmation that it changed but then it is actually called **X2** something

Could lead to problems at:
https://github.com/jasp-stats/jasp-desktop/blob/c7dbe8e2e3463364ce2fd082cddb96e2e61d7973/Desktop/data/datasetpackage.cpp#L1740-L1745

https://github.com/jasp-stats/jasp-desktop/blob/c7dbe8e2e3463364ce2fd082cddb96e2e61d7973/Desktop/data/importers/importer.cpp#L140-L141

https://github.com/jasp-stats/jasp-desktop/blob/c7dbe8e2e3463364ce2fd082cddb96e2e61d7973/Desktop/analysis/analysis.cpp#L480-L486

Desktop/components/JASP/Widgets/DataTableView.qml
```diff
@@ -257,15 +259,15 @@ FocusScope
...
+								theView.copy(false);
 								event.accepted = true;
 							}
 							break;
 
 						case Qt.Key_X:
 							if(controlPressed)
 							{
-								theView.cut(shiftPressed);
+								theView.cut(false);
```
JorisGoosen: Shouldnt you also do this for https://github.com/jasp-stats/jasp-desktop/blob/c7dbe8e2e3463364ce2fd082cddb96e2e61d7973/Desktop/components/JASP/Widgets/JASPDataView.qml#L86-L99 ?

And maybe also just remove the argument if we arent using it anymore

Desktop/data/columnmodel.cpp
```diff
@@ -258,7 +283,8 @@ void ColumnModel::setChosenColumn(int chosenColumn)
 	//This only works as long as we have a single dataSet but lets not go overboard with rewriting stuff atm
 	DataSet * data = DataSetPackage::pkg()->dataSet();
 	
-	subNodeModel()->selectNode(data ? data->column(chosenColumn) : nullptr);
+	_virtual = data ? (chosenColumn >= data->columnCount()) : true;
```
JorisGoosen:
```suggestion
	_virtual = !data || data->columnCount() <= chosenColumn;
```

CommonData/column.cpp
JorisGoosen: Well, it turns out it mostly doesnt matter, because all of the places this gets called eventually already make sure that the columnnames are in fact unique.

CommonData/column.cpp
```diff
@@ -89,10 +89,11 @@ void Column::setName(const std::string &name)
...
+
 	if(_title.empty() || _title == _name)
```
JorisGoosen: You broke this check by overriding `_name`...

So before when you would change (lets say) `V1` to `Tsjakka` the title would change with it, but now it doesnt...

JorisGoosen (changes requested):

JorisGoosen:
So changes requested:
- columntitle should stay the same as the columnname unless the user specifically edited it
- JASPDataView should behave the same as the main dataviewer thing
- There shouldnt be multiple functions doing the same thing in JASP. So Either just use the existing "UnicifyColumnNames" (Why did you not use it @boutinb ? ) or remove it entirely if it isnt necessary anymore.

boutinb:
> So changes requested:
> 
> * columntitle should stay the same as the columnname unless the user specifically edited it
> * JASPDataView should behave the same as the main dataviewer thing
> * There shouldnt be multiple functions doing the same thing in JASP. So Either just use the existing "UnicifyColumnNames" (Why did you not use it @boutinb ? ) or remove it entirely if it isnt necessary anymore.

I think I have added most of your comments.

Also cut rows should removed rows when done from row header.

JorisGoosen:
Sorry, its almost good but:
image

I changed the name for `contExpon` to `contExpon!!!` (title changed with it), but when copying it it you get `contExpon!!! 2` as name but it keeps the title as is... Sorry, label or long name I mean!

JorisGoosen (approved):

PR #5266, opened 2023-09-25
JASP cooperative window

JorisGoosen:
made the texts translatable
split off the url from the texts to mke sure translators cant mess up the urls
adds the extr button to the hamburgermenu "Cooperative" as well

The defines for the supporters on the welcomepage and the educators tier of the cooperative
have been placed conveniently in Desktop/cooperativedefs.h
so even non-technical contributors can update them easily
if we add other tiers later on they can be placed there in a similar manner

### DIFF
--- a/Desktop/components/JASP/Widgets/ContactWindow.qml
+++ b/Desktop/components/JASP/Widgets/ContactWindow.qml
@@ -18,17 +18,26 @@ WavyWindow
 		id:				contactText
 		textFormat:		Text.RichText
 		text:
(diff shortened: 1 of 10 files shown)


JorisGoosen:
image

Like it @EJWagenmakers ?

RensDofferhoff (approved):

boutinb (approved):
Maybe next page we can put the list of the educatorsTiers in our static jasp server (like version and knownIssue).

JorisGoosen:
> Maybe next page we can put the list of the educatorsTiers in our static jasp server (like version and knownIssue).

Not a bad idea, but also maybe not necessary so much?

PR #5245, opened 2023-09-08
Make sure upgrade to conan2

shun2wang:
My conan was upgrade to 2.0 version but not really work so let's make sure it. the benefit of version 2.0 is that we can greatly reduce build configuration time on local builds without configuring dependencies again.

Using `pip install --upgrade conan` to upgrade local conan version to 2.0.x, and then using `conan version` to check install status.
maybe not wise but have to bump some pkg version to resolve conflict on conan center.

- Windows:

Visual Studio on my local is VS2022(i.e. VS17 193）in docs:
```
compiler.version=193
tools.microsoft.msbuild:vs_version = 17
```

- Linux:

N/A

- macOS:

I'm not on mac so cannot check it, anyone on mac can help.

PR #5269, opened 2023-09-27
New module installation mechanism

vandenman:
@RensDofferhoff I have only tested on Windows and Linux, and Windows was a few weeks ago. Could you give this a try on macOS?

I would recommend testing this PR in a new build folder because this _will_ mess with the module library and renv-cache.

Also, in `processhelper.cpp` it's possible that macOS needs some additional changes. I just haven't made them.

I still need to write some dev-docs and add more comments, so just see this as a first MWE.

All feedback is welcome!


RensDofferhoff:
I will test it on Mac!

JorisGoosen:
Ah, so this does not work on macos.

JASPengine depends on R-Interface which depends on RInside/Rcpp which depends on JASPEngine...
I think the only solution is to split off the otoolstuff into a separate application.
And outside of JASP in pure R land it isnt necessary either.

JorisGoosen:
Well, turns out that is not necesarily the problem, as the Patch.cmake would still be run.
But another problem is that `RScript` is used, which I think adds back this warning:
```
WARNING: ignoring environment value of
  R_HOME
```
Which then messes up this code:
```
if(NOT EXISTS ${RCPP_PATH})
	message(FATAL_ERROR "'Rcpp' installation has failed! ${RCPP_PATH}")
endif()
if(NOT EXISTS ${RINSIDE_PATH})
    message(FATAL_ERROR "'RInside' installation has failed!")
endif()

```


So I will rewrite the calls to RInside back to R

JorisGoosen:
Im getting closer but running into trouble like:
```
[1] "Installed jaspGraphs to '/Users/joris/Broncode/JASP/build-jasp-desktop-Qt_6_5_3_for_macOS-Debug/Modules/jaspTTests/.renv/1/jaspGraphs', now running post install fixes."
dyld[54036]: Library not loaded: libRInside.dylib
  Referenced from: <9C088926-30D4-3379-9AE5-3DB8D02E9EB5> /Users/joris/Broncode/JASP/build-jasp-desktop-Qt_6_5_3_for_macOS-Debug/Desktop/JASPEngine
  Reason: tried: 'libRInside.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OSlibRInside.dylib' (no such file), 'libRInside.dylib' (no such file), '/Users/joris/Broncode/JASP/build-jasp-desktop-Qt_6_5_3_for_macOS-Debug/Frameworks/R.framework/Versions/4.3-arm64/Resources/libRInside.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/Users/joris/Broncode/JASP/build-jasp-desktop-Qt_6_5_3_for_macOS-Debug/Frameworks/R.framework/Versions/4.3-arm64/Resources/libRInside.dylib' (no such file), '/Users/joris/Broncode/JASP/build-jasp-desktop-Qt_6_5_3_for_macOS-Debug/Frameworks/R.framework/Versions/4.3-arm64/Resources/libRInside.dylib' (no such file)
Error: install of package 'jaspGraphs' failed
In addition: Warning messages:
1: In generateBasicLockfile(lockfilePath, modulePkg, moduleLibrary) :
  Creating a basic lockfile by installing the module and unnecessarily installing GitHub dependencies. Please manage this lockfile yourself!
2: In getRecordsFromPkgdepends(modulePaths) :
  getRecordsFromPkgdepends failed!
3: In stop(e, domain = NA) : additional arguments ignored in stop()
```

Ive rebase the code on upgradeConan2 branch from #5245 and pushed it to: https://github.com/JorisGoosen/jasp-desktop/tree/newModuleInstall

JorisGoosen:
Well, I found some more problems, Rinside now actually gets patched and linked into jaspEngine.
However, the module folders are still empty for some reason and also Rcpp cant be found during runtime.
But that is something for next week

JorisGoosen:
Hmm, Ive taken your branch as it is in [1d52f92a5d5ff362cb0bc5f5fb589cfa5cdd0f1d](https://github.com/jasp-stats/jasp-desktop/pull/5269/commits/1d52f92a5d5ff362cb0bc5f5fb589cfa5cdd0f1d) , squashed it and added some mac fixes on top to make it compile and pushed that here: https://github.com/JorisGoosen/jasp-desktop/tree/newModuleInstall

however it does not run, maybe this is where I need to make some proceshelper changes?
But I have no good oversight on what all exactly changed :p

```
Engine#0:	Entering jaspRCPP_init.
Creating RInside.
Error: function 'Rcpp_precious_remove' not provided by package 'Rcpp'
In addition: Warning message:
In library(package, lib.loc = lib.loc, character.only = TRUE, logical.return = TRUE,  :
  there is no package called ‘Rcpp’
Execution halted
```

I guess it isnt finding Rcpp?
But im passing the main library and all, I guess there are libraries hidden somewhere else?

`R_CPP_INCLUDES_LIBRARY = /Users/joris/Broncode/JASP/build-jasp-desktop-Qt_6_5_3_for_macOS-Debug/R/R_cpp_includes_library` is probably it yeah
alright, Ill see how to get it in the processhelper

JorisGoosen:
I guess we should move the stuff under `build/R` to `build/Modules` at least on macOs
because otherwise we have to figure out exactly how to fit this into the bundle/framework structure all over again

JorisGoosen:
@vandenman want to integrate my changes? Or I can push to your branch perhaps

vandenman:
@JorisGoosen sorry for the delay, feel free to push directly to this branch!

JorisGoosen:
Ok, so the problem is that Rcpp is now not in the main R library of the framework.

So in jaspRcpp:
```
	rinside = new RInside();
```

Fails because it really depends on Rcpp (and maybe RInside haha)
In any case, the easiest solution would be to install Rcpp to the main library anyway.

But that might make @vandenman sad... So if he has a better solution?
`.libPaths` cant work cause R isnt loaded yet...

vandenman:
Doesn't RInside look at an environment variable that we can set? I'll take a look tomorrow.

vandenman:
Ok, so for me the current commit just works... I have an error with installing xml2 but that hopefully resolves itself. The main difference I think is that I have this when I print the libPaths:

image

we probably just need to set an env variable somewhere.

JorisGoosen:
Ok so the code at https://github.com/jasp-stats/jasp-desktop/tree/lockfiletesting does compile for macos and locally it works.
But the x86 dmg nightly gives this on startup:
afbeelding

vandenman:
superseded by https://github.com/jasp-stats/jasp-desktop/pull/5629

PR #5274, opened 2023-10-03
Fix data paste crash from some editors

RensDofferhoff:


boutinb:
Why to merge to buildRequest branch?

PR #5276, opened 2023-10-04
Variable Window redesign

### LINKED ISSUE #2355: [Bug]: File doesn't save using Command-S when changes are made only to column/variable names
### JASP Version

0.18.0

### Commit ID

_No response_

### JASP Module

Other

_No response_

macOS Intel

### Bug Description

If I open a JASP file, click on Edit Data, and then edit a variable/column name, the file will not save the updated name change using Command-S. If If I I close the file, it does not ask if I wish to save the changes. And when you reopening the file, you can see the changes have not been recorded. 

However, if you use the Main Menu --> Save command, the changes to the column name are saved. 

### Expected Behaviour

When variable names are edited, it should be noted as a change to the file, and 1) pressing Command-S should save the change as usual, OR 2) the user should be prompted to save when closing the file. 

### Steps to Reproduce

1. Open File that contains some data
2. Click Edit data
3. Click any column name
4. Click in the Name box, and make a name change
5. Press X to exit the column name editor
6. Press the Analyses button to return to the Analyses window
7. Press Command-S...there should be no response
8. Close the window (there will be no prompt to save changes)
9. Reopen the file and navigate to the edited variable name, and the change will not have been saved. 
...

### Log (if any)

_No response_

### Final Checklist

boutinb:
This PR redesigns the Variable Window with several purposes:

- When clicking on a 'virtual' column appears a popup window to make a computed column. This is not really user-friendly. In place the Variable window is now opened, where the user can set a name for the new column. If a name is set, the column is then created. The user can then change it in a computed column (and switch between R & drag-and-drop mode).
This gives also the user the possibility to create computed column and to switch it to a normal column.
- The column. type is also added in the Variable window, since it would be strange to be able to set all properties of a column, but not its type.
- A new tab is added, to make it possible to set custom empty values for a column.

Moreover, as the empty values must be stored in a JASP files, and reloaded when loading a JASP file, these empty values are maybe not the same as the ones of the user preferences. So a Workspace settings window has been added in the data edit mode (that replaces the Module menu). In this workspace settings, a description field has also been added, so the user can set more information.

Fixes https://github.com/jasp-stats/jasp-issues/issues/2355


CommonData/column.cpp
```diff
@@ -319,7 +349,7 @@ bool Column::_resetEmptyValuesForNominal(intstrmap &emptyValuesMap)
 			if(emptyValuesMap.count(row))
 			{
 				std::string orgValue = emptyValuesMap[row];
-				if (!ColumnUtils::isEmptyValue(orgValue))
+				if (!ColumnUtils::isEmptyValue(this, orgValue))
```
JorisGoosen: Better on column.cpp...

boutinb:
Not yet ready

JorisGoosen:
Just change it back to PR when it *is ready*

JorisGoosen:
image
❤️

CommonData/column.cpp
```diff
@@ -1684,10 +1727,10 @@ bool Column::setStringValueToRowIfItFits(size_t row, const std::string & value,
 						int integer;
 						if(     (
 									!oldLabel ||
-									!ColumnUtils::convertValueToIntForImport(oldLabel->label(),	integer)
+									!ColumnUtils::convertValueToIntForImport(oldLabel->label(),	integer, this)
 								)
 								&&
-								ColumnUtils::convertValueToIntForImport(value,				integer))
+								ColumnUtils::convertValueToIntForImport(value,				integer, this))
```
JorisGoosen: Ach....

Ill change it then

JorisGoosen:
Ok Ive made a bunch more changes I think its good like this, imma merge it ;)

PR #5281, opened 2023-10-09
Add dataSetInfo properties

boutinb:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2413

The QML files use dataSetModel.rowCount() and mainWindow.dataAvailiable. There are 2 problems with this: 

- First, QML analysis files should not use objects that are not defined outside the QMLComponents library (which will become a submodule), since these QML files will be used outside JASP (in R Studio).
- Second, dataSetModel.rowCount()  is a function, not a property, so it cannot signal its changes. So it should be made as a property.

For this, an object dataSetInfo is made available in the root context with 2 properties dataAvailable & rowCount. The provider (here the ColumnsModel) has to take care that the signals are well propagated.

When this PR is merged, the QML files will have to be updated.

### DIFF
--- a/QMLComponents/variableinfo.cpp
+++ b/QMLComponents/variableinfo.cpp
@@ -1,5 +1,7 @@
 #include "variableinfo.h"
 #include "jasptheme.h"
+#include "QQmlContext"
+#include "QTimer"
 
 VariableInfo* VariableInfo::_singleton = nullptr;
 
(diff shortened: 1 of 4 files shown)


QMLComponents/variableinfo.cpp
```diff
@@ -1,13 +1,24 @@
...
 
 VariableInfo* VariableInfo::_singleton = nullptr;
 
 VariableInfo::VariableInfo(VariableInfoProvider* providerInfo) :
 	QObject(providerInfo->providerModel()), _provider(providerInfo)
 {
 	if (_singleton == nullptr)
+	{
 		_singleton = this;
+		QTimer::singleShot(0, [&]() { _setDataSetInfoInContext(); });
```
JorisGoosen, line 14: Why not set it from mainwindow loadQML?

boutinb: I find more logical that the QMLComponents module itself registers the VariableInfo/dataSetInfo object.
The QMLComponents module should be as independent as possible of its environment, only a tiny interface (VariableInfoProvider) defines what is needed from the environment.
Now it is clear that other environments than JASPDesktop should provide the QML root Context object to QMLComponents so that the module can work: the module is not dependent on the VariableInfo object being registered elsewhere.

JorisGoosen: Ok cool!

boutinb: Can I merge this PR then?

JorisGoosen: No sorry I still want to test it but I first want to build development

JorisGoosen (approved):

PR #5293, opened 2023-10-23
Feat: Add video iframe support in annotation

shun2wang:
Close https://github.com/jasp-stats/INTERNAL-jasp/issues/2402

Let's evaluate the security of doing this together: the current approach is a domain whitelist mechanism, and we can also choose the URL entered by the user on the js side (I plan to do this but not now) to prevent XSS attacks when it becomes a web program in the future. any other security issues that should be more special attention?

### DIFF
--- a/Desktop/components/JASP/Widgets/MainPage.qml
+++ b/Desktop/components/JASP/Widgets/MainPage.qml
@@ -309,7 +309,7 @@ Item
 				else
 				{
 					request.reject();
-					console.log("Navigation requeste rejected:", requestedURL.hostname)
+					console.log("Navigation requested rejected:", requestedURL.hostname)
 				}
 			}
 
--- a/Desktop/components/JASP/Widgets/MainPage.qml
+++ b/Desktop/components/JASP/Widgets/MainPage.qml
@@ -268,7 +268,7 @@ Item
 					} 
 					else if (pattern.indexOf('*') !== -1)
 					{
-						const regex = new RegExp(`^${pattern.replace(/\./g, '\\.').replace(/\*/g, '.*')}$`);
+						var regex = new RegExp(`^${pattern.replace(/\./g, '\\.').replace(/\*/g, '.*')}$`);
 						if (regex.test(hostname))
 						{
 							return true;
--- a/Desktop/components/JASP/Widgets/MainPage.qml
+++ b/Desktop/components/JASP/Widgets/MainPage.qml
@@ -270,9 +270,12 @@ Item
 			onFullScreenRequested: function(request) 
 			{
 				request.accept()
-				analysesModel.visible = false
-				minimizeDataPanel()
-				mainWindowRoot.toggleFullScreen()
+
+				if(request.toggleOn)
+				{
+					analysesModel.visible = false
+					minimizeDataPanel()
+				}
 			}
 
 			onNavigationRequested: (request)=>
--- a/Desktop/components/JASP/Widgets/MainPage.qml
+++ b/Desktop/components/JASP/Widgets/MainPage.qml
@@ -299,6 +299,8 @@ Item
 				{
 					request.reject();
 					console.log("Navigation requeste rejected:", requestedURL.hostname)
+
+					messages.showWarningQML(qsTr("Video link not accepted"), qsTr("JASP only allows the following videoservices: \n\n%1\n\nContact the JASP team to request adding another videoservice to the list.".arg(urlWhitelist.join(", "))));
 				}
 			}
 
--- a/Desktop/components/JASP/Widgets/MainPage.qml
+++ b/Desktop/components/JASP/Widgets/MainPage.qml
@@ -255,14 +255,25 @@ Item
 				}
 			}
 
-			property var urlWhitelist : ['www.youtube.com', 'www.youtu.be', 'player.vimeo.com', 'www.bilibili.com', 'v.qq.com'];
+			property var urlWhitelist : ['www.youtube.com', 'www.youtu.be', '*.vimeo.com', '*.vimeocdn.com', '*.akamaized.net', '*.bilibili.com', '*.hdslb.com', '*.qq.com', '*.smtcdns.com'];
 
-			function isURLInWhitelist(hostname) 
+			function isURLInWhitelist(hostname)
 			{
-				for (var i = 0; i < urlWhitelist.length; i++) 
+				for (var i = 0; i < urlWhitelist.length; i++)
 				{
-					if (hostname === urlWhitelist[i]) 
+					var pattern = urlWhitelist[i];
+					if (pattern === hostname)
+					{
 						return true;
+					} 
+					else if (pattern.indexOf('*') !== -1)
+					{
+						const regex = new RegExp(`^${pattern.replace(/\./g, '\\.').replace(/\*/g, '.*')}$`);
+						if (regex.test(hostname))
+						{
+							return true;
+						}
+					}
 				}
 				return false;
 			}
@@ -299,8 +310,6 @@ Item
 				{
 					request.reject();
 					console.log("Navigation requeste rejected:", requestedURL.hostname)
-
-					messages.showWarningQML(qsTr("Video link not accepted"), qsTr("JASP only allows the following videoservices: \n\n%1\n\nContact the JASP team to request adding another videoservice to the list.".arg(urlWhitelist.join(", "))));
 				}
 			}
 
@@ -308,6 +317,7 @@ Item
 			{
 				resultsJsInterface.resultsLoaded = loadRequest.status === WebEngineView.LoadSucceededStatus;
 				setTranslatedResultsString();
+				runJavaScript(`window.sendUrlWhitelist(${JSON.stringify(urlWhitelist)})`); //sent urlWhitelist to js side
 			}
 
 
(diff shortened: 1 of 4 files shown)


JorisGoosen:
Ok this is pretty hilarious: 
image
that is after choosing "fullscreen"

shun2wang:
well, seems I should also hide the data pannel while fullscreen requested....but even so, it's not really a "fullscreen mode" the video will olny fullscreen in results page. Is there a way to make webengine jump out of the application main window?

JorisGoosen:
The changes look good and it seems to work well.
Also, because the qml contains the whitelist and is baked into the executable, I think the safety of the proposed mechanism is probably good enough.
Not sure who would try to hack someone through a jaspfile if they already have control over youtube and co :p

As far as the whitelist goes, youtube and vimeo seem like a good idea.
But 'www.bilibili.com', 'v.qq.com' I have no idea about, which makes sense as they seem to be aimed at quite a different part of the world. I assume they are rather large and/or wellknown?

JorisGoosen:
> well, seems I should also hide the data pannel while fullscreen requested....but even so, it's not really a "fullscreen mode" the video will olny fullscreen in results page. Is there a way to make webengine jump out of the application main window?

Well, it actually makes the JASP application full screen right now...
I think it would be better if we just show the video covering the entire resultwindow.
Would be good to keep some gui element of JASP in view.

JorisGoosen:
image
Also the gui elements themselves are negatives. The video are not though.

shun2wang:
>Not sure who would try to hack someone through a jaspfile if they already have control over youtube and co :p

Yes, not for desktop applications, but we have a plan to run jasp as a cloud service (in the long term), right? For web applications, I don’t think it’s too much to handle it from the input side.

>But 'www.bilibili.com', 'v.qq.com' I have no idea about, which makes sense as they seem to be aimed at quite a different part of the world. I assume they are rather large and/or wellknown?

Yes, they are more widely used than youtube in China and be credible.Once a user requests support for something else we can add it later.

shun2wang:
>I think it would be better if we just show the video covering the entire resultwindow.

haha, I guess it had coverd entire resultwindow already. but just data pannel not hidden.
My understanding of full screen was that the video covers all windows (just like chrome browser behavior), and the current full screen is within the page. But as you said, we should keep the JASP GUI, which will the analysis menu at the top.

should like:
图片
right?

JorisGoosen:
Yes it should look like that indeed, but then without the JASP GUI itself going fullscreen too :p

JorisGoosen:
> > Not sure who would try to hack someone through a jaspfile if they already have control over youtube and co :p
> 
> Yes, not for desktop applications, but we have a plan to run jasp as a cloud service (in the long term), right? For web applications, I don’t think it’s too much to handle it from the input side.

Hmm, yeah, although when we do that we will need to change a lot of things in the GUI.
As in, I dont think it makes a lot of sense to have a byte-compiled version of Chromium (webengine) running in webassembly (which the rest of jasp would have to be compiled to) inside chrome or firefox :p

> > But 'www.bilibili.com', 'v.qq.com' I have no idea about, which makes sense as they seem to be aimed at quite a different part of the world. I assume they are rather large and/or wellknown?
> 
> Yes, they are more widely used than youtube in China and be credible. Once a user requests support for something else we can add it later.

Alright seems fine!

JorisGoosen:
So Ive changed the behaviour of JASP to not make the application go fullscreen when youtube does.
If someone wants to fullscreen JASP they can just use Ctrl/Cmd+M or F11

Ive also added a feedbackwarning if a link that isnt accepted is pasted:
image

However there is still a big empty space in the note afterwards so maybe there needs to be a check on whether the navigationrequest is accepted?

Its pretty awesome to have a tutorial video integrated into your jasp file (or a rickroll hahaha)

shun2wang:
Yes, it's good to check on js side to waring this and remove this empty block. because quill always insert iframe into editor. I'm trying to figure out how to do this by hacking quill...you know we use it not `normal` hahaha

your feedback warning is great i think.

shun2wang:
and we should also custom the url because user usually insert no-embed link, quill did it for youtube only( means you inset `youtube.com/watch=...` and custom it to `youtube.com/embed/...` so I should do it for other links. but this is may not now? maybe later?

JorisGoosen:
> and we should also custom the url because user usually insert no-embed link, quill did it for youtube only( means you inset `youtube.com/watch=...` and custom it to `youtube.com/embed/...` so I should do it for other links. but this is may not now? maybe later?

Well I first tried pasting the embed link for youtube but it didnt work :p 
But yeah I guess normal users wouldnt do that and it works pretty easily now.

Its probably nice to do it for the other services but then again, I didnt even test those.
Ill at least try out vimeo now then.

JorisGoosen:
Ok I cant get vimeo in there at all, after manually copying a tiny piece from the vimeo embed share link I got to this:
image

link was: `https://player.vimeo.com/video/873167245`

How did you get a working link or you did not try either?

Id be happy with just youtube working, despite the fact that we might accidentally introduce advertisements into JASP this way...

shun2wang:
I tried everything haha and I know it will open entire page if not using a embed link. yes the `player.vimeo.com` is the embed already that not problems.

shun2wang:
Let's keep this PR open for a few days?  I'll try if I can fix this as soon as possible.

a quick way maybe passing the urlWhiteLlist in js side and then check it then do something.

shun2wang:
hmmmm, vimeo video still broken, not sure what happend...I once successed on a PC with proxy but now i cannot reproduce it.

now we give message from js side because browser will not request unsupported url.

Desktop/components/JASP/Widgets/MainPage.qml
```diff
@@ -255,14 +255,25 @@ Item
 				}
 			}
 
-			property var urlWhitelist : ['www.youtube.com', 'www.youtu.be', 'player.vimeo.com', 'www.bilibili.com', 'v.qq.com'];
+			property var urlWhitelist : ['www.youtube.com', 'www.youtu.be', '*.vimeo.com', '*.vimeocdn.com', '*.akamaized.net', '*.bilibili.com', '*.hdslb.com', '*.qq.com', '*.smtcdns.com'];
```
JorisGoosen: is `smtcdns.com` a chinese content delivery network or something?
The `akamaized` one I can find some info on, just like for `hdslb.com`

Desktop/components/JASP/Widgets/MainPage.qml
```diff
@@ -299,15 +310,14 @@ Item
...
-
-					messages.showWarningQML(qsTr("Video link not accepted"), qsTr("JASP only allows the following videoservices: \n\n%1\n\nContact the JASP team to request adding another videoservice to the list.".arg(urlWhitelist.join(", "))));
 				}
 			}
 
 			onLoadingChanged: (loadRequest)=>
 			{
 				resultsJsInterface.resultsLoaded = loadRequest.status === WebEngineView.LoadSucceededStatus;
 				setTranslatedResultsString();
+				runJavaScript(`window.sendUrlWhitelist(${JSON.stringify(urlWhitelist)})`) //sent to js side
```
JorisGoosen: Wouldnt it be much easier to change the whitelist once its loaded into JS in the browser?
Although I guess it isnt possible to store any JS in a jasp-file so I guess it doesnt matter?

Desktop/components/JASP/Widgets/MainPage.qml
shun2wang: This is some video cdn server which vimeo or other sever used. some main sites redirect to them.

Desktop/components/JASP/Widgets/MainPage.qml
shun2wang: These will not be stored in jasp files, the whitelist will be sent to the js side via this signal (getter/setter in browser but not in node),I also used to send translation strings in qml like this, or am I missing some other way?

JorisGoosen: Well, I mean tht if someone manages to hijack the JS a little bit they could then overwrite:
```
function isURLInWhitelist(hostname) {
	for (var i = 0; i < videoUrlList.length; i++) {
		var pattern = videoUrlList[i];
		if (pattern === hostname) {
			return true;
		} else if (pattern.indexOf('*') !== -1) {
			const regex = new RegExp(`^${pattern.replace(/\./g, '\\.').replace(/\*/g, '.*')}$`);
			if (regex.test(hostname)) {
				return true;
			}
		}
	}
	return false;
}
```
with 
```
function isURLInWhitelist(hostname) { return true; }
```
No?
And then load whatever

Desktop/html/js/jaspwidgets.js
```diff
@@ -1263,42 +1335,42 @@ JASPWidgets.ProgressbarView = JASPWidgets.View.extend({
...
 		this.initialize();
 	},
 
-	isActive: function() {
+	isActive: function () {
 		return -1 < this.model.get("value") && this.model.get("value") < this.model.get("maxValue");
 	},
 
-	isComplete: function() {
+	isComplete: function () {
```
JorisGoosen: Maybe undo this linter-action?
As in, every `function` and `if` got a space...

Desktop/components/JASP/Widgets/MainPage.qml
JorisGoosen: image
Ah yeah for `qq`

Desktop/components/JASP/Widgets/MainPage.qml
JorisGoosen: aaaah wait, of course, there is 2 checks now, so yeah no that is fine!

Desktop/html/js/jaspwidgets.js
shun2wang: ah,yeah,It was an accident by vscode.

shun2wang:
well, maybe I know what happend to Vimeo video. the browser what we using are [not support](https://www.chromium.org/audio-video/) H.264 video encode. unless we compiled qt with an enable flag by ourself (I guess will not do that). other videoserver suported `AV1` or `VP8/9` that's why Youtube video have no problems. as not all video supported that we can support youtube only?

There are two options: remove vimeo support completely. Or leave it and let the request open in the user's browser or exported as result.html to open in other browsers (but this is not as convenient as embedding the URL directly).

EDIT: For a user experience, we only support YouTube now.

Desktop/components/JASP/Widgets/MainPage.qml
```diff
@@ -255,22 +255,69 @@ Item
...
+				{
+					request.accept();
+					console.log("Navigation requeste accepted:", requestedURL.hostname)
+				}
 				else
 				{
-					if(request.navigationType === WebEngineNavigationRequest.LinkClickedNavigation)
-						Qt.openUrlExternally(request.url);
 					request.reject();
+					console.log("Navigation requeste rejected:", requestedURL.hostname)
```
JorisGoosen:
```suggestion
					console.log("Navigation requested rejected:", requestedURL.hostname)
```

JorisGoosen (approved):

PR #5294, opened 2023-10-23
Add option runtime

boutinb:
During the code signing, to add the option runtime with a variable containing the string "--options runtime" does not work.
So just use an if else construct to add this option.


Tools/CMake/Patch.cmake
```diff
@@ -42,7 +42,7 @@ else()
   file(
     GLOB_RECURSE
     LIBRARIES
-    #FOLLOW_SYMLINKS #Turned this off because it got infinitely regressed
+    #FOLLOW_SYMLINKS #Turned this off because it got infinitely regressed, but Joris turned it back on cause it breaks things
```
JorisGoosen:
```suggestion
    #FOLLOW_SYMLINKS #Turned this off because it got infinitely regressed
```

except here he didn'

JorisGoosen (approved):
❤️

PR #5298, opened 2023-10-31
Add possibility to make editable or not any cell in TableView

boutinb:
Add the function getEditable(columnIndex, rowIdex) that can be overriden to set a cell editable or not.
I will add a test in jaspTestModule


JorisGoosen:
you have `flags()` for that

edit: so you want this for qml inheritance?

JorisGoosen (approved):

PR #5302, opened 2023-11-02
jaspColumn improvements

JorisGoosen:
set analysisId to -1 in database on creation
make errormessage from jaspHtml show
use checkForFix branch of jaspBase

part of the fix for https://github.com/jasp-stats/INTERNAL-jasp/issues/2427 (AddColumnField and Distributions together can mess up the data)

### DIFF
--- a/CommonData/column.cpp
+++ b/CommonData/column.cpp
@@ -6,8 +6,7 @@
 
 Column::Column(DataSet * data, int id) 
 	: DataSetBaseNode(dataSetBaseNodeType::column, data->dataNode()), _data(data), _id(id)
-{
-}
+{}
 
 void Column::dbCreate(int index)
 {
@@ -21,7 +20,8 @@ void Column::dbLoad(int id, bool getValues)
 {
 	JASPTIMER_SCOPE(Column::dbLoad);
 
-	assert(_id != -1 || id != -1);
+	assert(_id == id || (id != -1 && _id == -1) || _id != -1);
+
 	if(id != -1)
 		_id = id;
 
@@ -45,7 +45,9 @@ void Column::dbLoadIndex(int index, bool getValues)
 {
 	JASPTIMER_SCOPE(Column::dbLoadIndex);
 
-	dbLoad(db().columnIdForIndex(_data->id(), index), getValues);
+	_id = db().columnIdForIndex(_data->id(), index);
+
+	dbLoad(_id, getValues);
 }
 
 void Column::dbDelete(bool cleanUpRest)
--- a/CommonData/column.cpp
+++ b/CommonData/column.cpp
@@ -205,9 +205,11 @@ void Column::setCodeType(computedColumnType codeType)
 		_constructorJson["formulas"] = Json::arrayValue;
 	}
 
-	_codeType = codeType;
-	_isComputed = _codeType != computedColumnType::notComputed;
+	if(_codeType == computedColumnType::analysisNotComputed)
+		_analysisId = -1;
 
+	_codeType	= codeType;
+	_isComputed = _codeType != computedColumnType::notComputed && _codeType != computedColumnType::analysisNotComputed;
 	
 	dbUpdateComputedColumnStuff();
 }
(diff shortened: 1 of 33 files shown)


JorisGoosen:
I think I might have accidentally dropped a commit from this?

JorisGoosen:
found it

JorisGoosen:
image

For computed columns from analyses only.
Columns from analyses that arent computed (from Distributions) get a dropdown that allows you to cut it free now.

JorisGoosen:
Ok, so I couldnt let go apparently, but Ive now also added some support for https://github.com/jasp-stats/INTERNAL-jasp/issues/694
Not quite computed columns, but you can now add columns from R as shown in: https://github.com/jasp-stats/jaspTestModule/blob/2ca3c9f5428d3a33356fa1b711ec7323e0f577a3/R/testAnalysisColumns.R#L23-L42

This should make it possible for https://github.com/jasp-stats/jasp-issues/issues/101 and other things to be implemented finally

JorisGoosen:
we should probably test undo/redo well with this

shun2wang:
If pacehoders in "long name" and "description" field will never showing😜

the field width seems (from qml code) has a good fix for multiple languages.

JorisGoosen:
also fixed labelwidth for @shun2wang ;)
image

> If pacehoders in "long name" and "description" field will never showing😜 

?

shun2wang:
I mean:
`placeholderText:	qsTr("<Fill in a more descriptive name of the column>")`

for long name field will never show because it always repalced by default value? but also the `description` of '...' will not showing.

but it doesn't matter...Or we can use tooltip here.

JorisGoosen:
Aah yeah, well, thats ok :p

Im happy if it works, but some tooltips might be good.
right now im just gonna make sure it all aligns well and works

JorisGoosen:
Ok I already found a bug... 
You can undo going from a analysisNotComputed column to a normal one. 
But that move resets the stored analysisID in the column so if you undo the "show analysis button" doesnt work anymore.
Its getting kind of late though so im gonna call it a night

boutinb:
By testing it, I could find some issues:

- In a Distribution analysis, you can now overwrite an existing column. I thought this should be forbidden.
- If a computed column is added, a 'Show parent analysis' is displayed in the Variable window, but this button does not work.
- In Data Mode, if you click on a empty column (to create it), you get in the Variable window the 'Show parent analysis'
- In order to better see the missing values of a column, you have made the Variable Window higher. But if I make this height smaller (in another tab, eg. with the Label Editor tab), then if I switch to the Missing Values, I don't see the values anymore. To prevent this, we could increase the Variable Window minimum height.

JorisGoosen:
> In a Distribution analysis, you can now overwrite an existing column. I thought this should be forbidden.

Only if the analysis itself made it!
Did you do `git submodule update` and start from a fresh buildfolder?

> If a computed column is added, a 'Show parent analysis' is displayed in the Variable window, but this button does not work.

"Added" ? By what? Manually or from an analysis?

> In Data Mode, if you click on a empty column (to create it), you get in the Variable window the 'Show parent analysis'

Oh ja? Ill check it out

> In order to better see the missing values of a column, you have made the Variable Window higher. But if I make this height smaller (in another tab, eg. with the Label Editor tab), then if I switch to the Missing Values, I don't see the values anymore. To prevent this, we could increase the Variable Window minimum height.

Or you just make it bigger again?

Ill just increase the minimum height for now or maybe split it

JorisGoosen:
Ok I fixed all the mentioned problems

boutinb:
Much better, but...
- Add column from a Destribution analysis, close the Analysis pane, so that you only see the data panel. Open the Variable Window, and click on the 'Show parent analysis' button: nothing happens
- Add a column from a Distribution analysis. Change the Computed type to 'Not computed'. Save and reload the JASP file: the column header has a Computed column icon, and the cells of this column are not editable
Screenshot 2023-11-08 at 14 34 22

JorisGoosen:
> Add column from a Destribution analysis, close the Analysis pane, so that you only see the data panel. Open the Variable Window, and click on the 'Show parent analysis' button: nothing happens

This is not true, except for when you are in data-editing-mode.
Ill add a switch-to-analysis mode if you click that button

boutinb (approved):

PR #5304, opened 2023-11-03
Misc: enable auto-build test on Linux

shun2wang:
Now we re-enable build testing on Linux, currently trigger was set by PR or manually.

Automatic testing will install all modules,not sure why but I have set `JASP_TEST_BUILD=ON` but it didn't work, full build will be 3-4h otherwise the test build only takes about 30-40 minutes. 

The long term idea is that if we use docker to package a container or use QEMU we might be able to test the GUI event or .jasp file outputs in the CLI.

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -1,32 +1,75 @@
-name: Ubuntu
+name: 'Ubuntu'
 
 on:
-  push:
-    branches:
-      - github-action-debug
+   pull_request:
+   workflow_dispatch:
 
 jobs:
-  qt:
-    name: Ubuntu Build
-    if: false
+      
+   build-jasp:
+   
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        qt-version: [6.2]
-
+    env:
+       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
+  
     steps:
-      - name: Qt Requirements
+      - uses: actions/checkout@v3
+      
+      - uses: r-lib/actions/setup-r@v2
+        with:
+          r-version: '4.3.1'
+      
+      - name: Install Qt6
+        uses: jurplel/install-qt-action@v3
+        with:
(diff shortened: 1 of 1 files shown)


JorisGoosen:
Ok changes look good, retriggered the build, lets see whether it compiles, if so we can merge otherwise...

shun2wang:
it will not success because some packages not install like Matrix or something. i need to let  `JASP_TEST_BUILD=ON` env enable not sure why it not work now.


I'll also add a comment trigger for it because we don't always need to build tests right? PR trigger is not good idea(I'm afraid we will reached the workflow times limit), and but manual trigger requires permissions

JorisGoosen:
> I'll also add a comment trigger for it because we don't always need to build tests right? PR trigger is not good idea(I'm afraid we will reached the workflow times limit), and but manual trigger requires permissions

Yeah maybe we should just run it every week or something?

> t will not success because some packages not install like Matrix or something. i need to let JASP_TEST_BUILD=ON env enable not sure why it not work now.

Ah, well that might be because JASP_TEST_BUILD is not an environment variable but a CMAKE variable.
So you should add it to the cmake config step as `-DJASP_TEST_BUILD=ON` or something like that

shun2wang:
>Ah, well that might be because JASP_TEST_BUILD is not an environment variable but a CMAKE variable.

yeah, I tried it once but seems still build all pkg.
https://github.com/shun2wang/jasp-desktop/commit/5b302a54d053ddb15b92cad7d0c5c0a7adab91a7

and logs: https://github.com/shun2wang/jasp-desktop/actions/runs/6742743045/job/18384124352

EDIT: OK, then I added to config step not build setp,haha.

JorisGoosen:
Im curious to see how many minutes get used if we use this for every PR.
Very good to have  check on the build on every PR

JorisGoosen (approved):
I like it, later we can add some regression tests, but just having the check for building is great already

PR #5306, opened 2023-11-06
Fix long-name filed width under multiple languages

shun2wang:
Variable window variable long-name filed width should adapt to translated label text width. for test: change to Chinese language and see new design of variable window close button.
also small style changes of translation strings.

Before:
_the close button is nearly blocked and filed not same wide._
图片

After:
图片

### DIFF
--- a/Desktop/data/undostack.cpp
+++ b/Desktop/data/undostack.cpp
@@ -536,7 +536,7 @@ void SetRFilterCommand::redo()
 CreateComputedColumnCommand::CreateComputedColumnCommand(QAbstractItemModel *model, const QString &name, int columnType, int computedColumnType)
 	: UndoModelCommand(model), _name{name}, _columnType{columnType}, _computedColumnType{computedColumnType}
 {
-	setText(QObject::tr("Create a computed column with name '%1").arg(name));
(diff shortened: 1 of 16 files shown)


Desktop/components/JASP/Widgets/VariablesWindow.qml
```diff
@@ -184,7 +184,7 @@ FocusScope
 						value:				columnModel.columnTitle
 						onValueChanged:		if(columnModel.columnTitle !== value) columnModel.columnTitle = value
 						undoModel:			columnModel
-						controlLabel.width:	Math.max(columnTitleVariablesWindow.controlLabel.implicitWidth, descriptionLabel.implicitWidth)
+						controlLabel.width:	Math.max(columnTitleVariablesWindow.label.length, descriptionLabel.implicitWidth)
```
JorisGoosen: Hmm, this is using the length of the text in the label itself, while `implicitWidth` is supposed to be the actual width.

Im not saying we shouldnt use a different variable here or handle the change of the label after a languagechange.
Ill have a look as well

shun2wang: Then I will dorp that commit from this PR.

boutinb (approved):

PR #5307, opened 2023-11-07
Condition on source using components

boutinb:
If a source use a component like a dropdown for example:
    'source: myVariables.drop' // if drop is the dropdown name.
And is this source use some condition using another component:
    'source: [{name: "myVariables.drop", condition: "check" }] // if check the name of checkbox component of the source.
Then the condition should be applied for the values of the source model.

To understand better what happens, see in the jaspTestModule, the "Test Sources with special attributes" analysis, "Source with controls" section

### DIFF
--- a/CommonData/column.cpp
+++ b/CommonData/column.cpp
@@ -6,8 +6,7 @@
 
 Column::Column(DataSet * data, int id) 
 	: DataSetBaseNode(dataSetBaseNodeType::column, data->dataNode()), _data(data), _id(id)
-{
-}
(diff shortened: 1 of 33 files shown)


JorisGoosen:
you dont want this anymore?

boutinb:
Hmm.. Yes I want this, but I'm afraid I made a mistake...

boutinb:
I have made https://github.com/jasp-stats/jasp-desktop/pull/5314

PR #5308, opened 2023-11-08
change cooperative to community

JorisGoosen:
@EJWagenmakers @JohnnyDoorn ive changed cooperative to community everywhere, and added "of benefactors" to it in the "CommunityWindow" (cause community is of course a bit broader normally then we use it here)

image


EJWagenmakers:
Great. However, "of benefactors" should go, imo. First, it suggests that nothing is given in return, which is not true; second, we might want to reserve the term "benefactors" for companies/institutions that wish to contribute but do not have a 100% teaching objective. I know that "Community" is too broad, but this is what the lawyer wanted. So ..."part of the community of benefactors" ought to be replaced with "part of the JASP community", I think.

JorisGoosen:
Ok, Ill change it.

Also `https://jasp-stats.org/community-how-to-join/` does not exist yet, just like the `https://jasp-stats.org/community-vision-and-goals/` 

I also just made a change to make the clickable area on the welcomepage a little bit smaller (the same size as the "link") on @koenderks 's request

boutinb (approved):

PR #5312, opened 2023-11-09
Beter handling of allowed types

boutinb:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2429

There was some confusions with the displayed 'allowed types' icons in a Variables List, and the real allowed types.

The Test Variables Form (suggested vs allowed Section) in jaspTestModule explains exactly what the the suggested & allowed properties should work.

### DIFF
--- a/QMLComponents/components/JASP/Controls/AssignButton.qml
+++ b/QMLComponents/components/JASP/Controls/AssignButton.qml
@@ -61,23 +61,7 @@ Button
 	function setIconToLeft()	{ if (rightSource.activeFocus)	leftToRight = false; setState(); }
 	function setState()
 	{
-		var isEnabled = sourceM.enabled && targetM.enabled && sourceM.model && sourceM.model.selectedItems().length > 0;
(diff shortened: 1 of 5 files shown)


JorisGoosen:
having ordinal as suggested still doesnt allow me to put scale in it...

boutinb:
> having ordinal as suggested still doesnt allow me to put scale in it...

Yes that is true. The rules with suggest are:

- if suggested contains the scale type, then nomincal & ordinal types are then also allowed.
- if suggested contains the nomincal type, then nominalText & ordinal types are also allowed.

JorisGoosen:
- If suggest is ordinal I cannot put nominal...
- If suggest is scale I *can* put nominal, and also ordinal?

So the implicit rule is that suggest ordinal is the same as allow ordinal?

JorisGoosen (approved):

JorisGoosen:
If that is so feel free to merge

PR #5316, opened 2023-11-13
Misc: Build check workflow update

shun2wang:
We now trigger it while C++ and build system changes.

enabled cache for qt installer, first run will take more times but will speed up in the next running.


JorisGoosen (approved):

JorisGoosen:
❤️

PR #5322, opened 2023-11-15
JASP on the small screen

JorisGoosen:
- FileMenu should be scrollable
- OSF Login now has a minified version for tiny windows
- MainWindow now can be minimum 400 * scaling insteead of 640 * scaling
- VariablesWindow column info now moves into a tab if it gets too small (Allows computed columns drag&drop to actually be usable)

OSF Login:
Big:
afbeelding

Small:
afbeelding

Variables:
Small:
afbeelding

Small computed column Drag&drop:
afbeelding

big:
afbeelding

### DIFF
--- a/Desktop/components/JASP/Widgets/ColumnBasicInfo.qml
+++ b/Desktop/components/JASP/Widgets/ColumnBasicInfo.qml
@@ -1,0 +1,197 @@
+//
+// Copyright (C) 2013-2023 University of Amsterdam
+//
+// This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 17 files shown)


JorisGoosen:
Ah I should also make the module menu scrollable

JorisGoosen:
afbeelding

And as you can see with a different language the sizes are now different too:
afbeelding

Preferences is now mostly usable on tiny screens (although for japanese it might get a bit cramped due to the font+characterset)
afbeelding

boutinb (approved):

PR #5340, opened 2023-12-04
Wrong variable setting after Column Insert before

boutinb:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/2512


shun2wang (approved):
Tested and looks good to me.

PR #5345, opened 2023-12-08
Table title gets APA styles

### LINKED ISSUE #25: Make tables 100% APA format
From the twitterverse:
"Sorry, but that's not fully APA-formatted. Column headings only capitalize first letter, & title of table is italicized." 
@raviselker Let's discuss whether we want to do this, because it might conflict with our other formatting conventions

shun2wang:
Fixes: https://github.com/jasp-stats/jasp-issues/issues/25
Fixes: https://github.com/jasp-stats/INTERNAL-jasp/issues/2123

**Not For 0.18.2**

Not a really 100% strict mode, for example this will also capitalize short conjunctions (e.g., “and,” “as,” “but,” “for,” “if,” “nor,” “or,” “so,” “yet”) where according to APA rules these shouldn't. But I don't think we need to care about this because one can always edit this minor point in their paper?

### DIFF
--- a/Desktop/html/js/jaspwidgets.js
+++ b/Desktop/html/js/jaspwidgets.js
@@ -94,10 +94,6 @@ JASPWidgets.Exporter = {
 		this.htmlImageFormat = JASPWidgets.ExportProperties.htmlImageFormat.temporary,
 		this.includeNotes = false;
 
-		this.isFormatted = function () {
-			return (this.format & JASPWidgets.ExportProperties.format.formatted) === JASPWidgets.ExportProperties.format.formatted
-		};
-
 		this.htmlOnly = function () {
 			return (this.format & JASPWidgets.ExportProperties.format.html) === JASPWidgets.ExportProperties.format.html
 		};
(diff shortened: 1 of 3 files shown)


Desktop/html/js/jaspwidgets.js
```diff
@@ -358,10 +358,10 @@ JASPWidgets.Exporter = {
 	},
 
 	getHeaderStyles: function (element, exportParams) {
-		if (exportParams.isFormatted())
-			return JASPWidgets.Exporter.getStyles(element, ["padding", "text-align", "margin", "display", "float", "vertical-align", "font-size", "font", "font-weight", "color"]);
-		else
-			return "";//JASPWidgets.Exporter.getStyles(element, ["display", "float"]);
+		//if (exportParams.isFormatted())
```
boutinb: Could you add some comments here, because I don't understand why the isFormatted is commented out.

EJWagenmakers:
Also: I agree it is only a small effort to change the capitalization; however, is is difficult to stipulate that words shorter than 4 characters should not start with a capital (unless they are the first word)?

Desktop/html/js/jaspwidgets.js
shun2wang: In fact, I don't fully understand why we keep `isFormatted` here, this will make exported html and copy-paste style inconsistent， I do this so that the style of the assignment and pasting into the document is consistent with the style we export.

shun2wang:
>Also: I agree it is only a small effort to change the capitalization; however, is is difficult to stipulate that words shorter than 4 characters should not start with a capital (unless they are the first word)?

The main intention of this PR is to have table titles italicized, and we actually already capitalize them in R analysis:

will looks like:
image

Desktop/html/js/jaspwidgets.js
shun2wang: Do you know why we use `isFormatted` here? @JorisGoosen

JorisGoosen: I have absolutely no idea what that even means to be honest...

JorisGoosen: But still, I agree a comment there saying something like "I commented this out because it breaks X" is probably good.
I see you also commented it out here:
```
	getNoteStyles: function (element, exportParams) {
		// if (exportParams.isFormatted())
			return JASPWidgets.Exporter.getStyles(element, ["margin", "padding", "max-width", "min-width", "display", "font-size", "font-weight", "font", "color", "border-top-style", "border-top-width", "border-top-color", "border-bottom-style", "border-bottom-width", "border-bottom-color"]);
		// else
		// 	return JASPWidgets.Exporter.getStyles(element, ["max-width", "min-width"]);
	},
```

But the last changes before that were all from 8 years ago... 
Im not even sure where and how I would see the difference between using `isFormatted` or not

shun2wang: The difference was  exported html files (by save results)and pasted html into a document(like LibreOffice or just paste as raw HTML) have inconsistent styles if use isFormatted.

I will have more test in different situations in detail.

JorisGoosen (approved):
I dont think there is any codepath where all that `isFormatted()` stuff ever gets turned on so lets rip it out.

PR #5347, opened 2023-12-12
EmptyValues updates should work (also for files from 0.18.1)

JorisGoosen:
Make sure changes are updated in database
Also prepare the upgradeDB function in databaseInterface for future changes

I stopped working on the refactor of EmptyValues, thatll be for post-0.18.2
This should at least make empty values per column etc work for files from 0.18.1 (and perhaps more)

### DIFF
--- a/Common/version.cpp
+++ b/Common/version.cpp
@@ -25,7 +25,17 @@ const char * Version::encodingError::what() const noexcept
 }
 
 
-Version::Version(std::string version)
+Version::Version(const char * version)
+{
+	fromString(version);
+}
+
+Version::Version(const std::string & version)
+{
+	fromString(version);
+}
+
+void Version::fromString(const std::string & version)
 {
 	const static std::regex parseIt("(\\d+)(\\.(\\d+))?(\\.(\\d+))?(\\.(\\d+))?"); //(sub)groups: 1 3 5 7//0 is whole match/line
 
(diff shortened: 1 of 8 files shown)


boutinb:
Could you provide a reproduction path of what this PR solves? 
There is indeed an archive version issue you solved. And you just removed a not used column 'emptyValuesJson' in the table column. But I don't see any problem you solved in the empty values handling.

JorisGoosen:
> Could you provide a reproduction path of what this PR solves? There is indeed an archive version issue you solved. And you just removed a not used column 'emptyValuesJson' in the table column. But I don't see any problem you solved in the empty values handling.

- open 0.18.1, open file from data library, save it somewhere
- build `development`
- load aforementioned file
- add missing values to a column (for instance '1' to contBinom)
- save file
- close file and load just saved file
- look at empty values in contBinom or whatever
- be amazed at the lack of them

For a bonus, I think it crashes if you try to then set empty values on the column but I might've misremembered.

JorisGoosen:
It also makes sure the database is upgraded at the right moment

boutinb:
There was still a problem: the workspace empty values are empty when a 0.18.1 JASP file is loaded. I have just added a commit to solve this.

JorisGoosen:
So approved and it can be merged?

boutinb (approved):

PR #5350, opened 2023-12-12
Fix syntax check on columns flux

### LINKED ISSUE #2462: [Bug]: Editing (inserting) columns confuses SEM module
### JASP Version

0.18.1

### Commit ID

_No response_

### JASP Module

SEM

Structural Equation Modeling

Windows 10

### Bug Description

See GIF. An SEM analysis seems to keep its own record of columns when the analysis is created, without updating itself when you edit the data (using the new Edit data panel). For example here I run a simple regression y~x, then I go in Edit data and insert a new column between x and y, and the analysis no longer matches the syntax (x has been changed for the new column that's now in its column index in the dataframe)

bug_SEM_insert_column2

### Expected Behaviour

The analysis should match the model syntax at all times, even when you insert or remove columns that are not included in the syntax. Here my model syntax was y\~x, but after inserting a column between y and x through Edit data, the analysis became y\~new_column (because the new column was inserted at the same place that x used to be).

### Steps to Reproduce

1. go to SEM > Structural Equation Modeling
2. run some model syntax, e.g. y~x
3. go in Edit data
4. insert a new column between some variables that are included in the model syntax

The new analysis will be using wrong variables even though the model syntax didn't change.

### Log (if any)

_No response_

### Final Checklist

JorisGoosen:
Fixes https://github.com/jasp-stats/jasp-issues/issues/2462

### DIFF
--- a/QMLComponents/controls/rowcontrols.cpp
+++ b/QMLComponents/controls/rowcontrols.cpp
@@ -48,6 +48,7 @@ void RowControls::init(int row, const Term& key, bool isNew)
 	context->setContextProperty("rowValue", key.asQString());
 
 	_rowObject = qobject_cast<QQuickItem*>(_rowComponent->create(context));
+	_rowObject->setParent(_parentModel);
 	_context = context;
 
 	if (_rowObject)	_setupControls();
--- a/QMLComponents/controls/textareabase.cpp
+++ b/QMLComponents/controls/textareabase.cpp
@@ -70,7 +70,12 @@ void TextAreaBase::setUp()
 			_separators.push_back(separator.toString());
 	}
 
-	connect(this, SIGNAL(applyRequest()), this, SLOT(checkSyntaxHandler()));
+	//If "rowCount" changes on VariableInfo it means a column has been added or removed, this means the model should be reencoded and checked
+	//Fixes https://github.com/jasp-stats/jasp-issues/issues/2462
+	connect(VariableInfo::info(),	&VariableInfo::rowCountChanged,		this,		&TextAreaBase::checkSyntaxHandler);
+
+	//Also do it on request of course ;)
+	connect(this,					&TextAreaBase::applyRequest,		this,		&TextAreaBase::checkSyntaxHandler);
 }
 
 void TextAreaBase::rScriptDoneHandler(const QString & result)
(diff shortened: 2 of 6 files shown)


QMLComponents/controls/textareabase.cpp
```diff
@@ -70,7 +70,12 @@ void TextAreaBase::setUp()
 			_separators.push_back(separator.toString());
 	}
 
-	connect(this, SIGNAL(applyRequest()), this, SLOT(checkSyntaxHandler()));
+	//If "rowCount" changes on VariableInfo it means a column has been added or removed, this means the model should be reencoded and checked
+	//Fixes https://github.com/jasp-stats/jasp-issues/issues/2462
+	connect(VariableInfo::info(),	&VariableInfo::rowCountChanged,		this,		&TextAreaBase::checkSyntaxHandler);
```
boutinb, line 75: Not really beautiful, but I don't see how to do that in another way.

QMLComponents/controls/rowcontrols.cpp
```diff
@@ -48,6 +48,7 @@ void RowControls::init(int row, const Term& key, bool isNew)
 	context->setContextProperty("rowValue", key.asQString());
 
 	_rowObject = qobject_cast<QQuickItem*>(_rowComponent->create(context));
+	_rowObject->setParent(_parentModel);
```
boutinb, line 51: This is needed:  the rowObject was not deleted.... and so all rowComponents.
This can make JASP crashing because these components can be connected to VariableInfo signals even if the analysis does not exist.

boutinb (commented):
Not really beautiful...
Could the column encoding not be a function of the column object directly? Then no need to cache the column names in the columnEncoder (and then no need to add all this code to catch up any change in the dataset)?

QMLComponents/controls/textareabase.cpp
JorisGoosen: well we couldve added more signals and all that but this should work fine

JorisGoosen:
> Not really beautiful... Could the column encoding not be a function of the column object directly? Then no need to cache the column names in the columnEncoder (and then no need to add all this code to catch up any change in the dataset)?

`Column` as parent doesnt make so much sense.
However we could make `ColumnEncoder` a child of `DataSet`, and have its values set via that.
Would make things cleaner indeed and work fine both in desktop and engine.
For R syntax mode it should still be separable anyway I think?

But I suggest we do that later, I could do it together with the refactor of empty values that im planning for after the release.

QMLComponents/controls/rowcontrols.cpp
JorisGoosen: OK

boutinb (approved):

PR #5357, opened 2023-12-18
Fix missing helpfiles on linux

### LINKED ISSUE #2374: [Bug]: Documentation on computing columns is gone
### JASP Version

0.18.1

### Commit ID

_No response_

### JASP Module

Unrelated

_No response_

Flatpak

### Bug Description

grafik

grafik

Hitting the blue help-buttons opened documentation on computed columns in former JASP versions. 

### Expected Behaviour

HItting the help buttons should show the documentation.

### Steps to Reproduce

1. create a computed column
2. click one of the help buttons

I also cleared all caches and config files to make sure my switching between beta and stable did not mess something up

### Log (if any)

_No response_

### Final Checklist

RensDofferhoff:
Fixes: https://github.com/jasp-stats/jasp-issues/issues/2374

Linux/unix is case sensitive so the incorrect casing of  filenames only worked on windows and mac and we never noticed.

### DIFF
--- a/Desktop/components/JASP/Widgets/ComputeColumnWindow.qml
+++ b/Desktop/components/JASP/Widgets/ComputeColumnWindow.qml
@@ -295,7 +295,7 @@ FocusScope
 				iconSource:		jaspTheme.iconPath + "info-button.png"
 				anchors.right:	parent.right
 				anchors.bottom: parent.bottom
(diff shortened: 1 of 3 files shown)


shun2wang (approved):
Tested on Ubuntu 22.04 LTS

JorisGoosen (commented):
I guess that must have been broken for a while then :p

PR #5358, opened 2023-12-18
Add explicit dependency for controls.

boutinb:
In the Process module a DropDown is created in a ComponentsList (of a TabView), and its option values depend on a VariablesList outside of the ComponentsList.
When loading a JASP file, JASP sets the values of each control. For this it first finds out which control depends on other controls, so that a control depending on another control will be set afterwards. The dependencies are mainly found via the `source` property of a control.
Problem is when controls are created dynamically (with a rowComponent): the controls are created only when the parent control is created, so the dependency of these controls cannot be determine before the parent control gets its value.
In the Process module, as the values of the DropDown depends on a VariablesList, when setting its current value (set in the JASP file), if the VariablesList did not get already its values, JASP will throw an error, since the current value of the DropDown does not appear to be one of its option values.
To handle this problem, I've just added an extra property `depends` with which you can set explicitly that the parent control depends on other control: in this way the values of the ComponentsList will be set (and the DropDown wil be created) only after the VariablesList will get its values. This is not a nice solution, since the QML developer must explicitly set this `depends` property, but most of them will not pay attention on this problem.
We should handle this problem in a more clever way later: the form should set the values to each control without caring of dependencies. The control self should be aware of its dependencies, and should wait for its dependent controls to be set, before setting self its value.

### DIFF
--- a/QMLComponents/analysisform.cpp
+++ b/QMLComponents/analysisform.cpp
@@ -215,6 +215,7 @@ void AnalysisForm::sortControls(QList<JASPControl*>& controls)
 {
 	for (JASPControl* control : controls)
 	{
+		control->addExplicitDependency();
 		std::vector<JASPControl*> depends(control->depends().begin(), control->depends().end());
 
 		// By adding at the end of the vector new dependencies, this makes sure that these dependencies of these new dependencies are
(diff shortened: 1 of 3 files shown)


JorisGoosen:
I guess you didnt make a testmodule case?

boutinb:
> I guess you didnt make a testmodule case?

Just made: in "Test Sources with Special attributes" analysis, last section.

JorisGoosen (approved):

PR #5363, opened 2023-12-27
Fix cells border showing in tableView

shun2wang:
Fixes: https://github.com/jasp-stats/INTERNAL-jasp/issues/2231
Fixes: https://github.com/jasp-stats/INTERNAL-jasp/issues/1759
Fixes: https://github.com/jasp-stats/INTERNAL-jasp/issues/1789

The reason is input field background component larger than the parent component so it will block the border of cells.


JorisGoosen (approved):
very nice, that was annoying me for a long time already

PR #5366, opened 2024-01-03
ENH: add cell cursor shape for jasp data view

shun2wang:
Now we can showing a cell cursor shape on data editor just like most spreadsheet editors. and I set to draft to ready for redraw mouse style if you want.

Not sure this will hacky or we can set on C++ side in `datasetview.cpp` like this ? because `qml Image` seems have performance problems on Windows...

```cpp
        QString imagepath = ":/icons/lightTheme/cell-cursor.png";
        QPixmap pix(imagepath);
        pix = pix.scaled(QSize(16,16), Qt::KeepAspectRatio);
        setCursor(QCursor(pix, -1, -1));
```
 but when I tried it will blocked by some GUI elements.


JorisGoosen:
Using an image as a mousecursor seems a bit excessive?

Isnt there a nice default cursor included with Qt https://doc.qt.io/qt-6/qt.html#CursorShape-enum ?

Edit: To add a custom cursor we could even use: `Qt::BitmapCursor`
But drawing an image manually instead of using builtin cursor seems a bit wrong somehow

shun2wang:
Yes, you are absolutely right, I will have another try.

PR #5369, opened 2024-01-04
chore: Supperss some qml warning in output window

shun2wang:
Supperss annoying wheel and qml->js signal warning in Qt output window, better developer's life:-)

### DIFF
--- a/Desktop/components/JASP/Widgets/MainPage.qml
+++ b/Desktop/components/JASP/Widgets/MainPage.qml
@@ -318,7 +318,8 @@ Item
 			{
 				resultsJsInterface.resultsLoaded = loadRequest.status === WebEngineView.LoadSucceededStatus;
 				setTranslatedResultsString();
-				runJavaScript(`window.sendUrlWhitelist(${JSON.stringify(urlWhitelist)})`); //sent urlWhitelist to js side
+				if(resultsJsInterface.resultsLoaded)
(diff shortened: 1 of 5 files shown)


Desktop/components/JASP/Widgets/MainPage.qml
```diff
@@ -318,7 +318,8 @@ Item
 			{
 				resultsJsInterface.resultsLoaded = loadRequest.status === WebEngineView.LoadSucceededStatus;
 				setTranslatedResultsString();
-				runJavaScript(`window.sendUrlWhitelist(${JSON.stringify(urlWhitelist)})`); //sent urlWhitelist to js side
+				if(resultsJsInterface.resultsLoaded)
```
boutinb, line 321: I understand you add this condition, but did you get a situation when this condition was false?

shun2wang: I never, the problem is `window.sendUrlWhitelist` from qml side always runs before defined in js side right, so qml will warning `window.sendUrlWhitelist is not a function`, am I wrong?

if I understand right if `WebEngineView` not loaded then browser will never be opened so I suppose `resultsLoaded` will also not loaded.

boutinb: I was afraid that window.sendUrlWhitelist was then never called. But apparently the onLoadingChanged is called 2 times: once when the resultsJsInterface.resultsLoaded is false, and once when it is true. So your code is perfect!

boutinb (approved):

PR #5371, opened 2024-01-04
Initialize controls only if their dependencies are already initialized

boutinb:
When setting values of the controls (when reading a JASP file or when duplicating an analysis), the order of setting the values is important, since some controls are dependent on other controls (via their source property or an assigned variables list is also dependent on its available variables list). 
This order can be set usually when the controls are statically build, but when they are dynamically build, the dependency is known only when they are build. Typically: when a ComponentList (or a VariableList with rowComponent) gets its value, the rows with their controls are then build, but these controls might depend on a source control which is not yet initialized. These row controls must then wait that the source control is first initialized before being initialized with a value: for example if a row control is a dropdown, with option values depending on another control, the dropdown cannot get its own value before knowing its option values, if not it will throw an error "this value is not possible".

This fixes this issue by telling the control to wait that all its dependencies must be initialized before getting its value.

Also some function and property renaming has been made, so that the code become (a bit) clearer.

Also, I took advantage of this PR to remove quite old ugly code... the Initialization of the available variablesList was quite ugly: a long comment in the bindTo method of the AnalysisForm object tried to explain how this initialization was done, but with this PR, I try to do it in the 'right way'.
The available VariablesLists have no bound value (they don't give values for the analysis), but they must be initialized. For this, as the setInitialized is called even for these controls, just override this method in VaraibelListBase, so that:
. when it has an avalable list model, then it should get its values from its sources.
. when it has an assigned list model, handle the particular case when the addAvailableVariablesToAssigned prperty is true, and it is initialized with no value, then it should take the values of its available list model (which might have default values).

### DIFF
--- a/QMLComponents/analysisform.cpp
+++ b/QMLComponents/analysisform.cpp
@@ -355,33 +355,11 @@ void AnalysisForm::_addLoadingError(QStringList wrongJson)
 
 void AnalysisForm::bindTo(const Json::Value & defaultOptions)
 {
-	QVector<ListModelAvailableInterface*> availableModelsToBeReset;
-
 	std::set<std::string> controlsJsonWrong;
 	
 	for (JASPControl* control : _dependsOrderedCtrls)
 	{
 		BoundControl* boundControl = control->boundControl();
-		JASPListControl* listControl = dynamic_cast<JASPListControl *>(control);
-
-		if (listControl && listControl->hasSource())
-		{
-			ListModelAvailableInterface* availableModel = qobject_cast<ListModelAvailableInterface*>(listControl->model());
-			// The availableList control are not bound with options, but they have to be updated from their sources when the form is initialized.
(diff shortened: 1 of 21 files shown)


QMLComponents/analysisform.cpp
```diff
@@ -357,60 +355,29 @@ void AnalysisForm::_addLoadingError(QStringList wrongJson)
...
-			if (availableModel)
-			{
-				if (defaultOptions.size() != 0 || _analysis->isDuplicate())
-					availableModel->resetTermsFromSources(false);
-				else
-					availableModelsToBeReset.push_back(availableModel);
-			}
-		}
-
+		Json::Value optionValue = Json::nullValue;
```
JorisGoosen, line 363: Haha very nice ^^

JorisGoosen:
Im using 
image
 to test and mentioning that here for posterity

JorisGoosen (approved):

PR #5377, opened 2024-01-08
Unknown computed column type 

### LINKED ISSUE #2517: [Bug]: .jasp file does not open / load after downloading 0.18.2
### JASP Version

0.18.2

### Commit ID

_No response_

### JASP Module

Unrelated

_No response_

Windows 11

### Bug Description

I cannot open my .jasp file after downloading the newest version of the tool.  When trying to open it in jasp 0.18.2.0, the loading page pops up and closes immediately. 

Could you please help me to understand how this can be fixed? I have attached the respective .jasp file for you as a zip file.
Thanks for your help!
[data_environmentalbehaviour__v8.zip](https://github.com/jasp-stats/jasp-issues/files/13854413/data_environmentalbehaviour__v8.zip)

### Expected Behaviour

file should have opened

### Steps to Reproduce

1.
2.
3.
...

### Log (if any)

_No response_

### Final Checklist

boutinb:
The computedColumnType 'unknown' does not exist anymore, but might be stored in the database. So we need to take care of this.
Fixes https://github.com/jasp-stats/jasp-issues/issues/2517

### DIFF
--- a/CommonData/column.cpp
+++ b/CommonData/column.cpp
@@ -65,12 +65,13 @@ void Column::loadComputedColumnJsonBackwardsCompatibly(const Json::Value & json)
 	//Columnname is used to find this:
 	//_column = DataSetPackage::pkg()->dataSet()->column(json["name"].asString());
 	
-	const std::string & rCode = json["rCode"].asString();
+	const std::string & rCode = json["rCode"].asString(),
+					  & codeTypeStr = json["codeType"].asString();
 	
 	setCompColStuff
 	(
 		json["invalidated"].asBool(),
-		computedColumnTypeFromString(json["codeType"].asString()),
+		(codeTypeStr.empty() || codeTypeStr == "unknown") ? computedColumnType::notComputed : computedColumnTypeFromString(codeTypeStr),
 		rCode,
 		json["error"].asString(),
 		json["constructorCode"]
--- a/CommonData/column.cpp
+++ b/CommonData/column.cpp
@@ -67,11 +67,20 @@ void Column::loadComputedColumnJsonBackwardsCompatibly(const Json::Value & json)
 	
 	const std::string & rCode = json["rCode"].asString(),
 					  & codeTypeStr = json["codeType"].asString();
-	
+
+	computedColumnType codeType = computedColumnType::notComputed;
+	if (!codeTypeStr.empty())
+	{
+		try { codeType = computedColumnTypeFromString(codeTypeStr); }
+		catch(...) {}
+	}
+
+	try { codeType = computedColumnTypeFromString(codeTypeStr); } catch(...) {}
+
 	setCompColStuff
 	(
 		json["invalidated"].asBool(),
-		(codeTypeStr.empty() || codeTypeStr == "unknown") ? computedColumnType::notComputed : computedColumnTypeFromString(codeTypeStr),
+		codeType,
 		rCode,
 		json["error"].asString(),
 		json["constructorCode"]
--- a/CommonData/column.cpp
+++ b/CommonData/column.cpp
@@ -27,8 +27,8 @@ void Column::dbLoad(int id, bool getValues)
 
 	db().transactionReadBegin();
 	
-					db().columnGetBasicInfo(	_id, _name, _title, _description, _type, _revision);
-	_isComputed =   db().columnGetComputedInfo(	_id, _analysisId, _invalidated, _codeType, _rCode, _error, _constructorJson);
+	db().columnGetBasicInfo(	_id, _name, _title, _description, _type, _revision);
+	db().columnGetComputedInfo(	_id, _analysisId, _invalidated, _codeType, _rCode, _error, _constructorJson);
 
 	db().labelsLoad(this);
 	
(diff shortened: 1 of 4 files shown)


CommonData/databaseinterface.cpp
```diff
@@ -1092,7 +1092,12 @@ bool DatabaseInterface::columnGetComputedInfo(int columnId, int &analysisId, boo
 		std::string constructorJsonStr	= _wrap_sqlite3_column_text(stmt,	5);
 					analysisId			= sqlite3_column_int(		stmt,	6);
 
-		codeType = codeTypeStr.empty() ? computedColumnType::notComputed : computedColumnTypeFromString(codeTypeStr);
+		codeType = computedColumnType::notComputed;
+		if (!codeTypeStr.empty())
+		{
+			try { codeType = computedColumnTypeFromString(codeTypeStr); }
+			catch(...) {}
+		}
```
JorisGoosen, line 1103:
```suggestion
		try { codeType = computedColumnTypeFromString(codeTypeStr); }
		catch(...) {}
```

There is no need for the if, if it is empty it will also fail and thus get caught

CommonData/column.cpp
```diff
@@ -65,12 +65,22 @@ void Column::loadComputedColumnJsonBackwardsCompatibly(const Json::Value & json)
...
+	}
+
+	try { codeType = computedColumnTypeFromString(codeTypeStr); } catch(...) {}
+
 	setCompColStuff
 	(
 		json["invalidated"].asBool(),
-		computedColumnTypeFromString(json["codeType"].asString()),
+		codeType,
 		rCode,
```
JorisGoosen: You dont need to make changes to column.cpp I think?

Just the change to `DatabaseInterface::columnGetComputedInfo` should be enough

JorisGoosen (changes requested):

CommonData/databaseinterface.cpp
boutinb: Yes but I saw that in the file that was given in the issue, that most of the columns have empty value. If there are hundreds of columns, that would make hundreds of catch exceptions, which can be a bit long.

CommonData/column.cpp
boutinb: I was not quite completely sure that this error could not arise here. For the case, it does not hurt I fimd.

CommonData/databaseinterface.cpp
JorisGoosen: Ok well I dont think a few hundred exceptions will make a difference but whatever.

CommonData/column.cpp
JorisGoosen: Yeah but if the changes are not necessary at all why make them?

JorisGoosen:
Ah while testing I found that loading the jasp file from the issue with the current PR gets:
image

So it is "not computed" yet it is computed according to the columnheader.

So this is not really a fix yet.

JorisGoosen:
the question is more, how can `unknown` even be written there? Or was that simply the default when we didnt set it before?
How is it possible that a column was stored with `unknown` as computed column type at all?

This has a rootcause and we need to figure it out how it happened.

Also, to expand this workaround it would perhaps be good to check in `DatabaseInterface::columnGetComputedInfo` whether the `computedColumnType` isnt `notComputed` before returning `true` (because that says: `isComputed`)

boutinb:
> the question is more, how can `unknown` even be written there? Or was that simply the default when we didnt set it before? How is it possible that a column was stored with `unknown` as computed column type at all?
> 
> This has a rootcause and we need to figure it out how it happened.

When I open the file in 0.18.1, then I see also a not computed column with an icon telling this is a computed column. So the situation was buggy in 0.18.1. The rcode of this column is also empty.
I have tried to reproduce this in 0.18.1 by undo/redo, copy/paste, but could not find a way to get such a situation.
But the code should be better in 0.18.2! So we can expect that this is not reproducible anyway.

The isComputed column in the database is not needed anymore. So I would just remove it.

JorisGoosen:
> The isComputed column in the database is not needed anymore. So I would just remove it.

Ok so remove it in `void DatabaseInterface::upgradeDBFromVersion(Version originalVersion)` 
By adding something to it like:
```
	   if(originalVersion < "0.18.2")
			   runStatements("ALTER TABLE DataSets ADD COLUMN description     TEXT;"                 "\n");

	   if(originalVersion < "0.18.3")
			   runStatements("ALTER TABLE Columns DROP COLUMN isComputed;"                 "\n");

```

And then remove it from all the functions that expect it...

You want this to be a hotfix or a potential place to create more bugs though? ;)

boutinb:
Hmm by removing isColumn in the database, we have then a downgrade issue: JASP file made in 0.18.3 won't be readable in 0.18.2 or lower... This is not expected for a minor release. So I will let the isColumn in the database. We can remove it in 0.19

JorisGoosen:
Exactly, however, you still need to fix the output of that function then. The best thing would be to make sure that if the computedColumnType is "notComputed" the `isComputed` thing is set FALSE

boutinb:
I have removed the _isColumn property of the Column class, as it can be deduced by the codeType property
However for downgrade purpose, the isColumn column in the database is kept so that JASP file made in 0.18.3 can be still read in 0.18.2 or 0.18.1.

JorisGoosen (approved):

PR #5378, opened 2024-01-08
some filter improvements

JorisGoosen:
The text will now look like (as suggested in https://github.com/jasp-stats/jasp-issues/issues/2515#issuecomment-1881235945):
```
# Add filters using R syntax here. By default the non-R filter(s) are passed with generatedFilter.
# To include generated filters, append clauses with "&": generatedFilter & ...
# Click the (i) icon in the lower right corner for further help.

generatedFilter
``` 
What do you think @EJWagenmakers @JohnnyDoorn ?

It is now also translatable, which it wasnt before.

This PR also fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2462 in that the statusbar now gives sensible feedback. It also is translatable now.

I was also annoyed that I couldnt run rcode in the filter and computed column windows by pressing ctrl/cmd+return.
So I added that too

And then there was the fact that filtererrors were being ignored by the desktop... This is now also fixed.

For the kicker I also added the `is.na` function to the filter/compcolconstructor function lists as suggested in https://github.com/jasp-stats/jasp-issues/issues/2493

### DIFF
--- a/CommonData/filter.h
+++ b/CommonData/filter.h
@@ -6,9 +6,9 @@
 #include <vector>
 #include "utils.h"
 
-#define DEFAULT_FILTER		"# Add filters using R syntax here, see info button for help.\n\ngeneratedFilter # by default: pass the non-R filter(s)"
 #define DEFAULT_FILTER_JSON	"{\"formulas\":[]}"
 #define DEFAULT_FILTER_GEN	"generatedFilter <- rep(TRUE, rowcount)"
+
 
 class DataSet;
 class DatabaseInterface;
@@ -61,7 +61,7 @@ private:
 	DataSet				*	_data				= nullptr;
 	int						_id					= -1,
 							_filteredRowCount	= 0;
-	std::string				_rFilter			= DEFAULT_FILTER,
+	std::string				_rFilter			= "",
 							_generatedFilter	= DEFAULT_FILTER_GEN,
 							_constructorJson	= DEFAULT_FILTER_JSON,
 							_constructorR		= "",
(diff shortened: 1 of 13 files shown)


boutinb (approved):

EJWagenmakers:
Maybe it is clear in context. For me, right now, I am a little confused about the following:
"By default the non-R filter(s) are passed with generatedFilter." -> OK This is already a little confusing to me. We are in the R filter functionality, right?
"To include generated filters, append clauses with "&": generatedFilter & ..." -> What are "generated filters"?
EJ

JorisGoosen:
@EJWagenmakers here, some context:
image

EJWagenmakers:
Is the drag-and-drop filter also a non-R filter?

JorisGoosen:
> Is the drag-and-drop filter also a non-R filter?

Yes

JorisGoosen:
image

->

image

tomtomme:
Would be cool if it could be explained like this:
generatedFilter = Filter created in JASP by point & click interface = non-R filters

Or just stick to one of the three, if possible. Becaus it is confusing to have 3 names for the same thing here.
Also, the latest screenshot shows the code of the generated filter in light grey on dark grey. This will be missed by most people. More contrast please. Thx!

JorisGoosen:
How about
```
# Above you see the code that JASP generates for both value filtering and the drag&drop filter.
# This default result is stored in 'generatedFilter' and can be replaced or combined with a custom filter.
# To combine you can append clauses using '&': 'generatedFilter & customFilter & perhapsAnotherFilter'
# Click the (i) icon in the lower right corner for further help.

generatedFilter
```

EJWagenmakers:
Yes, this is very clear!

JorisGoosen:
Haha ok nice

JorisGoosen:
Ill see how hard it is to get the comment highlighted as comment

JorisGoosen:
afbeelding

afbeelding

PR #5380, opened 2024-01-09
A computed column is not always computed when opening a JASP file

boutinb:
Make a JASP file with several analyses.
Add a computed column by using another column.
Save it, close it, and open it again: the computed column is not computed due to an engine crash.

This crash is due to the fact that when trying to compute the computed column, the ColumnEncoder in the Engine was not yet initilalized with the names of the columns.


JorisGoosen:
Ok @boutinb ive made the code a bit less duplicated and feisty, it should still work the same way though.
I cant get the problem occuring anymore.

If you agree with my changes we can merge it

JorisGoosen (approved):

PR #5381, opened 2024-01-09
Update version number to 0.18.3.0

RensDofferhoff:


chenrui333:
@RensDofferhoff 👋 why are there two version schemes for project version and release version? Thanks!

RensDofferhoff:
Ah you mean the _2 ? unfortunately there was a small issue we needed to resolve and quickly swapped out the file.
In future we shall try not to break the convention again.
Thanks!

PR #5409, opened 2024-01-31
Enh: Disable Windows GameDVR in msix

shun2wang:
to disable Windows GameDVR in msix.

Im just added the .reg file but not make it work with Msix installer, I leave it to you @RensDofferhoff and you may push some commit in this PR.

Or we can do it by cmd/powershell script like:

```cmd
reg add HKCU\Software\Microsoft\Windows\CurrentVersion\GameDVR /v AppCaptureEnabled /t REG_DWORD /d 0 /f
reg add HKCU\System\GameConfigStore /v GameDVR_Enabled /t REG_DWORD /d 0 /f
reg add HKLM\Software\Policies\Microsoft\Windows\GameDVR /v AllowGameDVR /t REG_DWORD /d 0 /f
```

Which is more easier?

### DIFF
--- a/Tools/windows/msix/windowsGameDVR.reg
+++ b/Tools/windows/msix/windowsGameDVR.reg
@@ -1,0 +1,10 @@
+Windows Registry Editor Version 5.00
+
+[HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\GameDVR]
+"AppCaptureEnabled"=dword:00000000
+
+[HKEY_CURRENT_USER\System\GameConfigStore]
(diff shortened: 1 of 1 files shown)


JorisGoosen:
When there is a working version of this @FBartos can be guinneapig

shun2wang:
According to this [document](https://learn.microsoft.com/en-us/windows/msix/desktop/desktop-to-uwp-behind-the-scenes#registry) create a .dat registry file to make some virtual registry entries will works for MSIX things.

and then close this pr.

PR #5372, opened 2024-01-05
chore: JASP notes refactoring

shun2wang:
Refactor jaspNotes into separate file, this facilitates maintenance and adding new features, and each module is independent. `jaspwidgets.js` file was too long and complex to maintenance. I push it now to make it independent of subsequent feature improvement/fix PRs for easy review.

_No feature and bugfix added!_

PR #5413, opened 2024-02-04
Enh: Latex editor refactor and improve

shun2wang:
Fixes: https://github.com/jasp-stats/INTERNAL-jasp/issues/2396

- Robustify & Clarify & Visibility LaTeX annotations editor. based on: https://github.com/jasp-stats/jasp-desktop/pull/5372
- Better tooltip for jaspNote editor.

### DIFF
--- a/Desktop/components/JASP/Widgets/MainPage.qml
+++ b/Desktop/components/JASP/Widgets/MainPage.qml
@@ -435,7 +435,8 @@ Item
 				"Remove Indent"   : qsTr("Remove Indent"), 	    "Font Size" : qsTr("Font Size"), "Clear Formatting" : qsTr("Clear Formatting"),			"Click here to add text" : qsTr("Click here to add text"),
 				"Copied to clipboard" : qsTr("Copied to clipboard"), "Citations copied to clipboard" : qsTr("Citations copied to clipboard"), 	"LaTeX code copied to clipboard" : qsTr("LaTeX code copied to clipboard"),
 				"Introduction:"		: qsTr("Introduction:"),  "Conclusion:" : qsTr("Conclusion:"), "Image" : qsTr("Image"), "Embed web video" : qsTr("Embed web video"), "Unsupported video services" : qsTr("Unsupported video services"), 
-				"JASP only allows the following videoservices:" : qsTr("JASP only allows the following videoservices:"), "Contact the JASP team to request adding another videoservice to the list." : qsTr("Contact the JASP team to request adding another videoservice to the list.")
+				"JASP only allows the following videoservices:" : qsTr("JASP only allows the following videoservices:"), "Contact the JASP team to request adding another videoservice to the list." : qsTr("Contact the JASP team to request adding another videoservice to the list."),
+				"Enter LaTeX here and click the preview box to apply." : qsTr("Enter LaTeX here and click the preview box to apply."), "Click to edit this formula" : qsTr("Click to edit this formula")
 
 			}
 
--- a/Desktop/components/JASP/Widgets/MainPage.qml
+++ b/Desktop/components/JASP/Widgets/MainPage.qml
@@ -436,7 +436,8 @@ Item
 				"Copied to clipboard" : qsTr("Copied to clipboard"), "Citations copied to clipboard" : qsTr("Citations copied to clipboard"), 	"LaTeX code copied to clipboard" : qsTr("LaTeX code copied to clipboard"),
 				"Introduction:"		: qsTr("Introduction:"),  "Conclusion:" : qsTr("Conclusion:"), "Image" : qsTr("Image"), "Embed web video" : qsTr("Embed web video"), "Unsupported video services" : qsTr("Unsupported video services"), 
 				"JASP only allows the following videoservices:" : qsTr("JASP only allows the following videoservices:"), "Contact the JASP team to request adding another videoservice to the list." : qsTr("Contact the JASP team to request adding another videoservice to the list."),
-				"Enter LaTeX here" : qsTr("Enter LaTeX here"), "Click to apply formula" : qsTr("Click to apply formula"), "Click to edit this formula" : qsTr("Click to edit this formula")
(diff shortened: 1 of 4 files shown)


JorisGoosen:
Alright, this looks very cool:
image

And it is real nice to see the result real-time while editing.
However, it isn't very clear that clicking the formula adds it into the note.
Clicking outside of it also just makes it dissappear, I guess I expected it to automatically add it then.

Right now pressing return/enter also just inserts a newline into the formula. 
Maybe we could do:
- shift-enter inserts newline
- enter applies formula and adds it to the note?

And I think that when the formula editor loses focus it should probably also just add it.

shun2wang:
>Clicking outside of it also just makes it dissappear, I guess I expected it to automatically add it then.

I kept the default behavior the same as the original quill.js formula editor, which "cancel" input when it loses focus(fewer buttons are more concise). However, it seems that the behavior of pressing <kbd>Enter</kbd> to apply should be added back, but in that case, we seem to have to add additional instructions to tell users how to use shortcut keys to newline？

JorisGoosen:
> I kept the default behavior the same as the original quill.js formula editor, which "cancel" input when it loses focus(fewer buttons are more concise). 

I like the lack of buttons, it looks good for sure!

> However, it seems that the behavior of pressing Enter to apply should be added back, but in that case, we seem to have to add additional instructions to tell users how to use shortcut keys to newline？

Well, I think we could also leave the behaviour like this (for Enter) and just apply it on losing focus.
That should be good. Maybe also add a Ctrl/Cmd+Enter for "Apply" ?

shun2wang:
>Maybe also add a Ctrl/Cmd+Enter for "Apply" ?

Great idea, will do!

shun2wang:
I've added the shortcut key bind, is this what you want?
- <kbd>Enter</kbd>/ <kbd>Ctrl</kbd> + <kbd>Enter</kbd>to apply latex, or click the previewer.
- <kbd>Shift</kbd> + <kbd>Enter</kbd> for new line;

Kucharssim:
Sorry that I am checking it only now! It looks great, the live preview and manual resizing works awesome! There are some quicks with editing:

- Sometimes it is tricky to go back to the edit mode; if I click on the rendered formula, it "jumps" away from my cursor but the editor won't open.
- When I go back to edit mode, the LaTeX syntax is lost. For example, create a formula `\frac{\alpha}{\beta}`, accept, and then open the edit mode again, what I see now in the editor now is `αβ`. Accepting that will render `αβ` in the output which is clearly not correct.
- When I create an equation, then press enter a couple of times (to make new paraghraph), then make another equation, sometimes the second equation "overwrites" the first equation, sometimes the second equation is then pasted in front of the first equation. 


> I've added the shortcut key bind, is this what you want?
> Enter/ Ctrl + Enterto apply latex, or click the previewer.
> Shift + Enter for new line;

I keep pressing "Enter" to make a new line and getting frustrated that it applies the changes. It would be awesome to let the user customise these key bindings in global settings...

JorisGoosen:
Ok, my intention was:
- Enter is newline
- Cmd/Ctrl+Enter is apply
- Clicking latex is apply

Ill check it out and maybe tweak it a bit so we can merge it.

JorisGoosen:
image
:p

JorisGoosen:
Ok, maybe its better if @shun2wang fixes it up, sorry it took me a while to look at the PR again.

But I see several issues:
- latex editor is lower z-level than analysis title, and pops up a bit low on the screen
- clicking on a formula is hard: because it also selects the note-editor, which makes the toolbar pop-in and -out of view, which moves the formula from under your mouse...
- Enter -> Add a new line to formula (Dont close the editor)
- Cmd/Ctrl+Enter -> Close the editor
- Clicking the formula -> close the editor
- Closing the editor, or losing focus should apply the formula!
- Escape -> close the editor but *do not apply* the changes

Also it seems to want to open the editor way below the note box, regardless of where my editor-cursor is in the notebox...

shun2wang:
Good catch! I will fix these asap. thank you all for review it.

shun2wang:
>Closing the editor, or losing focus should apply the formula!

I'm not sure about this behavior since we don't have a "cancel/close" button and losing focus to close editing is the default behavior of the whole quill editor.(for example also link editor)

JorisGoosen:
> > Closing the editor, or losing focus should apply the formula!
> 
> I'm not sure about this behavior since we don't have a "cancel/close" button and losing focus to close editing is the default behavior of the whole quill editor.(for example also link editor)

Well, in the link editor though there is this really clear "save" button.

I personally think that its better behaviour to simply apply the formula.
Given also that typing a formula and then accidentally clicking outside it would simply drop it all now...

I do agree that it is a bit confusing that not all behaviour within quill is consistent then. But I think most textfields in JASP simply apply the input without save-buttons or stuff like that.

JorisGoosen:
Very nice!

Works smoothly I must say.

Although there is 1 small quible:
I am on a macos machine, where one expects Cmd+Enter to be the command, but the only one that works is Ctrl+Enter...
I think it should simply apply and close the formula editor on both of them.

shun2wang:
>I am on a macos machine, where one expects Cmd+Enter to be the command, but the only one that works is Ctrl+Enter...

Ah, I'll add that in a minute...

shun2wang:
Sorry, could you test on macOS now(with also ctrl+enter and cmd+enter)? I don't have it.

JorisGoosen (approved):
Much nicer editor like this!

EJWagenmakers:
I am keen to try it out!

shun2wang:
@EJWagenmakers sure, nightly build:

- macOS arm64： https://static.jasp-stats.org/Nightlies/macOS-arm64/JASP-nightly-development-00c1f3555b7eacf4452fdd16a62b324b66f09520.dmg
- macOS x86_64：https://static.jasp-stats.org/Nightlies/macOS-x86_64/JASP-nightly-development-00c1f3555b7eacf4452fdd16a62b324b66f09520.dmg
- Windows zip：https://static.jasp-stats.org/Nightlies/WindowsZip/JASP-nightly-development-00c1f3555b7eacf4452fdd16a62b324b66f09520.zip

EJWagenmakers:
Awesome, this looks superclean!

PR #5424, opened 2024-02-15
Attempt to add missing system deps for Arch Linux

Joao-O-Santos:
This PR aims to start to address jasp-stats/jasp-issues#2586

Feel free to reject it, as I cannot confirm these are all the missing dependencies, and I've only updated it for Arch.

Any feedback is welcomed so I can make a better PR when I have more time.


RensDofferhoff (approved):
Thanks for your efforts!
I will soon add a list for Fedora/RHEL.

PR #5430, opened 2024-02-23
feat: Add debug info to about window

### LINKED ISSUE #766: Allow to copy version info from the About dialog
The About dialog of JASP does not allow to copy any info from there. But to report bugs, in discussions in forums etc. one needs the info about the version of JASP one is currently using.

Therefore it would be helpful to have a buttoon in the About dialog that coies all relevant info to the clipboard so that one can easily paste it elsewhere.

Here is an example how the OpenSource software "FreeCAD" solved this:
FreeCAD_Hkg9Z2vjvj

so I request such a button in JASP's About dialog.

shun2wang:
close: https://github.com/jasp-stats/INTERNAL-jasp/issues/2087
fix: https://github.com/jasp-stats/jasp-issues/issues/766

For reviewer: note I didnt test it on macOS!

Now we have obtained basic and additional useful information through system local commands(Linux/macOS->shell, Windows->powershell) and Qt's basic functions to provide debugging environment information while users feedback.I added it to about window as a button to copy it. used my poor C++ skills...

Application Info: where JASP build information
Basic Info: where OS and CPU information
Extra Info: depends on OS to give software/hardware info.

_btw: maybe we should display the development team below, similar to how most open source software do acknowledgments here.to commemorate whether they are former/current  teamers or translators._

### DIFF
--- a/Desktop/components/JASP/Widgets/AboutWindow.qml
+++ b/Desktop/components/JASP/Widgets/AboutWindow.qml
@@ -17,7 +17,7 @@ WavyWindow
 
 	title:					qsTr("About JASP")
 	width:					600 * jaspTheme.uiScale
-	height:					550 * jaspTheme.uiScale
+	height:					600 * jaspTheme.uiScale
 
 
 	property int labelwidth:	10 + Math.max(	jaspVersionLabel.implicitWidth,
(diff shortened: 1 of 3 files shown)


RensDofferhoff:
This is a good idea I think!
Not sure if the command call will work in containerized MSIX environment. 
But I will take a look.

RensDofferhoff:
Okay I have some builds running so I can check this on all platforms

RensDofferhoff:
Okay so the extra info section doesn't work for Flatpak and Mac dmg (which is fine).
It works for MSIX and MSI and gives the code page. It does give you local network IP but and timezone but that is probably fine.
I think we should merge this @JorisGoosen ?

shun2wang:
Not sure why not work on macOS shell，However, the hardware of macOS is very consistent, and we have obtained the CPU architecture in the basic information. for extra info the flatpak indeed not allow to run CLI but that OK as the container contains all dependencies.

JorisGoosen:
> Okay so the extra info section doesn't work for Flatpak and Mac dmg (which is fine). It works for MSIX and MSI and gives the code page. It does give you local network IP but and timezone but that is probably fine. I think we should merge this @JorisGoosen ?

I want to have a look at this before we merge it but I missed the previous emails.
I can look at it next week!

JorisGoosen:
> btw: maybe we should display the development team below, similar to how most open source software do acknowledgments here.to commemorate whether they are former/current teamers or translators.

Could be nice to have a rolling "thanks to all contributors" or something there.
Or just a link to the teampage ?

JorisGoosen (approved):

PR #5446, opened 2024-03-18
chore: Auto bump R release version in buildbot

shun2wang:
always install new release R version for build test.

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -18,7 +18,7 @@ jobs:
       
       - uses: r-lib/actions/setup-r@v2
         with:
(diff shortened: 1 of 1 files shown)


vandenman (approved):
I don't have any objections to doing this, but just as an aside I prefer to specify the version manually for the unit tests in the modules.

shun2wang:
I've thought about this, but I think catching build issues early on in the build of the main program might help us move to the next version of R easily, this build test will not get into nightly build for user side, since it might be related to underlying architectural changes.

>but just as an aside I prefer to specify the version manually for the unit tests in the modules.

for analysis test I think so:-)

PR #5450, opened 2024-03-20
Upgrade FactorsForm control with nested VariablesList feature

boutinb:
Add properties to the `FactorsForm` control so that it can be used with nested `VariablesList`: the values of the n'th `VariablesList` should be inside the (n+1)th `VariablesList` as undraggable items.

Change the `_copyTermsWhenDropped` flag in `ListModelDraggable` to `_removeTermsWhenMoved`: `_copyTermsWhenDropped` was not used to determine the behaviour of its own `VariablesList`, but the one from whom the terms is taken from. That was quite difficult to understand and lead also to issue for the `FactorForm`.
`_removeTermsWhenMoved` indicates that the `VariablesList` of this model should remove the/its variables when they are moved to another `VariablesList`. 

The source, typically an `AvailableVariablesList`, of an assigned `VariablesList` with interactions or a `VariablesList` with paired variables (eg Paired Samples T-Test) should not lose their variables when they are moved to these assigned variableslists.

A new Test `FactorsForm` analysis has been added in the Test Module

### DIFF
--- a/QMLComponents/components/JASP/Controls/FactorsForm.qml
+++ b/QMLComponents/components/JASP/Controls/FactorsForm.qml
@@ -18,7 +18,8 @@ FactorsFormBase
 	property alias	availableVariablesList: availableVariablesList
     property bool   allowAll: false
 	property int    listWidth:			parent.width * 2 / 5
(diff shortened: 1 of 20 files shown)


JorisGoosen (approved):
Nice improvements!

PR #5458, opened 2024-04-02
Avoid looping in FactorsForm and handle interaction in options

boutinb:
The options sent to the analyses should give the interaction terms in the right form, that as an array (only if allowInteraction property is true, so that CFA that uses also FactorsForm gets still the same options):
```
"nestedFactors" : 
        [
            {
                "components" : [],
                "name" : "model0",
                "title" : "Model 0"
            },
            {
                "components" : 
                [
                    ["facExperim"],
                    ["facFive"],
                    ["facExperim", "facFive"]
                ],
                "name" : "model1",
                "title" : "Model 1"
            }
        ]
```

Also the code ensuring the nestingness could lead to a loop. This is also solved.

### DIFF
--- a/QMLComponents/controls/factorsformbase.cpp
+++ b/QMLComponents/controls/factorsformbase.cpp
@@ -17,8 +17,6 @@
 //
 
 #include "factorsformbase.h"
-#include "boundcontrols/boundcontrolterms.h"
-#include "analysisform.h"
 #include "utilities/qutils.h"
 #include "variableslistbase.h"
 #include "log.h"
@@ -54,11 +52,25 @@ void FactorsFormBase::bindTo(const Json::Value& value)
 
 	for (const Json::Value& factor : value)
 	{
-		vector<string> indicators;
-		for (const Json::Value& indicator : factor["indicators"])
-			indicators.push_back(indicator.asString());
+		Terms initTerms;
+		for (const Json::Value& termsJson : factor[fq(_optionKey)])
+		{
+			if (allowInteraction())
+			{
+				// For interaction, each term is an array of strings
+				std::vector<std::string> term;
+				for (const Json::Value& elt : termsJson)
+					if (elt.isString())
+						term.push_back(elt.asString());
+				initTerms.add(Term(term));
+			}
+			else
+				// If not, each term is just a string
+				initTerms.add(termsJson.asString());
+		}
+
 		
-		factors.push_back(make_tuple(factor["name"].asString(), factor["title"].asString(), indicators));
+		factors.push_back(ListModelFactorsForm::Factor(tq(factor["name"].asString()), tq(factor["title"].asString()), initTerms));
 	}
 	
 	_factorsModel->initFactors(factors);
@@ -73,7 +85,7 @@ Json::Value FactorsFormBase::createJson() const
 		Json::Value row(Json::objectValue);
 		row["name"] = fq(baseName() + QString::number(i + startIndex()));
 		row["title"] = fq(baseTitle() + " " + QString::number(i + startIndex()));
-		row["indicators"] = Json::Value(Json::arrayValue);
+		row[fq(_optionKey)] = Json::Value(Json::arrayValue);
 
 		result.append(row);
 	}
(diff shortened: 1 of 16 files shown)


JorisGoosen:
You are rewriting some things, right?

boutinb:
> You are rewriting some things, right?

Yes, but still some tricky stuff.... nearly done

QMLComponents/controls/factorsformbase.cpp
```diff
@@ -54,11 +52,25 @@ void FactorsFormBase::bindTo(const Json::Value& value)
...
 	{
-		vector<string> indicators;
-		for (const Json::Value& indicator : factor["indicators"])
-			indicators.push_back(indicator.asString());
+		Terms initTerms;
+		for (const Json::Value& termsJson : factor[fq(_optionKey)])
+		{
+			if (allowInteraction())
+			{
+				// For interaction, each term is an array of strings
```
JorisGoosen, line 60: Why dont we have classes with a hierarchy or something for these terms?

QMLComponents/controls/factorsformbase.cpp
```diff
@@ -54,11 +52,25 @@ void FactorsFormBase::bindTo(const Json::Value& value)
...
+				initTerms.add(Term(term));
+			}
+			else
+				// If not, each term is just a string
+				initTerms.add(termsJson.asString());
+		}
+
 		
-		factors.push_back(make_tuple(factor["name"].asString(), factor["title"].asString(), indicators));
+		factors.push_back(ListModelFactorsForm::Factor(tq(factor["name"].asString()), tq(factor["title"].asString()), initTerms));
```
JorisGoosen, line 73: this looks like it already ^^

QMLComponents/controls/variableslistbase.cpp
```diff
@@ -248,73 +248,55 @@ void VariablesListBase::moveItemsDelayedHandler()
...
 void VariablesListBase::moveItems(QList<int> &indexes, ListModelDraggable* targetModel, int dropItemIndex)
 {
-	if (targetModel && indexes.size() > 0)
+	if (!targetModel || !indexes.size()) return;
+
+	if (form()) form()->blockValueChangeSignal(true);
+
+	std::sort(indexes.begin(), indexes.end());
+	ListModelDraggable* sourceModel = _draggableModel;
+	if (sourceModel == targetModel)
```
JorisGoosen, line 257: Id expect it to not make any sense to move something from a model to itself?
Am I missing something?

JorisGoosen: Or is this to move them within the list?

JorisGoosen:
```suggestion
	if (sourceModel == targetModel) //User is reorganising the order of the terms manually
```

QMLComponents/controls/variableslistbase.cpp
```diff
@@ -248,73 +248,55 @@ void VariablesListBase::moveItemsDelayedHandler()
...
-			sourceModel->moveTerms(indexes, dropItemIndex);
-		else
+		if (!targetModel->keepTerms()) // If the target keeps its terms anyway, don't (re-)add new terms (because they probably came from target)
 		{
-			bool refreshSource = false;
-			Terms termsAdded;
-			Terms removedTermsWhenAdding;
-			QList<int> indexAdded = indexes;
+			// Check which terms can be added in the target model (especially terms that have not the right types might be refused).
+			Terms terms = sourceModel->termsFromIndexes(indexes);
```
JorisGoosen: Why dont you define a function analogous to `termsFromIndexes` called `indexesFromTerms`

Also its technically `indices` in english not `indexes`

JorisGoosen: but apparently we call it that all over jasp anyway haha

JorisGoosen (approved):
ive made some changes

JorisGoosen:
I think I didnt break it, but maybe @boutinb can give the new code a go?
If you agree it still works we can merge it

PR #5459, opened 2024-04-03
Determines type for empty columns based on what user types

JorisGoosen:
when a user enters a new value in data-editing mode this code will:
- check whether no data is in the column yet (so no values at all, or only empty/missing values)
- is the first non-missing-value thing a user enters a number? convert/keep scalar type
- is the first non-missing-value thing a user enters a string? convert/keep nominal type
- implements request from https://github.com/jasp-stats/INTERNAL-jasp/issues/2490

### DIFF
--- a/CommonData/column.cpp
+++ b/CommonData/column.cpp
@@ -1238,7 +1238,7 @@ bool Column::setStringValueToRow(size_t row, const std::string & userEntered)
 			oldDouble		= _dbls[row];	
 	bool	itsADouble		= ColumnUtils::getDoubleValue(userEntered, newDoubleToSet),
 			nothingThereYet	=	_ints.size() == std::count_if(_ints.begin(), _ints.end(), [&](int i)	{ return i == Label::DOUBLE_LABEL_VALUE || i == EmptyValues::missingValueInteger || labelByIntsId(i)->isEmptyValue(); }) 
-							&&	_dbls.size() == std::count_if(_dbls.begin(), _dbls.end(), [](double d)	{ return std::isnan(d); });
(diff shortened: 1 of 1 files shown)


koenderks (approved):

CommonData/column.cpp
```diff
@@ -1236,15 +1236,29 @@ bool Column::setStringValueToRow(size_t row, const std::string & userEntered)
 	
 	double	newDoubleToSet	= EmptyValues::missingValueDouble,
 			oldDouble		= _dbls[row];	
-	bool	itsADouble		= ColumnUtils::getDoubleValue(userEntered, newDoubleToSet);
+	bool	itsADouble		= ColumnUtils::getDoubleValue(userEntered, newDoubleToSet),
+			nothingThereYet	=	_ints.size() == std::count_if(_ints.begin(), _ints.end(), [&](int i)	{ return i == Label::DOUBLE_LABEL_VALUE || i == EmptyValues::missingValueInteger || labelByIntsId(i)->isEmptyValue(); })
```
boutinb: Nice but not really efficient: it has to go through the whole column, but in most of the cases already with the first values, it can see that there are already non-empty values here.

JorisGoosen: fair enough, let me see

JorisGoosen:
Ive changed it for you, but I dont think it really matters. This was only ever run for user interaction stuff and looping over a bunch of ints and dbls in a vector is pretty fast actually. But luckily std lib has got me covered for easy efficient rewrites ;)

boutinb (approved):

PR #5460, opened 2024-04-04
Improvements to column header, window and more

JorisGoosen:
I was annoyed by a few things and thought there was room to make things a bit prettier.

In the columnheader:
Ive changed the computed column error icon to a red exclamation mark. It is now also displayed instead of the compute-column-symbol when there is an error. Same for the loading icon.
The column header should now be a bit more compact.

The buttons in the label editor have been moved to the left, as they are then close to the values instead of far away.
Ive also made them scrollable (when it gets too small) and anchored them nicely.

Then there are some changes to the computed column backend where ive made a dedicated code path for it so I could reduce the amount of boilerplate that shows up in the errormessages.


boutinb:
Nice!
I just found an error, but probably has nothing to do with this commit: open the debug.csv, Edit mode, open contBinom variable Window, click op label 1, change order of labels, click on a cell of the contBinom column: all cells become 1 !

JorisGoosen:
> Nice! I just found an error, but probably has nothing to do with this commit: open the debug.csv, Edit mode, open contBinom variable Window, click op label 1, change order of labels, click on a cell of the contBinom column: all cells become 1 !

While it is good that you found a bug it is indeed present in `development` (which I discovered by trying it) and has absolutely nothing to do with this PR...

So maybe just make an issue for that instead of posting about it here?
Ill do it for you now though and then Ill go fix the problem.

PR #5461, opened 2024-04-05
Fix some errors when moving variables in QML

boutinb:
Several fixes:

- In a VariablesList with interactions (eg. in an ANOVA analysis), move in Model Terms an interaction term, a variable may disappear (fix in indexOf method in terms.cpp)
- In Repeated Measures ANOVA, in the Repeated Measures Cell, if you add some variables, and try to remove some of them: the wrong variables are removed (ListModelMeasuresCellsAssigned::indexesFromTerms is added)
- Some overriden termsFromIndexes functions are useless, and are removed.

### DIFF
--- a/QMLComponents/controls/variableslistbase.cpp
+++ b/QMLComponents/controls/variableslistbase.cpp
@@ -262,7 +262,7 @@ void VariablesListBase::moveItems(QList<int> &indexes, ListModelDraggable* targe
 				termsRejected;
 
 		//if a model keeps terms we dont need to bother adding or removing anything
-		if (!targetModel->keepTerms())	termsToAdd				=	targetModel->canAddTerms(	sourceModel->termsFromIndexes(indexes)		);	// Check which terms can be added in the target model (especially terms that have not the right types might be refused).
+		if (!targetModel->keepTerms())	termsToAdd				=	targetModel->canAddTerms(	termsToAdd									);	// Check which terms can be added in the target model (especially terms that have not the right types might be refused).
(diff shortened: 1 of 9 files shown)


QMLComponents/models/listmodelinteractionassigned.cpp
```diff
@@ -94,19 +94,6 @@ void ListModelInteractionAssigned::removeTerms(const QList<int> &indices)
 	setTerms();
 }
 
-Terms ListModelInteractionAssigned::termsFromIndexes(const QList<int> &indexes) const
```
JorisGoosen, line 97: I guess you didnt like my solution?

QMLComponents/controls/variableslistbase.cpp
```diff
@@ -262,7 +262,7 @@ void VariablesListBase::moveItems(QList<int> &indexes, ListModelDraggable* targe
 				termsRejected;
 
 		//if a model keeps terms we dont need to bother adding or removing anything
-		if (!targetModel->keepTerms())	termsToAdd				=	targetModel->canAddTerms(	sourceModel->termsFromIndexes(indexes)		);	// Check which terms can be added in the target model (especially terms that have not the right types might be refused).
+		if (!targetModel->keepTerms())	termsToAdd				=	targetModel->canAddTerms(	termsToAdd									);	// Check which terms can be added in the target model (especially terms that have not the right types might be refused).
```
JorisGoosen, line 265: this is exactly the same as before...

QMLComponents/models/listmodelinteractionassigned.cpp
JorisGoosen: Or did I make the other ones? 
Pfooh

PR #5464, opened 2024-04-09
support reversing values for one or more selected columns

JorisGoosen:
This adds a button to both the columnheader menu in the data viewer and the label-editor.

Implements https://github.com/jasp-stats/jasp-issues/issues/2325 as far as the reversing of values goes.
This is pretty straightforward:
- it collects all values and checks if they (convertible to) `double`
- orders them and makes a mapping from highest to lowest, highest-1 to lowest+1, etc
- writes them back to the labels

Also supports this for multiple columns a la change columntype


JohnnyDoorn:
What's a good way to test this if I don't build JASP? Is it straightforward to turn this into a nightly? Or should I just test it on one of your computers?

JorisGoosen:
> What's a good way to test this if I don't build JASP? Is it straightforward to turn this into a nightly? Or should I just test it on one of your computers?

Ah ok, well I can let a nightly build for you, sure, arm64 macos right?
Im currently merging the undo-commands for columntype and reverse values into one though.

Because otherwise it takes quite a lot of Cmd+Z to get back to the previous state after changing all columns :p

JohnnyDoorn:
Awesome, thanks <3

boutinb (approved):

boutinb:
Just added a commit so that the Undo text gets the right column name

JorisGoosen:
Ill just merge it, @JohnnyDoorn will see it in the development nightly tomorrow.
Im also making an extra change where you can order the labels by value, should be done in a few minutes

PR #5466, opened 2024-04-11
Better selecting!

JorisGoosen:
started work on jasp-stats/INTERNAL-jasp#2495

this PR will:
- allow for selecting disjunct sections of the data
- actions on selected stuff should only do things for those selected
- copy pasting within jasp will only copy those cells selected
- to spreadsheet it will copy nonselected cells as ""
- cmd/shift clicking now should work

I couldnt get the nice shift-select on hover to work properly for now but that is way less important than being able to change columntypes or reversing levels so here we are.
Ive turned on a build of `betterSelecting` for @JohnnyDoorn to review, if the static server with the nightlies does not go down in the meantime

### DIFF
--- a/Common/timers.cpp
+++ b/Common/timers.cpp
@@ -26,8 +26,17 @@ void _printAllTimers()
 {
 	if(timers == nullptr)
 		return;
+	
+	typedef std::pair<std::string, boost::timer::cpu_timer *> nameTimerPair;
+	
+	std::vector<nameTimerPair> sortMe(timers->begin(), timers->end());
+	
+	std::sort(sortMe.begin(), sortMe.end(), [](const nameTimerPair & l, const nameTimerPair & r)
+	{
+		return l.second->elapsed().user > r.second->elapsed().user;
+	});
 
-	for(auto keyval : *timers)
+	for(const nameTimerPair & keyval : sortMe)
 		std::cout << keyval.first << " ran for " << keyval.second->format() << std::endl;
 }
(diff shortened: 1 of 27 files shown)


boutinb:
Several issues found:

- An extra insert negative rows in the Undo stack appears when copy paste some values. I solved it in ExpandDataProxyModel::_expandIfNecessary (https://github.com/jasp-stats/jasp-desktop/pull/5466/commits/f5d507065e21953c21b99848a411038bf7ca7efc).
- The VariableWindow does not show the selected variable in the Data view. It shows the right variable only if the header of the column is clicked, not if a cell is clicked.
Screenshot 2024-04-19 at 15 34 53

- If you set a value on an extra row, this value disappears.
- If you copy-paste some values on a column, and some of these values add some labels, when undoing the values are reset to their original values, but the labels stay even if no value uses them. This is not the case when just setting a new value (by typing directly in the cell), and undoing it.

JorisGoosen:
Also a point I found: doing selext all on certain data (probably just big, as I used the data from https://github.com/jasp-stats/INTERNAL-jasp/issues/2498) is crazy slow.

JorisGoosen:
Alright I guess this is ready for review again, ive turned on a build for macos arm  for Johnny in case he wants to have a look

JorisGoosen:
I will work on a speedup for big copies in a separate branch though

boutinb (approved):

PR #5468, opened 2024-04-15
Fix TabView issues

### LINKED ISSUE #2661: [Bug]: QML TabView previous tab gets cut off for multiple tabs
### JASP Version

0.18.3

### Commit ID

_No response_

### JASP Module

Process

Classical Process

Windows 10

### Bug Description

When specifying a second (or third) model in the dropdown interface in the tabview component and switching back to the tab of the first model the interface is cut off in the bottom.

image
[tabview_bug.zip](https://github.com/jasp-stats/jasp-issues/files/14930439/tabview_bug.zip)

### Expected Behaviour

The full interface should be displayed.

### Steps to Reproduce

1. Specify two models in separate tabs using the dropdown interface.
2. Go back to the first tab.

### Log (if any)

_No response_

### Final Checklist

### LINKED ISSUE #2662: [Bug]: Dropdowns reset in TabView component when new tab is added
### JASP Version

0.18.3

### Commit ID

c8c4acf1dcb12480b97fe7e07583b903366768df

### JASP Module

Process

Classical Process

Windows 10

### Bug Description

When adding a second tab in the tab view component in the `Models` section, all dropdowns in the first tab reset.

First tab with options selected in dropdowns:

image

Second tab with emtpy dropdowns:

image

Now, when switching back to the first tab, all dropdowns are empty too:

image

This also happens when adding a third tab (and so on).

### Expected Behaviour

The dropdowns in the first tab should keep their original values.

### Steps to Reproduce

1. Specify the dropdowns in the first tab.
2. Add a second tab.
3. Go back to the first tab.

### Log (if any)

_No response_

### Final Checklist

boutinb:
Fixes https://github.com/jasp-stats/jasp-issues/issues/2662 
Fixes https://github.com/jasp-stats/jasp-issues/issues/2661

The height of a TabView was not correctly set, especially when the height could vary dynamically.
Also the Tab buttons got a a new look so that selected button becomes more visible.

Remove ShouldShowFocus property in jaspControl (it has no use), and rename some functions.

### DIFF
--- a/QMLComponents/components/JASP/Controls/TabView.qml
+++ b/QMLComponents/components/JASP/Controls/TabView.qml
@@ -25,7 +25,6 @@ ComponentsListBase
 {
 	id						: tabView
 	background				: rectangleItem
-	focusIndicator			: null
 	implicitWidth 			: parent.width
 	implicitHeight			: itemStack.y + itemStack.height
 	shouldStealHover		: false
@@ -34,8 +33,7 @@ ComponentsListBase
 	minimumItems			: 1
 	newItemName				: qsTr("New tab")
 	controlType				: JASPControl.TabView
-
-	onFocusIndicatorChanged	: focusIndicator = null
+	focusOnTab				: false
 
 	property alias	label				: tabView.title
 	property bool	showAddIcon			: addItemManually
--- a/QMLComponents/components/JASP/Controls/TabView.qml
+++ b/QMLComponents/components/JASP/Controls/TabView.qml
@@ -1,5 +1,5 @@
 //
-// Copyright (C) 2013-2018 University of Amsterdam
+// Copyright (C) 2013-2024 University of Amsterdam
 //
 // This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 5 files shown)


JorisGoosen:
Ok, the issue https://github.com/jasp-stats/jasp-issues/issues/2662 is solved now.


However, I still get the wrong height for switching models ( https://github.com/jasp-stats/jasp-issues/issues/2661 ):
afbeelding

afbeelding

After adding model 3, 1 and 2 both have the wrong height.  In fact it seems that only the last model has the correct height.
Also after, for instance, removing the last one the then last one gets a correct height.

JorisGoosen (changes requested):
Uh yeah, so I request that the height gets set properly ;)

boutinb:
I see with the pictures you show, that you did not use the right branch to test: the look & feel of the tab buttons is not the same:
Screenshot 2024-04-18 at 18 12 50

JorisGoosen (approved):

PR #5478, opened 2024-04-23
labels should be written batched during load

JorisGoosen:
should speed up loading certain files
Might fix or improve https://github.com/jasp-stats/INTERNAL-jasp/issues/2498


shun2wang (approved):
6-10 times speed improvement!

PR #5483, opened 2024-04-29
Generate properly the markdown help

boutinb:
The JASPControl::helpMD function has been simplified, so that all specific display needed for a control, has been moved to the overriden control function. 

The info property is changed into a QVariant, so that more information can be set. 
For controls: if it is a string, this is, as before, just the text of the info. But it can be used as an object like this: 
`{ text: "....", label: "....", useControlType: true, isHeader: true}` 
In this way the developer can specify whether the label can have another value than the title, whether the type of the control should be displayed, and whether the info should be displayed as an header. Other kind of information can be later easily added.

For Form, the info can be also just a string but also an object: 
`{ top: "..",, bottom: "...", references: ["ref1, "ref2"], RPackages: ["pack1", "pack2"], examples: ["example1", "example2"] }`

### DIFF
--- a/QMLComponents/analysisform.cpp
+++ b/QMLComponents/analysisform.cpp
@@ -901,13 +901,13 @@ QString AnalysisForm::helpMD() const
 	markdown << metaHelpMD();
 	
 	if(!_info.bottom.isEmpty())
-		markdown << _info.bottom  << "\n";
+		markdown << "\n\n---\n" << _info.bottom  << "\n";
 
 	auto printList = [&markdown](const QStringList& list, const QString& title) -> void
 	{
 		if(list.length() > 0)
 		{
-			markdown << ("\n\n---\n# " + title + "\n");
+			markdown << ("\n\n---\n### " + title + "\n");
 			for (const QString& elt : list)
 				markdown << "- " << elt << "\n";
 		}
--- a/QMLComponents/analysisform.cpp
+++ b/QMLComponents/analysisform.cpp
@@ -884,7 +884,7 @@ std::set<string> AnalysisForm::usedVariables()
 ///Generates documentation based on the "info" entered on each component
 QString AnalysisForm::helpMD() const
 {
-	if(!_analysis) return "";
+	if(!_analysis || !initialized()) return "";
 
 	QStringList markdown =
 	{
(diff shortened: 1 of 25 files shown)


QMLComponents/analysisform.cpp
```diff
@@ -308,12 +307,31 @@ QString AnalysisForm::msgsListToString(const QStringList & list) const
...
 		return;
 
-	_info = info;
+	_info.var = info;
+
+	if (info.canConvert<QString>())
+		_info.top = info.toString();
+	else if (info.canConvert<QMap<QString, QVariant> >())
+	{
+		QMap<QString, QVariant> map = info.toMap();
```
JorisGoosen: Maybe you should make a proper type for this instead of a variantmap...

boutinb: Hmm I have tried now for some hours, but I cannot make it.
If I have a look at the Qt doc, I should use Q_GADGET and QML_VALUE_TYPE
https://doc.qt.io/qt-6/qtqml-typesystem-valuetypes.html
https://doc.qt.io/qt-6/qtqml-cppintegration-definetypes.html
But it gives errors either in compilation at run-time not recognizing the type... There is also no real example how to use QML_VALUE_TYPE.

JorisGoosen: https://doc.qt.io/qt-6/qtqml-cppintegration-definetypes.html#registering-an-instantiable-object-type

This no?

JorisGoosen: I can have a look tomorrow

QMLComponents/analysisform.cpp
```diff
@@ -308,12 +307,31 @@ QString AnalysisForm::msgsListToString(const QStringList & list) const
...
+		if (map.contains("top"))
+			_info.top = map["top"].toString();
+		if (map.contains("bottom"))
+			_info.bottom = map["bottom"].toString();
+		if (map.contains("references"))
```
JorisGoosen: Did you see https://github.com/jasp-stats/INTERNAL-jasp/issues/445 ?

QMLComponents/analysisform.cpp
```diff
@@ -308,12 +307,31 @@ QString AnalysisForm::msgsListToString(const QStringList & list) const
...
+			_info.references = map["references"].toStringList();
+		if (map.contains("RPackages"))
```
JorisGoosen: Why dont we get the packages from `DESCRIPTION` instead?

PR #5484, opened 2024-04-29
implements working directory from data

JorisGoosen:
implements https://github.com/jasp-stats/jasp-issues/issues/895


RensDofferhoff (approved):
works fine

PR #5486, opened 2024-04-30
Handle comments in ODS better & more!

JorisGoosen:
I tried to make the importers all a bit more generic by moving some handy functions to Importer from Readstat and reusing that for ODS etc. Although I didnt integrate he jaspImporter and jaspImporterOld. As they arent even derived from Importer... But I left that this way as those "importers" is rather different anyway.

This should make sure that the title of a column is derived from the label in readstat or a comment in ODS
Any cell-comments in ODS are used as labels now

Should implement https://github.com/jasp-stats/jasp-issues/issues/352
Also implements https://github.com/jasp-stats/jasp-issues/issues/2660

### DIFF
--- a/Desktop/data/importers/csvimporter.cpp
+++ b/Desktop/data/importers/csvimporter.cpp
@@ -25,7 +25,6 @@ using namespace std;
 
 CSVImporter::CSVImporter() : Importer()
 {
-    DataSetPackage::pkg()->setIsJaspFile(false);
 }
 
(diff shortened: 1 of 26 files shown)


RensDofferhoff:
Tested on a few files.
Aside from the sync issue it seems fine to me.

RensDofferhoff (dismissed):

JorisGoosen:
Ill have a look if I can solve the sync thing now

EDIT: the synch thing seems to be broken in development as well..

PR #5487, opened 2024-04-30
Fix ribbon column inserts

JorisGoosen:
just specify which datasetview is the "main" one instead of hoping its the last or the first created...
this should link the right one to the ribbon and then fixes jasp-stats/INTERNAL-jasp#2516


RensDofferhoff (approved):
Works!

PR #5490, opened 2024-05-02
make synching great again

JorisGoosen:
This makes synchronisation work and adds some nuance:
- csvs are assumed to only have values, so users are allowed to enter labels for those values and those labels will then be used for new entries of those values.
- ODS (and readstat stuff) however can have labels, so there a user that overwrites a label will disable synchronisation
- Columns are also reordered to reflect the order they have in the datafile
- on synch any duplicate labels (value and display are the same) are merged into the first one
- also after a synch (or manual entering of data in editor) any orphaned labels (not in the data anymore) are removed

I did not test with readstat files because I do not have any of those proprietary messes installed on any of my systems.
But I assume that works fine as they all use the same interface internally anyway

### DIFF
--- a/CommonData/column.cpp
+++ b/CommonData/column.cpp
@@ -485,6 +485,10 @@ columnType Column::setValues(const stringvec & values, const stringvec & labels,
 		}
 	}
 	
+	if(labelsRemoveOrphans() && aChange)
+		(*aChange) = true;
+	
+	
 	dbUpdateValues(false);
 	
 	//Now determine what the most logical columntype would be given the current values AND empty values!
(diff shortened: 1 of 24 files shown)


RensDofferhoff:
Found one issue until now:
- Start libreoffice with :
[Book.ods](https://github.com/jasp-stats/jasp-desktop/files/15199723/Book.ods)
- Open the same file in jasp.
- right click on the empty column and insert a new column in libre.
- Save

Crashes with std::exception

RensDofferhoff:
Also happens for CSVs

JorisGoosen:
Cannot reproduce on macos...

JorisGoosen:
Ok, ill merge this now so we can build some nightlies of it, but then ill fix linux tomorrow

PR #5492, opened 2024-05-06
fix variables name display in variablesList

### LINKED ISSUE #2703: [Bug]: Variable display error
### JASP Version

0.18.3

### Commit ID

_No response_

### JASP Module

Unrelated

General display of variables for all analyses

Windows 10

### Bug Description

Hi, 
I am opening a .csv file which I transformed from .xlsx. All data is shown normally, but when I want to do analyses, the variables are not displayed correctly (see attached file).
I tried different .csv formats, installed JASP anew, restarted computer, but nothing changed the issue - what am I missing? 
I would be very happy about suggestions/comments!

image

### Expected Behaviour

Normal display of variables.

### Steps to Reproduce

1. Just opening the file.
2.
3.
...

### Log (if any)

_No response_

### Final Checklist

shun2wang:
Fix: https://github.com/jasp-stats/jasp-issues/issues/2703

Empty characters at both ends should be removed. in the sample file there are some newlines  at the end of variables name, I'm not sure we should allow this, it should be removed everywhere completely to be on the safe side, but only for display purposes this seems to be enough.

### DIFF
--- a/QMLComponents/components/JASP/Controls/VariablesList.qml
+++ b/QMLComponents/components/JASP/Controls/VariablesList.qml
@@ -488,7 +488,7 @@ VariablesListBase
 					id:						colName
 					anchors.left:			variablesList.showVariableTypeIcon ? icon.right : itemRectangle.left
 					anchors.leftMargin:		jaspTheme.generalAnchorMargin
-					text:					model.name
+					text:					model.name.trim()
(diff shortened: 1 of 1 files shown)


JorisGoosen:
Its probably better to simply not allow any newlines in the columnnames at all ever.
I can fix this easily on the import side

shun2wang:
great!

PR #5497, opened 2024-05-09
try delay the check version action while app start

shun2wang:
make a test branch and build please.

### DIFF
--- a/Desktop/gui/jaspversionchecker.cpp
+++ b/Desktop/gui/jaspversionchecker.cpp
@@ -26,7 +26,7 @@
 
 JASPVersionChecker::JASPVersionChecker(QObject *parent) : QObject(parent)
 {
-	QTimer::singleShot(500, this, &JASPVersionChecker::checkForJaspUpdate);
+	QTimer::singleShot(500000, this, &JASPVersionChecker::checkForJaspUpdate);
 }
 
 void JASPVersionChecker::checkForJaspUpdate()
--- a/Desktop/gui/jaspversionchecker.cpp
+++ b/Desktop/gui/jaspversionchecker.cpp
@@ -26,6 +26,8 @@
 
 JASPVersionChecker::JASPVersionChecker(QObject *parent) : QObject(parent)
 {
+    std::cout << "Checking JASP update" << std::endl;
+
 	QTimer::singleShot(500000, this, &JASPVersionChecker::checkForJaspUpdate);
 }
 
--- a/Desktop/main.cpp
+++ b/Desktop/main.cpp
@@ -440,7 +440,7 @@ int main(int argc, char *argv[])
 				QQuickWindow::setGraphicsApi(QSGRendererInterface::OpenGL);
 				args.push_back("--disable-gpu");
 				char dst[] = "LIBGL_ALWAYS_SOFTWARE=1";
-				_putenv(dst);
+				putenv(dst);
 			}
 
 			if(hideJASP)
--- a/Desktop/main.cpp
+++ b/Desktop/main.cpp
@@ -19,6 +19,7 @@
 
 #include <QDir>
 
+#include "qquickwindow.h"
 #include "utilities/application.h"
 #include "utilities/settings.h"
 #include <QtWebEngineQuick/qtwebenginequickglobal.h>
--- a/Desktop/main.cpp
+++ b/Desktop/main.cpp
@@ -434,10 +434,12 @@ int main(int argc, char *argv[])
 			if(safeGraphics)
 			{
 				std::cout << "Setting special options for software rendering (aka safe graphics)." << std::endl;
-				QCoreApplication::setAttribute(Qt::AA_UseSoftwareOpenGL);
+				//QCoreApplication::setAttribute(Qt::AA_UseSoftwareOpenGL);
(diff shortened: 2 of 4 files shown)


Desktop/gui/jaspversionchecker.cpp
```diff
@@ -46,7 +46,7 @@ void JASPVersionChecker::downloadKnownIssues()
 void JASPVersionChecker::downloadVersionFinished()
 {
 	QString version			= _networkReply->readAll().trimmed(),
-			downloadfile	= "https://jasp-stats.org/download/";
+			downloadfile	= "http://jasp-stats.org/download/";  //Do not do use https here because on windows some PC has ssl verify problems.
```
JorisGoosen: This is only used when one clicks the download button.
Id rather not remove https here, as that could allow mitm to lead people to the wrong (possibly spoofed) downloadpage

shun2wang: Ah, yes!

JorisGoosen:
Ive pushed them to jasp-stats and turned on nightlies, but I think I broke windows so maybe only those for macos will actually build

JorisGoosen:
@RensDofferhoff  will build a version now (thanks!)

Desktop/main.cpp
```diff
@@ -434,10 +434,12 @@ int main(int argc, char *argv[])
 			if(safeGraphics)
 			{
 				std::cout << "Setting special options for software rendering (aka safe graphics)." << std::endl;
-				QCoreApplication::setAttribute(Qt::AA_UseSoftwareOpenGL);
+				//QCoreApplication::setAttribute(Qt::AA_UseSoftwareOpenGL);
+				QCoreApplication::setAttribute(Qt::AA_UseDesktopOpenGL);
+				QQuickWindow::setGraphicsApi(QSGRendererInterface::OpenGL);
```
shun2wang, line 440: I imagine if we could try different rendering backends by passing startup parameters in order to debug and solve the problem of qtWebengine init/crashing on certain graphics cards and rendering backends. any ideas? @JorisGoosen

shun2wang:
Sorry I missed include "qquickwindow.h" so maybe we should re-trigger it again.

RensDofferhoff:
@shun2wang I will first make a build of development branch and then this one. Could you rebase on development or should I?

shun2wang:
ah yes please do it. thanks

Desktop/main.cpp
```diff
@@ -434,10 +435,12 @@ int main(int argc, char *argv[])
...
 				args.push_back("--disable-gpu");
 				char dst[] = "LIBGL_ALWAYS_SOFTWARE=1";
-				putenv(dst);
+				_putenv(dst);
```
JorisGoosen: this breaks unix builds

Desktop/utilities/imgschemehandler.cpp
```diff
@@ -12,6 +13,8 @@ using namespace Modules;
 ImgSchemeHandler::ImgSchemeHandler(QObject *parent)
 	: QWebEngineUrlSchemeHandler{parent}
 {
+    std::cout << "ImgSchemeHandler for QtWebengine" << std::endl;
```
shun2wang, line 16: now the application stoped here to start and continue initialize...any ideas to go debug here? @JorisGoosen 

or build a release with debug symbols version?

shun2wang: in logs now we can see:
```
Setting special options for software rendering (aka safe graphics).
Making new argument list for Application startup: JASP.exe --logToFile --disable-gpu
Starting JASP 0.19 from commit 0e4d561737223e223956950a6c9f6d5a5e1b3f6d and branch lockfiletesting
QtWebEngineQuick initialized
Application initialized
PlotSchemeHandler for QtWebengine
```

Desktop/main.cpp
JorisGoosen: Reading https://www.qt.io/blog/new-graphics-integration-in-qt-webengine-6.6-and-even-6.5.1 it looks like you can set it via an environment variable as `QSG_RHI_BACKEND=opengl` and thus do not need the changes you made (for testing anyway).

Not sure if there are other settings that make sense. In any case, we just switched to 6.7 and I guess theyve now removed the old opengl backend. So the new nightly should be behaving differently even in opengl mode if I understand correctly.

With the environment variable it is possible to test if unsetting it makes a difference in 0.18.3 (it was built with 6.5.3 I think)

Desktop/utilities/imgschemehandler.cpp
```diff
@@ -12,6 +13,8 @@ using namespace Modules;
...
+
 	QQuickWebEngineProfile::defaultProfile()->installUrlSchemeHandler("img", this);
 }
```
JorisGoosen, line 19:
```suggestion
	std::cout << "scheme installed" << std::endl;
}
```

JorisGoosen: same at the other one, the we know it breaks in there. But the thing is, if it does then what?

shun2wang: Yes, we can do nothing when I find it‘s a Qt bug(maybehttps://bugreports.qt.io/browse/QTBUG-122947). We can only hope for future Qt solutions. Then I will leave it for now.

PR #5503, opened 2024-05-15
make sure cmake find jaspColumnEncoder source

shun2wang:
I cannot build neither on Windows and Linux for a clean build if not change `CMAKE_SOURCE_DIR` to `PROJECT_SOURCE_DIR`.

but I'm not really sure this make sense because in this dir `CMAKE_SOURCE_DIR` and `PROJECT_SOURCE_DIR` point to same path.

### DIFF
--- a/Common/CMakeLists.txt
+++ b/Common/CMakeLists.txt
@@ -25,7 +25,7 @@ endif()
 target_include_directories(
 	Common
 	PUBLIC # JASP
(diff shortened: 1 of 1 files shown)


JorisGoosen (approved):

JorisGoosen:
I dont get it either, although there is indeed some small difference between the two.
This is also fine by me though and if it helps building on your sys ^^

PR #5507, opened 2024-05-16
Fix typo in eye movements data

JohnnyDoorn:
@RensDofferhoff so if I change stuff here, also the online library gets updated right? and does that include the html file that your fancy script extracts from the .jasp file?


RensDofferhoff:
It should run it everyday so lets see what happens :p

RensDofferhoff (approved):

PR #5508, opened 2024-05-16
Some fixes

JorisGoosen:
- rcommander: align dropdown with buttons and make them rectangular 
- welcomepage: make logo jump out a bit on hover 
- combobox/dropdown cant refer to form because it probably doesnt exist. 
- remove Matrix workaround 
- some repository related stuff:

we were getting R always from https://cran.r-project.org/, completely ignoring the setting of `R_REPOSITORY`. This is now used instead

Also, the repository itself wasnt getting properly loaded for certain installs. because it would try http://cloud.r-project.org/bin/macosx/contrib/4.4/ instead http://cloud.r-project.org/bin/macosx/big-sur-arm64/contrib/4.4/.
This is rectified by setting the binary type for r to `mac.binary.arm64` or `mac.binary.x86_64` instead of `mac.binary`

### DIFF
--- a/Desktop/components/JASP/Widgets/RCommanderWindow.qml
+++ b/Desktop/components/JASP/Widgets/RCommanderWindow.qml
@@ -252,7 +252,7 @@ Window
 				}
 			}
 
-			JC.RoundedButton
+			JC.RectangularButton
 			{
 				id:				runButton
 				text:			qsTr("Run code")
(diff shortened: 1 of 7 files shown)


JorisGoosen:
It might be that the `mac.binary` thing only happens for `install.packages`, but I guess these are still good improvements

boutinb (approved):

PR #5511, opened 2024-05-16
update development docs

shun2wang:
we should specify the tag of ReadStat while clone it. also update relared build tools.


RensDofferhoff (approved):
Great!

PR #5512, opened 2024-05-17
fix linux build action broken

shun2wang:


RensDofferhoff:
Goodmorning.
I'd rather update the Qt version of the github action. I have done so now lets see if that works

shun2wang:
Okay, thanks.

shun2wang:
but maybe it's better to add the ninja tools back because it was failed action so I removed it.
and you can just push it.

RensDofferhoff:
did that and it seems to work now! Thanks!

PR #5519, opened 2024-05-21
Show (implicitly) allowed columntypes in variableslist clearly and sorted

JorisGoosen:
fixes jasp-stats/INTERNAL-jasp#2532 sort of

Makes sure that whatever ends up being "allowed" in a variableslist is what the user sees.
Because who could guess the logic of:
- scale suggested -> scale, ordinal, nominal allowed
- ordinal suggested -> ordinal allowed
- nominal suggested -> ordinal, nominal allowed
- whatever allowed -> thatever allowed

So I made some code to nicely show suggested icons transparently and allowed columns clearly.
But then, it is kind of hard to know what to drop in the boxes in the end...
So I removed that code and just show the implicitely allowed types.

Im wondering whether I should also just remove the `scale, ordinal, nominal` entirely, because that is what an empty one does anyway.

### DIFF
--- a/QMLComponents/components/JASP/Controls/VariablesList.qml
+++ b/QMLComponents/components/JASP/Controls/VariablesList.qml
@@ -248,7 +248,7 @@ VariablesListBase
 			z:			2
 			mipmap:		true
 			smooth:		true
-			opacity:	1.0
+			opacity:	0.5
 			anchors
 			{
 				bottom:			itemRectangle.bottom;
--- a/QMLComponents/components/JASP/Controls/VariablesList.qml
+++ b/QMLComponents/components/JASP/Controls/VariablesList.qml
@@ -238,7 +238,7 @@ VariablesListBase
 		
 	Repeater
 	{
-		model: allowedColumnsIcons.length > 0 ? allowedColumnsIcons : suggestedColumnsIcons
+		model: allowedColumnsIcons
 
 		Image
 		{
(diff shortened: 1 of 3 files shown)


boutinb:
Then the suggested types have no sense really... But as I have the impression that it was never really clear for the researchers what was the differences between allowed & suggested types, it is was good to make it simpler.
We should then replace the suggested type from the analyses (step by step with other PR's).
I would set back the opacity to .8 for the icons, to make a difference with the icon type of a variable item in the Variables list.

boutinb (approved):

JorisGoosen:
> Then the suggested types have no sense really... 

Well no
The only way it wouldve made sense to me is that "suggested" actually suggests to the user what to type of vars to put and then allows everything anyway.

> But as I have the impression that it was never really clear for the researchers what was the differences between allowed & suggested types, it is was good to make it simpler. We should then replace the suggested type from the analyses (step by step with other PR's). 

Yes, or instead we make suggested more logical, so with the behaviour where it *really is* a suggestion.
Maybe this would be a more logical interpretation? @JohnnyDoorn @vandenman @Kucharssim @EJWagenmakers ?

Edit: Not for now though!
Because that would be quite a difference from how it worked, but perhaps not from how it was interpreted?

> I would set back the opacity to .8 for the icons, to make a difference with the icon type of a variable item in the Variables list.

Before this PR it was the "inactive" symbol with an opacity of 0.8 ? Or what?
Im not sure why we would want the inactive icon, so I used the normal one, looks brighter and more saturated.
Ill make it clearer that its not the variables themselves.

RensDofferhoff:
This should either be merged now or after the release.

JorisGoosen:
ok!

RensDofferhoff:
:)

PR #5520, opened 2024-05-22
fix help doc string escapes

shun2wang:
fix: https://github.com/jasp-stats/INTERNAL-jasp/issues/2530

### DIFF
--- a/Desktop/utilities/helpmodel.cpp
+++ b/Desktop/utilities/helpmodel.cpp
@@ -22,6 +22,7 @@ void HelpModel::runJavaScript(QString renderFunc, QString content)
 	Log::log() << "Help is sending content: '" << content << "'" << std::endl;
 #endif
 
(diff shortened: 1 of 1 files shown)


Desktop/utilities/helpmodel.cpp
```diff
@@ -22,6 +22,7 @@ void HelpModel::runJavaScript(QString renderFunc, QString content)
 	Log::log() << "Help is sending content: '" << content << "'" << std::endl;
 #endif
 
+	content.replace("\\", "\\\\");  //The order here is important, replace the escape characters first.
```
JorisGoosen:
```suggestion
	content.replace("\\", "\\\\");  //The order here is important, replace the escaped escape characters first. Helps with latex, see: https://github.com/jasp-stats/INTERNAL-jasp/issues/2530
```

JorisGoosen (approved):

PR #5527, opened 2024-05-23
Make sure we dont leak crazy amounts of memory on analyses/edit switch

### LINKED ISSUE #2733: [Bug]: choppy panel dragging after entering data
with 0.18.3 and latest 0.19 beta

entering data

### Bug Description

After entering some data, dragging panels via the 3 vertical grey dots gets extremely choppy, like 1fps slideshow.
This does not happen when starting with data from the data library! Only when I enter values myself. The amount in the screenshot is enough to trigger it. 
Might be intel+wayland related? But strange that it does not occur with loaded data...

### Expected Behaviour

JASP should stay fluent as always

### Steps to Reproduce

1. open jasp
2. enter some columns of data
3. drag an analysis panel

JorisGoosen:
Also fixes https://github.com/jasp-stats/jasp-issues/issues/2733


Desktop/components/JASP/Widgets/DataTableView.qml
```diff
@@ -78,7 +78,7 @@ FocusScope
 			itemVerticalPadding:	8 * jaspTheme.uiScale
 
 			model:					dataSetModel
-			cacheItems:				!ribbonModel.dataMode
+			cacheItems:				true //!ribbonModel.dataMode
```
JorisGoosen, line 81: switching both caching and expanding the dataset is probably not a great idea

and I turned caching off before because there were weird interactions with items editing all over the place...
That seems to have been fixed though

Desktop/qquick/datasetview.cpp
```diff
@@ -120,11 +120,29 @@ QSizeF DataSetView::getRowHeaderSize()
 	return getTextSize(text);
 }
 
+void DataSetView::clearCaches()
```
JorisGoosen, line 123: I rewrote this to be a correct function, it is now however not used

RensDofferhoff (approved):

boutinb (approved):

PR #5532, opened 2024-05-27
skip empty values labels when getting value/label

JorisGoosen:
fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2537 (issues with missing values)

### DIFF
--- a/CommonData/column.h
+++ b/CommonData/column.h
@@ -165,14 +165,15 @@ public:
 			const Labels		&	labels()																				const	{ return _labels; }
 			bool					labelsMergeDuplicates();
 			bool					labelsRemoveOrphans();
-			Label				*	labelByDisplay(			const std::string	&	display)								const; ///< Might be nullptr for missing label, returns the first of labelsByDisplay
-			Label				*	labelByValue(			const std::string	&	value)									const; ///< Might be nullptr for missing label, returns the first of labelsByValue
-			Label				*	labelByIntsId(			int						intsId)									const; ///< Might be nullptr for missing label
 			Labelset				labelsByDisplay(		const std::string	&	display)								const; ///< Might be nullptr for missing label
 			Labelset				labelsByValue(			const std::string	&	value)									const; ///< 
-			Label				*	labelByValueAndDisplay(	const std::string	&	value, const std::string &	label)		const; ///< Might be nullptr for missing label, assumes you ran labelsMergeDuplicates before
-			Label				*	labelByRow(				int						row)									const; ///< 
 			int						labelIndex(				const Label			*	label)									const;
+			Label				*	labelByRow(				int						row)									const; ///< 
+			Label				*	labelByValue(			const std::string	&	value)									const; ///< Might be nullptr for missing label, returns the first of labelsByValue
+			Label				*	labelByIntsId(			int						intsId)									const; ///< Might be nullptr for missing label
+			Label				*	labelByDisplay(			const std::string	&	display)								const; ///< Might be nullptr for missing label, returns the first of labelsByDisplay
+			Label				*	labelByIndexNotEmpty(	size_t					index);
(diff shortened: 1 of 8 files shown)


CommonData/column.h
```diff
@@ -165,14 +165,15 @@ class Column : public DataSetBaseNode
...
 			Labelset				labelsByValue(			const std::string	&	value)									const; ///< 
-			Label				*	labelByValueAndDisplay(	const std::string	&	value, const std::string &	label)		const; ///< Might be nullptr for missing label, assumes you ran labelsMergeDuplicates before
-			Label				*	labelByRow(				int						row)									const; ///< 
 			int						labelIndex(				const Label			*	label)									const;
+			Label				*	labelByRow(				int						row)									const; ///< 
+			Label				*	labelByValue(			const std::string	&	value)									const; ///< Might be nullptr for missing label, returns the first of labelsByValue
+			Label				*	labelByIntsId(			int						intsId)									const; ///< Might be nullptr for missing label
+			Label				*	labelByDisplay(			const std::string	&	display)								const; ///< Might be nullptr for missing label, returns the first of labelsByDisplay
+			Label				*	labelByIndexNotEmpty(	size_t					index);
+			Label				*	labelByValueAndDisplay(	const std::string	&	value, const std::string &	label)		const; ///< Might be nullptr for missing label, assumes you ran labelsMergeDuplicates before
```
JorisGoosen: I know this looks terrible but it just adds `Label				*	labelByIndexNotEmpty(	size_t					index);` and makes it look prettier ;)

boutinb:
It seems to work. Just one thing: if you have an empty value with NA for example, the first time it displays in gray NA. If you click on it and goes away, the NA disappears: it has lost the information that the empty value was an NA.

boutinb:
Oeps... It is worse than I thought. Just load this file:
[descriptivesbug.csv](https://github.com/jasp-stats/jasp-desktop/files/15456742/descriptivesbug.csv)

Click on the NA cell of Headbanging intensity (row 7)
Click Enter -> the NA disappears
Click again on the empty value cell of Headbanging intensity.
Enter NA -> the 4th row of this column becomes also NA !!!!

JorisGoosen:
Ok this problem is also fixed.

There is some sensible, but perhaps counter-intuitive, behaviour when entering data in a field that cant handle it (so text in `scalar` fields). This is now a bit hard to clear, but "Clear cells" works so...

Lets talk about this tomorrow. But with this fix at least it (seems) to work consistently

boutinb:
It works now much better. Only one thing: If you change a value to 'NA' eg, in a row that has a value which is in other rows, it updates all the rows to 'NA'.
I have added a commit for this. If you find it correct, then we can merge this PR, and make new nightly.

JorisGoosen:
> It works now much better. Only one thing: If you change a value to 'NA' eg, in a row that has a value which is in other rows, it updates all the rows to 'NA'. I have added a commit for this. If you find it correct, then we can merge this PR, and make new nightly.

Ill test it out!

JorisGoosen:
Your change disallows entering missing values entirely...

Ill look for a different solution

Edit: I cannot reproduce what you say happens in https://github.com/jasp-stats/jasp-desktop/pull/5532#issuecomment-2134523581
Edit2: So ive removed your commit.

JorisGoosen:
Ok, ive got a repropath with 
[descriptivesbug.csv](https://github.com/jasp-stats/jasp-desktop/files/15466761/descriptivesbug.csv)

Replace the `NA` at the end of "Headbanging intensity" with '2' and then '2' with 'NA'

JorisGoosen:
I noticed some code duplication sneaked in so I removed some stuff I dont think is necessary in c8262670533d9ad94720a5329bc3d8cd33d3e1cf "een poging setValue zoveel mogelijk te gebruiken"

Also automatically adds MAKEFLAGS to all R pkg builds, which should speedup from source compilation a lot (like jaspBase etc)

boutinb (approved):

PR #5536, opened 2024-05-30
Make autoSort toggleable

JorisGoosen:
Implements https://github.com/jasp-stats/INTERNAL-jasp/issues/2555 and makes the alignment of the labeleditor better as well

### DIFF
--- a/CommonData/column.cpp
+++ b/CommonData/column.cpp
@@ -41,7 +41,7 @@ void Column::dbLoad(int id, bool getValues)
 	
 	Json::Value emptyVals;
 	
-	db().columnGetBasicInfo(	_id, _name, _title, _description, _type, _revision, emptyVals);
+	db().columnGetBasicInfo(	_id, _name, _title, _description, _type, _revision, emptyVals, _autoSortByValue);
 	db().columnGetComputedInfo(	_id, _analysisId, _invalidated, _forceTypes, _codeType, _rCode, _error, _constructorJson);
 	
 	_emptyValues->fromJson(emptyVals);
@@ -243,6 +243,22 @@ bool Column::setConstructorJson(const std::string & constructorJson)
 	Json::Reader().parse(constructorJson, parsed);
 	
 	return setConstructorJson(parsed);
+}
+
+void Column::setAutoSortByValue(bool sort)
+{
+	JASPTIMER_SCOPE(Column::setAutoSortByValue);
+
+	if(sort == _autoSortByValue)
+		return;
+	
+	_autoSortByValue = sort;
+	
+	db().columnSetAutoSort(_id, _autoSortByValue);
+	
+	labelsHandleAutoSort();
+	
+	return;	
 }
 
 bool Column::setConstructorJson(const Json::Value &constructorJson)
@@ -599,6 +615,8 @@ void Column::_dbUpdateLabelOrder(bool noIncRevisionWhenBatchedPlease)
 			incRevision();
 		return;
 	}
+	
+	labelsHandleAutoSort(false);
 
 	intintmap orderPerDbIds;
 
@@ -1210,7 +1228,14 @@ void Column::labelValueChanged(Label *label, double aDouble)
 			_dbls[r] = aDouble;
 		}
 	
+	labelsHandleAutoSort();
 	dbUpdateValues();
+}
+
+void Column::labelsHandleAutoSort(bool doDbUpdateEtc)
+{
+	if(_autoSortByValue)
+		labelsOrderByValue(doDbUpdateEtc);	
 }
 
 void Column::labelDisplayChanged(Label *label)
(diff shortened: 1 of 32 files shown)


boutinb:
I have still some strange things... Open debug.csv:

- Open contBinom Label Editor 
- Unset 'Automatically order labels by their numeric values'
- Undo: The values get a decimal format (0.0 and 1.0)

boutinb:
Also:

- Open contBinom Label Editor
- Unset 'Automatically order labels by their numeric values'
- Select the first row (value 0)
- Click on the 'Move selected labels down'
- The row goes down, but the buttons to move the labels up or down are disabled.

JorisGoosen:
those label problems have been fixed (there was an issue where label deserialisation went wrong because the original-value-json was being forced to string instead of it just staying as it should be (a json:value!).

Also fixed a bug caused by turning caching of items in the datasetview, where a random edit item would be opened when trying to change editposition by clicking.

Desktop/components/JASP/Widgets/JASPDataView.qml
```diff
@@ -16,6 +16,7 @@ FocusScope
 				property alias mouseArea:				wheelCatcher
 				property bool  doubleClickWorkaround:	true
 				property alias isMainDataViewer:		theView.mainData
+	readonly	property alias editCoordinates:			theView.editCoordinates
```
boutinb, line 19: I don't see where editCoordinates is used. Can be removed?

Desktop/data/undostack.h
```diff
@@ -58,35 +58,44 @@ class SetWorkspacePropertyCommand: public UndoModelCommand
...
-	ColumnModel*			_columnModel = nullptr;
-	int						_colId		= -1,
-							_labelIndex = -1;
+	int						_labelIndex = -1;
 	QString					_newLabel,
 							_oldLabel;
 };
 
-class SetLabelOriginalValueCommand: public UndoModelCommand
+class SetLabelOriginalValueCommand: public UndoModelCommandLabelChange
```
boutinb, line 90: I think you forgot to also undo value changes in the data. The undo method must be also overriden.
If you change the value of a label, and undo it, it does not undo the values. The old value/label still exists. For example: In contBinom, change Value 1 to 10, Undo: the 10 are still in the data. Change column, and come back to contBinom: you see still the label 10/10.

CommonData/column.cpp
```diff
@@ -1811,20 +1859,54 @@ Json::Value Column::serialize() const
...
+						labelJson["intsId"]			.asInt(), 
+						labelJson["label"]			.asString(), 
+						labelJson["filterAllows"]	.asBool(), 
+						labelJson["description"]	.asString(), 
+						labelJson["originalValue"], 
+						labelJson["order"]			.asInt(), 
+						-1
+			);
+}
+
```
boutinb, line 1966: Should the deserialize labels not be done inside a batch?

CommonData/databaseinterface.cpp
```diff
@@ -21,14 +21,17 @@ void DatabaseInterface::upgradeDBFromVersion(Version originalVersion)
...
-			"ALTER TABLE DataSets ADD COLUMN description     TEXT;"			"\n");
+			"ALTER TABLE DataSets ADD COLUMN description     TEXT;");
 
 	
 	if(originalVersion < "0.19.0")
 		runStatements(		
-			"ALTER TABLE Columns  ADD 	COLUMN emptyValuesJson TEXT;"		"\n"
-			"ALTER TABLE Columns  ADD 	COLUMN forceSourceColType INT NULL;""\n"
-			"ALTER TABLE Columns  DROP 	COLUMN isComputed;"					"\n"); // was removed in 0.18.3
+			R"C++20IsGreat(
```
boutinb: It's c++11 feature!

boutinb (changes requested):
Still one bug: undo value change in Label Editor does not work. Cf below

Desktop/data/undostack.h
JorisGoosen: aaah fuck yes

CommonData/databaseinterface.cpp
JorisGoosen: even more amazing ^^

Desktop/components/JASP/Widgets/JASPDataView.qml
JorisGoosen: Could be removed, I just added it because it might be useful at some point and its not too much extra code anyway.
I first thought the variable-editor used this too...

CommonData/column.cpp
JorisGoosen: Good one

Desktop/components/JASP/Widgets/JASPDataView.qml
JorisGoosen: Im not removing it, it could turn out to be useful

boutinb (approved):

PR #5540, opened 2024-06-04
Add dropdown types

boutinb:


JorisGoosen:
So @boutinb ive added the icons and the totalNumerics path for variableslist.
I think we should probably still have:
- properties: `minNumericLevels`, `maxNumericLevels`.
- And perhaps: `minLevels` and `maxLevels`

These both can then be the check for the drag->drop, instead of the type.

JorisGoosen:
Ok awesome!

JorisGoosen (approved):

PR #5542, opened 2024-06-04
make sure to also set type properly when pasting text vs  number a la hand entering

JorisGoosen:
fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2500

### DIFF
--- a/CommonData/column.h
+++ b/CommonData/column.h
@@ -153,7 +153,7 @@ public:
 			void					labelValueChanged(Label * label, int	anInteger) { labelValueChanged(label, double(anInteger)); }
 			void					labelDisplayChanged(Label * label);
 			
-			bool					setStringValue(				size_t row, const std::string & value, const std::string & label = "", bool writeToDB = true); ///< Does two things, if labelButOnlyFromSpreadsheetPaste=="" it will handle user input, as value or label depending on columnType. Otherwise it will simply try to use userEntered as a value. But this will trigger the setting of type
+			bool					setStringValue(				size_t row, const std::string & value, const std::string & label = "", bool writeToDB = true); ///< Does two things, if label=="" it will handle user input, as value or label depending on columnType. Otherwise it will simply try to use userEntered as a value. But this will trigger the setting of type
 			bool					setValue(					size_t row, const std::string & value, const std::string & label,	bool writeToDB = true);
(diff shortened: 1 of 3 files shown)


Desktop/data/datasetpackage.cpp
```diff
@@ -1916,21 +1916,23 @@ void DataSetPackage::pasteSpreadsheet(size_t row, size_t col, const std::vector<
 	{
 		Column	*	column		= _dataSet->column(c + col);
 		columnType	desiredType	= coltypes.size() > c ? columnType(coltypes[c]) : column->type();
+		bool		noType		= desiredType == columnType::unknown;
```
boutinb: This variable is not used..

CommonData/column.h
```diff
@@ -153,7 +153,7 @@ class Column : public DataSetBaseNode
 			void					labelValueChanged(Label * label, int	anInteger) { labelValueChanged(label, double(anInteger)); }
 			void					labelDisplayChanged(Label * label);
 			
-			bool					setStringValueToRow(		size_t row, const std::string & value,								bool writeToDBAndSetType = true);
+			bool					setStringValue(				size_t row, const std::string & value, const std::string & label = "", bool writeToDB = true); ///< Does two things, if labelButOnlyFromSpreadsheetPaste=="" it will handle user input, as value or label depending on columnType. Otherwise it will simply try to use userEntered as a value. But this will trigger the setting of type
```
boutinb: The comment specifies labelButOnlyFromSpreadsheetPaste, but the method has label as argument...

boutinb (changes requested):
Just 2 small remarks

Desktop/data/datasetpackage.cpp
```suggestion
```

CommonData/column.h
JorisGoosen: image
Then:
image
Then:
image

Then:
```suggestion
			bool					setStringValue(				size_t row, const std::string & value, const std::string & label = "", bool writeToDB = true); ///< Does two things, if label=="" it will handle user input, as value or label depending on columnType. Otherwise it will simply try to use userEntered as a value. But this will trigger the setting of type
```

JorisGoosen: image

PR #5556, opened 2024-06-11
Renv Lockfiles + Flatpak Dev modules

RensDofferhoff:
Switches to renv lockfiles installed using jaspModuleInstaller.

Renv lockfiles in each jasp Module contain all dependencies of the module.
jaspModuleInstaller will add/alter records to this lockfile to produce _processedLockFile.lock.
renv restore is called on this lockfile.
If MODULE_INSTALL_MODE is localizeAll every jaspModule source in the lockfile will be altered to those in the install folder. This mode is used in compilation of jasp.
If it is localizeModuleOnly (as for dev module installation) only a record to source of module to be installed is added. 

Flatpak installation now uses an renv cellar. Cellars are generated using the same mechanisms as normal installation. This way no special gathering script is required.

See https://static.jasp-stats.org/Nightlies/ for the latest lockfiletesting nightlies

Todo:
- Write some documentation somewhere

Known Issues:
jaspProcess is not installed correctly on Mac because blavaan compilation fails. I have no clue why but if we just use the CRAN binaries its all fine. I have a PR for this: https://github.com/jasp-stats/jaspProcess/pull/102
Maybe @vandenman  has any clue why this is going on. Seems to fail on rstantools call.


RensDofferhoff:
When this is merged we can remove module install code from jaspBase.
And we can add runners to automatically update lockfiles daily/weekly

JorisGoosen:
Nice!

I can build and run it perfectly well on windows on release.
Edit: That was the wrong branch

JorisGoosen:
Alright, the good branch does work, however I still cant install dev modules.
Ill look into it later more

Modules/Rcpp_RInside.lock
```diff
@@ -0,0 +1,23 @@
+{
+  "R": {
+    "Version": "4.3.3",
```
shun2wang, line 3: Why we lock all R version to 4.3.3?

RensDofferhoff: This is ignored. We will have runners to automatically update these lockfiles in the future

RensDofferhoff:
> Alright, the good branch does work, however I still cant install dev modules. Ill look into it later more

Any logs?

PR #5557, opened 2024-06-12
anonymized github_pat in user logs

shun2wang:
It seems that we should not log the user's github_pat in plain text. Although most of the time this is not a big deal, but it is unnecessary if we need users to upload logs.

### DIFF
--- a/Engine/engine.cpp
+++ b/Engine/engine.cpp
@@ -195,8 +195,14 @@ bool Engine::receiveMessages(int timeout)
 			// Log::log() << "Parsing request failed on:\n" << err << std::endl;
 		}
 
-		//Clear send buffer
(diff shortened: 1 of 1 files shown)


JorisGoosen:
This wont fix it for when the user has it set in the preferences, because then it will be printed in what the desktop sends/logs as well I think?

Great idea though!

shun2wang:
No I think this just hide the github_pat in logs if users set it. because it just used for log() stream I guess?

shun2wang:
we could not record it in logs but should pass or showing inside JASP anyway.

shun2wang:
ah sorry maybe I missed your point, you mean the Desktop.log right? I just changed the engine logs indeed.

JorisGoosen (approved):
Ah no that is also only printed in the engine it turns out, so your fix is sufficient!

PR #5560, opened 2024-06-13
Scroll faster

JorisGoosen:
Makes dataviewer go faster


RensDofferhoff (approved):
no major memory leaks. Definitely an improvement

boutinb (approved):

PR #5565, opened 2024-06-15
fix data table edit popup menu

shun2wang:
maybe a typo here? cause it not popup...

### DIFF
--- a/Desktop/components/JASP/Widgets/DataTableViewEdit.qml
+++ b/Desktop/components/JASP/Widgets/DataTableViewEdit.qml
@@ -194,8 +194,7 @@ Item
 
             onClicked: (mouse) =>
 			{
-				if(mouse.buttons & Qt.RightButton)
-					dataTableView.showPopupMenu(editItemRoot, mapToGlobal(mouse.x, mouse.y), rowIndex, columnIndex);
+				dataTableView.showPopupMenu(editItemRoot, mapToGlobal(mouse.x, mouse.y), rowIndex, columnIndex);
 			}
 		}
 	}
--- a/Desktop/components/JASP/Widgets/DataTableViewEdit.qml
+++ b/Desktop/components/JASP/Widgets/DataTableViewEdit.qml
@@ -194,7 +194,8 @@ Item
 
             onClicked: (mouse) =>
 			{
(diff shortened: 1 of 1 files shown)


Desktop/components/JASP/Widgets/DataTableViewEdit.qml
```diff
@@ -194,8 +194,7 @@ Item
 
             onClicked: (mouse) =>
 			{
-				if(mouse.buttons & Qt.RightButton)
-					dataTableView.showPopupMenu(editItemRoot, mapToGlobal(mouse.x, mouse.y), rowIndex, columnIndex);
+				dataTableView.showPopupMenu(editItemRoot, mapToGlobal(mouse.x, mouse.y), rowIndex, columnIndex);
```
boutinb:
```suggestion
				if(mouse.button === Qt.RightButton)
					dataTableView.showPopupMenu(editItemRoot, mapToGlobal(mouse.x, mouse.y), rowIndex, columnIndex);
```

shun2wang: I think it is unnecessary ? because:
```
acceptedButtons:	Qt.RightButton
```

boutinb (approved):
Yes this was a type, but the if statement is still needed.

Desktop/components/JASP/Widgets/DataTableViewEdit.qml
boutinb: > I think it‘s is unnecessary ? because:
> 
> ```
> acceptedButtons:	Qt.RightButton
> ```

Oh you are right! I was too quick. But it's not bad either.

JorisGoosen:
Anyhow, `buttons` also exists right?
https://doc.qt.io/qt-6/qml-qtquick-mouseevent.html

But whatever, if this works it works

shun2wang:
>Anyhow, buttons also exists right?

yes, but `buttons` is `0` in this case, and the `Qt::RightButton` is `2` so the `&` operator always get '0'? I dont know ...maybe it should just used for `onPressed`.

JorisGoosen:
Well, maybe there was a problem with it on your platform?
In any case, if it works with this fix its all good as far as Im concerned

PR #5568, opened 2024-06-17
Popup of dropdown is not set on the right place

boutinb:
The popup of a ComboBox shows up per default at the height of the ComboBox control. Since Qt 6.7, It is apparently not needed anymore to specify the y of the ComboBox control.


JorisGoosen (approved):

JorisGoosen:
Yes they also fixed things in the height of the popup which I made use of in: https://github.com/boutinb/jasp-desktop/commit/6b20ac88b09040c0706a02f4db107f62ebb7d812

PR #5572, opened 2024-06-18
Analysis constraint error should not be closeable

boutinb:
When an error occurs because for example a variable does not have enough levels, this error should not be closeable: the user must remove it from the VariablesList first. As long such an error exists, the analysis cannot be run.

### DIFF
--- a/Desktop/analysis/analysis.cpp
+++ b/Desktop/analysis/analysis.cpp
@@ -222,6 +222,9 @@ void Analysis::run()
 
 void Analysis::refresh()
 {
+	if (form() && form()->hasError())
+		return;
+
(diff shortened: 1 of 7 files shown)


JorisGoosen:
Looks real good but if I press `refresh` it still reruns the analysis

As in Cmd/ctrl+R

boutinb:
> Looks real good but if I press `refresh` it still reruns the analysis
> 
> As in Cmd/ctrl+R

OK, I prevent now refreshing if the analysis has an error.

JorisGoosen (approved):

JorisGoosen:
my first smartphonemerge ever ;)

PR #5586, opened 2024-06-25
Set the control width of Dropdown by setting the fieldWidth property

### LINKED ISSUE #2404: [Feature Request]: Add option to QML dropdown to prevent width from adjusting to longest element name
### Description

Currently, the QML dropdowns automatically adjust their width to the longest element label they have. While this is usually desirable, it can in some cases lead to the dropdown being cut off by the parent element.

### Purpose

In cases, where the space is limited and the length of the longest element is not known, there should be an option to prevent the width from adjusting.

### Use-case

In jaspProcess, the user can use their own variable names as elements for the dropdowns. For long variable names, the dropdowns are cut off.

_No response_

Other

An option in the `DropDown` component that disables the automatic width adjustment.

The options `fieldWidth` and `minControlWidth` don't solve the problem.

### Additional context

image

boutinb:
Add fieldWidth property to Dropdown, to fix the width of the dropdown. If the length of the value is then smaller, it should be elided with '...'. The popup menu should be as wide as necessary so that all values are clearly seen.
A new test Test Dropdown has been added to the Test module.

Fixes https://github.com/jasp-stats/jasp-issues/issues/2404

### DIFF
--- a/Desktop/components/JASP/Widgets/FileMenu/PrefsUI.qml
+++ b/Desktop/components/JASP/Widgets/FileMenu/PrefsUI.qml
@@ -50,19 +50,44 @@ ScrollView
 		{
 			id:		fontGroup
 			title:	qsTr("Fonts")
-			
-			property real maxText: Math.max(Math.max(interfaceTxt.implicitWidth, codeTxt.implicitWidth), Math.max(resultTxt.implicitWidth, languageTxt.implicitWidth));
 
-			Row
+			Item
 			{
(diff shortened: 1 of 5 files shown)


JorisGoosen:
Should this be reviewed and merged?

boutinb:
> Should this be reviewed and merged?

Yes it can be reviewed now

JorisGoosen (approved):
Lovely, ive also added auto-aligning for the labels of dropdowns and used it in PrefsUI

PR #5595, opened 2024-07-03
Suboption not always updated in componentsList

boutinb:
In a componentsList, the sub-controls can be created before the parent control is completed. When creating the json option, for this sub-control, the parent object does not exist. So it must be created. 
Also the meta information is maybe not yet available, so to check whether the option has types, we cannot trust the meta information, but we need to check whether the json is an object with "value" and "types" as member.

Fixes https://github.com/jasp-stats/jasp-test-release/issues/2686


boutinb:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/2622

PR #5623, opened 2024-07-14
bugfix: fix tooltip in the checkBox and RadioButton

shun2wang:
Fixes: https://github.com/jasp-stats/INTERNAL-jasp/issues/2525
Fixes: https://github.com/jasp-stats/INTERNAL-jasp/issues/2366

The reason is that sometimes the checkbox and radioButton cannot actually detect the mouse hover event.

### DIFF
--- a/QMLComponents/components/JASP/Controls/CheckBox.qml
+++ b/QMLComponents/components/JASP/Controls/CheckBox.qml
@@ -44,6 +44,7 @@ CheckBoxBase
 			property alias	fontInfo:				label.fontInfo
 			property alias	label:					control.text
 			property alias	labelTextFormat:		label.textFormat
+			property string toolTip:                ""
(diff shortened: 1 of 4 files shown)


boutinb:
@shun2wang There was already a tooltip in jaspControl but it did not work. Handling attached property in c++ can be tricky: the attached property can be called directly if there is an import statement in the QML file without label. But by setting the tooltip directly in the MouseArea it works perfectly.

boutinb:
I have added also a test in the jaspTestModule.

shun2wang:
Great! so then we should remove all tooltip settings from other jaspControls component and just reuse it from C++ side?

I rebased this PR to head.

shun2wang:
@boutinb This should get into 0.19.2 I gusess.

boutinb (approved):

PR #5630, opened 2024-07-18
Auto Sort labels bug and other label handling issues

boutinb:
This solves several issues:

- The "Auto Sort Label per value" setting (true per default in the preferences) is not saved in the database. Now this setting is saved in the database and thus also in the JASP file.
- The "Auto Sort Label per value" works also for non numeric values. Now the labels are per default also alphabetically ordered.
- The Variable Window does not show always the selected column, especially when navigating in the Data view with the keyboard: the Variable Window was showing still the old selected column.
- The up/down/reverse buttons were not visible or/and enabled in the right way. Now if Auto Sort is on, all the other buttons are removed, except the 'Reverse order of all numerical values' if at least 2 labels have numerical values. If Auto Sort is off, all other buttons are displayed, and are enabled only if they really can change the order (the up button is disabled for example if the the selected row is the first one).

### DIFF
--- a/CommonData/column.cpp
+++ b/CommonData/column.cpp
@@ -17,7 +17,7 @@ void Column::setAutoSortByValuesByDefault(bool autoSort)
 	_autoSortByValuesByDefault = autoSort;
 }
 
-Column::Column(DataSet * data, int id)
+Column::Column(DataSet * data, int id, columnType colType, computedColumnType computedType, bool autoSort)
 :	DataSetBaseNode(dataSetBaseNodeType::column, data->dataNode()),
 	_data(				data),
 	_id(				id),
@@ -29,19 +29,32 @@ Column::Column(DataSet * data, int id)
 		db().columnSetAutoSort(_id, _autoSortByValue); //Store autosort in db
 }
 
+
+Column* Column::addColumn(DataSet * data, int index, const std::string& name, columnType colType, computedColumnType computedType, bool alterDataSetTable)
+{
+	int id = data->db().columnInsert(data->id(), index, colType, computedType, Column::autoSortByValuesByDefault());
+	Column* col = new Column(data, id, colType, computedType, Column::autoSortByValuesByDefault());
+
+	if (!name.empty())
+		col->setName(name);
+
+	return col;
+}
+
+Column* Column::loadColumn(DataSet * data, int index)
+{
+	Column* col = new Column(data, index, columnType::unknown, computedColumnType::notComputed, Column::autoSortByValuesByDefault());
+	col->dbLoadIndex(index, false);
+
+	return col;
+}
+
+
 Column::~Column()
 {
 	labelsTempReset();
 	delete _emptyValues;
 	delete _doubleDummy;
-}
-
-void Column::dbCreate(int index)
-{
-	JASPTIMER_SCOPE(Column::dbCreate);
-
-	assert(_id == -1);
-	db().columnInsert(_id, index);
 }
 
 void Column::dbLoad(int id, bool getValues)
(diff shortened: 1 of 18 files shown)


boutinb:
This is now ready for review.

CommonData/column.h
```diff
@@ -40,14 +40,17 @@ class Analysis;
...
 class Column : public DataSetBaseNode
 {
-public:
-									Column(DataSet * data, int id = -1);
+private:
+									Column(DataSet * data, int id, columnType colType, computedColumnType computedType, bool autoSort);
+
+public:									
+			static Column *			addColumn(DataSet* data, int index = -1, const std::string & name = "", columnType colType = columnType::scale, computedColumnType computedType = computedColumnType::notComputed, bool alterDataSetTable = true);
+			static Column *			loadColumn(DataSet* data, int index);
```
JorisGoosen, line 52: This function obviously should be on `DataSet`...

JorisGoosen:
Ok Ive added a fix already for the autosort not being stored: https://github.com/jasp-stats/jasp-desktop/commit/00545680c905c30b6130739467d80d172b6915c4

CommonData/column.cpp
```diff
@@ -1648,24 +1662,43 @@ void Column::labelsOrderByValue(bool doDbUpdateEtc)
...
-	std::map<double, int>	orderMap;
-	
-	for(size_t i=0; i<asc.size(); i++)
-		orderMap[asc[i]] = i;
-	
-	//and now to write them back into the data
-	for(Label * label : _labels)
+	doublevec asc = valuesNumericOrdered();
+
+	if (asc.empty())
```
JorisGoosen, line 1746: So with this change it either orders the numeric ones *or* the non-numerics if there are no numerics?

Why not order both?

CommonData/column.cpp
```diff
@@ -1694,6 +1727,20 @@ doublevec Column::valuesNumericOrdered()
...
+	stringset	values;
+
+	for(const Label * label : _labels)
+	{
+		std::string	aValue	= label->originalValueAsString();
+		if (!isEmptyValue(aValue))
+			values.insert(aValue);
+	}
+
+	return stringvec(values.begin(), values.end());
```
JorisGoosen, line 1741: You are implicitly sorting it by pushing it into a stringvec? or what?

JorisGoosen (commented):
I have some questions on the code and have the feeling you are changing quite a few things for little gain.

Maybe we can have a look together next week when you are back?

JorisGoosen:
lets talk about this tomorrow

Desktop/components/JASP/Widgets/LabelEditorWindow.qml
```diff
@@ -458,49 +458,49 @@ FocusScope
...
-					toolTip:		columnModel.autoSort ? qsTr("Reverse order of the labels with non-numeric values") : qsTr("Reverse order of all labels")
+					toolTip:		qsTr("Reverse order of all labels")
 	
 					height:			buttonColumnVariablesWindow.buttonHeight
 					implicitHeight: buttonColumnVariablesWindow.buttonHeight
 					width:			height
-					visible:		!columnModel.autoSort || columnModel.rowsTotal - columnModel.firstNonNumericRow > 1 //If there are at least 2 non numerics there is something to reverse
-					
+					visible:		!columnModel.autoSort
+					enabled:		columnModel.rowCount() > 1
```
JorisGoosen, line 475: dont use rowCount() in a binding because it wont update on changes...

JorisGoosen:
All those changes to column, dataset and databaseinterface seem to be useless and change the logic in a bunch of places.
I dont think this is necessary and another way is here: https://github.com/jasp-stats/jasp-desktop/pull/5700

PR #5631, opened 2024-07-25
fix: auto set CMAKE_GENERATOR for conan

shun2wang:
I cannot make a clean build boost if dont set it manually so let's set it automatically in conan.

and on some windows locale codepage the msvc cannot build without `utf-8` codepage(for example my cpdepage 936), it good also for build bot to build and run JASP with utf-8 I guess.


JorisGoosen (approved):
Doesnt look like this will hurt anything although im not sure what problem it fixes exactly

PR #5633, opened 2024-07-29
bugfix: ribbon menu translations hot update

shun2wang:
Fix: https://github.com/jasp-stats/INTERNAL-jasp/issues/2372

So we just remove/renew the menu ribbon button to "reload" it to apply the translations.

### DIFF
--- a/Desktop/modules/ribbonmodel.cpp
+++ b/Desktop/modules/ribbonmodel.cpp
@@ -38,6 +38,7 @@ RibbonModel::RibbonModel() : QAbstractListModel(DynamicModules::dynMods())
 	connect(DynamicModules::dynMods(), &DynamicModules::dynamicModuleUninstalled,	this, &RibbonModel::removeDynamicRibbonButtonModel			);
 	connect(DynamicModules::dynMods(), &DynamicModules::dynamicModuleReplaced,		this, &RibbonModel::dynamicModuleReplaced					);
 	connect(DynamicModules::dynMods(), &DynamicModules::dynamicModuleChanged,		this, &RibbonModel::dynamicModuleChanged					);
+    connect(PreferencesModel::prefs(), &PreferencesModel::languageCodeChanged,      this, &RibbonModel::reloadModulesMenu   );
 }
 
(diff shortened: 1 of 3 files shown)


boutinb:
Really nice!
Just remove the 6 Modules/pkgdepends.lock of your PR.

shun2wang:
>Just remove the 6 Modules/pkgdepends.lock of your PR.

Well, it should be updated all because it block build on Linux.

Or someone can update all packages in `pkgdepends.lock`.

RensDofferhoff:
@shun2wang current pkgdepends does not build for you?

shun2wang:
yeah, you can also find it from any failed actions: https://github.com/jasp-stats/jasp-desktop/actions/runs/10598393408

old packages has build problems with gcc  version maybe.

RensDofferhoff:
Interesting! I have not seen these issues on my machines

@boutinb  The updates are alright

Desktop/modules/ribbonmodel.cpp
```diff
@@ -110,6 +111,27 @@ void RibbonModel::loadModules(std::vector<std::string> commonModulesToLoad, std:
...
+
+    for (const auto& button : allButtons)
+    {
+        removeRibbonButtonModel(button);
+    }
+
+    loadModules();
+}
+
 void RibbonModel::addRibbonButtonModelFromDynamicModule(Modules::DynamicModule * module)
```
boutinb, line 135: I don't think this is the right way to do this. The loadModules(I) will recreate the modules, make some memory leakage (_entriesInsert, _entriesDelete, etc...), this is quite dangerous I think.
I don't understand why the 'special ribbon entries' are not translated, and the module ribbon buttons are. I remember that @JorisGoosen knew why. Maybe it is better to wait for his return of his holidays.

JorisGoosen:
Thanks for making this @shun2wang but Ive made another PR for this at https://github.com/jasp-stats/jasp-desktop/pull/5653/
That one obviates the need for destroying and recreating the buttons. 
And also doesnt lead to possible problems with (re)loading the modules etc. Which could lead to problems with development modules, it also feels a bit more straightforward this way.

PR #5636, opened 2024-08-05
Fix max column

### LINKED ISSUE #2837: [Bug]: Failed to import data big data file
### JASP Version

0.19.0

### Commit ID

_No response_

### JASP Module

Unrelated

_No response_

Windows 11

### Bug Description

Failed to import data from .sav and .csv file. There are >1500 attributes (columns).
I reduced the characters to 967, import from a .csv file, save as Research result DB_am_1d.jasp, but the program crashes during startup, reports that the file location is bad.
There are 444 entries. What's the problem that I can't run Jasp? I can send the data file.

### Expected Behaviour

Failed to import data
[Tyrimo rezultatu DB_am_1d.zip](https://github.com/user-attachments/files/16335908/Tyrimo.rezultatu.DB_am_1d.zip)

### Steps to Reproduce

1. Sumažinau požymių iki 967
2.
3.
...

### Log (if any)

_No response_

### Final Checklist

RensDofferhoff:
Should fix https://github.com/jasp-stats/jasp-issues/issues/2837
But loading is still very slow

### DIFF
--- a/Desktop/data/importers/csv/csv.cpp
+++ b/Desktop/data/importers/csv/csv.cpp
@@ -402,6 +402,7 @@ void CSV::determineNumRows()
 
 		if (i >= _utf8BufferEndPos - 1)
 		{
+			_utf8BufferEndPos = 0;
 			bool success = readUtf8();
 			if (success)
 				i = -1;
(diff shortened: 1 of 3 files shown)


shun2wang:
Expend max_column in sqlite maybe not the best solution but yes it re-compiled will work as a temporary solution. so we can still track it in https://github.com/jasp-stats/INTERNAL-jasp/issues/2421#issuecomment-2097752605

Also this fix only works on macOS and Windows because we don't use conan on Linux? so a document and buildbot(for flatpak) update would be great.

RensDofferhoff:
I'm adding a sqlite compile step to org.jaspstats.JASP separately now.

PR #5638, opened 2024-08-15
feat: implement importing excel (.xls .xlsx) files

### LINKED ISSUE #2134: [Feature Request]: Open Excel (.xls/.xlsx) data files
### Description

Currently, only able to open .CSV files

### Purpose

Avoid having to export data files from .XLSX to .CSV, which removes all formulas, formatting, etc.

### Use-case

Opening .xlsx files directly, like JAMOVI does, saves time and reduces errors

Often, we do a lot of data processing in Excel (computing variables etc.). That file then has to be exported to CSV to be opened in JASP. Then, any time we need to make a change, we have to update the Excel file, re-export, then update the JASP file.

_No response_

ability to directly open .xlsx Excel files

None

### Additional context

_No response_

shun2wang:
Close: https://github.com/jasp-stats/jasp-issues/issues/2134
Close: https://github.com/jasp-stats/INTERNAL-jasp/issues/1390
Close: https://github.com/jasp-stats/INTERNAL-jasp/issues/412

JASP can now import excel files from first workbook and the first sheet by default.

I introduced a third-party library [freexl](https://www.gaia-gis.it/fossil/freexl/index) to parse the excel files. but a known limitation is that you cannot import files with more than 705 columns (this is a bug of freexl), I made a patch to lifte this limitation temporarily.

- [x] init implement import excel files (on Linux)
- [x] work for Windows
- [x] work for macOS (someone could help?)
- [x] update development document


boutinb:
What nice!!! I'll try it tomorrow on Mac.

boutinb:
I wonder why you did not use conan to load the freexl library.
I have done this for Mac (see my last commit), and it works perfect. we don't need to copy the freexl.h file in this way.
Can you try this on Windows and Linux?

conanfile.txt
```diff
@@ -18,6 +18,7 @@ brotli/1.0.9
 sqlite3/3.46.0
 gmp/6.3.0
 mpfr/4.2.1
+freexl/2.0.0
```
shun2wang, line 22: official FreeXL has a bug, so we using a patch version from my fork.

shun2wang:
> I wonder why you did not use conan to load the freexl library. I have done this for Mac (see my last commit), and it works perfect. we don't need to copy the freexl.h file in this way. Can you try this on Windows and Linux?

because it cannot read more than 700 cols Excel files so I made a patch and build it manually. see my first comment.

see: https://github.com/shun2wang/freexl

shun2wang:
>see: https://github.com/shun2wang/freexl

Do we maintain a repo of JASP, I am just using my fork for now. @boutinb

boutinb:
But if I see your PR, you are using a freeXL version set in rtools.
We could use your freeXL version on GitHub and compile it, but for maintenance reason (JASP depends on many libraries, and its build is becoming quite complex) I would prefer to lie on conan which takes care on portability and compiling issues.
That the current freeXL cannot load more than 700 columns is not a big issue I find. I would ask freeXL to merge your fix (maybe you did it already), so that the next version will include your fix.

shun2wang:
>But if I see your PR, you are using a freeXL version set in rtools.

Yes, we using rtools msys2 build toolchain to build it on Windows and the `ReadStat` also build manually by this way. maybe @JorisGoosen know more details here.

>That the current freeXL cannot load more than 700 columns is not a big issue I find. I would ask freeXL to merge your fix (maybe you did it already), so that the next version will include your fix.

We have received complaints from users about multi-column importing, and Rens has applied a fix temporarily for that. Yes I had filed bugs to that project but I dont think it will be merged for quite some time, because freeXL is updated very infrequently to ensure stability, and the last release was a year ago.

Your concerns are valid, ~~I would try to find a way to apply patches manually from conan instead of waiting for the official master branch release.~~

EDIT: no, I used a dev branch of FreeXL which including some bugfix important. I think maybe build it manually is best choice for now.

JorisGoosen:
I agree with Bruno that adding more manual stuff to the build is not very maintainable...
And Im already unhappy with the fact that for some systems we need to manually build readstat on some systems.

> > But if I see your PR, you are using a freeXL version set in rtools.
> 
> Yes, we using rtools msys2 build toolchain to build it on Windows and the `ReadStat` also build manually by this way. maybe @JorisGoosen know more details here.
> 
> > That the current freeXL cannot load more than 700 columns is not a big issue I find. I would ask freeXL to merge your fix (maybe you did it already), so that the next version will include your fix.

I agree that staying under 700 columns until they merge your fix is not at all a problem.
Just making it possible to import from xlsx files etc is going to be a highly appluaded feature in any case.
So very nice you made it.

> We have received complaints from users about multi-column importing, and Rens has applied a fix temporarily for that. Yes I had filed bugs to that project but I dont think it will be merged for quite some time, because freeXL is updated very infrequently to ensure stability, and the last release was a year ago.
>
> Your concerns are valid, ~I would try to find a way to apply patches manually from conan instead of waiting for the official master branch release.~
> 
> EDIT: no, I used a dev branch of FreeXL which including some bugfix important. I think maybe build it manually is best choice for now.

What bug was fixed?
Maybe it is ok to just use the version with bug that is in conan?

I will now do an indepth review of your code ^^

JorisGoosen:
Also, if we *are* going to use our modified version I would like to move that code to jasp-stats and use that one.

JorisGoosen:
Already a thing: in qml and c++ files we use tabs not spaces. So perhaps you could clean up your indentation to be tabs of 4 instead of the spaces?

Edit: I see that you do use tabs in a lot of places, just not everywhere?

Desktop/data/importers/excel/__freexl.h
```diff
@@ -0,0 +1,736 @@
+// fetch from: https://www.gaia-gis.it/fossil/freexl/dir?ci=tip
```
JorisGoosen: Cant you just add freexl to https://github.com/jasp-stats/jasp-desktop/blob/3d01a0986f4485a19874a16426174ecfcb7342a6/Desktop/CMakeLists.txt#L92 and then do a normal include?

Desktop/data/importers/excel/excel.cpp
```diff
@@ -0,0 +1,118 @@
...
+using namespace std;
+
+Excel::Excel(const string &locator)
+{
+	_path = locator;
+}
+
+void Excel::open()
+{
+	_fileSize = Utils::getFileSize(_path);
```
JorisGoosen: You could also use `QFileInfo` for this.

Desktop/data/importers/excel/excel.cpp
```diff
@@ -0,0 +1,118 @@
...
+		throw runtime_error("Could not access file");
+
+	if (_fileSize == 0)
+		throw runtime_error("File is empty");
+}
+void Excel::openWorkbook() 
+{
+	QString xlsFilePath = QString::fromStdString(_path);
+	const char* utf8Path = xlsFilePath.toUtf8();
+	QString extension = QFileInfo(xlsFilePath).suffix().toLower();
```
JorisGoosen:
```suggestion
	QString xlsFilePath = tq(_path);
	const char* utf8Path = _path.c_str(); //But it would be better to just use _path.c_str() directly if you need it. It is in utf8 in any case.
	QString extension = QFileInfo(xlsFilePath).suffix().toLower();
```

Desktop/data/importers/excel/excel.h
```diff
@@ -0,0 +1,54 @@
...
+//
+
+#ifndef EXCEL_H
+#define EXCEL_H
+
+#include <vector>
+#include <string>
+#include <stdint.h>
+
+#include <freexl.h>
```
JorisGoosen, line 24: So here you include "freexl.h" from the include paths, but why then add `__freexl.h' ?

Desktop/data/importers/excel/excelimportcolumn.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+ExcelImportColumn::ExcelImportColumn(ImportDataSet *importDataSet, std::string name, long reserve) : ImportColumn(importDataSet, name)
+{
+	_data.reserve(reserve);
+}
+
+ExcelImportColumn::~ExcelImportColumn()
+{
+	JASPTIMER_SCOPE(ExcelImportColumn::~ExcelImportColumn());
+	_data.clear();
```
JorisGoosen: This not really necessary, as the vector will get cleaned up in any case by the destructor.

Desktop/data/importers/excelimporter.cpp
```diff
@@ -0,0 +1,116 @@
...
+//
+
+#include "excelimporter.h"
+#include "data/importers/excel/excel.h"
+#include "data/importers/excel/excelimportcolumn.h"
+#include <string>
+#include <QFileInfo>
+#include <QDebug>
+
+using namespace std;
```
JorisGoosen: Im not a big fan of `using namespace std;` we still have it in some older code.
But adding `std::` in front of functions is not so bad and it avoids possible misdirections for functions like `max()` or `min()` and other, perhaps unexpected, collisions.

And I see that `std::` is being used anyway so it shouldnt be a problem to remove it anyway.

Desktop/data/importers/excelimporter.cpp
```diff
@@ -0,0 +1,116 @@
...
+				} else 
+				{
+					try 
+					{
+						if (std::to_string(std::stoi(colName)) == colName)
+							colName = "V" + colName;
+					} catch (...) {}
+				}
+				if (std::find(colNames.begin(), colNames.begin() + i, colName) != colNames.begin() + i)
+					colName += "_" + std::to_string(i + 1);
```
JorisGoosen:
```suggestion
				string colName = colNames[i];
				if (colName.empty()) 
					colName = "V" + std::to_string(i + 1);
				else if(ColumnUtils::isIntValue(colName))
					colName = "V" + colName;

				//What is the purpose of the `find` here? Maybe explain in a comment here
				if (std::find(colNames.begin(), colNames.begin() + i, colName) != colNames.begin() + i)
					colName += "_" + std::to_string(i + 1);
```

Desktop/data/importers/excelimporter.cpp
```diff
@@ -0,0 +1,116 @@
...
+				importColumns.push_back(new ExcelImportColumn(data, colName, rows - 1));
+			}
+		}
+		else 
+		{
+			for (int i = 0; i < importColumns.size(); ++i) 
+			{
+				importColumns[i]->addValue(i < lineValues.size() ? lineValues[i] : "");
+			}
+		}
```
JorisGoosen:
```suggestion
		else 
			for (int i = 0; i < importColumns.size(); ++i) 
				importColumns[i]->addValue(i < lineValues.size() ? lineValues[i] : "");
```

Desktop/data/importers/excelimporter.cpp
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	for (ExcelImportColumn* col : importColumns) 
+	{
+		data->addColumn(col);
+	}
```
JorisGoosen:
```suggestion
	for (ExcelImportColumn* col : importColumns) 
		data->addColumn(col);
```

Desktop/data/importers/excelimporter.cpp
```diff
@@ -0,0 +1,116 @@
...
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+//
+
+#include "excelimporter.h"
+#include "data/importers/excel/excel.h"
+#include "data/importers/excel/excelimportcolumn.h"
+#include <string>
+#include <QFileInfo>
+#include <QDebug>
+
```
JorisGoosen, line 25:
```suggestion
#include "columnutils.h"

```

Desktop/data/importers/excel/__freexl.h
boutinb: My commit to use conan was not complete: I just renamed freexl.h to __freexk.h, to be sure that this file won't be used. I did not remove it, because I wanted Shun to test first whether the conan version compiles and works also on Windows. The CMakefile.txt, Install.cmake and Libraries.cmake must be changed also.

JorisGoosen: But I also dont see why this file would be needed without conan?

I know I did something like this with readstat but that was to make it play nice with c++ I think

shun2wang:
Thanks for your reviews! OK let's use conan version to make build easier. and ignore 700 cols limitation. then I will following the official release. 🍻

Desktop/data/importers/excel/__freexl.h
shun2wang: Because when using cmake to reference `freexl.h` in the msys2(Rtools) include path, it will automatically link some other standard libraries, which will conflict with the msvc compiler and linker during compilation. However, once we use conan, we do not need to introduce this header.

boutinb (approved):
It works really well on Mac.
I think we can merge this for 0.19.2

JorisGoosen:
I agree!

Very nice @shun2wang  😃

PR #5651, opened 2024-09-03
Speed up loading many columns

JorisGoosen:
add some pragmas to speed it all up
dont add each column separately but just create the table from scratch, making sure that the filter-column is still first


RensDofferhoff (approved):
Huge improvement!

PR #5656, opened 2024-09-09
bugfix: replace newlines with _ in CSV cell values

shun2wang:
Other data file will not have this problems because newlines in values were not allowed.

I think `_` is better than a space, because space invisible sometimes, so that we should also change: https://github.com/jasp-stats/jaspColumnEncoder/blob/3425813804558ea0246351e3b361c19fffcf73f4/stringutils.h#L188 ?

### DIFF
--- a/Desktop/data/importers/csv/csv.cpp
+++ b/Desktop/data/importers/csv/csv.cpp
@@ -536,7 +536,7 @@ bool CSV::readLine(vector<string> &items)
     for (size_t index = 0; index < items.size(); index++)
 	{
         string item = items.at(index);
-		boost::algorithm::replace_all(item, "\n", "_"); // so we should not newlines in values right?
+		boost::algorithm::replace_all(item, "\n", " "); // so we should not allow newlines in values right?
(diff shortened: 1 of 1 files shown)


JorisGoosen:
But what is the problem with newlines in values?
Right now they are read and used as they are.

While I agree users shouldnt really do that I think replacing them by '_' is a bit brusque.
Is there some bug you are trying to solve with this?

shun2wang:
The vaule with newlines will overflow from cells after imported CSV, perhaps adjusting the cell height is an option, but I feel that there is a risk of line breaks appearing in the data anyway, which might cause confusion on the unpredictable R code side. I don't know, but overflowing the data view is an obvious problem right now.

shun2wang:
I mean:
image

After:
image


Even we dont allow user input 'newline' inside cells right:-)

JorisGoosen:
Well, that is why I replaced the newlines with spaces :p

JorisGoosen (approved):

PR #5661, opened 2024-09-11
syntaxhighlighting improvements

JorisGoosen:
adds support for highlighting columnnames from JASP (including names with spaces that are allowed within JASP) also makes sure to do the comment rule only at the end to make sure comments are entirely commented to avoid confusing users

Looks like:
image
image


shun2wang (commented):
Cols highlight looks good to me.

But I dont think the output of R console window should be highlighted, output may including dataframe, tables...it shouldn't be applied right?

boutinb (approved):

PR #5662, opened 2024-09-12
remove WAL from sqlite to avoid bugs

JorisGoosen:
also log how long it took to load or synch a file

Why remove WAL?
It was faster yes, but often the engine would not have the up-to-date info and then crash. Adding walcheckpoints all over the place didnt seem to help

the slowdown is a bit noticeable with >3000k columns, but given that we didnt support them before shouldnt be a big problem.

For speed comparisons:
```
        WAL     NO-WAL
3KC      4s        10s
10KC    28s        54s
200KR   30s        30s
1MR    146s       147s
1KR5KC 347s       356s
```
Where the "R" sets have 2 columns and the "C" sets have a single row. As can be seen, there *is* a delay for 1K cols and above, but it seems constant. And any fixes for the synching-db problem with the engine most likely includes adding new delays... So lets just turn it off

### DIFF
--- a/CommonData/databaseinterface.cpp
+++ b/CommonData/databaseinterface.cpp
@@ -1589,8 +1589,6 @@ void DatabaseInterface::create()
 	}
 	else
 		Log::log() << "Opened internal sqlite database for creation at '" << dbFile() << "'." << std::endl;
-
-	dbStartUpPragmas();
 	
 	transactionWriteBegin();
 	runStatements(_dbConstructionSql);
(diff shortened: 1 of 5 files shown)


boutinb (approved):

JorisGoosen:
Also, the bug that is being solved is this:
with WAL enabled:
- open a datafile
- close it
- open a datafile (for instance debug.csv)
- see filtercrash
- rerun filter with no differences (like, add a space or something), see how it works now...

RensDofferhoff (approved):

PR #5635, opened 2024-08-03
dependencies: upgrade marked.js to fix markdown render

shun2wang:
Fix: https://github.com/jasp-stats/jasp-issues/issues/2847

It's a bug of [marked.js](https://marked.js.org/) and was fixed from [v5.0.4](https://github.com/markedjs/marked/releases/tag/v5.0.4)  so let's just upgrade it.

PR #5663, opened 2024-09-15
Post highlighter tweaks

shun2wang:


JorisGoosen:
Why no highlight in the output?

It also prints what users entered, could be nice to see highlighting there no?
(I mean, I doubted if I should add it but I figured it was nicer with than without but I can def be convinced)

And the change for the JS: I have no idea what you changed there or why and cnnot guess from the code alone. Could you explain for my benefit? ^^

shun2wang:
1. for the R command output window will print dataframe, tables, error code...which should not be "highlight" cause complex type. we should only highlight r code in "input" window.
image
A bit strange, but no big deal...I'm not very persistent in my opinion haha

2. that was mistake I introduced from https://github.com/jasp-stats/jasp-desktop/pull/5635, this was a js module but we should use js script directly, a error you can see in the devtools😉

shun2wang:
OK, as we talked in mattermost, let's add it back.

JorisGoosen (approved):

PR #5666, opened 2024-09-17
Fix bug where having empty values messes up labeledits

JorisGoosen:
This is done by resetting labelsTemp

@kylelang this solves the problem you mentioned in the meeting?

For other (future) reviewers: 
Problem is that when you have a column like 'contBinom' and you change some values (in data editing) to a missing value (like NA) then editing the displayvalue of the label in the labeleditor will not update properly.
And from that point on further editing will make a mess of the data.

https://github.com/jasp-stats/INTERNAL-jasp/issues/2648

### DIFF
--- a/Desktop/data/datasetpackage.cpp
+++ b/Desktop/data/datasetpackage.cpp
@@ -250,8 +250,9 @@ QModelIndex DataSetPackage::index(int row, int column, const QModelIndex &parent
 				
 			case dataSetBaseNodeType::column:
 			{
-				Column * col = dynamic_cast<Column*>(parentNode);
-				pointer = dynamic_cast<const void*>(col->labels().size() > row ? col->labels()[row] : col->labelDoubleDummy());
+				Column	* col	= dynamic_cast<Column*>(parentNode);
+				Label	* lab	= col->labelByIndexNotEmpty(row);
+				pointer			= dynamic_cast<const void*>(lab ? lab : col->labelDoubleDummy());
 				break;
 			}
 				
--- a/CommonData/column.cpp
+++ b/CommonData/column.cpp
@@ -1213,7 +1213,7 @@ std::map<double, Label*> Column::replaceDoubleWithLabel(doublevec dbls)
 		if(!std::isnan(_dbls[r]) && doubleIntIdMap.count(_dbls[r]))
 			_ints[r] = doubleIntIdMap[_dbls[r]]; 
 	
-	dbUpdateValues();
+	dbUpdateValues(false);
 	
 	return doubleLabelMap;
 }
@@ -1226,7 +1226,6 @@ Label *Column::replaceDoublesTillLabelsRowWithLabels(size_t row)
 	if(labelByIndexNotEmpty(row))
 		return labelByIndexNotEmpty(row);
 	
-	bool		labelsTempUpToDate = _revision == _labelsTempRevision; //If so we can just update it at the end. We dont not yet need to recreate them
 	doublevec	dbls;
 	double		dbl;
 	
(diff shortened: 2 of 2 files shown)


kylelang:
I was just about to hit the submit button on jasp-stats/INTERNAL-jasp#2648, when you submitted this PR.

kylelang:
I don't think this PR fully fixes the bug. If I build JASP from your feature branch, recoding factor levels other than the first seems to work correctly now, but recoding the first factor level still breaks.

The types of variables that trigger the issue seem informative:

- The problem only arises when the column contains numeric-looking values and missing values represented as NA in the CSV.
- Missing values coded as NA don't cause problems when the rest of the values in the column are character labels.
- Numeric-looking columns with missing values coded some other way don't cause problems.

I spent some time in the debugger trying to locate the source of the issue.

The problem seems to stem from the way the `_labels` field of a `Column` object is populated and used.

In the problematic case, the call to `Column.setValues()` at [datasetpackage.cpp:1471](https://github.com/JorisGoosen/jasp-desktop/blob/8b57ec7807c96ce8ced16465853eeda975e212c5/Desktop/data/datasetpackage.cpp#L1471) creates a length-one entry in the `_labels` field of `column` where "NA" is recorded as the only label for the variable.

Since the other values are numeric-looking, the data ingress functions seem to assume that we'll be able to automatically generate labels from the numeric values when necessary. However, some of the downstream functions that trigger when editing labels seem to assume that either every value has a label or no labels exist. So, our Franken-variable breaks these functions.

As near as I can tell, the critical failure happens at [datasetpackage.cpp:850](https://github.com/JorisGoosen/jasp-desktop/blob/8b57ec7807c96ce8ced16465853eeda975e212c5/Desktop/data/datasetpackage.cpp#L850). The pointer defined here ends up pointing to the `_labels` field for "NA". The `index` object used here is generated from the (correct) row number in the label editor view, but the contents of `_labels` field don't match the label editor view.

kylelang (changes requested):
I don't think this PR fully fixes the bug. If I build JASP from your feature branch, recoding factor levels other than the first seems to work correctly now, but recoding the first factor level still breaks.

The types of variables that trigger the issue seem informative:

- The problem only arises when the column contains numeric-looking values and missing values represented as NA in the CSV.
- Missing values coded as NA don't cause problems when the rest of the values in the column are character labels.
- Numeric-looking columns with missing values coded some other way don't cause problems.

I spent some time in the debugger trying to locate the source of the issue.

The problem seems to stem from the way the `_labels` field of a `Column` object is populated and used.

In the problematic case, the call to `Column.setValues()` at [datasetpackage.cpp:1471](https://github.com/JorisGoosen/jasp-desktop/blob/8b57ec7807c96ce8ced16465853eeda975e212c5/Desktop/data/datasetpackage.cpp#L1471) creates a length-one entry in the `_labels` field of `column` where "NA" is recorded as the only label for the variable.

Since the other values are numeric-looking, the data ingress functions seem to assume that we'll be able to automatically generate labels from the numeric values when necessary. However, some of the downstream functions that trigger when editing labels seem to assume that either every value has a label or no labels exist. So, our Franken-variable breaks these functions.

As near as I can tell, the critical failure happens at [datasetpackage.cpp:850](https://github.com/JorisGoosen/jasp-desktop/blob/8b57ec7807c96ce8ced16465853eeda975e212c5/Desktop/data/datasetpackage.cpp#L850). The pointer defined here ends up pointing to the `_labels` field for "NA". The `index` object used here is generated from the (correct) row number in the label editor view, but the contents of `_labels` field don't match the label editor view.

JorisGoosen:
```
1. Load the attached dataset.
2. Try to change the label for column = "b" value = "1" to some character string.
```

[toy.csv](https://github.com/user-attachments/files/17095499/toy.csv)

JorisGoosen:
Fixed it

Desktop/data/datasetpackage.cpp
kylelang, line 1: This change seems to fix the remaining problems with the label indexing.

kylelang (approved):
As far as I can tell, the bug looks fixed. Thanks!

kylelang:
> Fixed it

Looks good. Thanks!

PR #5667, opened 2024-09-17
Preload data

JorisGoosen:
Also merge: https://github.com/jasp-stats/jaspColumnEncoder/pull/9 and https://github.com/jasp-stats/jaspBase/pull/152
Ive also made some small changes to jaspDescriptives: https://github.com/jasp-stats/jaspDescriptives/pull/314

for jasp-stats/INTERNAL-jasp#2583

**Don't forget** `git submodule update` ;)

This adds support for automatic preloading of data for analyses. By default Ive turned it *off* to ease the transition but ideally that would be *on*. This is done through the `preloadData` property on `Description` and `Analysis` in Description.qml.
If disabled it will act like 0.19(.1). Basically adding an "optionname.types" and flattening "optionname" option into just the names.

It encodes all the columnname-options to their columnname+requested type if they have one. (Which is the case for the variableslists etc). And then gives the analysis a flat list of encoded columnnames that will refer to the column in the dataset with the specific version of the data requested.

There is now also jaspRcpp support for the columnencoder where it can give the type back. This has not been implemented all the way to R though.

The confusing "Keeping types" and "Converting types" button in computed column window has now been removed.
This because the columns in filterconstructor and computed columnconstructor now have support for transforming their datatype locally (including showing the preview a la variables list).
This is also available in R-code as "V1.ordinal" etc.
The syntaxhighlighting will also help indicate this.


JorisGoosen:
JAGS en SEM are not supported properly

vandenman:
Specifically, no support for:

- `.readDataSetToEnd(all.columns = TRUE)`.
- analyses whose columns are determined by parsing a TextArea.

JorisGoosen:
Im currently finishing support for changing type with dropdown in filterconstructor.

Then I need to add a property to modules and analyses that states "I can handle datasets being passed as an argument" aka "preloadData" or something.

And further I want to remove the "keeping types" and "converting types" cause it really isnt necessary anymore.

JorisGoosen:
All the changes have been made, Ill update the PR description to reflect the changes.

boutinb:
It works nearly perfect. But now if I close the Variable Window, it says nearly always "Would you like to save your changes to the Computed Column":
. If I just add a new (not computed) column, and I click on the red cross icon, I get this message.
. If I open a computed column, change something and compute the column, if I close the variables window, I get this message. I can the 'Apply' or 'Discard' : if I open the Variable Window and close it just afterwards I get the same message.

The computed column help should be updated.

JorisGoosen:
The help is updated and the apply thing should now only bother you when you made changes ;)

boutinb (approved):

boutinb:
@vandenman Do you have time to review this PR today?

vandenman (approved):
I have not looked at the code in depth, but I did try this out and it build and worked fine.

PR #5671, opened 2024-09-19
Data entry redux

JorisGoosen:
Speed up the filtered data entry model massively by not making gigantic jsons for everything.

This will add  "named filters" to jasp, which can be linked to jaspControls.
We could later extend on this functionality for analyses etc.

It also will make the filtered date entry model and r-side directly interface with the dataset and thus force the component to always be linked to a computed column.

### DIFF
--- a/Common/enginedefinitions.h
+++ b/Common/enginedefinitions.h
@@ -8,7 +8,7 @@
 /// Using enumutilities templates to make sure we can easily and quickly go from enum -> string -> enum for json communication
 ///
 
-DECLARE_ENUM(engineState,			initializing, idle, analysis, filter, rCode, computeColumn, moduleInstallRequest, moduleLoadRequest, pauseRequested, paused, resuming, stopRequested, stopped, logCfg, settings, killed, reloadData);
+DECLARE_ENUM(engineState,			initializing, idle, analysis, filter, filterByName, rCode, computeColumn, moduleInstallRequest, moduleLoadRequest, pauseRequested, paused, resuming, stopRequested, stopped, logCfg, settings, killed, reloadData);
 DECLARE_ENUM(performType,			run, abort, saveImg, editImg, rewriteImgs);
 DECLARE_ENUM(analysisResultStatus,	validationError, fatalError, imageSaved, imageEdited, imagesRewritten, complete, running, changed, waiting);
 DECLARE_ENUM(moduleStatus,			initializing, installNeeded, loading, installModPkgNeeded, readyForUse, error);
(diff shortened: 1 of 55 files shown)


koenderks:
Don't forget the classical workflow (I see now that you adjusted only the Bayesian workflow?)

JorisGoosen:
Its a draft and non-functional yet, I just opened it to have an easy way to see all the changes.

I still need to integrate the computed column into the dataentry thing all the way. Hopefully It'll be done next week

JorisGoosen:
Only jaspAudit has some changes related to this PR

JorisGoosen:
@koenderks @boutinb this is ready for review

for koen: I think it still works, maybe you can confirm :p
Should be much quicker for large datasets alround now.

koenderks:
I'm currently trying this but I get:

image

I don't think this has to do with your changes, right? It is only in the dev module, not the non-dev audit. I can only test this once I can access the workflow, so I'll wait a bit.

JorisGoosen:
You can build `development` right @koenderks ?
If you checkout this jasp-desktop branch (dataEntryRedux) and run git submodule update it ought to have the changes in jaspAudit as well. No need for development modules if you can build jasp.

This error of yours looks like you might need to clean up the buildfolder a bit?
(Unless you can build and run `development` with audit because then there is some other problem)

koenderks:
Nice, it works fine. However, for a brief moment (long enough to make a screenshot however) you get this visual error:

image

Furthermore, I notice that the `selected` column is immediately made and visible but the `auditResult` column is not?

image

Works fast with dataset of 100000 rows!

JorisGoosen:
> Nice, it works fine. However, for a brief moment (long enough to make a screenshot however) you get this visual error:

This is caused by the way Audit right now sets the filter-code and "auditResult" columnname only when you click "Fill Table". Only then is `auditResult created`. But `selected` isnt filled yet, so the filter returns nothing... So the error/warning.
Then on such an error I simply let it run again.
Now `selected` is filled and thus the next filter-return it shows stuff in the table.

This could be avoided by filling `selected` earlier.

Also, the initial values of the component are taking from `bookValue` but because they should only be set on the selected rows I have to wait until the filter returns to know which rows. So:
- Fill table, creates auditResults and sets filter to something like "selected > 0"
- Filter runs, but selected is empty so gets nothing. Mentions this to user.
- analysiscode runs, fills selected but gets no values from performAudit.
- filter runs again because of the earlier not-getting-anything
- now table knows which rows it should show. 
- first time it gets a usable filter it takes "initialValues" and sets it to the rows
- runs analysiscode again and now it gets all the info
 
> Furthermore, I notice that the `selected` column is immediately made and visible but the `auditResult` column is not?

Yes, the computed column is now created by the FilteredDataEntry thing. If you want it created earlier you should set `colName` earlier.

> Works fast with dataset of 100000 rows!

Right? :D

boutinb (approved):
Just added a commit to set the connections in the right object.

PR #5672, opened 2024-09-20
Dont loop through the labels to avoid collisions

### LINKED ISSUE #2910: Can't import TXT file
### JASP Version

0.19.0

### Commit ID

_No response_

### JASP Module

Unrelated

_No response_

Windows 11

### Bug Description

Whenever I try to import a file it never opens. About halfway it will get stuck loading and no matter how long I let it run it won't open the file. I've tried uninstalling and installing JASP again, I've tried the repair function, I've tried different kind of TXT files and rebooting my device but nothing seems to work.
See the image for more clarification:

### Expected Behaviour

It should've opened the file(s).

### Steps to Reproduce

1. Open jasp
2.select open
3. Select a file
4. the file never stops loading
...

### Log (if any)

_No response_

### More Debug Information

_No response_

### Final Checklist

JorisGoosen:
Instead a map is used to know what labels have what original value and display text, as they are uniquely defined by this. Before was naively looping through the labels to check and merge them in a preceding step. Same goes for intsId, it is simply tracked now instead of found by looping.

Both loops are now lookups at the cost of a little maintenance.

Fixes https://github.com/jasp-stats/jasp-issues/issues/2910 (extremely slow loading of semibig text file)

Basically now any LLL (label lookup loop) I could see in Column has been replaced by a map, which makes the linked file in the issue actually usable. 
As far as I can see the data is still read correctly.

### DIFF
--- a/CommonData/column.cpp
+++ b/CommonData/column.cpp
@@ -493,10 +493,8 @@ columnType Column::setValues(const stringvec & values, const stringvec & labels,
 	bool	onlyDoubles = true, 
 			onlyInts	= true;
 	
-	//Make sure we have only 1 label per value and display combo, because otherwise this will get too complicated
-	if(labelsMergeDuplicates() && aChange)
-		(*aChange) = true;
-	
+	//Weve already made sure we have only 1 label per value and display combo, because otherwise this will get too complicated
+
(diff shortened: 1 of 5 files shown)


JorisGoosen:
Maybe this will also allow us to remove the speedupworkarounds from labelsTemp... Lets see

boutinb (approved):

PR #5674, opened 2024-09-21
chore: add .exe installer on Windows

shun2wang:
Let's throw the msi away!

Now we will have a quick, simple and intuitive .exe installer:
图片

图片

- Both install for Administrator and per-users
- Multiple language chooses while installing
- Installing from the command line
- Better scalability and maintainability

- [ ] Set for build bot
- [ ] Uninstal clean up
- [ ] Maybe shour run with some parameters after installed.
- [ ] Signature


JorisGoosen:
We were thinking of just throwing away msi and using only msix + zip.

What is the advantage of also doing this?

shun2wang:
Yes, my main goal is to deprecate MSI.

>What is the advantage of also doing this?

Maybe custom install path by users( not everyone want install to C disk).

JorisGoosen:
You can just install windows store to a different drive too...

Deprecating msi is a great idea though ;)

But going back to installshield... We deprecated that in favour of msi like 6 years ago :p

tomtomme:
At my university they distribute the msi via https://opsi.org - will it be possible to download the MSIX without a microsoft account and integrate it in software like opsi? I cannot test this on my linux machines because the microsoft store blocks downloads even if I am logged in...

If not, the msi or this new exe stuff might still be needed to deploy jasp on universities / companies. 
The zip is not a good solution for this I think, because it lacks file-type / mime integration etc.

shun2wang:
>But going back to installshield... We deprecated that in favour of msi like 6 years ago :p

Haha history always repeats itself. 

But I think we need to consider the module store in the future, maybe we need to select the analysis module at the installation stage. This is the most important. Based on .exe it can be installed selectively, or online (I mean the installation stage, installing/update modules from within JASP is another matter)

JorisGoosen:
Good points both, and well, I dont particularly like msi so lets see :p

shun2wang:
Rebasing to resolve merge conflicts requires a lot of steps, closed temporarily.

JorisGoosen:
You could also turn it into a draft?

PR #5676, opened 2024-09-24
Computed column fixes

### LINKED ISSUE #2927: [Bug]: Computing data using drag-and-drop
### JASP Version

0.19.0

### Commit ID

_No response_

### JASP Module

Other

Creating new variable from existing data

Windows 11

### Bug Description

In the previous version of JASP, we can easily create indicator variable using drag-and-drop, for example, if I want to create an indicator variable "young" based on my existing data "age", I can create a new column, make it have data type **Scale**, and drag-and-drop "age<30", the computed column will **take value of 1 and 0**. Now this functionality does not exist any more, and instead shows **"TRUE" and "FALSE"** in the computed column which are not usable in any subsequent regression analyses. How do I change this without resorting to R code?

### Expected Behaviour

Use drag-and-drop to compute indicator variable easily and show them as numerical data that can be used for analysis directly. This is the main reason that I am using JASP for my class that it **does not require coding**, but now it is not usable for an entry-level data analysis course. At the very least tell people how to change these TRUE/FALSE variables into 1/0 with proper documentation.

### Steps to Reproduce

1. add a column
2. compute with drag and drop
3. put in some logical statement
...

### Log (if any)

_No response_

### More Debug Information

_No response_

### Final Checklist

JorisGoosen:
Fixes jasp-stats/jasp-issues#2927 (getting TRUE/FALSE instead of 1/0) and https://github.com/jasp-stats/INTERNAL-jasp/issues/2653 (computed columns dont work in development...)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -81,3 +81,6 @@ Desktop/utilities/simplecryptkey.h
 Desktop/modules/activemodules.h
 Common/appinfo.cpp
 QMLComponents/utilities/appdirs.h
+Engine/jaspBase_distributionSamplers.h
+Engine/jaspBase_friendlyConstructorFunctions.h
+Engine/jaspBase_transformFunctions.h
--- a/Engine/CMakeLists.txt
+++ b/Engine/CMakeLists.txt
@@ -15,6 +15,10 @@ if(LINUX)
 
 endif()
 
+make_includable(jaspBase/R/distributionSamplers.R			jaspBase_distributionSamplers.h)
+make_includable(jaspBase/R/friendlyConstructorFunctions.R	jaspBase_friendlyConstructorFunctions.h)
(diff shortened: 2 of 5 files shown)


boutinb:
This does not work for Ordinal or Nominal computed column: if you change the computed column to Ordinal or Nominal it still gives NA's. If you also make directly an ordinal computed column (with the + button in Analysis mode), it also gives NA's.
Screenshot 2024-09-24 at 15 18 03

JorisGoosen:
Disable "Converting types", or in any case, make sure all the types make sense

I will remove that functionality once https://github.com/jasp-stats/jasp-desktop/pull/5667 gets merged

boutinb (approved):

PR #5677, opened 2024-09-24
Add levels property to VariablesList

boutinb:
This was a request from @koenderks 
You can now asks the levels of variable in a (single) VariablesList:
```
		DropDown
		{
			name: "myDrop"
			label: "Labels with " + variableWithLabels.levels.length + " values"
			values: variableWithLabels.levels
		}
```
It is then also possible to set a property according to the number of levels:
`property bool enoughLevels: variableWithLabels.levels.length > 2`

### DIFF
--- a/CommonData/column.cpp
+++ b/CommonData/column.cpp
@@ -537,7 +537,7 @@ columnType Column::setValues(const stringvec & values, const stringvec & labels,
 			if(label && !label->isEmptyValue())
 				labelsNonNA.insert(label);
 		}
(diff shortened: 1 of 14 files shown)


JorisGoosen (approved):
Ok, looks good!

Ive made sure it takes empty values into account, both on the column and the workspace. Will now also reset property levels when workspace or column emptyvalues change.

Added a onLevelsChanged `messages.log` to "test variables form" in jaspTestModule for easy debuggin

Its all OK by me now, so merge it at your leisure.

PR #5686, opened 2024-10-08
Remote cellar Mac

RensDofferhoff:
Changes: 

1. adds a remote cellar for rpkgs to jasp module compilation. 
2. add cellar creation script
3. add auto download on cmake run

Notes: not yet active on Windows

I disabled it on windows for now because strange things still happen there.
But this way we can get around blavaan issue on mac.

See  the cellars here
https://static.jasp-stats.org/development/cellars/

### DIFF
--- a/Tools/CMake/JASP.cmake
+++ b/Tools/CMake/JASP.cmake
@@ -124,7 +124,7 @@ if(R_PKG_CELLAR_DOWNLOAD_REMOTE AND R_PKG_CELLAR_PATH STREQUAL "${CMAKE_SOURCE_D
         SET(R_PKG_CELLAR_REMOTE "https://static.jasp-stats.org/development/cellars/cellar_macOS_arm64_latest.tar.gz")
       endif()
     elseif(WIN32)
-      SET(R_PKG_CELLAR_REMOTE "https://static.jasp-stats.org/development/cellars/cellar_Windows_x86_64_latest.tar.gz")
+      # SET(R_PKG_CELLAR_REMOTE "https://static.jasp-stats.org/development/cellars/cellar_Windows_x86_64_latest.tar.gz")
     endif()
   endif()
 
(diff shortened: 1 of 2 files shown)


Tools/createCellarFromRenvCache.py
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/env python3
```
vandenman, line 1: so I get this was perhaps easier to write, but this could also just be an R script that uses libarchive? jaspTools already uses an R interface to libarchive, for example.

RensDofferhoff: What would be the advantage?

Tools/CMake/JASP.cmake
```diff
@@ -73,8 +73,16 @@ set(MSIX_SIGN_CERT_PASSWORD
   CACHE STRING "Password selfsign cert for Nightlies")
 
 set(R_PKG_CELLAR_PATH
+  "${CMAKE_SOURCE_DIR}/cellar/"
+  CACHE STRING "Set the path for an renv package cellar to be used during build phase. Changing this disables the use of a remote cellar")
+
+set(R_PKG_CELLAR_REMOTE
   ""
-  CACHE STRING "Set the path for an renv package cellar to be used during build phase")
+  CACHE STRING "URL to downloand a remote package cellar from (Make it https). If not set a platform default will be set by cmake")
```
JorisGoosen, line 81:
```suggestion
  CACHE STRING "URL to downloand a remote package cellar from (Make it https!). If not set a platform default will be set by cmake")
```

Tools/CMake/JASP.cmake
```diff
@@ -106,6 +114,29 @@ endif()
 
 message(STATUS "CRAN mirror: ${R_REPOSITORY}")
 
+#this handles the cellar download.
+if(R_PKG_CELLAR_DOWNLOAD_REMOTE AND R_PKG_CELLAR_PATH STREQUAL "${CMAKE_SOURCE_DIR}/cellar/" AND NOT EXISTS "${CMAKE_SOURCE_DIR}/cellar/")
+  #set appropriate default remote if needed
+  if(R_PKG_CELLAR_REMOTE STREQUAL "")
```
JorisGoosen, line 120: So yeah, would be nice at a later stage if there is some more IDing of the the cellar downloads.
Something with hashes or so.

JorisGoosen (approved):

Tools/createCellarFromRenvCache.py
JorisGoosen: Then python doesnt become a builddep, although given that buildbot already runs its hardly a problem

PR #5697, opened 2024-10-14
fix renv install on mac

RensDofferhoff:


boutinb (approved):

Tools/CMake/R.cmake
```diff
@@ -838,14 +838,40 @@ message(STATUS "Setting up renv, Rcpp, RInside, and jaspModuleInstaller")
...
+if(APPLE)
+  # Patch renv
+  message(CHECK_START "Patching ${RENV_LIBRARY}")
+  execute_process(
+    COMMAND_ECHO STDOUT
+    #ERROR_QUIET OUTPUT_QUIET
+    WORKING_DIRECTORY ${R_HOME_PATH}
+    COMMAND
+      ${CMAKE_COMMAND} -D
+      NAME_TOOL_PREFIX_PATCHER=${PROJECT_SOURCE_DIR}/Tools/macOS/install_name_prefix_tool.sh
```
vandenman, line 859: @RensDofferhoff just for my understanding, which .so file needed to get patched here? Something from Rcpp of RInside?

RensDofferhoff: See the renv changelog: https://rstudio.github.io/renv/news/index.html

There now exist a renv.so in libs/ when it is build
This needs to be patched and signed so it may be loaded later on.
before this version 1.0.11 it didn't exist.

PR #5699, opened 2024-10-15
If no intialValuesSource, use defaultValue

JorisGoosen:
Implements https://github.com/jasp-stats/INTERNAL-jasp/issues/2661


koenderks (approved):
Seems to work like a charm! It requires https://github.com/jasp-stats/jaspAudit/pull/477 so that the filter is filled before the data viewer is opened.

PR #5704, opened 2024-10-15
DropDown should also be able to set the column type

boutinb:
Move allowed type code from VariablesList to JASPControlList, so that allowedColumns property and all the contraints are also possible to DropDown control.

Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2662


JorisGoosen (approved):
Works smoothly

PR #5710, opened 2024-10-18
Fix preloading data for interaction variables

boutinb:
In Meta Analysis, if you add a variable in predictions, the analysis gives an error that it cannot find this variable.


boutinb:
Small fix needed for Franticek. So I need to merge it now...

PR #5721, opened 2024-10-25
Handle also paired variablesList

boutinb:
The Bain Paired T-Tests is in preloadData mode, and should work.

To test, load the bainPairedSamples JASP file in the Data Library/Bain folder, and refresh it.
Test also Paired T-Tests of the TTests module.

Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2678

### DIFF
--- a/QMLComponents/boundcontrols/boundcontrolmultiterms.cpp
+++ b/QMLComponents/boundcontrols/boundcontrolmultiterms.cpp
@@ -27,11 +27,13 @@ BoundControlMultiTerms::BoundControlMultiTerms(ListModelMultiTermsAssigned* list
 
 void BoundControlMultiTerms::bindTo(const Json::Value& value)
 {
-	BoundControlBase::bindTo(value);
+	Json::Value adjustedValue = _isValueWithTypes(value) ? value["value"] : value;
+
+	BoundControlBase::bindTo(adjustedValue);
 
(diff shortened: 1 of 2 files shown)


koenderks (approved):
Seems to me like it is working correctly

QMLComponents/boundcontrols/boundcontrolmultiterms.cpp
```diff
@@ -71,3 +73,36 @@ void BoundControlMultiTerms::resetBoundValue()
...
+{
+	Json::Value newValue;
+
+	if (_control->encodeValue())
+	{
+		if (_isValueWithTypes(value))
+			newValue = value;
+		else
+		{
+			Json::Value types(Json::arrayValue);
```
JorisGoosen, line 88: So this "else" is *just* for loading pre-0.19.2 versions no?

JorisGoosen: If so maybe add a comment here a la 
```suggestion
			// Else we are loading from a jasp version before "preloadData" was on or var.types were added to the options
			Json::Value types(Json::arrayValue);
```

JorisGoosen (approved):

JorisGoosen:
It all works fine and the code seems good but I think maybe a comment might be warranted in the code where I mention this ;)

PR #5723, opened 2024-10-29
Encode rowValue of a componentList when it contains variables

boutinb:
The MixedModels analyses do not work because the random effects ComponentsList has for source the "Random effects grouping factors" Variables List. So the rowValue of the componentsList (the key of a row in the ComponentsList) is a variable (with possible interactions), and it must be also encoded with the right type.

For this the Term object should contain not only the type when it is a single variable, but also the types of each interaction when it contains several variables.

Another test has been added in the jaspTestModule encoding analysis.

NB: The Mixed Models still do not work after this fix, but @FBartos knows why. He needs first this fix to be able to fix completely the Mixed Models.

### DIFF
--- a/QMLComponents/models/term.h
+++ b/QMLComponents/models/term.h
@@ -47,8 +47,11 @@ public:
 	bool						isDraggable()	const			{ return _draggable; }
 	void						setDraggable(bool draggable)	{ _draggable = draggable; }
 
-	columnType					type()			const			{ return _type; }
-	void						setType(columnType type)		{ _type = type; }
+	// If a term has several components, its type self is unknown, but the components have maybe a type.
+	columnType					type()			const			{ return _types.size() == 1 ? _types[0] : columnType::unknown; }
+	void						setType(columnType type)		{ _types = {type}; }
+	columnTypeVec				types()			const			{ return _types; }
+	void						setTypes(columnTypeVec types)	{ _types = types; }
 
 	typedef QStringList::const_iterator const_iterator;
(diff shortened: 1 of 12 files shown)


QMLComponents/models/terms.cpp
```diff
@@ -415,11 +419,15 @@ Terms Terms::ffCombinations(const Terms &terms)
...
-
-			term.append(newTerm);
-			newTerms.add(Term(term));
+			QStringList termComponents = _terms.at(i).components();
+			termComponents.append(combos.at(j).components());
+			columnTypeVec types = _terms.at(i).types();
+			columnTypeVec coTypes = combos.at(j).types();
+			types.insert(types.end(), coTypes.begin(), coTypes.end());
+
+			Term newTerm(termComponents);
```
JorisGoosen, line 428: Why not expand the constructor?

QMLComponents/models/term.h
```diff
@@ -81,7 +84,7 @@ class Term
 	QStringList		_components;
 	QString			_asQString;
 	bool			_draggable = true;
-	columnType		_type = columnType::unknown;
+	columnTypeVec	_types = {columnType::unknown};
```
JorisGoosen, line 87: _types should be settable in the constructor (and can then have `columnType::unknown' as the default argument.

JorisGoosen (approved):
I guess it works (a bit hard to see without a real testcase) and I wouldve like to see the Terms constructor expanded to have 'type' as as an argument.
But If I merge it now it ends up in the nightly...

PR #5733, opened 2024-11-12
Order of variables should be kept in formula

boutinb:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2689

This issue is reproduceable with Variables list with interactions: in ANOVA or MIxed Models for example.

Also in the Mixed Models, if several random effects are added, some effects are sometimes suddenly added or removed when applying the R Syntax. This was due to the fact that the controls dynamically created by the ComponentsList were not removed before setting the new options. 

At last, avoid generating R Syntax during the initialization of the form: each time a control gets a new value, the R Syntax is generated: this is not needed when the form is being initialized.

### DIFF
--- a/QMLComponents/rsyntax/formulasource.cpp
+++ b/QMLComponents/rsyntax/formulasource.cpp
@@ -257,24 +257,25 @@ QString FormulaSource::_generateRandomEffectsTerms(const Terms& terms) const
 	return result;
 }
 
-QString FormulaSource::generateInteractionTerms(const Terms& tterms, const Json::Value& changedTypes)
+QString FormulaSource::generateInteractionTerms(const Terms& terms, const Json::Value& changedTypes)
 {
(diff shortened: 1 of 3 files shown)


QMLComponents/rsyntax/formulasource.cpp
```diff
@@ -294,14 +295,27 @@ QString FormulaSource::generateInteractionTerms(const Terms& tterms, const Json:
...
+		}
+	}
+
+	// Use the original order of the terms to add them in the formula
+	for (const Term term : terms)
+	{
+		if (termsToAdd.contains(term))
+		{
+			if (!first)	result += " + ";
+			result += FormulaParser::transformToFormulaTerm(term, termsToAdd[term], FormulaParser::interactionSeparator);
```
JorisGoosen, line 317: lazy solution :p

JorisGoosen (approved):

PR #5734, opened 2024-11-13
Fix Variables List with layers for contingency table analysis

boutinb:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2683

I'm not sure if a Layer Variables List with several layers has ever worked properly. It was at least not possible to load a JASP file with a contingency table analysis with several layers. To reproduce the problem, you can also just duplicate the analysis.

### DIFF
--- a/QMLComponents/models/listmodellayersassigned.cpp
+++ b/QMLComponents/models/listmodellayersassigned.cpp
@@ -17,8 +17,7 @@
 //
 
 #include "listmodellayersassigned.h"
-#include "boundcontrols/boundcontrollayers.h"
-#include "utilities/qutils.h"
+#include "controls/jasplistcontrol.h"
 
 using namespace std;
 
@@ -32,13 +31,13 @@ void ListModelLayersAssigned::initLayers(const std::vector<std::vector<std::stri
 {
 	beginResetModel();
 	
-	_variables.clear();
+	_variablesPerLayer.clear();
 	for (const std::vector<std::string>& variables : allVariables)
 	{
 		QList<QString> layer;
 		for (const std::string& variable : variables)
 			layer.push_back(QString::fromStdString(variable));
-		_variables.push_back(layer);
+		_variablesPerLayer.push_back(layer);
 	}
 
 	_setTerms();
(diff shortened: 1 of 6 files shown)


QMLComponents/models/listmodellayersassigned.cpp
```diff
@@ -114,23 +104,23 @@ Terms ListModelLayersAssigned::addTerms(const Terms& terms, int dropItemIndex, c
 	
 	beginResetModel();
 	
-	int layer = _variables.length();
+	int layer = _variablesPerLayer.length();
 	int indexInLayer = 0;
 	if (dropItemIndex >= 0)
-		layer = _getLayer(dropItemIndex, indexInLayer, true);
+		std::tie(layer, indexInLayer) = _getLayer(dropItemIndex, true);
```
JorisGoosen:
```suggestion
		auto [layer, indexInLayer] = _getLayer(dropItemIndex, true);
```

In other places you already use structured bindings right?
Why use std::tie in only two occasions then? 
I had to look up what it does only to find: https://stackoverflow.com/a/43780252

QMLComponents/models/listmodellayersassigned.cpp
```diff
@@ -143,15 +133,15 @@ void ListModelLayersAssigned::moveTerms(const QList<int> &indexes, int dropItemI
 {	
 	beginResetModel();
 	
-	int layerDrop = _variables.length();
+	int layerDrop = _variablesPerLayer.length();
 	int indexInLayerDrop = 0;
 	if (dropItemIndex >= 0)
-		layerDrop = _getLayer(dropItemIndex, indexInLayerDrop, true);
+		std::tie(layerDrop, indexInLayerDrop) = _getLayer(dropItemIndex, true);
```
JorisGoosen:
```suggestion
		auto [layerDrop, indexInLayerDrop] = _getLayer(dropItemIndex, true);
```

JorisGoosen (changes requested):
It seems to work fine and if you are ok with my replacing the std::tie by the auto [] thing it can be merged.

Consistency is always appreciated ;)

PR #5735, opened 2024-11-15
std::tie is not equal to the unpack operator []

boutinb:
std::tie assigns to existing variables
[] operator defines and assigns new variables.

But with a conditional assignment operator we can still remove the std::tie.


JorisGoosen:
Oops!

JorisGoosen (approved):

JorisGoosen (approved):

PR #5738, opened 2024-11-20
Fix some issues found in the Test Module

boutinb:
- In Test Variable Encoding, Simple Variables Form Section, set some variables so that the dropdown gets some values. Duplicate the analysis: error telling that an option cannot be set. The isJsonValid function of the ComboBox control was not up-to-date, and should allow object Json value. 
- In Test Variable Encoding, if you drag & drop 2 variables in an Interaction VariablesList, it adds automatically the interaction term. If there is an allowed type, the terms should get automatically this type even if they have another type. This does not work for the interaction term: in the result, we can still see that it keeps the original types.
- In Test Sources with special attributes analysis, in the source with levels section. If you set a nominal/ordinal variable, you see the levels of the variable in the dropdown. If you duplicate the analysis, it loses these values. This is due to the fact that when the values of a dropdown is directly set to the levels of another control, there is here no way to know the dependency of the controls between them: during the initialization of the form, the Dropdown can get its value before the Variable list (it actually depends on), and as the binding value is not a possible value of the dropdown, it loses. For this a special if the binding value does not correspond of the possible values, this value is kept in the _lostValue member. In case that the possible values of the dropdown change (because the Variables list is finally set, the dropdown checks whether the _lostValue could fit. This must be done only before the form is completely initialized.
- Add textFormat propery to controls. A Test Format analysis is also added in the Test Module.

### DIFF
--- a/QMLComponents/controls/comboboxbase.cpp
+++ b/QMLComponents/controls/comboboxbase.cpp
@@ -83,7 +83,11 @@ void ComboBoxBase::bindTo(const Json::Value& value)
 		}
 	}
 	else if (!selectedValue.empty())
-		_lostValue = selectedValue;
+		// The control is bound with a value, but its model is empty.
+		// Probably the values are set with a direct reference of a property of another control, like varList.levels, and this control is not yet initialized.
+		// (as the combobox has no direct reference to the varList self, it cannot add a dependency in _depends).
+		// So keep this value, and use it if the model is reset during the initialization of the form.
+		_unusedInitialValue = selectedValue;
 
 	_setCurrentProperties(index);
 
@@ -155,7 +159,7 @@ void ComboBoxBase::setUp()
 	if (form())
 	{
 		connect(form(), &AnalysisForm::languageChanged,			[this] () { _model->resetTermsFromSources(); }	);
-		connect(form(), &AnalysisForm::analysisChanged,			[this] () { _lostValue = ""; });
+		connect(form(), &AnalysisForm::analysisChanged,			[this] () { _unusedInitialValue = ""; });
 	}
 }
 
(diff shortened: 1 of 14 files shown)


JorisGoosen (approved):

QMLComponents/controls/comboboxbase.cpp
```diff
@@ -82,6 +82,8 @@ void ComboBoxBase::bindTo(const Json::Value& value)
 			index = int(std::distance(values.begin(), itr));
 		}
 	}
+	else if (!selectedValue.empty())
+		_lostValue = selectedValue;
```
JorisGoosen: Bedoel je "lost" value of "last" value?

JorisGoosen:
Im guessing the name of lostValue should be lastValue and also you need to manually rebase this so we can merge it

JorisGoosen (changes requested):

QMLComponents/controls/comboboxbase.cpp
boutinb: I called it lostValue because of the issue it solves: the dropdown is initialized with a value, but has no possible values (its model is empty). This is due to the fact that the VariablesList it depends on, gets initialized afterwards. In this case (when the dropdown has for values the levels of a VariablesList control), the dropdown has no way of guessing on which control it depends (`values : varList.levels`: the values are set with a list of strings, the levels, but it has no reference of the varList control self). This means that during the initialization of the form, VariablesList might be initialized after the dropdown.
Usually, if the dropdown is set with a non-existing value, it neglects this value (probably it comes from an old version of the analysis, and sets a warning), but in this case we want to keep this value until the form is completely initialized: if in the meantime, the dropdown gets a signal that its model is updated, it should try again to set this `lostValue`.
Maybe `lostValue` is not a good choice, but `lastValue` is definitly not he right name. This value should be used only during the initialization of the form. Some comments in the code would be nice also....

boutinb: OK I replaced _lostValue by _unusedInitialValue. And add some comments

PR #5739, opened 2024-11-22
bugfix: fix missing html content in notes

shun2wang:
Fix: https://github.com/jasp-stats/jasp-test-release/issues/2776

This is Quill's issue, after upgrading to Quill v2, converting HTML should to be inserted as an object.

Note: Should update all data libraries from 0.19.1 to avoid description loss issues introduced by the test version.In addition, we may need to consider why automated testing does not cover this point.

### DIFF
--- a/Desktop/html/js/jaspNotes.js
+++ b/Desktop/html/js/jaspNotes.js
@@ -540,7 +540,7 @@ JASPWidgets.NoteBox = JASPWidgets.View.extend({
 				html = this.model.get("text");
 			}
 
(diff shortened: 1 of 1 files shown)


JorisGoosen:
> we may need to consider why automated testing does not cover this point

Automated testing now only looks at the tables and their contents. But they ignore the notes.
I guess if the notes are indeed in html preview then we could definitely also check for their existence after refreshing/loading

boutinb (approved):

PR #5742, opened 2024-11-26
Prevent setting an empty column name

boutinb:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/2808

### DIFF
--- a/Desktop/data/datasetpackage.cpp
+++ b/Desktop/data/datasetpackage.cpp
@@ -1634,13 +1634,12 @@ void DataSetPackage::setColumnName(size_t columnIndex, const std::string & newNa
 
 	std::string oldName = getColumnName(columnIndex);
 
-	if(!column->setName(newName))
-		return;
-	
-	
-	setManualEdits(true);
-	emit datasetChanged({}, {}, QMap<QString, QString>({{tq(oldName), tq(newName)}}), false, false);
-	refresh();
+	if(column->setName(newName))
+	{
+		setManualEdits(true);
+		emit datasetChanged({}, {}, QMap<QString, QString>({{tq(oldName), tq(newName)}}), false, false);
+	}
+	refresh(); //We do refresh in any case because then the emptied name of the column in variableswindow will get filled again
 }
 
 void DataSetPackage::setColumnTitle(size_t columnIndex, const std::string & newTitle)
--- a/Desktop/data/datasetpackage.cpp
+++ b/Desktop/data/datasetpackage.cpp
@@ -1623,7 +1623,7 @@ stringvec DataSetPackage::getColumnDataStrs(size_t columnIndex)
 	return out;
 }
 
-void DataSetPackage::setColumnName(size_t columnIndex, const std::string & newName, bool resetModel)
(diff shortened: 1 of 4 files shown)


Desktop/data/datasetpackage.cpp
```diff
@@ -1634,16 +1634,17 @@ void DataSetPackage::setColumnName(size_t columnIndex, const std::string & newNa
...
-	if(resetModel)
-		endResetModel();
+	endResetModel();
 
-	setManualEdits(true);
-	emit datasetChanged({}, {}, QMap<QString, QString>({{tq(oldName), tq(newName)}}), false, false);
+	if (change)
+	{
+		setManualEdits(true);
+		emit datasetChanged({}, {}, QMap<QString, QString>({{tq(oldName), tq(newName)}}), false, false);
```
JorisGoosen, line 1640: You now still give an `changed name from X to ""` signal no?
As in, if change == false because the new name is "" the new name is not newName...

JorisGoosen (changes requested):

JorisGoosen:
Also, to solve @tomtomme 's issue all you really need to do is call "delete column" properly if someone erases the name. Because that was the original intended behaviour right?
You erase the name to erase the column?

Desktop/data/datasetpackage.cpp
shun2wang: I think return to last valid name and make a tooltip is makes sense.

tomtomme:
> Also, to solve @tomtomme 's issue all you really need to do is call "delete column" properly if someone erases the name. Because that was the original intended behaviour right? You erase the name to erase the column?

I would consider this dangerous behaviour IF the undo button does not work. Because it is not intuitive that deleting the name would delete also my data. Deleting the name is something a user most probably would do by accident. So the best intended behaviour from my point of view, would be, to restore the original name OR just do nothing.
If you disagree and think the column should be deleted on erasing the name please make sure that the UNDO button works. Thx!

JorisGoosen:
> please make sure that the UNDO button works

Indeed, this is what I mean by "make sure it just calls deleteColumn" cause that would include the undo/redo

boutinb:
But I also disagree that making the name empty should remove the column: most of the time, making the name empty will be just a bad manip of the user. Trying to make it empty should reset the old name.

Desktop/data/datasetpackage.cpp
JorisGoosen: Im not saying it doesnt make sense, im saying that hes sending a signal `dataSetChanged({}, {}, "someName", "")` despite the name *NOT* changing

JorisGoosen:
Ok wahtever but Im fairly sure you @boutinb are the one who added this functionality in the first place.

Desktop/data/datasetpackage.cpp
boutinb: No: if newName is empty, change is false, and datasetChanged signal is not emitted.

JorisGoosen (approved):
Ill approve it now, despite it being very ugly that the name stays empty:
image

PR #5745, opened 2024-11-27
Do not synchronize uncessarily with the datafile

boutinb:
When a JASP file is loaded, it tries always to synchronize with its data file, even if the datafile was not changed: this is because the right dataFile timestamp was not correctly set in the JASP file.


JorisGoosen (approved):

JorisGoosen:
Never noticed that timestamp

PR #5747, opened 2024-11-28
Show the right implicit empty values in tooltip

boutinb:
If you set a nominal or ordinal variable in a scale VariablesList, a tooltip tells you how the values are transformed into scale. It gives also a list of 'implicit' empty values: that is the not empty values that could not be converted into a double value. This list was not correct, it gave all the convertable values.

### DIFF
--- a/CommonData/column.cpp
+++ b/CommonData/column.cpp
@@ -2381,13 +2381,15 @@ stringvec Column::previewTransform(columnType transformType)
 			for(Label * label : _labels)
 				if(!label->isEmptyValue() && !ColumnUtils::isDoubleValue(label->originalValueAsString()))
 				{
-					if(count++ < showThisMany)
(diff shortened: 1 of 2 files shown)


JorisGoosen:
Could you give me an example of the problem and how this solves it?
Because Ive been looking in my nightly for weird tooltip with "wrong" info and cant see it?

boutinb:
This is what I got with this test:
Screenshot 2024-11-28 at 12 11 57

boutinb:
And this is what I get after my fix:
Screenshot 2024-11-28 at 12 14 52

boutinb:
Here are my file: 
[simple.csv](https://github.com/user-attachments/files/17947451/simple.csv)

JorisGoosen:
@boutinb I think the double check should be on the value, not the label.

PR #5761, opened 2024-12-16
R Syntax project: split the Engine object

boutinb:
Split Engine object with an EngineBase object. The latter can then be used for the R Syntax: this EngineBase object has all the necessary functions used by the RBridge object.
It can be also instantiated with a In memory sqlite database.

### DIFF
--- a/CommonData/databaseinterface.cpp
+++ b/CommonData/databaseinterface.cpp
@@ -53,13 +53,13 @@ void DatabaseInterface::upgradeDBFromVersion(Version originalVersion)
 	transactionWriteEnd();
 }
 
-DatabaseInterface::DatabaseInterface(bool createDb)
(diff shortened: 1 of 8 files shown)


CommonData/databaseinterface.cpp
```diff
@@ -1603,49 +1609,49 @@ void DatabaseInterface::_runStatementsRepeatedly(const std::string & statements,
...
 }
 
-void DatabaseInterface::create()
+void DatabaseInterface::create(bool inMemory)
 {
 	JASPTIMER_SCOPE(DatabaseInterface::create);
 	assert(!_db);
 
-	if(std::filesystem::exists(dbFile()))
+	if(!inMemory && std::filesystem::exists(dbFile(false, inMemory)))
```
JorisGoosen: I think you should make `_inMemory` and just set that in the constructor.

PR #5764, opened 2024-12-18
R Syntax: split analysis and analysisForm

boutinb:
As described in the JASP Architecture for R Syntax document, we need to split the analysis and analysisForm objects. For this the AnalysisFormBase is added (AnalysisBase existed already), the Base objects contain all calls from Analysis to AnalysisForm and vice-versa.

The initColumnWithString is function is moved to the DataSet object so that the RDataFrame importer can use it.

The parseOptions function in the AnalysisForm is already added.

### DIFF
--- a/QMLComponents/analysisform.cpp
+++ b/QMLComponents/analysisform.cpp
@@ -30,8 +30,6 @@
 #include <QQmlContext>
 #include <QQmlEngine>
 #include <QTimer>
-#include <QJsonDocument>
-#include <QJsonObject>
 #include "preferencesmodelbase.h"
 
 using namespace std;
@@ -277,8 +275,8 @@ QString AnalysisForm::parseOptions(QString options)
 	Json::Value	 jsonOptions;
 	Json::Value jsonResult(Json::objectValue);
 
-	QJsonDocument doc = QJsonDocument::fromJson(options.toUtf8());
-	jsonReader.parse(doc.toJson().toStdString(), jsonOptions, false);
+	
+	jsonReader.parse(fq(options), jsonOptions, false);
 
 	if (!_analysis)
 		setAnalysis(new AnalysisBase(this)); // Create a dummy analyis object
--- a/QMLComponents/analysisform.cpp
+++ b/QMLComponents/analysisform.cpp
@@ -133,9 +133,9 @@ void AnalysisForm::runScriptRequestDone(const QString& result, const QString& co
 				// Some controls generate extra controls (rowComponents): these extra controls must be first destroyed, because they may disturb the binding of other controls
 				// For this, bind all controls to null and wait for the controls to be completely destroyed.
 				QTimer::singleShot(0, [=](){
-				bindTo(options);
+					bindTo(options);
 					blockValueChangeSignal(false, false);
-				_analysis->boundValueChangedHandler();
+					_analysis->boundValueChangedHandler();
 				});
 			}
 		}
--- a/QMLComponents/analysisform.cpp
+++ b/QMLComponents/analysisform.cpp
@@ -20,7 +20,6 @@
 #include "knownissues.h"
 #include "boundcontrols/boundcontrol.h"
 #include "utilities/qutils.h"
-#include "models/listmodeltermsavailable.h"
 #include "controls/jasplistcontrol.h"
 #include "controls/expanderbuttonbase.h"
 #include "log.h"
@@ -31,7 +30,8 @@
 #include <QQmlContext>
 #include <QQmlEngine>
 #include <QTimer>
-#include "controls/variableslistbase.h"
+#include <QJsonDocument>
+#include <QJsonObject>
 #include "preferencesmodelbase.h"
 
 using namespace std;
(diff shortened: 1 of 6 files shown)


QMLComponents/analysisform.cpp
```diff
@@ -272,16 +273,28 @@ void AnalysisForm::setHasVolatileNotes(bool hasVolatileNotes)
...
-		options = _analysis->boundValues();
-		return true;
+		bindTo(jsonOptions);
+		jsonOptions = _analysis->boundValues();
 	}
 
-	return false;
+	jsonResult["options"] = jsonOptions;
+	jsonResult["error"] = fq(getError());
+	return tq(jsonResult.toStyledString());
```
JorisGoosen, line 292: Why not return the Json value then?

boutinb: In the jaspQMLR example, the QML form is just a QuickItem object, so I call the parseOptions with the QMetaObject::invokeMethod method:
https://github.com/jasp-stats/jaspQmlR/blob/a7a93a2f6fa103d47560c1ed541e9c252eed3671/src/qmlR.cpp#L208
In this way I don't need to load the AnalysisFormBase header in the jaspQMLR package.
I don't think I can pass a Json::Value argument in the InvokeMethod, so I thought just send the string, and reset it as Json object afterwards.

QMLComponents/analysisform.cpp
```diff
@@ -272,16 +270,28 @@ void AnalysisForm::setHasVolatileNotes(bool hasVolatileNotes)
...
-bool AnalysisForm::parseOptions(Json::Value &options)
+
+QString AnalysisForm::parseOptions(QString options)
 {
-	if (_rSyntax->parseRSyntaxOptions(options))
+	Json::Reader jsonReader;
+	Json::Value	 jsonOptions;
+	Json::Value jsonResult(Json::objectValue);
+
+	QJsonDocument doc = QJsonDocument::fromJson(options.toUtf8());
```
JorisGoosen: Why is this step here?
It seems that the whole QJsonDocument

QMLComponents/analysisform.cpp
```diff
@@ -135,11 +133,10 @@ void AnalysisForm::runScriptRequestDone(const QString& result, const QString& co
 				// Some controls generate extra controls (rowComponents): these extra controls must be first destroyed, because they may disturb the binding of other controls
 				// For this, bind all controls to null and wait for the controls to be completely destroyed.
 				QTimer::singleShot(0, [=](){
-					bindTo(options);
+				bindTo(options);
```
JorisGoosen:
```suggestion
					bindTo(options);
```

QMLComponents/analysisform.cpp
```diff
@@ -135,11 +133,10 @@ void AnalysisForm::runScriptRequestDone(const QString& result, const QString& co
...
 					blockValueChangeSignal(false, false);
-					_analysis->boundValueChangedHandler();
+				_analysis->boundValueChangedHandler();
```
JorisGoosen:
```suggestion
					_analysis->boundValueChangedHandler();
```

QMLComponents/analysisform.cpp
```diff
@@ -567,7 +575,7 @@ void AnalysisForm::setTitle(QString title)
 		_analysis->setTitle(fq(title.simplified()));
 }
 
-void AnalysisForm::setOptionNameConversion(const QVariantList & conv)
+void AnalysisForm::setOptionNameConversion(const QVariantList &conv)
```
JorisGoosen:
```suggestion
void AnalysisForm::setOptionNameConversion(const QVariantList & conv)
```

QMLComponents/analysisform.h
```diff
@@ -145,6 +145,7 @@ public slots:
 	Q_INVOKABLE void		refreshAnalysis();
 	Q_INVOKABLE bool		initialized()			const	{ return _initialized; }
 	Q_INVOKABLE QString		generateWrapper()		const;
+    Q_INVOKABLE QString     parseOptions(QString options);
```
JorisGoosen:
```suggestion
	Q_INVOKABLE QString		parseOptions(QString options);
```

QMLComponents/analysisform.h
```diff
@@ -180,9 +181,8 @@ public slots:
 	void					sortControls(QList<JASPControl*>& controls);
 	QString					getSyntaxName(const QString& name)				const;
 	void					setHasVolatileNotes(bool hasVolatileNotes);
-	bool					parseOptions(Json::Value& options);
-	void					setActiveJASPControl(JASPControl* control, bool hasActiveFocus);
-	JASPControl*			getActiveJASPControl()	{ return _activeJASPControl; }
+    void					setActiveJASPControl(JASPControl* control, bool hasActiveFocus);
+    JASPControl*			getActiveJASPControl()	{ return _activeJASPControl; }
```
JorisGoosen:
```suggestion
	void					setActiveJASPControl(JASPControl* control, bool hasActiveFocus);
	JASPControl*			getActiveJASPControl()	{ return _activeJASPControl; }
```

JorisGoosen (approved):

PR #5770, opened 2025-01-03
add tooltip for TabView name editor

shun2wang:
Even I just found out it can be edited😜, so let's add a tooltip for this.


JorisGoosen (approved):
This doesnt use the attached tooltip so we could style this, but given that almost no tooltips are now styled (except for on the handles of the splitview) this is OK

PR #5774, opened 2025-01-08
Variable type change in CFA does not rerun the analysis

boutinb:
The Factors Form widget does not store the types of its variables in the options.  So this information is lost. But this information is needed in the CFA analysis, since it is possible in this analysis to change the variable type. Also if you change the type, the analysis is not rerun.
This fix takes care of backward compatibility: a JASP file from 0.19.3 or older containg a CFA analysis (or a Regression Linear or Logistic Regression analysis that uses also a Factors Form widget) can be read.

The jaspTestModule has been updated to test also type change in Factors Form.

### DIFF
--- a/QMLComponents/controls/factorsformbase.cpp
+++ b/QMLComponents/controls/factorsformbase.cpp
@@ -48,12 +48,17 @@ void FactorsFormBase::setUpModel()
 
 void FactorsFormBase::bindTo(const Json::Value& value)
 {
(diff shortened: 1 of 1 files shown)


JorisGoosen (approved):
Wouldve been nice if the Terms *always* have types so we dont need to do so many checks everywhere all the time

juliuspfadt (approved):

PR #5776, opened 2025-01-10
R Syntax: Load QML Controls as QML Module

boutinb:
Make the QML Components as a QML Module (by using the qt_add_qml_module cmake function).
This makes also possible to use this module as a plugin: this is needed for the R Syntax project

All versions in QML import statement must be removed.

No qmldir needed (it is generated automatically), and no need to register all the base QML objects: they are automatically registered via the QML_ELEMENT macro.

~~The module is linked to the CommonData and has access to all DataSet objects, but it should not use static members (as the sqlite database is instantiated in another translation unit, any call to this instance would crash). So a DatasetInterface must be created to load the sqlite database, and a new DataSet must use this DatasetInterface object: (as in the Engine). The Audit analyses creates Filter object: this object must be created via this DataSet object.
The Consumer/Provider pattern that was used to read data for the QML controls, could be in the future replaced by calls to this DataSet object.~~


QMLComponents/plugin.cpp
```diff
@@ -0,0 +1,62 @@
...
+    Q_PLUGIN_METADATA(IID QQmlEngineExtensionInterface_iid)
+
+
+	void initializeEngine(QQmlEngine *engine, const char *uri) override
+	{
+		QQmlEngineExtensionPlugin::initializeEngine(engine, uri);
+
+		QLocale::setDefault(QLocale(QLocale::English)); // make decimal points == .
+
+		QmlUtils::setGlobalPropertiesInQMLContext(engine->rootContext());
```
shun2wang, line 28: Build on Linux failed, see: https://github.com/jasp-stats/jasp-desktop/actions/runs/12710965693/job/35433254997?pr=5776#step:9:230

boutinb: Oh I see, thanks for the warning!

shun2wang:
Sorry, but version things still not be removed in `Description.qml` `Upgrades.qml` in submodules I think?

boutinb:
> Sorry, but version things still not be removed in `Description.qml` `Upgrades.qml` in submodules I think?

That's true, I forgot these ones also.

CommonData/filter.cpp
```diff
@@ -3,6 +3,8 @@
 #include "dataset.h"
 #include "databaseinterface.h"
 
+std::map<std::string, Filter*>	_filterMap;
```
JorisGoosen: This is a terrible Idea, a global really?

boutinb: OK, I have changed this, by creating a new DatabaseInterface that loads the sqlite database. This is indeed much cleaner.

CommonData/dataset.cpp
```diff
@@ -619,3 +619,9 @@ bool DataSet::initColumnWithStrings(int colIndex, const std::string & newName, c
 	return anyChanges || column->type() != prevType;
 }
 
+bool DataSet::isFilterNameFree(const std::string& filterName)
+{
+	return -1 == db().filterGetId(filterName);
```
JorisGoosen: Makes more sense to have this on `Filter` because the name is independent of the DataSet in question.

CommonData/filter.cpp
```diff
@@ -170,11 +170,6 @@ bool Filter::checkForUpdates()
 		return false;
 }
 
-bool Filter::filterNameIsFree(const std::string &filterName)
-{
-	return -1 == DatabaseInterface::singleton()->filterGetId(filterName);
```
JorisGoosen: Which is also where it was...

CommonData/dataset.cpp
JorisGoosen: Why did you make it non-static?

QMLComponents/variableinfo.cpp
```diff
@@ -68,5 +70,12 @@ bool VariableInfo::dataAvailable()
 
 DataSet *VariableInfo::dataSet()
 {
-	return _provider ? reinterpret_cast<DataSet*>(_provider->provideInfo(VariableInfo::DataSetPointer).value<void*>()) : nullptr;
+	static DataSet* singleDataSet = nullptr;
```
JorisGoosen: This is not going to work is it?
How do you reflect the dataset getting destroyed?

JorisGoosen: Yeah, so as I suspected:
- open dataset, use a sampling audit analysis, it works
- close dataset, dont close jasp
- open other dataset, use a sampling audit analysis, its broken...

Ill just make sure the datasetpointer works, or Ill fix it some other way

boutinb: Hmm I forgot this case.

JorisGoosen:
So ive reverted a bunch of the changes you made to cpp code that was unnecessary.
It now runs fine on macos, I will proceed to test Windows and Linux.

PR #5778, opened 2025-01-14
chore: build check workflow fix for readstat

shun2wang:
@JorisGoosen this could be merge to fix workflow.

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -60,7 +60,7 @@ jobs:
         run: |
             wget https://github.com/WizardMac/ReadStat/releases/download/v1.1.9/readstat-1.1.9.tar.gz
             tar -xzf readstat-*.tar.gz && cd readstat-*/
-            ./configure
+            ./configure --disable-werror
             make
             sudo make install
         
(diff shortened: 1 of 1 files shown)


.github/workflows/ubuntu.yml
```diff
@@ -61,7 +61,7 @@ jobs:
             wget https://github.com/WizardMac/ReadStat/releases/download/v1.1.9/readstat-1.1.9.tar.gz
             tar -xzf readstat-*.tar.gz && cd readstat-*/
             ./configure
-            make
+            make CFLAGS='-Wno-error' CXXFLAGS='-Wno-error'
```
JorisGoosen: Uhm, I turn this on on purpose usually?
Shouldnt we just do: `-Wno-error=use-after-free` instead?

JorisGoosen:
```suggestion
            make CFLAGS='-Wno-error=use-after-free' CXXFLAGS='-Wno-error=use-after-free'
```

shun2wang: Turning it on will make the build fail, we only do this in the workflow for build test and not any local builds. 
Fully fixing it would require upstream code changes (I remember you fixed this for upstream once, but not sure why it's happening again)

JorisGoosen: Maybe some compiler got updated to be more strict or something, but this fixes it so looks good ^^

JorisGoosen (approved):

PR #5783, opened 2025-01-22
Data set q etc

### LINKED ISSUE #1330: [Feature-Request]: use filters in the analyses rather than in the data
* Enhancement: Currently, filtering is done in the data sheet, and applies to all analyses in a JASP file. Often, however, one may want to perform different t-tests on different subgroups in the same JASP file, as follow-ups to an ANOVA. It would be fantastic, if one could include a filter in the analysis, so that it applies only to that analysis.
* Purpose: Followup statistics restricted to specific groups
* Use-case: Imagine you have a mixed ANOVA, with 3 groups of subjects, each with a 2x2 within subject design. You ran an ANOVA with a group x something effect. Now you want to run follow up analyses separately in each group. 
 
Currently, I must filter the data for group=1, then I can run what-ever analysis I wish on the other factors, then I can refilter the data for group=2, and I see the results update. However, my final JASP file, will never contain the results of the analysis for all the subgroups, and its thus hard for me to upload my analyses to open science forum in an easily readable way.
 
 I see two solutions to do that. (A) include a filter option in the analysis. Now I could filter for Group=1, duplicate the analysis, and change the filter to Group2, etc., and have that all in one neat JASP file. 
**Describe alternatives you've considered**
 
(B) having a categorical grouping field in t-tests, with the results window then showing the result separatly for each level of the grouping field.
**Additional context**

JorisGoosen:
image

Also does minore visible tweaks like making some more things round, mostly in preferences (but all dropdowns too!):
image

Now has nice tabs for selecting datasets and filters alike:
image
The currently selected filter/dataset's title can be edited right inside the tabbutton. Just like it can be deleted, or new ones added.
image

To do:
- [x] Use tabs above filter like suggested by Rens in https://github.com/jasp-stats/jasp-issues/issues/1330#issuecomment-1452259849
- [x] And also use tabs for the dataset selection.
- [x] Furthermore it could be nice if the application did not crash on editing any of the not-first-datasets.
- [x] The overview of the filter also needs to have grouptitles for the datasets so the different filters can be distinguished
- [x] Allowing the name/title of a filter to be edited would also be nice thing
- [x] Same for the DataSet
- [x] Fix all the bugs remaining
- [x] Fix FilteredDataEntryListModel when switching datasets!
- [x] Computed DataSets
- [x] Make sure analyses have shown somewhere in  the results on which filter they are actually based. Might require adding a "title" or something for Filters as well. See feature request: https://github.com/jasp-stats/jasp-issues/issues/4314


JorisGoosen:
image
Not the prettiest but it works for now. It can be made better later

PR #5785, opened 2025-02-04
R Syntax: Load QML Controls as QML Module

boutinb:
Make the QML Components as a QML Module (by using the qt_add_qml_module cmake function).
This makes also possible to use this module as a plugin: this is needed for the R Syntax project

All versions in QML import statement must be removed.

No qmldir needed (it is generated automatically), and no need to register all the base QML objects: they are automatically registered via the QML_ELEMENT macro.

The QMLComponents module is a static library. To make it a shared library would be really painful on Windows: all the library (Common, CommonData), should be also dynamic, and export their object with dll_export macros, in order to solve all the problems correctly of static members. Also an extra dynamic library would be necessary to collect all the objects that are used in QMLComponents and Desktop.

As static library, there is no initialization of the module with the QML Engine, so the context objects needed by the QML Controls (jaspTheme, preferencesModel), some variables (DEBUG_MODE...), and the registration of the enum types must be done outside the module. This is done by the MainWindow class, and this must be done by the R package that will call the QML form.


JorisGoosen (approved):
Had to make a small tweak to fix some issue on windows when adding a column (through audit sampling for instance). With qt 6.8.2 if it makes a difference.


Works lovely now though!
Nice you figured out how to make it behave on Windows ^^

PR #5793, opened 2025-02-19
bugfix: fix to cell value vertical-align in tables

shun2wang:
Fix: https://github.com/jasp-stats/INTERNAL-jasp/issues/2794

Not sure why we set to 'top', the browser user agent stylesheet auto set to middle/inherit by default so we just remove it.
Or we can specify it to be fixed to center alignment, any preference?

Note: Maybe the unit test .jasp files also need to be updated because the CSS style string will be changed.

### DIFF
--- a/Desktop/html/css/jasp.css
+++ b/Desktop/html/css/jasp.css
@@ -114,6 +114,7 @@ td.symbol {
 
 td {
 	text-align: right ;
+	vertical-align: middle ;
 }
 
 thead tr:nth-child(2) th:not(.separator) {
--- a/Desktop/html/css/jasp.css
+++ b/Desktop/html/css/jasp.css
@@ -114,7 +114,6 @@ td.symbol {
 
 td {
 	text-align: right ;
-	vertical-align: top ;
 }
 
 thead tr:nth-child(2) th:not(.separator) {
--- a/Desktop/html/css/jasp.css
+++ b/Desktop/html/css/jasp.css
@@ -114,8 +114,8 @@ td.symbol {
 
 /* Customed superscripts and subscripts to maintain consistent vertical alignment and line height */
 td sub, td sup {
-	font-size: 75%;
-	line-height: 0;
+	font-size: 70%;
+	line-height: 1em;
 	position: relative;
 	vertical-align: baseline;
 }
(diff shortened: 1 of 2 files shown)


JorisGoosen:
I think I would prefer it to specify what we want there. Cause maybe whatever we inherit changes later?

JorisGoosen:
Ok, I am on a weird arch linux setup here and cannot make screenshots but exponents are still not aligned with other numbers...

JorisGoosen:
Ah @FBartos gave me the great idea to do as the young people do:
screenshot

JorisGoosen:
Maybe bottom alignment is better?

shun2wang:
Well, this is actually due to the fact that the cell elements with exponent and the numeric elements without indices are not the same height, a height difference of 1-2 pixels by auto adjusted.

shun2wang:
I also adjusted latex img vertical align in tables, test case in jaspTestModule-> test jaspHtml.

JorisGoosen:
Well, its a bit better now, but now the exponent slightly overlaps the row above it...

signal-2025-02-24-151845_002

I think right now in general all the rows in the resulttables are a bit cramped.

As far as the latex thing goes, im not sure what you changed or where I would see this in the testcase?
Am I overlooking some exponents somewhere?

shun2wang:
ah! seems there will be differences on different screens. I'll think about it...

for latex things: the latex image is not vertical align middle in table rows with numbers, the commit will change it,not related with exponents.

shun2wang:
I suspect I'm overcomplicating things because it doesn't follow the ”KISS principle“...🤔

JorisGoosen:
Isnt there some way to anchor on the middle horizontal line of a cell?

shun2wang:
Any idea @boutinb ? 🤔

Desktop/html/css/jasp.css
```diff
@@ -112,9 +112,27 @@ td.symbol {
...
+
+td sub {
+	bottom: -0.25em;
+}
+
 td {
 	text-align: right ;
-	vertical-align: top ;
+	vertical-align: middle ;
+	padding-top: .1em;
```
boutinb, line 134: I don't understand why you have added this padding-top and padding-bottom.

shun2wang:
Replaced by: https://github.com/jasp-stats/jasp-desktop/pull/5994

PR #5794, opened 2025-02-19
Monetary columns and locale sensitive number display

JorisGoosen:
This PR implements https://github.com/jasp-stats/INTERNAL-jasp/issues/1744 where a "monetary" format is added.
The format can just be "monetary" and then by default euro is chosen, to add a currency just do something like "monetary:USD" or "monetary:EUR".

This also implements almost everything in https://github.com/jasp-stats/jasp-issues/issues/39 as far as localizing numbers and money go. 

Date, time and or datetime colums are *not* added in this pr but the localisation stuff added now should help implementing it.

Ill also trigger some builds


CommonData/columnutils.h
```diff
@@ -3,30 +3,44 @@
```
shun2wang, line 3:
```suggestion

#include <functional>
```
I cannot compiling it without this with MSVC, and buildBot on Windows also failed.

Desktop/html/js/utils.js
```diff
@@ -1,3 +1,48 @@
+var currentLocale	= new Intl.Locale("en");
+var currentLocaleId = "en"
+
+function setCurrentLocaleID(id)
+{
+	console.log(`Setting current locale to id "${id}"`)
```
shun2wang:
```suggestion

```

The javaScript log cannot write to log files right? then I think it's unnecessary in production environment.

shun2wang (changes requested):
In addition to two small suggested changes, there are other potentially serious problems:
1. Label-Value combination change is not synchronously? when you change the locale, see that combination it's different.
2. Edit value with decimal will crash JASP, error msg: `replaceDoublesTillLabelsRowWithLabels choked on a temp-label that cant be converted to double???`
3.While import [SPPS_Debug.sav](https://static.jasp-stats.org/debugData/SPPS_Debug.sav) in `NumericRestrictedOrdinal` from debug.sav the decimal point mixed both `.` and `,`.
4. Maybe setting options should be moved to "Result" and put it with `Table options` together.

koenderks (changes requested):
Took me a while to realize that I had to put in `format = "monetary"` instead of `type = "monetary` but after that it looks quite nice. Except for the fact that Euro's (at least in most European countries) uses `,` as the decimal separator and `.` as the thousand separator. Furthermore it looks nice.

image

koenderks:
@shun2wang Sorry, I accidentally requested a re-review from you. You can ignore this.

EJWagenmakers:
Yes but do we even want to allow comma's as digital separators? Maybe we should, but it sseems like it could be terribly confusing

JorisGoosen:
> Took me a while to realize that I had to put in `format = "monetary"` instead of `type = "monetary` but after that it looks quite nice. Except for the fact that Euro's (at least in most European countries) uses `,` as the decimal separator and `.` as the thousand separator. Furthermore it looks nice.


Yeah so the dots/commas is dependent on locale not on the currency. Because otherwise you would get different formatting for lets say, a column of yen next dollars next to euros? Seems unwanted

JorisGoosen:
> Yes but do we even want to allow comma's as digital separators? Maybe we should, but it sseems like it could be terribly confusing

Yeah this is a more broad question actually, also see the following:
French:
image
German:
image
American:
image

This adding of thousands-separators is a sideeffect of making the numbers locale dependent. 
I think it looks pretty good to be honest.

I also added a "Use American English locale" for numbers checkbox in the settings.
This will do almost the same as what JASP always did (and the same as the english setting) but it *does* add these thousands separators.

Although if we dont want those maybe I can find some way to turn them off?

Desktop/html/js/utils.js
JorisGoosen: Oops!

CommonData/columnutils.h
JorisGoosen: Ill have a look

JorisGoosen:
> In addition to two small suggested changes, there are other potentially serious problems:
> 
>     1. Label-Value combination change is not synchronously? when you change the locale, see that combination it's different.

Well, the label is a stringified version of the value, so it has the locale dependent setting.
I setup the code in such a way that after a locale change it checks all the labels and converts them. 
Not sure why this isnt working for you.

> 
>     2. Edit value with decimal will crash JASP, error msg: `replaceDoublesTillLabelsRowWithLabels choked on a temp-label that cant be converted to double???`

Editing it in the labeleditor or in the data?

>        3.While import [SPPS_Debug.sav](https://static.jasp-stats.org/debugData/SPPS_Debug.sav) in `NumericRestrictedOrdinal` from debug.sav the decimal point mixed both `.` and `,`.

Hey that is a good one, ill have a look.

>     3. Maybe setting options should be moved to "Result" and put it with `Table options` together.

The language option?
the locale/language also has an effect on the qml components (textinput etc) and the dataset.

shun2wang:
>Editing it in the labeleditor or in the data?

In the data.

JorisGoosen:
> > Editing it in the labeleditor or in the data?
> 
> In the data.

Hmm, doesnt crash for me on linux, but Ill check later on a windows system

shun2wang:
So you can see the crash also on Linux Debug build while press <kbd>Enter</kbd>
[录屏 2025-02-20 22-23-52.webm](https://github.com/user-attachments/assets/bb12b5d4-0ef9-4f40-924a-c57097452265)

JorisGoosen:
> So you can see the crash also on Linux Debug build while press Enter [录屏 2025-02-20 22-23-52.webm](https://github.com/user-attachments/assets/bb12b5d4-0ef9-4f40-924a-c57097452265)

Ah yes, now I can reproduce...

JorisGoosen:
I think Ive fixed the crash shun encountered.

Ive also made the "alternative" locale option a bit more comprehensive, like crazy comprehensive :p

There are new builds running, where the default locale used is the one linked to the chosen language, unless the alternative is selected:
image
image
image
image
image
image

JorisGoosen:
Wel Ive also changed the sorting of the languages and territories a bit, although it will be missing from (some?) of the builds.

JorisGoosen:
Ok, so now the alternative locale actually gets stored...
So Ive started new builds anyway.

Desktop/components/JASP/Widgets/FileMenu/PrefsUI.qml
```diff
@@ -190,17 +190,46 @@ ScrollView
...
+				{
+					id:							alternativeLocaleTerritory
+					label:						qsTr("Alt. territory")
+					values:		 				languageModel.altTerritories
+					startValue:				 	languageModel.currentAltTerritory
+					onValueChanged:				if(value != "") languageModel.currentAltTerritory = value
+					addEmptyValue:				false
+					KeyNavigation.tab:			altnavcheckbox
+				}
+
```
shun2wang: Yes the crash was fixed.

Maybe if we use something like `Number(-12345.37).toLocaleString(Qt.locale(<local code>))` to have a preview following locale changed in the right side, will help users see expected changes visually.

everything else seems fine to me, except the thousandths side effect (I'm still not used to it)😜because I can think of a situation where the user changes the language but doesn't want to change the number representation.

shun2wang (approved):

Desktop/components/JASP/Widgets/FileMenu/PrefsUI.qml
JorisGoosen: > I can think of a situation where the user changes the locale but doesn't want to change the number representation

But then what do they chance the locale for?

And Im also still getting used to the thousands separators yeah :p

shun2wang: - From now: change language -> number format changed also, then if you don't want to you should Specifying an Alternative locale
- What I mean is: the format does not follow the language changes, but follows the alternative locale. and render a preview to the right side. Because we are changing the default user  old behavior.

I'm not particularly adamant about this though, since I can also change the language by changing the locale afterwards, it's just a matter of order. It would be nice to have some documentation though.

JorisGoosen: We could turn the "alternative" on by default and then have the american thing be default.
The preview could be nice too and I guess I should indeed write some documentation.

I figured the tooltip would be enough :p

Desktop/components/JASP/Widgets/FileMenu/PrefsUI.qml
```diff
@@ -185,15 +185,85 @@ ScrollView
...
+				checked:					languageModel.useAlternativeLocale
+				onCheckedChanged:			languageModel.useAlternativeLocale = checked
+				toolTip:					qsTr("Use the locale specified below for display of numbers and currency.") //dates and times will have to be added later
+				
+				KeyNavigation.tab:			alternativeLocaleTerritory
+			}
+			
+			RowLayout
+			{
+				spacing:		jaspTheme.generalAchorMargin
```
boutinb:
```suggestion
				spacing:		jaspTheme.generalAnchorMargin
```

Desktop/components/JASP/Widgets/FileMenu/PrefsUI.qml
```diff
@@ -185,15 +185,85 @@ ScrollView
...
+						startValue:				 	languageModel.currentAltTerritory
+						value:						languageModel.currentAltTerritory
+						onValueChanged:				if(value != "") languageModel.currentAltTerritory = value
+						addEmptyValue:				false
+						KeyNavigation.tab:			altnavcheckbox
+					}
+					
+				}
+				
+				Item{ width: jaspTheme.generalAchorMargin}
```
boutinb:
```suggestion
				Item{ width: jaspTheme.generalAnchorMargin}
```

Desktop/utilities/languagemodel.cpp
```diff
@@ -128,9 +227,52 @@ void LanguageModel::setCurrentLanguage(QString language)
...
+	};
+
+	static ColumnUtils::toIntF altFuncToInt = [&](const std::string & str, int & intVal)
+	{
+		bool isInt = false;
+		intVal = currentLocale().toInt(tq(str), &isInt);
+
+		return isInt;
+	};
+
```
boutinb: At first, I was quite puzzled about this code. A little comment could make it easier to read it. 
```suggestion
	// ColumnUtils is in CommonData library and doesn't access Qt (for instance for QLocale), so instead we use a callback.
```

Desktop/components/JASP/Widgets/FileMenu/PrefsUI.qml
```diff
@@ -185,15 +185,85 @@ ScrollView
...
 
-				KeyNavigation.tab: 			altnavcheckbox
+				KeyNavigation.tab: 			useAlternativeLocale
 				
 			}
 			
+			CheckBox
+			{
+				id:							useAlternativeLocale
+				label:						qsTr("Use an alternative locale for numbers")
```
boutinb: I think 'locale' is a concept that most users don't know
```suggestion
				label:						qsTr("Use an alternative language and territory dependent display for numbers")
```

Desktop/components/JASP/Widgets/FileMenu/PrefsUI.qml
```diff
@@ -185,15 +185,85 @@ ScrollView
...
+				}
+				
+				Item{ width: jaspTheme.generalAchorMargin}
+				
+				Rectangle
+				{
+					implicitWidth:		exampleFormattingText.implicitWidth
+					implicitHeight:		exampleFormattingText.implicitHeight
+					
+					color:				enabled ? jaspTheme.white : jaspTheme.whiteBroken
```
boutinb: This enabled does not work. It should be languageModel.useAlternativeLocale instead. But I don't think it should be grayed out if the alternative locale is not checked: the way the numbers are written is still depending on the language.

Desktop/html/js/utils.js
```diff
@@ -79,18 +119,23 @@ function formatColumn(column, type, format, alignNumbers, combine, modelFootnote
...
 
-	let formats = format.split(";");
-	let p = NaN;
-	let dp = NaN;
-	let sf = NaN;
-	let pc = false;
-	let approx = false;
-	let log10 = false;
-	dp = parseInt(window.globSet.decimals);  // NaN if not specified by user
+	///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
```
boutinb, line 122: I'm impressed that you dared to touch this code!

Desktop/main.cpp
```diff
@@ -461,7 +461,8 @@ int main(int argc, char *argv[])
 			QmlUtils::configureQMLCacheDir();
 		#endif
 
-			QLocale::setDefault(QLocale(QLocale::English)); // make decimal points == .
+			if(Settings::value(Settings::USE_ALT_LOCALE).toBool())
+				QLocale::setDefault(QLocale(QLocale::English)); // make decimal points == .
```
boutinb: Why do we need that here?

QMLComponents/rsyntax/rsyntax.cpp
```diff
@@ -379,7 +379,7 @@ QString RSyntax::transformJsonToR(const Json::Value &json)
 			result = QString::number(json.asUInt());
 		break;
 		case Json::realValue:
-			result = QString::number(json.asDouble());
+			result = QString::number(json.asDouble()); //This is not taking locale into account, but as its going to R this is ok I guess?
```
boutinb, line 382: Yes

QMLComponents/controls/textinputbase.cpp
```diff
@@ -58,9 +44,9 @@ QString TextInputBase::_getDoubleArrayValue(const std::vector<double>& doubleVal
 	for (double doubleValue : doubleValues)
 	{
 		if (!first)
-			value += ",";
+			value += ";";
```
boutinb, line 38: I think Descriptive needs an upgrade after this change

QMLComponents/controls/textinputbase.cpp
```diff
@@ -71,40 +57,40 @@ void TextInputBase::bindTo(const Json::Value& value)
...
-		else
-			_value = dblVal;
+		if (value.isNumeric())		
+			dblVal = value.asDouble();
+		
+		else if (value.isString() && !QColumnUtils::getDoubleValue(tq(value.asString()), dblVal))
+			dblVal = NAN;
+			
+		_value = dblVal; //Stored as the user enters (so 0-100), but sent in json / 100 through _getPercentValue
+		//This mean the "bound value" is 0...1 so:
```
boutinb, line 71: _getPercentValue is not used anymore and can be removed

QMLComponents/controls/jaspdoublevalidator.cpp
```diff
@@ -31,28 +32,34 @@ QValidator::State JASPDoubleValidator::validate(QString& s, int& pos) const
...
+	//bool isInt		= QColumnUtils::getIntValue(	s, intVal);
+	
+	//Maybe the number is formatted in some crazy way, because of locales
+	QString toEnglish = isNumber ? QString::number(value) : s;
+	
 	// check length of decimal places
-    QString point = locale().decimalPoint();
-	int indexPoint = s.indexOf(point);
+    QString point = ".";
+	int indexPoint = toEnglish.indexOf(point);
```
boutinb: Apparently this does not work: for an IntegerField you can now only have a number of 6 digits. And for a DoubleField you can have only 6 digits before the decimal sign.

boutinb (changes requested):
Great work!
I have some comments, and found 1 bug about the number of allowed digit (before the decimal sign for double): you can now have only 6 digits.

Desktop/components/JASP/Widgets/FileMenu/PrefsUI.qml
JorisGoosen:
```suggestion
					color:				jaspTheme.white
```

QMLComponents/controls/textinputbase.cpp
JorisGoosen: No it doesnt

Desktop/main.cpp
JorisGoosen: we dont, it is still there from when the option was called "use_usa_locale"

CommonData/columnutils.h
```diff
@@ -1,32 +1,53 @@
 #ifndef COLUMNUTILS_H
 #define COLUMNUTILS_H
 
+#include <functional>
```
shun2wang: Hi @JorisGoosen , seems now you add it twice.

JorisGoosen: Haha yeah I was afraid so, Im checking some last things :p

JorisGoosen:
@boutinb I think all mentioned problems have been solved.

boutinb (approved):

tomtomme:
@koenderks 
> Took me a while to realize that I had to put in `format = "monetary"` instead of `type = "monetary` but after that it looks quite nice. Except for the fact that Euro's (at least in most European countries) uses `,` as the decimal separator and `.` as the thousand separator. Furthermore it looks nice.

Wher do I put format = "monetary" ? Is this for computed columns only?

koenderks:
This is for analysis table columns only, not computed columns

tomtomme:
Thx! So the analysis has to support it. Currently only used in audit?

koenderks:
Indeed, both correct!

PR #5797, opened 2025-02-24
add nighly version notice to welcome page

shun2wang:
Fix: https://github.com/jasp-stats/INTERNAL-jasp/issues/2739

We using a branch name to detecting if it is a nightly build.

Now in the welcom page for both developers and users convenient to know whether they are working with a nightly under testing. because we did see in the issue that some users "accidentally" used the test version in the production environment.

### DIFF
--- a/Desktop/data/datasetpackage.cpp
+++ b/Desktop/data/datasetpackage.cpp
@@ -1637,6 +1637,7 @@ void DataSetPackage::setColumnName(size_t columnIndex, const std::string & newNa
 	{
 		setManualEdits(true);
 		emit datasetChanged({}, {}, QMap<QString, QString>({{tq(oldName), tq(newName)}}), false, false);
+		enginesReceiveNewData();
 	}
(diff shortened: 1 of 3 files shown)


JorisGoosen:
This could work if we only ever build releases from `stable` though. This happens often, but definitely not always...

shun2wang:
> This could work if we only ever build releases from `stable` though. This happens often, but definitely not always...

Because sometimes repo under a [Git detached HEAD](https://kodekloud.com/blog/git-detached-head/) then cmake here get "HEAD" as branch name. 
```cmake
  execute_process(
    COMMAND ${GIT_EXECUTABLE} rev-parse --abbrev-ref HEAD
    WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
    OUTPUT_VARIABLE GIT_BRANCH
    OUTPUT_STRIP_TRAILING_WHITESPACE)
```

so write access team should fix it first.I don't know how buildbot works with git but maybe `checkout -b` to local branch from a origin branch should always get branch names. such like `git checkout -b xxx(branchName) origin/xxxx(branchName)`

PR #5799, opened 2025-02-26
add nighly version notice to welcome page

shun2wang:
Reopened https://github.com/jasp-stats/jasp-desktop/pull/5797 which was closed by accidently force push...

Fix: https://github.com/jasp-stats/INTERNAL-jasp/issues/2739

We using a branch name to detecting if it is a nightly build. before merge this we should fix [Git detached HEAD](https://kodekloud.com/blog/git-detached-head/) of this repo.

Now in the welcom page for both developers and users convenient to know whether they are working with a nightly under testing. because we did see in the issue that some users "accidentally" used the test version in the production environment.

### DIFF
--- a/Desktop/components/JASP/Widgets/WelcomePage.qml
+++ b/Desktop/components/JASP/Widgets/WelcomePage.qml
@@ -105,6 +105,48 @@ FocusScope
 				}
 			}
 
+			Rectangle
+			{
+				id:					nightlyNoticeButton
+				color:				jaspTheme.rose
+				radius:				height / 2
+				height:				unstableNotice.height * 1.5
+				width:				unstableNotice.width  * 1.2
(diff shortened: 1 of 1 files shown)


JorisGoosen:
The nightlies page does start with a warning:
image
Maybe I should just make the "Use at your own risk" way bigger?

There is also the problem of `HEAD` indeed, the settings of buildbot look sensible to me and @RensDofferhoff but we have no idea why it keeps doing that.

Also there is still the problem where sometimes we did a special release for only flatpak from a different branch than stable/development which would not work well with this PR.

I do agree it would be awesome if the nightly would look different from the release.

I guess if the release always follows the pattern of:
- development changes go to stable
- we build the release from stable

@boutinb @RensDofferhoff

PR #5800, opened 2025-02-26
Tweak drag n drop and column type icons and tooltips

JorisGoosen:
bigger capture zones and easier clicking to edit the "..." in dragndrop.

also add description in tooltip of columns in variablelist and JASPColumn for https://github.com/jasp-stats/jasp-issues/issues/1190

### DIFF
--- a/Desktop/data/columnsmodel.cpp
+++ b/Desktop/data/columnsmodel.cpp
@@ -56,6 +56,11 @@ QString ColumnsModel::getColumnIcon(columnType colType) const
 	return VariableInfo::getIconFile(colType, VariableInfo::DefaultIconType);
 }
 
+QString ColumnsModel::getColumnDescription(const QString &name) const
+{
+	return provideInfo(VariableInfo::ColumnDescription, name).toString().trimmed();
+}
+
 QString ColumnsModel::getColumnIconTransform(int colType) const
 {
 	return getColumnIconTransform(columnType(colType));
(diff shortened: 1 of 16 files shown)


Desktop/data/columnsmodel.cpp
```diff
@@ -28,6 +28,7 @@ ColumnsModel::ColumnsModel(DataSetTableModel *tableModel)
 	connect(this, &ColumnsModel::labelsReordered,						info, &VariableInfo::labelsReordered	);
 	connect(this, &ColumnsModel::filterChanged,							info, &VariableInfo::filterChanged		);
 	connect(this, &ColumnsModel::dataSetChanged,						info, &VariableInfo::dataSetChanged		);
+	connect(this, &ColumnsModel::modelReset,							info, &VariableInfo::refresh			);
```
boutinb: This should not be necessary: the modelReset signal is already connected to the _resetModel in SourceItem, that resets the models bound to the columnsModel.
Why did you have to do this?

QMLComponents/controls/sourceitem.cpp
```diff
@@ -133,6 +133,7 @@ void SourceItem::connectModels()
 		connect(variableInfo,	&VariableInfo::labelsReordered,		controlModel, &ListModel::sourceLabelsReordered );
 		connect(variableInfo,	&VariableInfo::filterChanged,		controlModel, &ListModel::filterChanged );
 		connect(variableInfo,	&VariableInfo::columnsChanged,		controlModel, &ListModel::sourceColumnsChanged );
+		connect(variableInfo,	&VariableInfo::refresh,				controlModel, &ListModel::refresh );
```
boutinb, line 136: This should not be necessary...

boutinb (commented):
Just a small remark.
The rest is OK for me.

JorisGoosen:
At least one of those connections is needed to make sure that when  a column is changed the variablelist is refreshed so that the preview and or description of the column are shown in the tooltip.

So whenever a column changes in fact the variablelist needs to refresh.

Im not sure which connection was the right one because it isnt always easy to see which connection does what.
Maybe having them both is overkill, but it is in response to useractions anyway I think so I dont see the problem.

Desktop/data/columnsmodel.cpp
JorisGoosen: This one was apparently not the one I needed to add then :p

boutinb (approved):

PR #5803, opened 2025-02-28
Expose used varaibles in text area

boutinb:
Lavaan code in TextArea can use variables from the DataSet. These columns should be exposed, so that they can be used as source of other controls.

When adding this feature, I had to solve some issues:

- The resolution of the sources was not working properly when the source and target controls were created dynamically. The new test (Test TextArea) in the TestModule checks this with a TabView containing a TextArea wih Lavaan code which is source of a VariablesList (or a DropDown).
- The containsVariables and containsInteraction properties in List Controls (live VariablesList) were cached in each control. This was cumbersome and unneeded. The _containVariabels, _containInteraction and _termsAreInteractions are removed from the JASPListControl, and containsVariables and containsInteraction are functions that check each time the sources to see whether they contain variables or interactions.
- Also the width of TabView was not correctly set

@FBartos When this PR is merged, there is an example in jaspTestModule (Test Text Area) how to use the variables in a Lavaan TextArea as source for other controls. It gives also an example with a TabView with 2 columns.

### DIFF
--- a/QMLComponents/boundcontrols/boundcontrolrlangtextarea.cpp
+++ b/QMLComponents/boundcontrols/boundcontrolrlangtextarea.cpp
@@ -90,6 +90,13 @@ void BoundControlRlangTextArea::checkSyntax()
 		Log::log() << "Warning: no non prefixed entries returned from column encoder?";
 
 
+	if (!_textArea->initialized())
+	{
+		// Do not run the engine to check the script if the control in not yet initialized
+		_setBoundValues();
+		return;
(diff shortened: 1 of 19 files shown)


JorisGoosen (approved):
The translations for the preview are dropped because of this code. 
I think it is because parentlistview went from JASPControl to JASPListControl, but lets see.

Some other stuff isnt translated yet anyway so I hope it gets solved by translators and the autofill there

PR #5805, opened 2025-03-03
Use crash report

boutinb:
Use the crash report created in https://github.com/jasp-stats/jasp-issues/pull/3255

### DIFF
--- a/Desktop/mainwindow.cpp
+++ b/Desktop/mainwindow.cpp
@@ -1452,6 +1452,11 @@ void MainWindow::_openDbJson()
 
 void MainWindow::openGitHubBugReport() const
 {
+	static bool alreadyOpened = false;
+
+	if (alreadyOpened) return;
+	alreadyOpened = true;
+
 	bool openGitHubUserRegistration = false;
 
 	if(!Settings::value(Settings::USER_HAS_GITHUB_ACCOUNT).toBool())
--- a/Desktop/mainwindow.cpp
+++ b/Desktop/mainwindow.cpp
@@ -1457,7 +1457,8 @@ void MainWindow::openGitHubBugReport() const
 	if (alreadyOpened) return;
 	alreadyOpened = true;
 
-	bool openGitHubUserRegistration = false;
+	bool	openGitHubUserRegistration = false,
+			openBrowseFolder = false;
 
 	if(!Settings::value(Settings::USER_HAS_GITHUB_ACCOUNT).toBool())
 	{
@@ -1486,6 +1487,28 @@ void MainWindow::openGitHubBugReport() const
 	try			{ systemInfo << "* Commit used: " << AboutModel::commitUrl() << std::endl; }
 	catch(...)	{ systemInfo << "Commit couldn't be found\n"; }
 
+	try
+	{
+		if (!_preferences->logToFile())
+			debugInfo << tr("No log files are available. To get more information, please turn logging on. For this: open the file menu (the blue hamburger button left top), navigate to Advanced Preferences and check the 'Log to file' checkbox.") << std::endl;
+		else
+		{
+			QDir logDir(AppDirs::logDir());
+			QFileInfoList files = logDir.entryInfoList(QDir::Files, QDir::Time);
+
+			debugInfo << tr("Please drag and drop these log files into this issue: ") << std::endl;
+			for (const QFileInfo& file : files)
+			{
+				debugInfo << "* " << file.fileName() << std::endl;
+				if (file.fileName().contains("Desktop")) // The Engine log files are newer, the Desktop file is the oldest log file
+					break;
+			}
+			debugInfo << std::endl;
+			openBrowseFolder = true;
+		}
+	}
+	catch(...)	{ debugInfo << "No Log files path found"; }
+
 	try			{ debugInfo << "Debug information: " << _engineSync->currentStateForDebug() << std::endl; }
 	catch(...)	{ debugInfo << "No debug information found"; }
 
@@ -1505,6 +1528,9 @@ void MainWindow::openGitHubBugReport() const
 			{
 				QDesktopServices::openUrl(QUrl("https://github.com/join"));
 			});
+
+		if(openBrowseFolder)
+			MessageForwarder::browseOpenFile(tr("Drag and drop the log files"), AppDirs::logDir(), "Log files (*.log)");
 
 		emit exitSignal(1);
 	}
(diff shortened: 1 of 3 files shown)


Desktop/mainwindow.cpp
```diff
@@ -1470,42 +1470,28 @@ void MainWindow::openGitHubBugReport() const
...
 	{
-		QString percentEncodedIssue = QUrl::toPercentEncoding(tq(fillIt.str()));
+		QString systemInfoStr	= QUrl::toPercentEncoding(tq(systemInfo.str())),
+				debugInfoStr	= QUrl::toPercentEncoding(tq(debugInfo.str()));
 
-		const char * baseIssueUrl = "https://github.com/jasp-stats/jasp-issues/issues/new?labels=bug&title=JASP+crashed&body=";
+		QString baseIssueUrl = "https://github.com/jasp-stats/jasp-issues/issues/new?template=crash-report.yml&title=JASP+crashed";
 
-		QUrl issueUrl = baseIssueUrl + percentEncodedIssue;
+		QUrl issueUrl = baseIssueUrl + "&system-info=" + systemInfoStr + "&log=" + debugInfoStr;
```
shun2wang, line 1522: GitHub [rest api](https://docs.github.com/en/rest/using-the-rest-api/issue-event-types) does not support fill into issue body sub fields, maybe we can just also push .log files and debug info into clipboard and let user paste to here.

Desktop/mainwindow.cpp
```diff
@@ -1470,42 +1470,28 @@ void MainWindow::openGitHubBugReport() const
...
 
 		QDesktopServices::openUrl(issueUrl);
```
shun2wang, line 1524: This `QDesktopServices::openUrl(issueUrl);` will open 3 same issues(that's why we can saw 3 same issues reported by one user) so we should make sure it just open only once.

Desktop/mainwindow.cpp
boutinb: You can fill it with query parameters. Try this for example (I have generated this by forcing a crash in JASP):
https://github.com/jasp-stats/jasp-issues/issues/new?template=crash-report.yml&title=JASP+crashed&system-info=%2A%20JASP%20version%3A%200.19.3%0A%2A%20OS%20name%20and%20version%3A%20macOS%20Sequoia%20%2815.1%29%0A%2A%20Commit%20used%3A%20https%3A%2F%2Fgithub.com%2Fjasp-stats%2Fjasp-desktop%2Fcommit%2F3183d55fd4daea76858c14b3eefe25c1aeb05d07%0A&log=Debug%20information%3A%20%2A%2AEngine%20%23%200%2A%2A%20process%20%20is%20%2Arunning%2A%20and%20it%27s%20state%20is%20%2Aidle%2A.%0A%0A

Desktop/mainwindow.cpp
boutinb: I don't understand this... It opens only 1 issue...

shun2wang: I'm not sure where you created a crash, but the fact is that I assume our program crashes randomly. For example if I call `openGitHubBugReport()` inside the function of `MainWindow::setQmlImportPaths()` directly to simulate a crash qml slots , three identical issue reports will be opened, which is the same as many user-side crash reports.

Desktop/mainwindow.cpp
shun2wang: That's right, but the user still needs to actively find the .log file and upload it, but in fact we have not yet found any user who can effectively find and upload it. My intention is to automatically navigate to the folder where the .log is located (if the user has enabled logging) and push the file to the user's clipboard, or directly open that folder and guide the user to upload the log file. Otherwise we still receive invalid reports, which is no different from whether there is this PR or not😝

Desktop/mainwindow.cpp
shun2wang: So you can see: 
image

shun2wang: And maybe just something like this to wrapper the `openGitHubUrl` function to avoid to open github url too many times I guess.
```cpp

void MainWindow::openGitHubUrl(QUrl url) const
{
    static bool alreadyOpened = false;
    if (alreadyOpened) return; 
    alreadyOpened = true;

    QDesktopServices::openUrl(url);
}

```

Desktop/mainwindow.cpp
boutinb: Yes, that would be better if the log could be (semi) automatically added to the issue. But the first step is to use the right template, and this PR does this.

Desktop/mainwindow.cpp
boutinb: Oh I see... The setQmlImportPaths was apparently called 3 times: 1 during the main initialization, and 2 times in the initialization of the modules (you use apparently the JASP_TEST_BUILD setting, and you have only 2 modules?) 
But this is an artificial situation. The openGitHubBugReport is called only when an Engine crashed. Maybe the Desktop tried to restore the Engine, and it crashes just afterwards, or all the running Engines crashed simultaneously... In any case, we should get only 1 issue. I'll see what I can do to remove the redundant calls.

boutinb: Yes, your solution works, I have added it, thanks!

Desktop/mainwindow.cpp
shun2wang: How about this:
1. get path from `AppDirs::logDir`
2. read latest .log files as text stream
3. Push to clipboard or just append to issueUrl **OR** ask in the message box and auto open the `AppDirs::logDir` for user to drag drop to comments.

Desktop/mainwindow.cpp
JorisGoosen: I dont understand this conversation?

First the three issues was because of the artificial situation. Ok, so its not a problem.
But then Bruno says "remove redundant calls" and adds the code of Shun.

But why? What redundant calls?

Desktop/mainwindow.cpp
boutinb: We cannot copy the content of the log files, because the URL would be too long.
But I have tried to help the user to find the log files, by telling which files he should load, and by opening a finder at the logDir path.

Desktop/mainwindow.cpp
shun2wang: My example is not appropriate, but in fact there are indeed situations where multiple issues are opened at the same time, see my screenshot above from users.

JorisGoosen:
Ok this is already quite an improvement over before, ive made sure that the crashing of jasp-desktop itself also opens this now.

However, I dont get why we open a "open file" dialog for the logs?
Ill replace it with the same thing as in preferences advanced.

JorisGoosen (approved):

PR #5806, opened 2025-03-04
Use Group inside Checkbox in order to align sub-controls

boutinb:
If you have controls inside a Checkbox, these controls should be automatically aligned as if they were in a Group, that is that. the label and the field control (of TextFields and Dropdowns) should be aligned.


boutinb:
I have added in jastTestModule an analysis to test this (in Test Controls Layout)

JorisGoosen (approved):
Looks good! ^^

PR #5808, opened 2025-03-06
add thousands separators only on display

JorisGoosen:
when editing (or in a textinput in qml) it will not show thousands separators at all
left the deeuropeanise backup option in there.

results and data will still show those seps though

Also changed qmp to std::map

This should also fix the problems discovered in https://github.com/jasp-stats/INTERNAL-jasp/issues/2802 I think


QMLComponents/utilities/qutils.cpp
```diff
@@ -353,6 +354,11 @@ bool QColumnUtils::isDoubleValue(const QString &value)
 	return ColumnUtils::isDoubleValue(fq(value));
 }
 
+void QColumnUtils::setNumberOptionsOnQLocale(QLocale & locale)
+{
+	locale.setNumberOptions(QLocale::OmitGroupSeparator);
```
shun2wang, line 359: Now can we get rid of the thousands separator by default if we add this in the settings? otherwise I'm afraid our users still can't get used to the sudden appearance of thousands separators in the next version.

JorisGoosen: We could also add a setting for removing it completely yes.
But this seems like overkill?

JorisGoosen: im adding it

JorisGoosen:
Ok almost ready for review, just some small irregularities in the amount of decimals

JorisGoosen:
(which im fixing)

JorisGoosen:
~~ I cant fix it @koenderks the change to make the "ids" not have thousands separators at all by respecting "string" as an actual type and not secretely converting it to a number causes the machine learning tables to have a different output:
image (in "explained proportion bla bla") ~~

It used to be formatted according to a fixedFormat (bringing it down incidentally to 2 decimals) but apparently now it gets 3 decimals. While this isnt very wrong it does give a bunch of errors in the jasp-unit test.

~~Not only there though...~~

~~Probably could be fixed with a "mixed" type?~~
~~But the easiest would be to just try to convert any cell-content to number anyway, even if the r-analysis-writer wanted `string`. (Like it was before this PR, but it will then introduce thousands separators in the ID unless disabled as a setting) ~~

~~I think having thousands separators in the ID is not so bad. So I will push that change after testing that it actually lets more files pass the test but would be good to think about this~~

Ok that doesnt actually solve it.

JorisGoosen:
Hmmm...

This (transposed) column gets two formats "sf:4" and "dp:3"
It seems with the previous formatting code the `dp` was used preferentially over `sf` in that it would format `0.341682345` to `0.342` and now it does `0.3417`.

It seems more anlyses are hit by this.
But given that the formatting code is a mess it is maybe a good one to think about?
Like, 4 significant figures with a number `0<x<1` and 3 decimal points is a conflicting requirement as far as I can see, so which one has precedence?

JorisGoosen:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2810

boutinb:
Does not work fine:
Screenshot 2025-03-13 at 17 13 52


Also in QML the thousand separator does not appear anymore even if it is set in the preferences:
Screenshot 2025-03-13 at 17 15 15

JorisGoosen:
> Also in QML the thousand separator does not appear anymore even if it is set in the preferences:

This is on purpose.
Because I wanted to remove thousands separators everywhere we edit.
And to get the textinputbase to work correct is hard


But the gigantic contExpon numbers looks like a case where maybe significant figures ought to have some effect yeah :p

boutinb:
It works fine for me. But maybe @koenderks should check it, because I have some troubles to reproduce his issue in https://github.com/jasp-stats/INTERNAL-jasp/issues/2802.

JorisGoosen:
Ah yeah, he totally should!

But there are still other issues and I want to make sure I dont mess up anything foreseeable.

koenderks:
I only see one issue:

1. The value `0` gets displayed in scientific notation now?

image

JorisGoosen:
Yeah im trying to get it all to work but the code is a bit of a mess...

JorisGoosen:
Should fix https://github.com/jasp-stats/INTERNAL-jasp/issues/2816 too

JorisGoosen:
Ah there is still a problem with zero in display of significant figs

JorisGoosen:
Ok this is ready for review again.

boutinb:
It seems to work quite well, but if I check the Builtit_Monetory JASP file from Koen, then I don't get exactly the same output:
That was in 0.19.3:
Screenshot 2025-03-27 at 16 54 02
En that is now:

Screenshot 2025-03-27 at 16 55 05
This is OK for me, but @koenderks should say if he is happy with this change....

koenderks:
I don't really see what is different in that output @boutinb?

boutinb:
> I don't really see what is different in that output @boutinb?

The value column. Oh but this is a money type. So I suppose this is OK.
Let's merge this. Or do have @shun2wang & @vandenman still some comments?

JorisGoosen:
There might be some cases still where the zero is displayed slightly different then before, but im not sure what would be "correct" in any case.

PR #5809, opened 2025-03-06
A FormulaField may get a numeric value

boutinb:
The FormulaArrayType is never used, and is removed.

A FormulaField test is added in the Min/Max Analysis of the jaspTestModule

### DIFF
--- a/QMLComponents/controls/textinputbase.cpp
+++ b/QMLComponents/controls/textinputbase.cpp
@@ -88,8 +88,7 @@ void TextInputBase::bindTo(const Json::Value& value)
 	}
 	case TextInputType::FormulaType:
 	{
-		_value = value.isString() ? tq(value.asString())
-								  : (value.isNumeric() ? QColumnUtils::doubleToString(value.asDouble()) : QVariant());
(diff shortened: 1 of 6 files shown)


QMLComponents/controls/textinputbase.cpp
```diff
@@ -87,19 +87,14 @@ void TextInputBase::bindTo(const Json::Value& value)
...
 	}
 	case TextInputType::FormulaType:
-	case TextInputType::FormulaArrayType:
 	{
-		QString strValue;
-		if (value.isString())	
-			strValue = tq(value.asString());
-		_value = strValue;
+		_value = value.isString() ? tq(value.asString())
+								  : (value.isNumeric() ? QColumnUtils::doubleToString(value.asDouble()) : QVariant());
```
JorisGoosen: Why do you convert the double to a string?

JorisGoosen: Anyway, this breaks it once you switch the locale to french

JorisGoosen: Ive fixed it for you

JorisGoosen (approved):

PR #5821, opened 2025-03-20
Initial windows container

RensDofferhoff:
Sandboxing is off by default.
Select Developer Mode and the setting will appear in advanced settings.
Restart JASP.

Creates a folder in User documents for file IO in engines.
Throws a warning if users use something else

Needs a lot of testing due to the amount of changes to IPC


RensDofferhoff:
I have not yet seen any issues so I will make this a normal PR

RensDofferhoff:
It needs more testing obviously but If you are not unhappy with the code  we can merge it and just see over a long period of time

vandenman:
I can't quite test this since I don't use windows. Is there no sandbox on Linux? :disappointed:

RensDofferhoff:
> I can't quite test this since I don't use windows. Is there no sandbox on Linux? 😞

On Linux we would sandbox using flatpak.
On Mac it might be possible using a different mechanism and is always limited to dirs you have given permissions too already.

RensDofferhoff:
Maybe @FBartos could see if the latest windowsContainer nightly works for him?

shun2wang (commented):
I have Windows too. for me this PR stoped engine at `_channels.resize(maxEngineCount());` in `enginesync.cpp`, a error: `Fatal error: unable to initialize the JIT`, log: 
```
onecore\base\appmodel\appcontainerregistration\appcontainerregistration.cpp(624)\kernelbase.dll!00007FFAAF21DE3A: (caller: 00007FFAAF21AB41) ReturnHr(1) tid(1704) 800700B7
```

And according to https://learn.microsoft.com/en-us/uwp/schemas/appxpackage/uapmanifestschema/element-application#combinations-of-activation-info-attributes should we also add some settings to manifest while installed from MS-Store or run as .msix?

RensDofferhoff:
I will take a look at the msix now thanks! Might require a different solution.
Could you test the msi or zip?

JorisGoosen:
Im downloading the zip now

RensDofferhoff:
Maybe I should Throw up a message box to notify people the sandbox (fs) stuff is initializing. 
Just tried an laptop with hdd and that makes it rather slow

JorisGoosen:
It seems to work totally fine on my system, and if I try to access stuff somewhere in my home I get a nice permission denied in the r commander.

Awesome!

Ill have to check your changes in detail to come up with some sort ofd edgecase to test though.

RensDofferhoff:
@JorisGoosen  did first boot take a long time ?

Also could you verify `url.show("https://www.stats.ox.ac.uk/pub/datasets/csb/ch3a.txt")`
Also does not work for you ? (it should not).

Any feedback on the file transfer stuff for fileselector is welcome. Its a bit awkward for some users perhaps.

JorisGoosen:
> It needs more testing obviously but If you are not unhappy with the code we can merge it and just see over a long period of time

Ill be done with the review today or tomorrow I think

JorisGoosen:
> @JorisGoosen did first boot take a long time ?

I didnt really notice anything, there was a little window that popped up for a bit, but didnt take too long.
I did however install it on a SSD on a pretty fast machine.

> Also could you verify `url.show("https://www.stats.ox.ac.uk/pub/datasets/csb/ch3a.txt")` Also does not work for you ? (it should not).

Yes it does not work, although it doesnt work because its not allowed to change working directory (presumably to store the downloaded file). `system(dir)` also fails the same way.

Because I assume this sandbox also restricts tcp/ip ive checked:
```
uhm <- url("https://www.stats.ox.ac.uk/pub/datasets/csb/ch3a.txt")
open(uhm)
```

And that fails because its not allowed to connect to the URL, so perfect!

> Any feedback on the file transfer stuff for fileselector is welcome. Its a bit awkward for some users perhaps.

The file selector thing needs some love for sure. It would for instance be better if the user could just select whatever, then we let engine write to tempfiles. And then the engine copies it to the specified location. Definitely nicer that way.

I am also still getting some crashes on closing my workspace. But I havent really found a repro path yet.

Also cochrane managed to get stuck when trying to export a file from there through the fileselector. But now that I try this some more this might just be cochrane related. It also gets stuck if I dont export a file

Common/processinfo.cpp
```diff
@@ -90,3 +90,35 @@ bool ProcessInfo::isParentRunning()
...
+
+	if (!::GetTokenInformation(
+			tokenHandle,
+			TOKEN_INFORMATION_CLASS::TokenIsAppContainer,
+			&isAppContainer,
+			tokenInformationLength,
+			&tokenInformationLength
+			))
+	{
+		return false; //if this fails we just assume this is not a container
```
JorisGoosen, line 119: Add std::cerr print or something

Common/processinfo.cpp
```diff
@@ -90,3 +90,35 @@ bool ProcessInfo::isParentRunning()
...
+	HANDLE tokenHandle{};
+	DWORD isAppContainer{};
+	DWORD tokenInformationLength{ sizeof(DWORD) };
+
+	if (!::OpenProcessToken(
+			GetCurrentProcess(),
+			TOKEN_QUERY,
+			&tokenHandle))
+	{
+		return false; //if this fails we just assume this is not a container
```
JorisGoosen, line 108: print this

Common/processinfo.cpp
JorisGoosen: ah this is a check, ok maybe not then

Desktop/components/JASP/Widgets/FileMenu/PrefsAdvanced.qml
```diff
@@ -407,11 +407,24 @@ ScrollView
...
-				KeyNavigation.tab:	showEnginesWindow
+				KeyNavigation.tab:	engineSandbox
 				activeFocusOnTab:			true
 				text:				qsTr("Maximum number of engines: ")
 			}
 
+			CheckBox
+			{
+				id:					engineSandbox
+				visible:			Qt.platform.os === "windows" && preferencesModel.developerMode
```
JorisGoosen:
```suggestion
				visible:			Qt.platform.os === "windows" && (preferencesModel.developerMode || preferencesModel.engineSandbox)
```

Unless ofc it cant ever be enabled if developerMode is off

QMLComponents/utilities/messageforwarder.cpp
```diff
@@ -30,6 +30,55 @@ bool MessageForwarder::useNativeFileDialogs()
 	return DesktopCommunicator::singleton()->useNativeFileDialog();
 }
 
+bool MessageForwarder::engineSandbox()
+{
+	return DesktopCommunicator::singleton()->engineSandbox();
+}
+
+QString MessageForwarder::constrainToSandboxResult(const QString &selectedPath, bool file, bool save)
```
JorisGoosen, line 38: So I really think better behaviour is:

That path the user selects is what they get.
In jaspBase we add  `jaspFileWriter` or something. It simply stores whatever the R code gives it into our temp dir.
Then sends a special entry back in the results that isnt shown in the results at all, but will tell desktop about the file and lets it copy it there. It can have a reference to the option that contains the path, or the path or whatever.

I can help out with setting this up.

For now this would  be ok in `development` but if we release it like this I think its a bit too ugly.

Tools/CMake/R.cmake
```diff
@@ -669,6 +669,13 @@ elseif(WIN32)
       ${HASH_INDEX}
       R_PACKAGE_HASH)
 
+      # Remove this before merging
+      set(R_VERSION_NAME "R-4.4.3-win-containerable")
```
JorisGoosen, line 673: Did you manage to open an issue at the R tracker?

JorisGoosen (approved):
Overall looks good and works mostly fine (except for the random crashes at closing a workspace) but I think with a little bit of effort we can have it way nicer by doing the file copy.

Im not sure if Id wanna block the PR on those grounds though.

The crash that Shun experiences does sound bad though

Desktop/components/JASP/Widgets/FileMenu/PrefsAdvanced.qml
shun2wang: Also if user check it on/off should be better to prompt a message box to ask user restart JASP.

Desktop/components/JASP/Widgets/FileMenu/PrefsAdvanced.qml
```diff
@@ -407,11 +407,24 @@ ScrollView
...
+				enabled:			Qt.platform.os === "windows" && preferencesModel.developerMode
+				label:				qsTr("Sandbox engines")
+				checked:			preferencesModel.engineSandbox
+				onCheckedChanged:	preferencesModel.engineSandbox = checked
+				toolTip:			qsTr("Strengthen security on Windows by isolating Engines running R-code")
+
+				KeyNavigation.tab:		showEnginesWindow
```
shun2wang, line 425: If developerMode is off then tab navigation cannot jump to the next element.

shun2wang (commented):
I tried zip version, it not crash engine, but I can access and save all files everywhere, maybe because I am run in a admin user?

RensDofferhoff:
> I tried zip version, it not crash engine, but I can access and save all files everywhere, maybe because I am run in a admin user?

Thats correct! It just sandboxes the jaspEngine.

QMLComponents/utilities/messageforwarder.cpp
RensDofferhoff: We should discuss this tomorrow/today a bit.
I do agree with you this is a bit too jank for your average user.

shun2wang (approved):

FBartos:
I ran a couple of analyses and things seemed to work alright!

JorisGoosen:
I made a small change and am OK with the changes. Couldnt test on windows now though, but I trust its ok.

PR #5824, opened 2025-03-25
Add access to factors form variables lists and their titles

boutinb:
The Variables Lists (and their titles) inside a Factors Form, should be accessible to act as a source of other components.

See the Test Factors Form in the Test Module (Simple Factors Form Section)

This feature will be used by @juliuspfadt


JorisGoosen:
If I do:
- edit one factor title
- add some columns to it
- add a second factor
- select the second factor in the dropdown for one of the columns
- change the second factor title to something
I get:
image

It changed the selected option back to the first one btw

boutinb:
@JorisGoosen it should work now.

JorisGoosen:
Same problem, but now with the entries in the tab you are changing the name of.

I mean, this is not very important really. So I guess if you dont feel like fixing it you shouldnt and just merge it.

PR #5826, opened 2025-03-26
Rowwise functions added to Drag N Drop and auxiliary improvements

JorisGoosen:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2616 and also adds row SD vriance etc

Also tried to improve the behaviour of dropping stuff into the drag n drop and inserting stuff at the right (or left) place.
If a column wouldnt be accepted into something with its natural type it will use either the user selected type or whatever works.

This all mostly works as intended.

Although the `row*` functions in JASP will need some love.
Right now they work great for columns, but sadly columns only.
They cant handle getting single string/number inputs, so I guess some R-programmer should have a look?
@JohnnyDoorn  for instance?

Ill also trigger some builds.

### DIFF
--- a/Desktop/components/JASP/Widgets/ComputeColumnWindow.qml
+++ b/Desktop/components/JASP/Widgets/ComputeColumnWindow.qml
@@ -185,6 +185,15 @@ FocusScope
 					ListElement	{ type: "function";	friendlyFunctionName:	"";						functionName: "round";	functionParameters: "y,n";		functionParamTypes: "number,number";			toolTip: qsTr("rounds y to n decimals") }
 					ListElement	{ type: "function";	friendlyFunctionName:	"";						functionName: "length";	functionParameters: "y";		functionParamTypes: "string:number:boolean";	toolTip: qsTr("returns number of elements in y") }
 					ListElement	{ type: "function";	friendlyFunctionName:	"";						functionName: "median";	functionParameters: "values";	functionParamTypes: "number";					toolTip: qsTr("median") }
+					
+					ListElement	{ type: "rowfunction";	friendlyFunctionName:	"";	functionName: "rowMean";		toolTip: qsTr("Rowwise mean") }
+					ListElement	{ type: "rowfunction";	friendlyFunctionName:	"";	functionName: "rowSum";			toolTip: qsTr("Rowwise sum") }
+					ListElement	{ type: "rowfunction";	friendlyFunctionName:	"";	functionName: "rowSD";			toolTip: qsTr("Rowwise standard deviation") }
+					ListElement	{ type: "rowfunction";	friendlyFunctionName:	"";	functionName: "rowVariance";	toolTip: qsTr("Rowwise variance") }
+					ListElement	{ type: "rowfunction";	friendlyFunctionName:	"";	functionName: "rowMedian";		toolTip: qsTr("Rowwise median") }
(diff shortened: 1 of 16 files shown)


boutinb:
In the Filter constructor window, to make a formula is indeed easier, but throwing some elements in the bin does not seem to work anymore.

JorisGoosen:
What do you mean "some"?

JohnnyDoorn:
I'll gladly take a look - I saw there are some Windows nightlies with this feature but no Mac version, is that still on the way? Because I'm having some troubles building JASP at the moment.. 
And I also couldn't find where the R-code is for these functions in your changed files - where should I look?

JorisGoosen:
Its just the same R code as before: https://github.com/jasp-stats/jaspBase/blob/31bc23b340ebb9298c03b67108de300cdc3f9f57/R/friendlyConstructorFunctions.R#L154

Uhm, not sure why the mac nightlies didnt build. Ill have a look after the meeting

JorisGoosen:
Ive fixed the trashcan and triggered new builds. Probably the mac one will work this time

boutinb (approved):

JohnnyDoorn:
I've been playing around with [one of the latest nightlies](https://github.com/jasp-stats/jasp-desktop/commit/1034a2bf80a7eb2e40b081580ddbb7b13ff4b3b7) and looks really fantastic! I think this will really help the people that like computing lots of sumscores and whatnot.. What issue did you specifically refer to @JorisGoosen, with the single string input, because I haven't been able to break it with the debug data set :p any useless type gets converted to numeric and therefore NA and therefore filtered out it seems

JorisGoosen:
Yeah when you do: 
image
```
generatedFilter <- ((rowSumNaRm(V1.scale, 2) > 3))
```
Gives error:
```
data length [101] is not a sub-multiple or multiple of the number of rows [51]
```

JohnnyDoorn:
Ahhh on that bike.. Do we even want it to be possible to use a single value in those row-wise functions? Because what would that even mean/ be the user expectation? That the number 2 is repeated for every row? Because being able to specify a single value rather than a column causes the matrices that are created to be have fewer rows than there are rows (it concatenates all values, which here is 101, with ncol equal to the number of variables that are entered), and so also the output of rowSums will then yield a different number. So if you do
`rowSumNaRm(V1.scale, contNormal.scale, 2)`

You get a matrix with 67 rows (and so 67 row sums), which will not match the existing data, but you get no error (because you have 201 values, which can be neatly divided over 3 columns of 67 rows), just some missingness, which might be even more dangerous:
image

JorisGoosen:
> That the number 2 is repeated for every row?

This is what I expect, a naive non-statistician.

> Because being able to specify a single value rather than a column causes the matrices that are created to be have fewer rows than there are rows (it concatenates all values, which here is 101, with ncol equal to the number of variables that are entered), and so also the output of rowSums will then yield a different number.

This sounds a bit bonkers to me.
Which is to say: the functions we made are called "rowSomething" which I interpret as basically doing:
```
for loop on each rownumber:
	apply something to everything in this "row" 
```
It seems to me that if you enter a "constant" (a number or string) it would be part of every row.

And I sure as hell dont expect ity to output anything else then either an empty vector or a vector with the results of the "something" with exactly as many rows as the dataset.

JohnnyDoorn:
I agree it's bonkers, which is why I'd suggest to only allow variables to be specified in those row functions

JorisGoosen:
But why not allow a constant to be used across each row?

PR #5836, opened 2025-04-08
More threads while loading stuff

JorisGoosen:
Should speed up the loading of files with many columns quite a bit by using multithreading.
Also use a map for a particular lookup in the "setValue" of Column should make all loading quicker.

should do quite a bit towards https://github.com/jasp-stats/jasp-issues/issues/3038 and https://github.com/jasp-stats/INTERNAL-jasp/issues/2538

I do notice that editing in JASP with such a big file becomes a pain. 
This is probably because the undostack is taking too much memory.
And of course operations for that memory cause its all a big pile of json::value and classes.

This also messes with the JASP_TIMER profiling so I guess Ill remove that later.
And figure out how to set callgrind up.


shun2wang:
How many steps now our `undostack` will stored? Maybe we can just store only 10 or less setp in the stack.

JorisGoosen:
> How many steps now our `undostack` will stored? Maybe we can just store only 10 or less setp in the stack.

All the steps, yeah could be a good idea to do so.

But it would probably be better to see if doing each steps requires less operations.
Now for instance editing a cell in the data-editor can trigger some labelrelated operations on the background. So we sometimes need to store a lot or write entirely symmetric data-setting functionality. 
First ofcourse I should figure out where the problems are exactly.
But people cant edit if they cant load so this PR comes first I suppose.

JorisGoosen:
Ah this should probably be a draft given that I find an unhooked JaspEngine process on my laptop running 800% cpu without a jasp instance in sight...

At least I know why my lap is warm

JorisGoosen:
Ok, so im going to also improve editing speed and some of these things.
I am now testing if things run better if I dont do the whole labelsTemp dance, it seems to be ok so far.

But some more development on this is in order Im afraid...

JorisGoosen:
Okidoki, I think this might be ready for review again.

Maybe @koenderks is also interested in giving it a go?
This should give speedups ranging from a bit quicker to insanely much more quicker (with lots of columns for instance, or with editing a single value when there are hundreds of thousands in the column).

JorisGoosen:
Ill turn on some nightly builds as well!

Edit: ive turned them on alright but the buildbots aint building it seems.

CommonData/column.h
```diff
@@ -19,9 +19,9 @@ class Analysis;
 /// Anything with a label on it will have a Label-object in Column and the "intsId" of this label is entered in the corresponding _ints row.
 /// If the originalValue of that label is convertible to double/int it will *also* be stored in _dbls. (Which is different from how <0.19 JASPs did it)
 /// 
-/// If no label exists _ints simply contains Label::DOUBLE_LABEL_VALUE (-1) and it tells JASP that _dbl should be used.
+/// If no label exists _ints simply contains Label::NO_LABEL (-1) and it tells JASP that _dbl should be used.
 /// We do want users to be able to edit them, or to set "filter allows" or something on it.
-/// To this end labelsTempCount() can be called to get the total of "labels" a column has.
+/// To this end labelsNonEmptyCount() can be called to get the total of "labels" a column has.
 /// The shown labels are stored in a temporary internal representation (stringvec).
```
boutinb, line 25: This comment should be removed or updated

CommonData/dataset.cpp
```diff
@@ -639,23 +651,5 @@ stringset DataSet::findUsedColumnNames(std::string searchThis)
 	return columnsFound;
 }
 
-bool DataSet::initColumnWithStrings(int colIndex, const std::string & newName, const stringvec &values, const stringvec & labels, const std::string & title, columnType desiredType, const stringset & emptyValues, int threshold, bool orderLabelsByValue)
```
boutinb, line 642: Hmm: this function is still needed in the syntaxbridge.cpp!

CommonData/column.h
```diff
@@ -254,49 +244,48 @@ class Column : public DataSetBaseNode
...
+									_highestIntsId				= -1;
             stringvec				_nonFilteredLevels;
 			int						_nonFilteredNumericsCount	= -1;
-			bool					_invalidated		= false,
-									_autoSortByValue;
-			dropLevelsType			_dropLevels			= dropLevelsType::noChoice;
-			computedColumnType		_codeType			= computedColumnType::notComputed;
+			bool					_invalidated				= false,
+									_autoSortByValue,
+									_hasShadows					= false;
```
boutinb, line 260: I think this member is never set to true

boutinb (changes requested):
I have some questions about _labelsByValue and _labelsByDisplay: they map a string (value or display) to a vector of label objects. I can see the use of having the same label value (the display) for different values, so a same display can have multiple label objects, but a value should not be associated with multiple label objects, no?
Also if the value or display is changed, the vector of labels (of _labelsByValue or _labelsByDsiplay) can become empty. But the mapping is not removed. In which case this is needed?

boutinb:
Also I get some errors. Take the debug.csv, and use the contBinom column
. Add a new level '2' in the label editor, and then change the first row of contBinom from 1 to 10: the label 2 becomes 10 in the Lebel Editor.
. Load again debug.csv, change the first row of contBinom to 2: you get a new label 2 in the Label Editor. if you add a new level '3' in the label editor, and change back the 2 to 1 in the data, the labels 3 and 2 are removed (this happens with the keeping levels on or off). 
. Load again debug.csv, change the first row of contBinom from 1 to 2: the label 2 appears in the Label Editor. Set it back from 2 to 1: it disappears. Change the second row of contBinom from 0 to 2: you get the 63263438 value instead!

CommonData/dataset.cpp
JorisGoosen: No we just need to rewrite that code, also, there is a workaround function in datasetpackage that shows how.
But anyway, not using big vectors to copy shit is waaaaaay more efficient, so its better to use the "initWithLookup" functions

CommonData/column.h
JorisGoosen: No you are right, I forgot finishing this part of it, oops!

It turned out that recalculating the width of a column on each edit can actually take up quite a lot of processing time. But is totally unnecessary if proper admin is done.
However, given that I forgot this variable I guess it is now messed up if there are indeed shadows (different value from labeldisplay)

JorisGoosen:
> I have some questions about _labelsByValue and _labelsByDisplay: they map a string (value or display) to a vector of label objects.

Well theyve beend added because looking up labels was taking insane amount of times on many-row-datasets.
by display isnt really used right now I think, but it needed to be updated at the same time as the value one so I figured I can just add it.

> I can see the use of having the same label value (the display) for different values, so a same display can have multiple label objects, but a value should not be associated with multiple label objects, no?

Actually, righyt now in JASP a level  or label is defined by both value and display. This is why `_labelsByValDisplay` exists.
The reason the mapping is to multiple labels is that it is temporarily possible for having more than 1 label per key.

Suppose the user has two label: `0-f` `1-m` and they change the value to 0 for `m`: now two labels have the same value. 
Which is fine, if used as ordinal you will just have a different amount of levels then you would have as when it is as scale.
Same for the other way around.

The only thing ive been trying to prevent is having 2 levels with the same value and displaylabel. Because I cant see a scenario where that makes any sense. So after a manual edit `labelsMergeDuplicateInto` is called for each label that has duplicates is merged into the one given. Although now I look at that function it isnt using the valdisplay mapping. So maybe I can actually speed that up too. But it didnt seem to be too much of a drag for now.

> Also if the value or display is changed, the vector of labels (of _labelsByValue or _labelsByDsiplay) can become empty. But the mapping is not removed. In which case this is needed?

It isnt "needed" per se, but it is hardly a problem to have some empty vectors in the mapping. It will get reset at some point and it isnt leaking any memory.

and to clean it I would have to check everytime if I remove something if that was the last and then add an erase. But why?
I dont think those few bytes are going to be a problem.

JorisGoosen:
Ill fix the bug(s) you found btw, thanks for testing ^^

JorisGoosen:
@boutinb Ive fixed all the bugs you found ^^

boutinb:
It works better.
But now, if I add a level, I cannot see it in the result, even if I set/unset the 'Eye' icon 'Keeping levels':
Screenshot 2025-05-01 at 15 13 52

JorisGoosen:
Ah the order of newly added labels was also wrong.

It all seems good now, and ive made sure to add a kill timer to the busy loop of sqlite, not sure if we need it but I prefer an engine crashing over it getting stuck in the background I suppose.

Desktop/data/importers/importer.cpp
```diff
@@ -13,75 +14,127 @@ Importer::Importer()
...
 void Importer::syncDataSet(const std::string &locator, std::function<void(int)> progress)
 {
-	_synching = true;
-	long timeBeginS = Utils::currentSeconds();
-	
-	ImportDataSet *	importDataSet	= loadFile(locator, progress);
-	bool			rowCountChanged	= importDataSet->rowCount() != DataSetPackage::pkg()->dataRowCount();
+					_synching			= true;
+					_progressCallback	= progress;
+	long			timeBeginS		= Utils::currentSeconds();
```
shun2wang: I'm afraid that maximum size on MSVC and GCC is inconsistent, maybe which related to the engine timeout limit cause crash on Windows only(https://github.com/jasp-stats/INTERNAL-jasp/issues/2828)? anyway, I recommend using `int32_t/int64_t` or something everywhere.

I'm not referring in particular to here, but to the time function related to the engine.

JorisGoosen: I guess we could also just use the `std::chrono` datatypes now?

Interesting idea that it could be related to the timer., not sure if thats it though.

shun2wang: > I guess we could also just use the `std::chrono` datatypes now?
> 

I think so.

boutinb:
It works really fine for me now!

boutinb (approved):

JorisGoosen:
I want to change the chrono thing though!

JorisGoosen:
Ill do that now

shun2wang:
> I want to change the chrono thing though!

Once done then I will also want to play this PR.

JorisGoosen:
Go ahead!

Ill see if a nightly wants to build

JorisGoosen:
Ah, I missed some commit in columnEncoder it turns out...

Didnt realize utils was there

shun2wang:
I cannot build this on Windows for now.

EDIT: It is missing `initColumnWithStrings`.

CommonData/dataset.cpp
shun2wang: I cannot build on Windows without this functions in `syntaxbridge.cpp`, there are `fatal error LNK1120` on MSVC.

JorisGoosen:
Ill fix it!

JorisGoosen:
That should fix it

JorisGoosen:
This is ready for review and merge as far as im concerned, perhaps @shun2wang can let know if it now runs for him?

shun2wang:
> This is ready for review and merge as far as im concerned, perhaps @shun2wang can let know if it now runs for him?

I'm building this PR and will leave it for some more hours to see if engine running well, maybe tomorrow.

JorisGoosen:
Well shun, maybe we just merge it and be done with it.

If this doesnt solve it, then we can make a PR devoted to fixing the windows problem.

PR #5837, opened 2025-04-08
Add MetaSem syntax

boutinb:
Add TextTypeMetaSem type for TextArea
As for Lavaan or CSem TextArae, this will check the syntax of the text with the jaspMetaAnalysis::checkMetaModel function.

The source of the variable must be specified.
As the variables are composed by 2 (or more?) variables, they must be separated by a string: this string can be set the variableSeparator property.

@FBartos Should the variables be encoded? (this means they must be form the DataSet)


FBartos:
> @FBartos Should the variables be encoded? (this means they must be form the DataSet)
the separated variable names should not be encoded - their names would not correspond to anything in the data set, the original non-separed variable name is going to be encoded from the variable form already

boutinb:
@FBartos: the checkMetaSemSyntax branch is build on Windows

PR #5850, opened 2025-04-15
Add Key Label to Tab View

boutinb:
For a new SEM analysis build by @juliuspfadt, it becomes necessary to use key value and key label in Tab Views.
Until now, only Dropdown could use values that contain value and translatable label:
```
DropDown
{
...
values: [{value: "one", label: qsTr("One")}, {value: "two", label: qsTr("Two")} ]
} 
```
This kind of values would be in fact also handy for Tab Views, since the name of each tab should not automatically be the same as the displayed value (which should be translatable).
Moreover these kind of values should be transferable from one control to another one (by using a source).
For this I had to extend the Term object (the basic object handled by the models), to contain the value and label.
I took advantage of this change to rename some methods that needed some refresh, and to remove unnecessary models:
. the listModeLabelValueTerms was used specifically for DropDown, but it is not anymore relevant
. the listModelInteractionAvailable is not used anymore
. As the listModelTermsAvailable is then the only subclass of the listModelAvailableInterface, these 2 classes are merged (to listModelTermsAvailable).

An extra test has been added in the Tab View analysis of the jaspTestModule.


QMLComponents/boundcontrols/boundcontrolbase.cpp
```diff
@@ -117,44 +117,49 @@ std::string BoundControlBase::getName() const
...
+void BoundControlBase::_readTableValue(const Json::Value &value, const std::string& keyValue, const std::string& keyLabel, bool hasMultipleTerms, Terms& terms, ListModel::RowControlsValues& allControlValues, const Terms& sourceTerms)
 {
 	for (const Json::Value& row : value)
 	{
-		Json::Value keyValue = row[key];
-
-		Term term = Term::readTerm(keyValue);
+		Term term = Term::readTerm(row[keyValue]);
+		if (!keyLabel.empty() && row.isMember(keyLabel) && row[keyLabel].isString())
+			term.setLabel(tq(row[keyLabel].asString()));
```
JorisGoosen: This should be done in Term!

Also instead of Term::readTerm I suggest just making proper constructors for Term.
Then you can give json:Value or string or double.
The term class can then contain the logic for getting labels out.

Why is all that stuff done in this file?

QMLComponents/boundcontrols/boundcontrolmultiterms.cpp
```diff
@@ -66,7 +66,7 @@ void BoundControlMultiTerms::resetBoundValue()
 	for (const Terms& terms : tuples)
 	{
 		Json::Value rowValue;
-		for (std::string val : terms.asVector())
+		for (std::string val : terms.valuesAsVector())
```
JorisGoosen, line 69:
```suggestion
		for (const std::string & val : terms.valuesAsVector())
```

QMLComponents/boundcontrols/boundcontrolterms.cpp
```diff
@@ -284,8 +285,8 @@ bool BoundControlTerms::areTermsInOption(const Json::Value &option, Terms &terms
 	Terms termsToSearch = terms;
```
JorisGoosen: Terms is a class right?

Shouldnt this kind of search function be on that class then?

QMLComponents/boundcontrols/boundcontrolterms.cpp
```diff
@@ -297,8 +298,8 @@ Terms BoundControlTerms::_getTermsFromOptions(const Json::Value& option) const
 	Json::Value valueOption = _isValueWithTypes(option) ? option["value"] : option;
```
JorisGoosen, line 291: Maybe we should also consider making a proper Option type again?

QMLComponents/controls/comboboxbase.cpp
```diff
@@ -30,58 +30,36 @@ ComboBoxBase::ComboBoxBase(QQuickItem* parent)
...
 	int index = -1;
 
-	if (values.size() > 0)
+	if (terms.size() > 0)
 	{
 		if (selectedValue.empty())	index = 0;
 		else
 		{
-			auto itr = std::find(values.begin(), values.end(), selectedValue);
+			index = terms.indexOfValue(tq(selectedValue));
```
JorisGoosen, line 46: You already made these search functions :p

QMLComponents/models/term.cpp
```diff
@@ -28,22 +27,30 @@ const char * Term::separator =
...
 
 
-Term::Term(const std::vector<std::string>	components, const columnTypeVec&	types)	{ initFrom(tq(components),	types);		}
-Term::Term(const std::string				component,	columnType				type)	{ initFrom(tq(component),	type);		}
-Term::Term(const QStringList				components, const columnTypeVec&	types)	{ initFrom(components,		types);		}
-Term::Term(const QString					component,	columnType				type)	{ initFrom(component,		type);		}
+Term::Term(const std::vector<std::string>	& components,	const columnTypeVec	&	types)	{ initFrom(tq(components),	types);		}
+Term::Term(const std::string				& component,	const columnType		type)	{ initFrom(tq(component),	type);		}
+Term::Term(const QStringList				& components,	const columnTypeVec	&	types)	{ initFrom(components,		types);		}
+Term::Term(const QString					& component,	const columnType		type)	{ initFrom(component,		type);		}
```
JorisGoosen: Just add a `Json::Value` constructor here that reads the key and value fields youve defined here.

QMLComponents/models/term.h
```diff
@@ -34,25 +34,27 @@
...
-	std::string					asString()		const;
+	Term(const std::vector<std::string> &	components, const columnTypeVec	&	types	= { columnType::unknown }	);
+	Term(const std::string				&	component,	const columnType		type	= columnType::unknown		);
+	Term(const QStringList				&	components, const columnTypeVec	&	types	= { columnType::unknown }	);
+	Term(const QString					&	component,	const columnType		type	= columnType::unknown		);
+	Term(const QString					&	value,		const QString		&	label,	const QString	& info = QString());
+
+	const QStringList		&	components()			const;
+	const QString			&	label()					const { return _label;	}
+	const QString			&	value()					const { return _value;	}
```
JorisGoosen, line 46: I see you do `fq()` a lot on label and value, maybe it wouldve been better to store them as std:string hahaha

QMLComponents/models/terms.cpp
```diff
@@ -223,67 +223,57 @@ Term &Terms::at(size_t index)
...
 
-int Terms::indexOf(const QString &component) const
+int Terms::indexOfValue(const QString &value) const
 {
 	int i = 0;
-	for(const Term &term : _terms)
+	for (const Term& term : _terms)
 	{
-		if (term.contains(component))
+		if (term.value() == value)
```
JorisGoosen: values are always unique because of the _makeUnique() I guess?

And isnt this a bit slow if there are a lot of terms?
Like, maybe its worth keeping some maps in Terms?

JorisGoosen (changes requested):

QMLComponents/boundcontrols/boundcontrolterms.cpp
JorisGoosen: Just use indexOfValue here instead of a loop

JorisGoosen:
I do think it might be worth keeping a value and label map in Terms.
Maybe something like `std::map<std::string, std::set<Term>> valueToTerm` ?
And then it can be kept up to date in `Terms::add` and `Terms::insert/remove`.

And can Term get updated somewhere randomly?
While part of Terms?

QMLComponents/boundcontrols/boundcontrolbase.cpp
boutinb: Yes I'll make a Term constructor with a json argument.

QMLComponents/boundcontrols/boundcontrolterms.cpp
boutinb: I'll add a method in Terms to do this.

QMLComponents/boundcontrols/boundcontrolterms.cpp
boutinb: The BoundControl classes replace the old Option classes.

QMLComponents/models/term.h
boutinb: I think that are still more functions that need a QString...

QMLComponents/models/terms.cpp
boutinb: They are unique, only for Paired VariablesList (using the ListModelMultiTermsAssigned model) a term can be used multiple times in their model. But this model does not use the Terms object, it uses its own 'Tuples' object instead.
It's a pity, because I could reuse the 'components' in the Term object to store these tuples, but this PR is already too big for the small feature I wanted to add....

boutinb:
> I do think it might be worth keeping a value and label map in Terms. Maybe something like `std::map<std::string, std::set<Term>> valueToTerm` ? And then it can be kept up to date in `Terms::add` and `Terms::insert/remove`.
> 
> And can Term get updated somewhere randomly? While part of Terms?

I've added a map in Terms so that a Term can be found quickly with its value.

QMLComponents/boundcontrols/boundcontrolterms.cpp
JorisGoosen: Ah so the name is just a bit misleading, you know, cause it contains "control" but is actually not the control, cause that goes via JASPControl?

QMLComponents/models/terms.cpp
JorisGoosen: fair enough!

And it would have to be quite a lot of terms to get slow anyway

QMLComponents/boundcontrols/boundcontrolbase.cpp
```diff
@@ -117,44 +117,47 @@ std::string BoundControlBase::getName() const
...
 			}
 
-			allControlValues[term.asQString()] = controlMap;
+			allControlValues[term.value()] = controlMap;
 		}
 	}
 }
 
-Json::Value BoundControlBase::_getTableValueOption(const Terms& terms, const ListModel::RowControlsValues& componentValuesMap, const std::string& key, bool hasInteraction, bool keyHasVariables)
+Json::Value BoundControlBase::_getTableValueOption(const Terms& terms, const ListModel::RowControlsValues& componentValuesMap, const std::string& keyValue, const std::string& keyLabel, bool hasInteraction, bool keyHasVariables)
```
JorisGoosen: Why would you need to give the label to get the option?
Shouldnt the label be stored in the option and be accessible via the key?

Or is this function also responsible for adding stuff to the table?

QMLComponents/boundcontrols/boundcontrolterms.cpp
boutinb: Yes, the name is is not the best one, I agree. Maybe it should be back to Option

QMLComponents/boundcontrols/boundcontrolbase.cpp
boutinb: The label might be used by the analysis as well, as a title or in a header of a Table for example. The label is also translated, so it's better to translate it once in QML, and then pass the translated string to the analysis.

boutinb: This function generates the option for many kinds op control.

JorisGoosen: Yes, so it should really be called "_createTableValueOption" then right?
Because to me "get" feels more like granting access or copying a value that exists inside the containing structure.
While actually you don't get anything but create an option from scratch.

QMLComponents/boundcontrols/boundcontrolterms.cpp
JorisGoosen: I get confused every time looking through the inheritance tree :p

QMLComponents/models/listmodeltermsavailable.cpp
```diff
@@ -17,24 +17,243 @@
...
+
+void ListModelTermsAvailable::initTerms(const Terms &terms, const RowControlsValues&, bool)
+{
+	beginResetModel();
 	
+	_allTerms = _allSortedTerms = terms;
+	_setTerms(terms, _allSortedTerms);
+
+	if (currentSortType() != SortType::None)
+		Sortable::sortItems();
```
JorisGoosen, line 40: Why `Sortable::sortItems();` and not just `sortItems();`?

JorisGoosen (approved):

PR #5851, opened 2025-04-15
enh: scrolling view more intuitive guide

### LINKED ISSUE #3300: [Bug]: Advanced option in JASP cut off on small screens
### JASP Version

0.19.3

### Commit ID

_No response_

### JASP Module

Factor

_No response_

Windows 10

### Bug Description

I need to change the ML-estimator in CFA. But I cant find the Advanced option, it's not under Plots.

### Expected Behaviour

That I can find the Advance option, of course.

### Steps to Reproduce

I open Confirmatory factor analysis, enter some variables for a proposed factor. Then I go through models options etc, and finally I expect to find the Advanced option under the Plot options. But it is nowhere to see.

### Log (if any)

_No response_

### More Debug Information

_No response_

### Final Checklist

shun2wang:
Fix: https://github.com/jasp-stats/jasp-issues/issues/3300

I'm always annoyed by the default maximized windows, especially after the system memorizes my desktop window layout.I think it's also not good experience for users. 
so let's make scroll view more intuitive.

This is not the only solution, see also: https://ux.stackexchange.com/questions/38771/how-to-hint-that-theres-more-beyond-what-you-see-in-a-scroll-view

Looks now analysis form:
image
menu panel:
image


shun2wang:
Or should it be a reusable component?

boutinb (commented):
I find it quite a good solution.
I would use the jaspTheme.fileMenuColorBackground for the CustomMenu.qml
and jaspTheme.analysisBackgroundColor for the AnalysisForms.qml

shun2wang:
>  I would use the jaspTheme.fileMenuColorBackground for the CustomMenu.qml and jaspTheme.analysisBackgroundColor for the AnalysisForms.qml

I tried it, but it seems that the color depth contrast at the bottom is not strong enough, and there is still the illusion that "there is no content below"?or maybe just like a shadow?

boutinb:
> > I would use the jaspTheme.fileMenuColorBackground for the CustomMenu.qml and jaspTheme.analysisBackgroundColor for the AnalysisForms.qml
> 
> I tried it, but it seems that the color depth contrast at the bottom is not strong enough, and there is still the illusion that "there is no content below"?or maybe just like a shadow?

I don't have strong preferences here....
I'm also a bit annoyed by the default maximized window, so I really prefer this way. @JorisGoosen , @EJWagenmakers ?

JorisGoosen:
Ill add a preference option for open maximized that is off by default

(Because I love the maximized opening, saves me maximizing it every time I use it)
(Should just show helpwindow and engine window inside mainwindow and its all perfect.)

JorisGoosen:
Alright, this is almost ready for review again, by @boutinb  and by @shun2wang 

I did change the color to shadow, I think it looks pretty good in both light and dark theme.
While the previous color looked fine in darktheme it looked invisble in lighttheme.

I also think now we have scroll indicators everywhere?
The ribbon had them for a while already and I dont know what Im missing?

Aaah, the labeleditor!
Ok, ill add that.
Ah and the dropdowns...

Ok so maybe this is not ready for review yet.

JorisGoosen:
Ok, so I got a bit carried away, sorry for hijacking your PR @shun2wang 

But now we should have scroll indicators everywhere except the data.
Also it should make the labeleditor look a bit better (no horizontal scrollbar for instance).

shun2wang:
Behavior on drop-down now is it intentional? seems shadow always showing what ever if scrollbar is appeared.
image

Desktop/components/JASP/Widgets/FileMenu/DataLibrary.qml
```diff
@@ -70,6 +70,33 @@ Item
...
 	}
+	
+	ScrollMoreIndicator
+	{
+		anchors
+		{
+			top:			secondseparator.bottom
+			topMargin:		-secondseparator.height / 2
+			left:			parent.left
+			right:			parent.right
```
shun2wang, line 81: Should also set leftMargin? It seems like it should be anchored to the left side of the scrollbar, as it is done in the button? But it's not a big deal. I'm not really a stickler for this idea;-)

image

JorisGoosen:
> Behavior on drop-down now is it intentional? seems shadow always showing what ever if scrollbar is appeared. image

Ah looks like a bug!

Desktop/components/JASP/Widgets/FileMenu/DataLibrary.qml
JorisGoosen: Yeah, it wouldve been prettier, but there are some other issues there as well.
Like, it wont scroll all the way down for some reason.

But I was a bit distracted by the labeleditor :p

Ill have a look later

shun2wang:
1. The gradient you defined seems to be from the middle to the sides😜. I think it would be better to have a gradient from one side to the other, and the color gets darker as it gets closer to the bottom (top).
2. Should we also ensure that: 1. the scroll bar is displayed immediately when the input content height exceeds the viewport height (test case in testTextArea of ​​the test module) 2. the scroll bar is highlighted when the mouse hovers over the scrollable content.The benefit of this is that it always indicates to the user that the content is not fully displayed and can be scrolled.

QMLComponents/components/JASP/Controls/DropDown.qml
```diff
@@ -187,39 +186,71 @@ ComboBoxBase
...
+				
+				ScrollMoreIndicator
+				{
+					anchors
+					{
+						left:	parent.left
+						right:	parent.right
+						bottom:	parent.bottom
+					}
+					extraSpace:	popupView.contentHeight - popupView.contentY + popupView.height
```
shun2wang: A simple fix to drop down?

```suggestion
					anchors
					{
						top:	parent.top
						left:	parent.left
						right:	parent.right
					}

					upsideDown:	true
					extraSpace:	popupView.contentY
					visible:	popupView.contentY > 0
				}

				ScrollMoreIndicator
				{
					anchors
					{
						left:	parent.left
						right:	parent.right
						bottom: parent.bottom
					}
					extraSpace: popupView.contentHeight - popupView.contentY - popupView.height
					visible:	popupView.contentHeight > popupView.height
				}
```

JorisGoosen: Maybe thats too simple :p

I missed some parentheses 🤡

JorisGoosen:
> 1. The gradient you defined seems to be from the middle to the sides😜. I think it would be better to have a gradient from one side to the other, and the color gets darker as it gets closer to the bottom (top).

Uhm?

This is what I see:
image
(so, darker to the top/bottom depending on where it is)

Where you see it otherwise I dont know?

> 
>     2. Should we also ensure that: 1. the scroll bar is displayed immediately when the input content height exceeds the viewport height (test case in testTextArea of ​​the test module) 2. the scroll bar is highlighted when the mouse hovers over the scrollable content.The benefit of this is that it always indicates to the user that the content is not fully displayed and can be scrolled.

I like it!

Edit: I do like it, but it would require going through all the flickables JASPScrollBar is now linked to (through 'flickable'), but 'Flickable' does not tell us whether a mouse is hovering, so we would need to find a way around that.

A way would be to create a "ScrollMoreIndicator" that is not just the little gradient we made now.
But it is rather something that covers the entire "scrollable thing" and adds the gradients at top and bottom.
And it also includes a MouseArea that only checks for "containsMouse" with "hoverEnabled: true" and has "acceptedButtons: false".
This could then be a property we can link directly to the scrollbars.

While this is simple enough I dont have the time to get into that right now.
If you feel like it I can review your work!

However, I think this PR is ready to be merged now and will be already better than what we had before

JorisGoosen (approved):

PR #5863, opened 2025-04-26
enh: log more msg from Qt runtime

shun2wang:
To log more infomation from Qt runtime while running JASP. this may help us locate the error more.

This will of course increase the size of the log file, as there are warnings of partial duplicate writes, is this a burden on performance?

### DIFF
--- a/Desktop/main.cpp
+++ b/Desktop/main.cpp
@@ -401,6 +401,28 @@ void recursiveFileOpener(QFileInfo file, int & failures, int & total, int & time
 	}
 }
 
+void qtMessageHandler(QtMsgType type, const QMessageLogContext &context, const QString &msg)
+{
+	QByteArray localMsg = msg.toLocal8Bit();
+	const char *file	= context.file ? context.file : "";
+	const char *function = context.function ? context.function : "";
+
+	switch (type) {
+	case QtWarningMsg:
+		Log::log() << "Msg from Qt Warning: " << localMsg.constData() << " [" << file << ":" << context.line << ", " << function << "]" << std::endl;
+		break;
+	case QtCriticalMsg:
+		Log::log() << "Msg from Qt Critical: " << localMsg.constData() << " [" << file << ":" << context.line << ", " << function << "]" << std::endl;
+		break;
+	case QtFatalMsg:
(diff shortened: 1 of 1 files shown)


Desktop/main.cpp
```diff
@@ -401,6 +401,28 @@ void recursiveFileOpener(QFileInfo file, int & failures, int & total, int & time
...
+	case QtCriticalMsg:
+		Log::log() << "Msg from Qt Critical: " << localMsg.constData() << " [" << file << ":" << context.line << ", " << function << "]" << std::endl;
+		break;
+	case QtFatalMsg:
+		Log::log() << "Msg from Qt Fatal: " << localMsg.constData() << " [" << file << ":" << context.line << ", " << function << "]" << std::endl;
+		break;
+	case QtDebugMsg:
+	case QtInfoMsg:
+		break;
+	}
```
JorisGoosen:
```suggestion
	const char *file	= context.file ? context.file : "";
	const char *function = context.function ? context.function : "";

	switch (type) {
	case QtWarningMsg:
		Log::log() << "Msg from Qt Warning: " << msg << " [" << file << ":" << context.line << ", " << function << "]" << std::endl;
		break;
	case QtCriticalMsg:
		Log::log() << "Msg from Qt Critical: " << msg << " [" << file << ":" << context.line << ", " << function << "]" << std::endl;
		break;
	case QtFatalMsg:
		Log::log() << "Msg from Qt Fatal: " << msg << " [" << file << ":" << context.line << ", " << function << "]" << std::endl;
		break;
	case QtDebugMsg:
	case QtInfoMsg:
		break;
	}
```

Why not use `msg` directly?

shun2wang: This is Qt recommended writing: https://doc.qt.io/qt-6/qtglobal.html#qInstallMessageHandler

JorisGoosen: sure, but Log::log() already accepts QString anyway

JorisGoosen (approved):

JorisGoosen:
For performance I think there will be no impact unless loging is on. And if it turns out to be too bad for loging we could consider having multiple logging levels. But I think it will be ok and these msgs will be very useful for debuggin

PR #5866, opened 2025-04-28
add pageup and pagedown to datatableview

JorisGoosen:
also force active focus on dataviewer on a press or a flick

this does however work after youve scrolled even a tiny bit so I guess this counts as a: fix for https://github.com/jasp-stats/jasp-issues/issues/3388

You can now use page up and down in the dataviewer(s) and pressing shift with it makes it scrolls horizontally

If you wonder where this button is on a mac: fn+option+up/down

### DIFF
--- a/Desktop/components/JASP/Widgets/DataTableView.qml
+++ b/Desktop/components/JASP/Widgets/DataTableView.qml
@@ -222,7 +222,40 @@ FocusScope
 						event.accepted = true;
 						mainWindowRoot.changeFocusToFileMenu();
 						break;
-
+									
+				
+				case Qt.Key_PageUp:
+					event.accepted = true;
+					if(shiftPressed)	contentX = Math.max(0, contentX - width)
+					else				contentY = Math.max(0, contentY - height)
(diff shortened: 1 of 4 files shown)


Desktop/components/JASP/Widgets/DataTableView.qml
```diff
@@ -222,7 +222,40 @@ FocusScope
...
+					else				contentY = Math.max(0, contentY - height)
+					break;
+								
+				case Qt.Key_PageDown:
+					event.accepted = true; 
+					if(shiftPressed)	contentX = Math.min(contentWidth  - width,  contentX + width)
+					else				contentY = Math.min(contentHeight - height, contentY + height)
+					break;
+				
+				case Qt.Key_Down:
```
JorisGoosen, line 239: these dont seem to work, neither do the ones in the JASPDataView..

But this isnt a regression and I dont really want to spend more time on this issue.
In fact, the feature request is now implemented so that is something.

boutinb (approved):

PR #5867, opened 2025-04-28
Use typed columns in Lavaan syntax

boutinb:
When parsing Lavaan code, the variables found should be encoded with the typed variables.

### DIFF
--- a/Desktop/data/datasetpackage.cpp
+++ b/Desktop/data/datasetpackage.cpp
@@ -114,8 +114,7 @@ void DataSetPackage::enginesReceiveNewData()
 		else									emit enginesReceiveNewDataSignal();
 	}
 
-	ColumnEncoder::setCurrentColumnNames(	getColumnNames()); //Same place as in engine, should be fine right?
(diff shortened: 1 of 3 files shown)


Desktop/data/datasetpackage.cpp
```diff
@@ -1532,29 +1531,16 @@ void DataSetPackage::initializeComputedColumns()
 }
 
 
-stringvec DataSetPackage::getColumnNames()
+ColumnEncoder::colTypeMap DataSetPackage::getColumnNamesWithTypes()
```
JorisGoosen: So you renamed and moved the `getColumnTypesMap` around to become `getColumnNamesWithTypes`? :p

JorisGoosen (approved):
It looks like this entire PR is not needed, you just moved code around and renamed stuff...

JorisGoosen (changes requested):
ah clicked the wrong button

JorisGoosen (approved):

PR #5869, opened 2025-04-30
Added documentation on Configuration files, engine options and installing a module

JorisGoosen:
Also moved some stuff around in the QML.

@JohnnyDoorn  does this tell you all you need?
See https://github.com/JorisGoosen/jasp-desktop/blob/howToInstallADevModule/Resources/Help/preferences/PrefsAdvanced.md for a nicer render

### DIFF
--- a/Desktop/components/JASP/Widgets/FileMenu/PrefsAdvanced.qml
+++ b/Desktop/components/JASP/Widgets/FileMenu/PrefsAdvanced.qml
@@ -164,14 +164,171 @@ PrefsScrollView
                 toolTip:			qsTr("This will erase the 'renv', 'Modules' and development Module folders in the appdata.")
 				onClicked:			mainWindow.clearModulesFoldersUser();
 
-				KeyNavigation.tab:		useRemoteConf
+				KeyNavigation.tab:		directLibpathDevModEnabled
 				activeFocusOnTab:		true
 			}
 		}
 
+		
 		PrefsGroupRect
 		{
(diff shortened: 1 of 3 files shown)


JohnnyDoorn (approved):
Really nice, thanks!

boutinb (approved):

JohnnyDoorn:
After some more experience with this workflow, one thing thats still a bit unclear to me is how to best setup one central "JASP package library" for all modules, and would is a good practice for this library in terms of location (on mac :p).
Is `"~/jasp-renv-package-library"` acceptable or does it need to be in a specific subfolder?
and then the only thing I need to tweak in Kyle's code is
`renv::restore(clean = FALSE, lib = "~/jasp-renv-package-library")`
`renv::install('.', lib = "~/jasp-renv-package-library")`?


and so that would avoid me needing to install/setup jaspTools for every module?

JorisGoosen:
image
This part is supposed to explain that

You just do  `.libPaths("~/jasp-renv-package-library")` before you call `renv::restore` and `renv::install`

And you can then use the code examples I put in the documentation instead of those from Kyle's script

JorisGoosen:
I f this is not clear I can also spell it out in the documentation

JohnnyDoorn:
Yeah maybe just include another line in the code block with 
`.libPaths("~/jasp-renv-package-library") # use a dedicated project library`
to make it extra clear to any extreme n00bs

PR #5870, opened 2025-05-01
Fix Integer or double in a TextField

boutinb:
If a value is seen as an integer or a double, it is saved in the option as (json) integer or double. So it must be read as integer or double when reading a JASP file.


JorisGoosen (approved):

JorisGoosen:
Ah yes that makes sense

PR #5871, opened 2025-05-01
helpwindow tweaks

JorisGoosen:
fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2766 
also fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2685 probably

### DIFF
--- a/QMLComponents/controls/comboboxbase.cpp
+++ b/QMLComponents/controls/comboboxbase.cpp
@@ -305,6 +305,8 @@ void ComboBoxBase::_setCurrentProperties(int index, bool bindValue)
 
 QString	ComboBoxBase::generateMDHelp(int depth) const
 {
+	if(!isVisible())
+		return "";
 	QStringList markdown;
 
 	printLabelMD(markdown, depth);
@@ -317,15 +319,13 @@ QString	ComboBoxBase::generateMDHelp(int depth) const
 		{
 			QString label = term.label(),
 					info = term.info();
-			if (label.isEmpty())
+			if (label.isEmpty() || info.isEmpty())
 				continue;
 
-			markdown << "\n" << QString{depth * 2, ' '} << "- *" << label << "*";
-			if (!info.isEmpty())
-				markdown << (": " + info);
+			markdown << "\n" << QString{depth * 2, ' '} << "- *" << label << "*" << (": " + info);
 		}
 	}
-	else
+	else if(!hasInfo())
 	{
 		markdown << "\n" << QString{depth * 2, ' '};
 		// Display the options in one line separated by a comma.
--- a/QMLComponents/controls/comboboxbase.cpp
+++ b/QMLComponents/controls/comboboxbase.cpp
@@ -325,7 +325,7 @@ QString	ComboBoxBase::generateMDHelp(int depth) const
 			markdown << "\n" << QString{depth * 2, ' '} << "- *" << label << "*" << (": " + info);
 		}
 	}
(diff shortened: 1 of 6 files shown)


QMLComponents/controls/comboboxbase.cpp
```diff
@@ -317,15 +319,13 @@ QString	ComboBoxBase::generateMDHelp(int depth) const
...
 				continue;
 
-			markdown << "\n" << QString{depth * 2, ' '} << "- *" << label << "*";
-			if (!info.isEmpty())
-				markdown << (": " + info);
+			markdown << "\n" << QString{depth * 2, ' '} << "- *" << label << "*" << (": " + info);
 		}
 	}
-	else
+	else if(!hasInfo())
```
boutinb: I think this is to fix https://github.com/jasp-stats/INTERNAL-jasp/issues/2685, but this is not a good idea: the DropDown can have an info and still have as source the dataset variables (or a set of these variables).
I think better is to test whether the model has variables, and to use also an extra property 'generateOptionsHelp' or something like that. This will be set per default with the hasVariable property, but can be overriden: there are cases when a DropDown may have variables as values, but also extra values that need some info.

boutinb: I can implement this if you prefer.

JorisGoosen: Go ahead! 👍

QMLComponents/controls/expanderbuttonbase.cpp
```diff
@@ -35,7 +35,8 @@ void ExpanderButtonBase::setUp()
 
 QString ExpanderButtonBase::generateMDHelp(int depth) const
 {
-	if (!hasInfo()) return "";
+	if (!hasInfo() || !isVisible()) 
+		return "";
```
boutinb, line 39: This is also not good: for example a Section makes the controls invisible, but there should be still a help.

JorisGoosen: Well this is what the issue requested yes.

Im not sure if it is a good idea, but then we should say in the issue we arent doing it.

QMLComponents/controls/expanderbuttonbase.cpp
```diff
@@ -35,9 +35,14 @@ void ExpanderButtonBase::setUp()
...
 
-	// For Section, draw first a line, and reset the depth to 0.
-	return "\n---\n\n" + JASPControl::generateMDHelp(0);
+	QString label = (infoLabel().isEmpty() ? title() : infoLabel()).trimmed();
+	// For sub-section, draw first a line, and reset the depth to 0.
+	if (label.isEmpty() || depth > 0)
+		return "\n---\n\n" + JASPControl::generateMDHelp(0);
+
+	// Use collapsible section
+	return "<details>\n<summary><b>" + label + "</b></summary>\n" + JASPControl::generateMDHelp(0) + "\n</details>";
```
JorisGoosen, line 47: cool

QMLComponents/controls/comboboxbase.cpp
boutinb: I have just changed it so that it uses containsVariables to check whether it should list the options in a DropDown.
I have also made the section collapsible.

boutinb: I have added also a test in the Info Test analysis in the jaspTestModule

PR #5873, opened 2025-05-02
Infinite value in formula

boutinb:
Cannot set infinity value in FormulaField
If you set per default an infinity value in a formula field, or if you set manually an inifinity value, you get an error, because it cannot be parsed by R.
But as the parsing in unnecessary when the value is already a double, just skip this parsing.
A new test (Test Input Field) is added in the jaspTestModule

### DIFF
--- a/QMLComponents/controls/textinputbase.cpp
+++ b/QMLComponents/controls/textinputbase.cpp
@@ -88,20 +88,31 @@ void TextInputBase::bindTo(const Json::Value& value)
 	}
 	case TextInputType::FormulaType:
 	{
-		_value = value.isString() ? tq(value.asString()) : value.isNumeric() ? value.asDouble() : QVariant();
-
 		// If it is already numeric, no need to parse it.
 		// This also avoid parsing infinite value: a QVariant with an infinite value gives "inf" as string value,
 		// which gives an error when parsed by R.
-		if (value.isNumeric())
-			setProperty("realValue", _value);
-		else
-		{
-			setIsRCode();
+		bool setRealValue = true;
 
-			if (!_value.isNull())
+		if (value.isNumeric())
+			_value = value.asDouble();
+		else if (value.isString())
+		{
+			double dblVal = 0;
+			QString strValue = tq(value.asString());
+			if (!strValue.isEmpty() && !QColumnUtils::getDoubleValue(strValue, dblVal))
+			{
+				setIsRCode();
 				runRScript("as.character(" + _value.toString() + ")", true);
+				setRealValue = false;
+			}
+			else
+				_value = dblVal;
 		}
+		else
+			_value = QVariant();
+
+		if (setRealValue)
+			setProperty("realValue", _value);
 
 		break;
 	}
@@ -446,7 +457,10 @@ void TextInputBase::_setBoundValue()
 {
 	if (_inputType == TextInputType::FormulaType)
 	{
-		if (_value.metaType().id() == QMetaType::Double)
+		double valueDbl = 0;
+		bool isDbl = QColumnUtils::getDoubleValue(_value.toString(), valueDbl);
+
+		if (isDbl)
 		{
 			setProperty("realValue", _value);
 			setBoundValue(_getJsonValue(_value));
--- a/QMLComponents/controls/textinputbase.cpp
+++ b/QMLComponents/controls/textinputbase.cpp
@@ -253,6 +253,12 @@ void TextInputBase::rScriptDoneHandler(const QString &result)
 			setHasScriptError(true);
 			break;
 		}
+		else
+		{
+			clearControlError();
+			setHasScriptError(false);
+		}
+
 
 		if (!_formulaResultInBounds(val))
 		{
@@ -464,6 +470,9 @@ void TextInputBase::_setBoundValue()
 		{
 			setProperty("realValue", _value);
 			setBoundValue(_getJsonValue(_value));
+			clearControlError();
+			setHasScriptError(false);
+			emit formulaCheckSucceeded();
 		}
 		else
 		{
--- a/QMLComponents/controls/textinputbase.cpp
+++ b/QMLComponents/controls/textinputbase.cpp
@@ -89,10 +89,19 @@ void TextInputBase::bindTo(const Json::Value& value)
 	case TextInputType::FormulaType:
 	{
 		_value = value.isString() ? tq(value.asString()) : value.isNumeric() ? value.asDouble() : QVariant();
-		setIsRCode();
 
-		if (!_value.isNull())
-			runRScript("as.character(" + _value.toString() + ")", true);
+		// If it is already numeric, no need to parse it.
+		// This also avoid parsing infinite value: a QVariant with an infinite value gives "inf" as string value,
+		// which gives an error when parsed by R.
+		if (value.isNumeric())
+			setProperty("realValue", _value);
+		else
+		{
+			setIsRCode();
+
+			if (!_value.isNull())
+				runRScript("as.character(" + _value.toString() + ")", true);
+		}
 
 		break;
 	}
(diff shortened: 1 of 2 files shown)


boutinb:
I'm thinking now: why do we use R to parse the formulas, and not just JavaScript? We could use the QJSEngine Qt object. Or are the formulas sometimes used with typical R functions?

JorisGoosen:
> I'm thinking now: why do we use R to parse the formulas, and not just JavaScript? We could use the QJSEngine Qt object. Or are the formulas sometimes used with typical R functions?

The original use for R formula is also that you can check things with your r-package. because it runs there.

Nothing is stopping you from adding a "JSField" ;)

JorisGoosen:
Infinite and `Nan` seem ok.

But after fixing your testcase I saw this when opening it:
image

Then I change it and get:
image

then I change it back and get:
image

boutinb:
> > I'm thinking now: why do we use R to parse the formulas, and not just JavaScript? We could use the QJSEngine Qt object. Or are the formulas sometimes used with typical R functions?
> 
> The original use for R formula is also that you can check things with your r-package. because it runs there.
> 
> Nothing is stopping you from adding a "JSField" ;)

@FBartos told me that FormulaField are used also with typical R functions. So we should keep it so.
My fix at least prevents JASP from making a roundtrip to the Engine to check values that are just numbers and not formulas.

JorisGoosen:
> @FBartos told me that FormulaField are used also with typical R functions. So we should keep it so. 

Yes, obviously we should *keep it*, that is why I suggested making a new thing if you feel JS adds something.

> My fix at least prevents JASP from making a roundtrip to the Engine to check values that are just numbers and not formulas.

You couldve also just fixed it by simply converting whatever it was R didnt understand to something it does understand.
You state the error is:

> if you set manually an inifinity value, you get an error, because it cannot be parsed by R.

`nan` can be parsed right?

And `Inf` is supported in R right?

boutinb:
> Infinite and `Nan` seem ok.
> 
> But after fixing your testcase I saw this when opening it: image
> 
> Then I change it and get: image
> 
> then I change it back and get: image

Good I made this test!
I did not notice the error...
What's happening is that no R scripts are allowed to be run when the analysis is initializing (`blockValueChangeSignal(true)`), and when the analysis is initialized, the waiting R scripts are just ignored (see https://github.com/jasp-stats/jasp-desktop/blob/f780813a6a0bd4ad9a1cc096c6842af1773a679e/QMLComponents/analysisform.cpp#L746). I'm not sure why this is so, but it has nothing to do with my fix.
Apparently no one has set the default value of a FormulaField with a formula. Maybe we should make an issue for this

boutinb:
> Infinite and `Nan` seem ok.
> 
> But after fixing your testcase I saw this when opening it: image
> 
> Then I change it and get: image
> 
> then I change it back and get: image

Good I made this test!
I did not notice the error...
What's happening is that no R scripts are allowed to be run when the analysis is initializing (`blockValueChangeSignal(true)`), and when the analysis is initialized, the waiting R scripts are just ignored (see https://github.com/jasp-stats/jasp-desktop/blob/f780813a6a0bd4ad9a1cc096c6842af1773a679e/QMLComponents/analysisform.cpp#L746). I'm not sure why this is so, but it has nothing to do with my fix.
Apparently no one has set the default value of a FormulaField with a formula. Maybe we should make an issue for this

JorisGoosen:
Well... 

Is this not something we can fix now?

JorisGoosen:
In `void AnalysisForm::setAnalysisUp()` you unblock the signals `blockValueChangeSignal(false, false);` but without notifying.

Why not though?

And if you think this is really not used anywhere and wont lead to problems then make an issue by all means.
But this is quite an insidious bug that could lead to hard to detect bugs down the line.

boutinb:
> You couldve also just fixed it by simply converting whatever it was R didnt understand to something it does understand.

I find it better to test whether it is a number, and just if is not an number, to let it parse by R (as I said, this avoids a roundtrip to the engine).

> Is this not something we can fix now?

Hmm, I can have a look. The parameter 'notifyOnceUnblocked' is really old... but yes, it is maybe a good occasion to understand why it is set to false sometimes.



> And Inf is supported in R right?

Yes, but I think the problem is that it was sent 'inf' without capital letter.

JorisGoosen:
> > You couldve also just fixed it by simply converting whatever it was R didnt understand to something it does understand.
> 
> I find it better to test whether it is a number, and just if is not an number, to let it parse by R (as I said, this avoids a roundtrip to the engine).

Sure, if that isnt what is causing problems then whatever.

> > Is this not something we can fix now?
> 
> Hmm, I can have a look. The parameter 'notifyOnceUnblocked' is really old... but yes, it is maybe a good occasion to understand why it is set to false sometimes.

See the commit I just added, but it does mess up the outlining of labels for the inputs and makes the interface bounce...
 
> > And Inf is supported in R right?
> 
> Yes, but I think the problem is that it was sent 'inf' without capital letter.

Sure, but that might just be caused by converting to json through `Json::Value TextInputBase::_getJsonValue(QVariant value) const` because I see this there:

```
case TextInputBase::FormulaType:			return isDbl ? Json::Value(valueDbl) : fq(value.toString()); // Keep it as string and do not try to make it an integer or a double
```

And if for some reason you put "Inf" in the value and get here "inf" then where is it converted and why?
Because in testmodule you put "Inf" and in R "Inf" works...

JorisGoosen:
> Good I made this test! I did not notice the error...

Very happy with the test you added inderdaad!
Otherwise I wouldnt have noticed either

JorisGoosen:
@boutinb did you test with my added change?

boutinb:
> @boutinb did you test with my added change?

Yes it solves the issue, but I'd like to know why `notifyOnceUnblocked` was ever set to false. I can think that when reading a JASP file, you don't want any processing to change the options. We could argue that cancelling the R scripts at this level, is a dirty way to solve this.
I see that the R Filters are also blocked: the TableView with a FilteredDataEntry model runs a filter when the Audit analysis is initialized. This filter is now blocked, but with your fix, it will run... Not sure about the consequences...

JorisGoosen:
> > @boutinb did you test with my added change?
> 
> Yes it solves the issue, but I'd like to know why `notifyOnceUnblocked` was ever set to false. I can think that when reading a JASP file, you don't want any processing to change the options. We could argue that cancelling the R scripts at this level, is a dirty way to solve this.

I totally agree with you!

We should know the consequences, but I think you know better what is going on here.
We could also simply rerun the formula rscript only there?

> I see that the R Filters are also blocked: the TableView with a FilteredDataEntry model runs a filter when the Audit analysis is initialized. This filter is now blocked, but with your fix, it will run... Not sure about the consequences...

Will it though?
Doesnt that check for changes and only then runs?
And otherwise it only runs once anyway.

But if you could take the time to think about this and see if you can think of why it was like this and how we can solve this one?

QMLComponents/controls/textinputbase.cpp
```diff
@@ -89,10 +89,19 @@ void TextInputBase::bindTo(const Json::Value& value)
...
 	{
 		_value = value.isString() ? tq(value.asString()) : value.isNumeric() ? value.asDouble() : QVariant();
-		setIsRCode();
 
-		if (!_value.isNull())
-			runRScript("as.character(" + _value.toString() + ")", true);
+		// If it is already numeric, no need to parse it.
+		// This also avoid parsing infinite value: a QVariant with an infinite value gives "inf" as string value,
+		// which gives an error when parsed by R.
+		if (value.isNumeric())
```
JorisGoosen, line 96: Misschien kun je ColumnUtils gebruiken?

JorisGoosen:
Very nice, only one small problem. 
The errormsg when you enter a "wrong number" stays...
Even after correcting

JorisGoosen:
Ok I fixed the problem, you can review the change and if you are happy with it lets merge this

JorisGoosen (approved):

PR #5879, opened 2025-05-07
fix analyses without data

JorisGoosen:
they now work again, by setting DataSetpackage to Loaded also outlined some things nice and removed 2 unused declaration

### DIFF
--- a/Desktop/data/datasetpackage.cpp
+++ b/Desktop/data/datasetpackage.cpp
@@ -94,6 +94,9 @@ void DataSetPackage::setEngineSync(EngineSync * engineSync)
 	connect(this,	&DataSetPackage::enginesPrepareForDataSignal,	_engineSync,	&EngineSync::enginesPrepareForData,	Qt::QueuedConnection);
 	connect(this,	&DataSetPackage::enginesReceiveNewDataSignal,	_engineSync,	&EngineSync::enginesReceiveNewData,	Qt::QueuedConnection);
 
+	
+	
+
 	reset();
 }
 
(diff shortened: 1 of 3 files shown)


Desktop/data/datasetpackage.cpp
```diff
@@ -2624,3 +2627,10 @@ void DataSetPackage::setManualEdits(bool newManualEdits)
 
 	emit manualEditsChanged();
 }
+
+void DataSetPackage::setHasAnalysesWithoutData()							
+{ 
+	_hasAnalysesWithoutData		= true;
+	setLoaded();
```
JorisGoosen, line 2631: this is the fix ;)

Desktop/data/datasetpackage.cpp
```diff
@@ -94,6 +94,9 @@ void DataSetPackage::setEngineSync(EngineSync * engineSync)
 	connect(this,	&DataSetPackage::enginesPrepareForDataSignal,	_engineSync,	&EngineSync::enginesPrepareForData,	Qt::QueuedConnection);
 	connect(this,	&DataSetPackage::enginesReceiveNewDataSignal,	_engineSync,	&EngineSync::enginesReceiveNewData,	Qt::QueuedConnection);
 
+	
+	
+
```
```suggestion
```

boutinb:
In 0.19.3, it was possible to load a dataset, while JASP had already an analysis open without data.
With your fix, loading a dataset starts a new JASP session.
I remember that this was a request to be able to load a dataset in the same JASP session.

JorisGoosen:
Good point, I forgot about that

JorisGoosen:
Ok, lets do this differently

PR #5881, opened 2025-05-08
Add GroupTitleSmall to Description

JorisGoosen:
Implements https://github.com/jasp-stats/INTERNAL-jasp/issues/2317

image

image


shun2wang (commented):
Could you also make the highlight color rect right margin as same as it's left? OR not to set any margins on both sides also better.
also the Right-click menu in data editor.

koenderks:
It looks okay, but can you make it so that if no icon is specified it just shows the title without an icon?

image

JorisGoosen:
image
image

As you can see the grouptitles are not aligned with the normal entries, with or without icon. 
But I kind of like the clear indication of a break from just the outlining so I left it like this

koenderks (approved):
Looks okay to me! Maybe we could do with a little less space between the large header and the line below them, but that is preference.

image

boutinb (approved):

PR #5884, opened 2025-05-09
Try to make MD generation robust

boutinb:
Sub items that do not have labels or info, should be set at the same level as their parents. For this, the hierarchy of the items should be kept, before generating the MD string.

### DIFF
--- a/QMLComponents/controls/jaspcontrol.cpp
+++ b/QMLComponents/controls/jaspcontrol.cpp
@@ -44,8 +44,8 @@ JASPControl::JASPControl(QQuickItem *parent) : QQuickItem(parent)
 
 	connect(this, &JASPControl::titleChanged,			this, &JASPControl::helpMDChanged);
 	connect(this, &JASPControl::infoChanged,			this, &JASPControl::helpMDChanged);
-	connect(this, &JASPControl::visibleChanged,			this, &JASPControl::helpMDChanged);
-	connect(this, &JASPControl::visibleChildrenChanged,	this, &JASPControl::helpMDChanged);
+	//connect(this, &JASPControl::visibleChanged,			this, &JASPControl::helpMDChanged);
+	//connect(this, &JASPControl::visibleChildrenChanged,	this, &JASPControl::helpMDChanged);
 	connect(this, &JASPControl::backgroundChanged,		[this] () { if (!_focusIndicator)		setFocusIndicator(_background); });
 	connect(this, &JASPControl::infoChanged,			[this] () { if (_toolTip.isEmpty())	setToolTip(info());					});
 	connect(this, &JASPControl::toolTipChanged,			[this] () { setShouldStealHover(!_toolTip.isEmpty());					});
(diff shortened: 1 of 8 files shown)


QMLComponents/controls/jaspcontrol.h
```diff
@@ -63,6 +63,24 @@ class JASPControl : public QQuickItem
...
+		MDItem() {}
+		MDItem(const QString& _label) : label{_label} {}
+
+		bool	isEmpty()	{ return label.isEmpty() && info.isEmpty() && children.size() == 0; }
+		bool	hasHeader()	{ return !label.isEmpty() || !info.isEmpty(); }
+
+		QString print(int depth = 0) const;
+
+
+	} MDItem;
```
JorisGoosen: Why? Please why?

boutinb:
I have removed the MDItems structure.
In fact just a vector of JASPControl _MDSubItems is enough to define the hierarchy of the controls used for markdown generated help.
This _MDSubItems is first set correctly, then it makes much easier to generate the markdown help.

QMLComponents/controls/jaspcontrol.cpp
```diff
@@ -593,44 +594,49 @@ bool JASPControl::printLabelMD(QStringList& md, int depth) const
...
 {
-	if (!hasInfo()) return "";
-		
-	QStringList childMDs, markdown;
+	return !fullLabel().isEmpty() || !info().isEmpty();
+}
+
+void JASPControl::setMDSubItems()
+{
+	_MDSubItems.clear();
```
JorisGoosen: I dont like this at all...

Why do we need this function?
You can just do this determination in the generateMDHelp function...
Or let this return a list of `JASPControl*` 

But why store it?
Then we need to think about when to update it...

JorisGoosen (changes requested):
`_MDSubItems` should not exist.

The function that determines this should just return `std::vector<JASPControl*>` and not store anything.

boutinb:
@JorisGoosen It's ready to be reviewed again.

JorisGoosen:
Ok nice!

The changes look good, I will test it tomorrow.

One question, not particularly about this PR more of a "what do you think?" about:
```
QList<JASPControl*> JASPControl::getChildJASPControls(const QQuickItem * item, bool removeUnecessaryGroups)
```

If `removeUnecessaryGroups` is true basically "levels" of jaspcontrol are removed/collapsed because they are used for internal grouping and aren't supposed to be shown in helpMD?
And when false it just returns the controls as they are?

Should we maybe rename it to `collapseStructuralControls` or something?
Because why would we have "unnecessary groups" in the first place? :p

boutinb:
> Should we maybe rename it to `collapseStructuralControls` or something?

Yes it is more correct. I have just renamed it

JorisGoosen:
Ah oops I broke your commit after rebasing

JorisGoosen (approved):

PR #5889, opened 2025-05-13
Variety of QoL stuff

JorisGoosen:
Short paths should really not be necessary anymore given that we now use utf-8.

You can now tell JASP to specifically use or not use sandboxing on windows through `--sandbox` and `--noSandbox`.

And last but not least, if engines (keep) crash(ing) at the wrong times JASP now offers the option between a github report, exiting, or "salvage". Where the last one means: keep going so you can change preferences, or even try to save your data.
But because I dont want people complaining the engine doesnt run after skipping the msgbox Ive also made the data/welcome/analyses/results all red, so people cant possibly ignore that something is very much wrong indeed.

### DIFF
--- a/Desktop/main.cpp
+++ b/Desktop/main.cpp
@@ -121,19 +121,21 @@ bool runJaspEngineJunctionFixer(int argc, char *argv[], bool removeJunctions = f
 #endif
 
 
-void parseArguments(int argc, char *argv[], std::string & filePath, bool & newData, bool & unitTest, bool & dirTest, int & timeOut, bool & save, bool & logToFile, bool & hideJASP, bool & safeGraphics, Json::Value & dbJson, QString & reportingDir)
+void parseArguments(int argc, char *argv[], std::string & filePath, bool & newData, bool & unitTest, bool & dirTest, int & timeOut, bool & save, bool & logToFile, bool & hideJASP, bool & safeGraphics, bool & containerSettingForced, bool & container, Json::Value & dbJson, QString & reportingDir)
 {
-	filePath		= "";
-	unitTest		= false;
-	dirTest			= false;
-	save			= false;
-	logToFile		= false;
-	hideJASP		= false;
-	safeGraphics	= false;
-	newData			= false;
-	reportingDir	= "";
-	timeOut			= 10;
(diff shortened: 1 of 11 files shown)


Desktop/main.cpp
```diff
@@ -148,6 +150,8 @@ void parseArguments(int argc, char *argv[], std::string & filePath, bool & newDa
 		else if(args[arg] == "--safeGraphics")					safeGraphics			= true;
 		else if(args[arg] == "--newData")						newData					= true;
 #ifdef _WIN32
+		else if(args[arg] == "--sandbox")			{			containerSettingForced	= true;		container = true; }
+		else if(args[arg] == "--noSandbox")			{			containerSettingForced	= true;		container = false; }
```
shun2wang, line 154: Need to be carefully here that qtwebengine does also has an arguments of --sandbox, I'm not sure if this will effect on that.

shun2wang: ah,sorry qtwebengine called it "--no-sandbox".  https://doc.qt.io/qt-6/qtwebengine-debugging.html

JorisGoosen: Good to know!
But webengine will only use arguments that come after "--webEngineArgs" anyway I think?

shun2wang: Yes.

Desktop/utilities/processhelper.cpp
```diff
@@ -106,10 +106,29 @@ QProcessEnvironment ProcessHelper::getProcessEnvironmentForJaspEngine(bool bootS
...
+void ProcessHelper::fixPATHForWindows(QProcessEnvironment & env)
+{
+	const QString R_ARCH =
+#ifdef _WIN64
+		"x64";
+#else
+		"i386";
+#endif
+	
+	env.insert("PATH", AppDirs::programDir().absolutePath() + ";" + QDir(AppDirs::rHome()).absoluteFilePath("bin") + ";" + QDir(AppDirs::rHome()).absoluteFilePath("bin/" + R_ARCH)); // + rtoolsInPath);
```
JorisGoosen, line 130: should we maybe append the original PATH ?

boutinb (approved):

PR #5892, opened 2025-05-14
Update encoding when type change

boutinb:
In a R Script a column names are encoded with the encoding as the type-qualified column names (contNormal and contNormal.scale have the same enconding). But if the type of column is changed, the encoding should be changed also.

### DIFF
--- a/Desktop/data/datasetpackage.cpp
+++ b/Desktop/data/datasetpackage.cpp
@@ -51,12 +51,13 @@ DataSetPackage::DataSetPackage(QObject * parent) : QAbstractItemModel(parent)
 	_dataSet	= new DataSet(); //We create one here to make sure filter() etc can actually work
 	setDefaultWorkspaceEmptyValues();
 	
-	connect(this, &DataSetPackage::isModifiedChanged,	this, &DataSetPackage::windowTitleChanged);
-	connect(this, &DataSetPackage::loadedChanged,		this, &DataSetPackage::windowTitleChanged);
-	connect(this, &DataSetPackage::currentFileChanged,	this, &DataSetPackage::windowTitleChanged);
-	connect(this, &DataSetPackage::folderChanged,		this, &DataSetPackage::windowTitleChanged);
-	connect(this, &DataSetPackage::currentFileChanged,	this, &DataSetPackage::nameChanged);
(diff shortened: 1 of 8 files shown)


Desktop/analysis/analyses.cpp
```diff
@@ -39,6 +40,8 @@ Analyses::Analyses()
 	_singleton = this;
 
 	new KnownIssues(this);
+
+	connect(this, &Analyses::helpVisibleChanged, this, [this]() { if (_currentAnalysisIndex >= 0 && _currentAnalysisIndex < _orderedIds.size()) emit _analysisMap.at(_orderedIds[_currentAnalysisIndex])->helpVisibleChanged(); });
```
JorisGoosen: ❤️

Desktop/data/datasetpackage.cpp
```diff
@@ -1276,6 +1276,8 @@ bool DataSetPackage::setColumnTypes(intset columnIndexes, columnType newColumnTy
 			emit showWarning(tr("Changing column type failed"), tr("The column '%1' is generated by an analysis and its type is fixed.").arg(tq(col->name())));
 			continue;
 		}
+
+		ColumnEncoder::setCurrentColumnNames(	getColumnTypesMap());
```
JorisGoosen: You should do this in a slot connected to a signal instead of like this.

JorisGoosen:
Is this ready for review again?

Cause it looks much better already

boutinb:
> Is this ready for review again?
> 
> Cause it looks much better already

Yes

JorisGoosen:
Cool, I just finished working on my motorcycle, so I have some time to do this now ^^

JorisGoosen (approved):
Very nice!

Im going to approve it, so you can decide what to do with the following:

I love how the type of the variable in the variableslist switches with the data.
However, I think it would be nicer if it wouldnt change after the user decided what it wants it to be, instead the following happens:

image
image
image

The drag'n'drop constructor variables keep the userchosen type, even if the original type changes.
This is a nitpick, and not crucial but it feels like that behaviour respects the user's choices more.

boutinb:
> However, I think it would be nicer if it wouldnt change after the user decided what it wants it to be

I made an apart issue for this: https://github.com/jasp-stats/INTERNAL-jasp/issues/2833

PR #5900, opened 2025-05-22
Loading a data file should set automatically synchronization on

boutinb:
If you load a data file, the Synchronization is off: it should be by default on.

In the AsyncLoader::loadPackage, before loading the dataset, a DataSet object is created (line 239) if it does not exists: I'm not sure whether it is necessary to create this object here, because a new one is created in the Importer loadDataSet anyway.
So the Synchronization flag should be set after importer loadDataSet has been called, so that it is set to the right DataSet object.


JorisGoosen:
You need a dataset because otherwise references to Filter will crash JASP.

JorisGoosen (approved):

PR #5904, opened 2025-05-26
using latest cci of freexl fork sources

shun2wang:
Require: https://github.com/jasp-stats/conan-recipes/pull/1

PR #5906, opened 2025-05-27
0.19 files dont always have labels for all doubles we need to add them

JorisGoosen:
For instance to load the last two columns in the testfile in https://github.com/jasp-stats/jasp-issues/issues/3282

updates now work and use the same batched writing as the normal database writes for dataset

Without this fix, loading that file will show you no labels whatsoever for the last two columns at least.
With this fix they have a label for each double.

I can only build this on macos by rebasing it first on https://github.com/jasp-stats/jasp-desktop/pull/5904

### DIFF
--- a/Common/version.h
+++ b/Common/version.h
@@ -34,9 +34,9 @@ public:
 		const char* what() const noexcept override;
 	};
 
-	Version(const char * version);
-	Version(const std::string & version);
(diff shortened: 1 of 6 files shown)


boutinb:
When I load GSS2022-smaller, the Engine crashes. I could see that it in fact crashes when loading the jaspRegression module. Even in development It crashes with a small file when loading a jaspRegression...
I could load only the data from GSS2022-smaller, and it works fine!

boutinb (approved):

PR #5908, opened 2025-05-28
Load labels from database in the ordered way

boutinb:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2842

The row order of the labels in the database should not be trusted.
When loading the labels from the database the Column::labelsSet is called, but it is with the assumption that the row of the labels in the database is the same as the index in the _labels property.
Just add an "ORDER BY ordering" in the SQL statement to ensure that the selection is ordered as the _labels.
The call to _sortLabelsByOrder is then not needed anymore.

I find a bit weird that the labels were not querying in the database with an ORDER BY: @JorisGoosen is there any reason why the SELECT statement was done without ORDER BY, and afterwards ordering the labels with _sortLabelsByOrder?

I have found another issue: if you change manually (in the Label view) the order of the labels of a computed column, then recomputing the column sets also some wrong labels... This issue exists also in development, but this fix does not solve this issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -83,9 +83,6 @@ Modules/renv-root/
 Desktop/utilities/simplecryptkey.h
 Common/appinfo.cpp
 QMLComponents/utilities/appdirs.h
-CommonData/jaspBase_distributionSamplers.h
-CommonData/jaspBase_friendlyConstructorFunctions.h
-CommonData/jaspBase_transformFunctions.h
 
 # Bundle related
 Tools/ModuleBundleBuildDir
--- a/CommonData/.gitignore
+++ b/CommonData/.gitignore
@@ -1,1 +1,5 @@
 internalDbDefinition.h
+createIndexes.h
+jaspBase_distributionSamplers.h
+jaspBase_friendlyConstructorFunctions.h
+jaspBase_transformFunctions.h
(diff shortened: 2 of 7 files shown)


JorisGoosen:
The reason I didnt do the order by was to avoid sqlite doing too much unnecessary work.
But you have found a problem there!

I should not have used `row` there at all. Ill see if I can use that information.

boutinb:
> The reason I didnt do the order by was to avoid sqlite doing too much unnecessary work.

I think we can assume that the ordering made by sqlite is as efficient as the ordering made by our own c++ code.

JorisGoosen:
> > The reason I didnt do the order by was to avoid sqlite doing too much unnecessary work.
> 
> I think we can assume that the ordering made by sqlite is as efficient as the ordering made by our own c++ code.

Why would you assume this?

Maybe if there is an index on ordering, but is there?

JorisGoosen:
Im also having a look in another branch to see whether I can fix the problem in a less sqlite contigent way

JorisGoosen:
So there was no index on the order or columnindex of a label, which probably slowed down a bunch of thiings anyway. Ive now added this, both as upgrade or on creation of a fresh DB.

Ive also added an index to Columns, but just on colIdx and datasetId although I guess that won't make too much of a difference.

JorisGoosen (approved):

PR #5909, opened 2025-06-02
Conan fixes

RensDofferhoff:
Testing from clean system right now.

We need to be more carefull with this in future. Run any changes through the builbots minimum
Changes to conan broke every buildbot on all platforms.

### DIFF
--- a/Tools/CMake/Conan.cmake
+++ b/Tools/CMake/Conan.cmake
@@ -40,6 +40,23 @@ if(USE_CONAN)
     set(CONAN_RESULT_FILE "conanbuild.bat") #for windows
 
     message(STATUS "  ${CONAN_COMPILER_RUNTIME}")
+
+    if(freexl_POPULATED)
+      message(STATUS "Compiling freexl dependency ${freexl_SOURCE_DIR}")
+      execute_process(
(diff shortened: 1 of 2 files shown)


RensDofferhoff:
The Bison version you want is not  recipe that exist for Windows yet.
Newest boost doesn't play nicely with newer clang (resolved but its going to take time to roll out)

RensDofferhoff:
The cmake logic you made doesnt function when freexl isnt being build but gathered from local cache instead
I build it , delete it from buildfolder and then gather it from local cache.
This way all deps have one origin an one gather code path

RensDofferhoff:
Also gmp doesnt compile for me personally with CMAKE_OSX_DEPLOYMENT_TARGET 12.0
it does seem to work for target 13 however.
image


12 is unsupported by Apple, Should we continue supporting such an old version? @boutinb what is the latest/running you can get on your old machine

JorisGoosen (approved):

JorisGoosen:
> 12 is unsupported by Apple, Should we continue supporting such an old version? @boutinb what is the latest/running you can get on your old machine

We do not support oses not supported by their producer.

PR #5911, opened 2025-06-03
Use qt in Engine so we can do same formatting there for doubles

JorisGoosen:
fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2840


boutinb (approved):
Seems good.
I have added a commit to simplify Column::setValue

JorisGoosen:
Cool looks good!

Ill merge this, but there is still a bug where double formatting is not properly updated after changing locale/language

PR #5913, opened 2025-06-03
Tweaks

JorisGoosen:
Add a critical icon to crash msgbox and change loop a bit.

Also try to make JASP a bit less prone to crashing or getting stuck.

now uses same communication protocol on all three oses.

Stop calling "EngineSync::process" once exit is chosen or a fatal error occurs.

Set stdout/cerr etc to null for newly started detached JASPs. (to avoid weird crashes in secondary jasp's engines after closing the prime jasp).

### DIFF
--- a/Common/tempfiles.cpp
+++ b/Common/tempfiles.cpp
@@ -165,6 +165,8 @@ void TempFiles::deleteOrphans()
 
 					if (now - modTime > outOfDateDelta)
 					{
+						Log::log() << "Removing folder because status indicates it is out of date: " << statusFile.string() << std::endl;
+						
 						std::filesystem::remove_all(p, error);
 						if (error)
 							Log::log() << "Error when deleting directory: " << error.message() << std::endl;
--- a/CommonData/databaseinterface.cpp
+++ b/CommonData/databaseinterface.cpp
@@ -2290,15 +2290,13 @@ void DatabaseInterface::transactionWriteEnd(bool rollback)
 {
 	JASPTIMER_SCOPE(DatabaseInterface::transactionWriteEnd);
 	
-	assert(_transactionWriteDepth > 0);
-	
 	if(rollback)	
 	{
 		runStatements("ROLLBACK");
 		_transactionWriteDepth = 0;
 		throw std::runtime_error("Rollback!"); //Might be better to use a subclass of std::runtime_error but for now this isnt even used anyway.
 	}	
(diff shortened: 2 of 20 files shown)


boutinb (approved):

JorisGoosen:
Ah yeah I just noticed that now windows doesnt work anymore, but Ive almost got it fixed

shun2wang:
Hi @JorisGoosen May I ask a stupid question: What does this "Salvage" button mean? It is indeed uncommon, and I couldn't find a suitable corresponding word when I translated the interface. Is it means "recovery" or "ignore" or "try continue" in this case?

JorisGoosen:
Yeah "salvage" basically means to recover things from a wreckage. 
I choose that over "recovery" because recovery implies you can continue doing stuff afterwards, which is not true.
And "ignore" covers it a bit, but doesn't strongly imply that things are broken.

shun2wang:
> Yeah "salvage" basically means to recover things from a wreckage. 
> I choose that over "recovery" because recovery implies you can continue doing stuff afterwards, which is not true.
> And "ignore" covers it a bit, but doesn't strongly imply that things are broken.


How about we add a little explanation text here? Let the user know happens when they click this button,I believe I am not only one confused about this.

JorisGoosen:
Yeah, I suppose, but I cant put a text in the messagebox.
I guess I could show a secondary warning describing whats up?

shun2wang:
> Yeah, I suppose, but I cant put a text in the messagebox.
> I guess I could show a secondary warning describing whats up?

hummm, why cant put text in this, bbecause user will never/fear to click it if they doesn't know will happened.

JorisGoosen:
Because it is a sometimes a standard box, but see: https://github.com/jasp-stats/INTERNAL-jasp/issues/2860

JorisGoosen:
> > Yeah, I suppose, but I cant put a text in the messagebox.
> > I guess I could show a secondary warning describing whats up?
> 
> hummm, why cant put text in this, bbecause user will never/fear to click it if they doesn't know will happened.

Well, I was sort of assuming that salvage is a known english word and doesnt imply anything getting lost.

shun2wang:
> Because it is a sometimes a standard box, but see: https://github.com/jasp-stats/INTERNAL-jasp/issues/2860

ah, then I think "continue" is better,and secondary message box then.

PR #5915, opened 2025-06-06
Loading JASP File without data should not crash

boutinb:
Load Data Library/5. Frequnecies/Three-Sided Coin JASP file.
This should not crash.

### DIFF
--- a/CommonData/dataset.cpp
+++ b/CommonData/dataset.cpp
@@ -82,9 +82,12 @@ void DataSet::endBatchedToDB(std::function<void(float)> progressCallback, Column
 		_writeBatchedToDBDepth--;
 	}
 	
-	if(_writeBatchedToDBDepth == 0 && columns.size() > 0)
+	if(_writeBatchedToDBDepth == 0)
 	{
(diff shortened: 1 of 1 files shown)


CommonData/dataset.cpp
```diff
@@ -82,7 +82,7 @@ void DataSet::endBatchedToDB(std::function<void(float)> progressCallback, Column
 		_writeBatchedToDBDepth--;
 	}
 	
-	if(_writeBatchedToDBDepth == 0)
+	if(_writeBatchedToDBDepth == 0 && columns.size() > 0)
 	{
 		db().dataSetBatchedValuesUpdate(this, columns, [&progressCallback](float f){ progressCallback(0.75 + (f * 0.25));});
```
JorisGoosen:
```suggestion
	if(_writeBatchedToDBDepth == 0)
	{
		if(columns.size())
			db().dataSetBatchedValuesUpdate(this, columns, [&progressCallback](float f){ progressCallback(0.75 + (f * 0.25));});
		else
			progressCallback(1);
```

JorisGoosen (approved):

PR #5928, opened 2025-06-17
make sure to also copy libbz2 and liblzma so rdata works

JorisGoosen:
also make a macro to find these dlls because copy-pasting is for chumps

A nightly is building


FBartos (approved):

JorisGoosen:
Ill just wait the last minute for my nightly download to finish, but if the problem is solved there also ill merge this.

PR #5931, opened 2025-06-18
make sure any csv import dbl is converted to locale before comparing to datasetvalues

JorisGoosen:
Also not every column gets reloaded every time single time you synch


JorisGoosen:
Oops, still crashes with a computed column. Fixed it!

JorisGoosen:
Fixes https://github.com/jasp-stats/jasp-issues/issues/3321 too

JorisGoosen:
There are still issues with synching column changes.

PR #5938, opened 2025-06-24
Try to speed up adding computed columns to big datasets

JorisGoosen:
Some problems with loading a long file (2000000 rows) were:
- the filtereddataentry was trying to make show all the cells at once, despite the viewport width being -1
- it was deleting all labels from the database for a just created column... Just wasted time after all
- and if an engine started loading the data already, or an analysis and that analysis adds a computed column... It first waits, the does the computed column and then reloads.

Given that sampling added 3 computed columns and als has the filtereddataentry does indeed make things pretty slow...

Ive made all of this much quicker. The only thing now remaining is that the filtered data entry component still is a bit slow but its not obvious to the user that something is happening:
- first it tries to run the filter before the computed columns are filled, so a bunch of time waited for nothing. (Well, showing a warning that "everything was filtered out" 
- computed columns are then filled by running the analysis, this takes a while all on its own
- then it runs the filter by name again and actually selects the rows for filtereddataentry
- and it finally shows them.

It does work and with some patience it isnt so bad.
However, for user friendlieness I would love to have feedback in the component/analysis. But there is really no time for this now.

If a manual for auditing/sampling is created maybe it could mention opening the engineswindow to keep an eye on whats going on:
Ctrl+Shift+Alt+E
Cmd+Shift+Option+E
Opening it from the button in advanced preferences.

This PR probably fixes https://github.com/jasp-stats/jasp-test-release/issues/2998 as far as I can now

### DIFF
--- a/CommonData/column.cpp
+++ b/CommonData/column.cpp
@@ -713,6 +713,7 @@ bool Column::overwriteDataAndType(stringvec colData, columnType colType)
 	nonFilteredCountersReset();
 	labelsHandleAutoSort();
 	
+	
 	return changes;
 }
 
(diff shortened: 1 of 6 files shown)


JorisGoosen:
Also might fix https://github.com/jasp-stats/jasp-test-release/issues/3000 as I cannot reproduce it with this branch at all

JorisGoosen:
(I found some weird behaviour in scalar computed columns, ill fix that too)

koenderks (approved):
The filter is still very slow yes but at least the analysis does start up faster. With the 2 million rows, starting up was not a problem.

JorisGoosen:
@boutinb a review is welcome but dont merge it yet.
I did comment that code out before so maybe it wasnt very stable or something.

JorisGoosen:
Now its ready for review @boutinb and also to be merged.

boutinb (approved):
Just one question about updating _labelByNonEmptyIndex and _labelNonEmptyIndexByLabel

JorisGoosen:
> Just one question about updating _labelByNonEmptyIndex and _labelNonEmptyIndexByLabel

well, they are for the labeleditor only, and I think they get updated from there.

PR #5946, opened 2025-06-27
initial msix fix (#5945)

RensDofferhoff:
* generalize Temp Path finding (#5933)

* intial msix fix

### DIFF
--- a/Common/dirs.cpp
+++ b/Common/dirs.cpp
@@ -57,14 +57,21 @@ string Dirs::tempDir()
 		return p;
 
 	if(localAppDataDir() == "") { //we are an engine
-		char* buf = nullptr;
+#ifdef _WIN32
+        char* buf = nullptr;
 		size_t sz = 0;
 		if(_dupenv_s(&buf, &sz, "JASP_TMP_DIR") == 0 && buf != nullptr) {
-			p = std::string(buf);
+            p = std::string(buf);
 			free(buf);
 			return p;
 		}
-	}
+#else 
+        char* buf =  getenv("JASP_TMP_DIR");//safe in unix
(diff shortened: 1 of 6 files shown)


Common/dirs.cpp
```diff
@@ -55,10 +56,7 @@ string Dirs::tempDir()
...
 		return p;
 
-#ifdef _WIN32
-	//lets just read it from the environment if the anwser is found there
-	if(ProcessInfo::inWinContainer()) {
-		Log::log() << "In sandbox win container!" << std::endl;
+	if(localAppDataDir() == "") { //we are an engine
 		char* buf = nullptr;
 		size_t sz = 0;
 		if(_dupenv_s(&buf, &sz, "JASP_TMP_DIR") == 0 && buf != nullptr) {
```
boutinb, line 63: _dupenv_s is a microsoft function.
This cannot be compiled on Mac (and probably on Linux).

Common/dirs.cpp
```diff
@@ -81,24 +78,10 @@ string Dirs::tempDir()
...
-		if ( ! SUCCEEDED(SHGetFolderPathA(NULL, CSIDL_LOCAL_APPDATA, NULL, 0, buffer)))
-			throw Exception("App Data directory could not be retrieved");
-
-		dir = std::string(buffer);
-		dir += "/JASP/temp";
-#else
-
-		dir = string(getpwuid(getuid())->pw_dir);
-		dir += "/.JASP/temp";
-#endif
```
JorisGoosen, line 98: Why not just keep the unix #else as is ?

boutinb (approved):

PR #5951, opened 2025-07-01
Generated help should not show control with a debug flag

boutinb:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/3017
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2901

Also if a row component in a VariablesList with has an info property, the generated help should be generated only once (and not for every occurence of the row component).

Replace also '<hx></hx>' HTML tags by markdown bold tag '**'. These HTML tags render sometimes badly.


JorisGoosen (approved):
I made it so that turning developer mode on will show the debug controls and helpitems immediately and the opposite for turning it off.

JorisGoosen:
If @boutinb approves of my changes he can merge this at his leisure

PR #5952, opened 2025-07-01
Update R Syntax at the right moment

boutinb:
Remove _rSyntaxText and use properly QML

Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2899

### DIFF
--- a/QMLComponents/analysisform.cpp
+++ b/QMLComponents/analysisform.cpp
@@ -47,10 +47,10 @@ AnalysisForm::AnalysisForm(QQuickItem *parent) : QQuickItem(parent)
 	connect(this,									&AnalysisForm::formCompletedSignal,			this, &AnalysisForm::formCompletedHandler,	Qt::QueuedConnection);
 	connect(this,									&AnalysisForm::analysisChanged,				this, &AnalysisForm::knownIssuesUpdated,	Qt::QueuedConnection);
 	connect(KnownIssues::issues(),					&KnownIssues::knownIssuesUpdated,			this, &AnalysisForm::knownIssuesUpdated,	Qt::QueuedConnection);
-	connect(this,									&AnalysisForm::showAllROptionsChanged,		this, &AnalysisForm::setRSyntaxText,		Qt::QueuedConnection);
-	connect(PreferencesModelBase::preferences(),	&PreferencesModelBase::showRSyntaxChanged,	this, &AnalysisForm::setRSyntaxText,		Qt::QueuedConnection);
+	connect(this,									&AnalysisForm::showAllROptionsChanged,		this, &AnalysisForm::rSyntaxTextChanged,	Qt::QueuedConnection);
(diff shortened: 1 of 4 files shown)


QMLComponents/analysisform.cpp
```diff
@@ -660,7 +660,7 @@ void AnalysisForm::setAnalysisUp()
 	_initialized = true;
 
 	// Don't bind boundValuesChanged before it is initialized: each setup of all controls will generate a boundValuesChanged
-	connect(_analysis,					&AnalysisBase::boundValuesChanged,		this,			&AnalysisForm::setRSyntaxText,				Qt::QueuedConnection	);
+	connect(_analysis,					&AnalysisBase::boundValuesChanged,		this,			&AnalysisForm::rSyntaxTextChanged,				Qt::QueuedConnection	);
```
JorisGoosen, line 663: So when an option changes we get a analysis to run and also immediately the rsyntax rscript?

Cant the analysis just return the syntax as part of its results or something like that?
This just seems like asking for trouble (race conditions, weird engine kills etc)

boutinb: I think the R Syntax window should rather reflect the state of the options. If the options give an analysis error, the R Syntax should still display these faulty options, whatever the results of the analysis.
Also an analysis can run for quite long: there is no reason to wait for the end of the analysis to display the R Syntax. I don't see race conditions here: the R Syntax is just depending on the state of the options, not on the state of the engine.

JorisGoosen: You dont have to wait for the end of the analysis, you can just sent it as the first result... 
You know, because we always send back everything during the run of an analysis?
You just call the syntax gen function in runJaspResults.

Or am I missing something?

JorisGoosen: And the race condition is between the engine starting, then getting killed, then getting started, then getting killed again. Because if you change a bound value an engine will start running an analysis. 
But if another option changes the engine probably gets killed.

Anyway, the analysis already gets the options, so we can just have it return the syntax right?

You add a property to jaspResults that is called "rSyntax" and make sure it  gets sent first.

JorisGoosen: > R Syntax is just depending on the state of the options

But you run a script on the engine right?

boutinb: The script is sent only when the R Syntax is applied with form.sendRSyntax: at this time the options are computed according to the new R syntax (formula's are decrypted, and options depending on other options are set, etc...). The controls get their new values, and then the analysis is run with these new options.
If the user change a state of a control, the options are changed, then the R Syntax is updated and the analysis is run: these 2 last actions can be done in parallel.

shun2wang (commented):
~~I tried your PR but version in jaspDesctiptive still shows `version =  '0.20'`, but in module now version everywhere is 0.95. it is a bit messy for user cause we never release this version.~~

ab Sorry, never mind I used a old version Linux locally module build.

JorisGoosen (approved):

PR #5953, opened 2025-07-01
Display p-values properly and generic formatting fixes

JorisGoosen:
Aims to have the formatting of numbers in the result be more like before, by giving some precedence to decimal points requested over significant figures. This helps interpret the jasp unittests a bit better.

Also changes the way p values are displayed according to https://github.com/jasp-stats/jasp-issues/issues/3374
And some general fixes to disply of p values.

They will now look as follows (in french, so a comma):
image
image
image
image
image
image
image
image

### DIFF
--- a/Desktop/html/js/utils.js
+++ b/Desktop/html/js/utils.js
@@ -156,7 +156,7 @@ function formatColumn(column, type, format, alignNumbers, combine, modelFootnote
 			if (window.globSet.pExact) {
 				sf = 4;
 			} else {
-				p = Number(f.substring(2));
+				p = fixDecimals ? 1/Math.pow(10,dp) : Number(f.substring(2));
 				noZeroLead = true;
 			}
 		}
--- a/Desktop/html/js/utils.js
+++ b/Desktop/html/js/utils.js
@@ -51,6 +51,9 @@ function formatPrecision(number, precision, noZeroLead=false) {
 }
 
 function formatPrecisionWithRespectForFixedDecimals(number, sf, dp, noZeroLead=false) {
+	if(isNan(dp))
+		return formatPrecision(number, sf, noZeroLead);
+	
 	const formatter = new Intl.NumberFormat(currentLocaleId, { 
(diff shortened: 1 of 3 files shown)


JorisGoosen:
@JohnnyDoorn @vandenman @FBartos do the above pvalues formatted look good?

JohnnyDoorn:
Great stuff! Maybe before merging we can a second opinion from @FBartos or @vandenman? Don is teaching a workshop today so might be slow to respond.

FBartos:
I guess it's ok?

btw, it would be easier to post english screenshots of GUI in the future :)

JorisGoosen:
Maybe next time Ill use german or chinese :p

Desktop/html/js/utils.js
```diff
@@ -27,18 +27,45 @@ function formatMoney(_currency='EUR', amount) {
...
-	return formatter.format(number)
+	var result = formatter.format(number)
+	
+	if(number >= 0 && number < 1 && noZeroLead)
+		return result.substring(1);
+	return result;
+}
+
+function formatPrecisionWithRespectForFixedDecimals(number, sf, dp, noZeroLead=false) {
+	if(isNan(dp))
```
shun2wang:
```suggestion
	if(isNaN(dp))
```

shun2wang (changes requested):

shun2wang:
In test module "table formatting" seems "numbers" not following the fix decimal?
image

with 5 fix decimal set above.

JorisGoosen:
> In test module "table formatting" seems "numbers" not following the fix decimal?

yeah, i figured those might often be integers and then it would get kind of ugly if you put decimals.
Im assuming that all cases where people actually want decimals they also specify at least a format of some kind, and then  I think fixed decimals is applied.

shun2wang (approved):
OK, it's also better to add p value test case into test module. all seems great to me.
btw:I'm an advocate of reducing the use of `var` in js but this doesn't matter haha.

JorisGoosen:
Good idea yeah, but for now I will just merge this ^^

PR #5956, opened 2025-07-02
Open file by drag-and-dropping to JASP Window

boutinb:
From a file explorer, drag and drop a file to a JASP window: the file is opened either in this JASP instance if no file was not loaded yet, or start another JASP instance with this file.

### DIFF
--- a/Desktop/mainwindow.cpp
+++ b/Desktop/mainwindow.cpp
@@ -869,7 +869,7 @@ void MainWindow::openURLFile(QString fileURLPath)
 		return;
 	}
 
-	if (!FileTypeBaseValidName(fileInfo.suffix().toStdString()))
+	if (!FileTypeBaseValidName(fileInfo.suffix().toLower().toStdString()))
 	{
 		MessageForwarder::showWarning(tr("Open file"), tr("JASP does not support this file type %1.").arg(filePath));
 		return;
--- a/Desktop/mainwindow.cpp
+++ b/Desktop/mainwindow.cpp
@@ -853,7 +853,7 @@ void MainWindow::showLogFolder() const
 
 void MainWindow::openURLFile(QString fileURLPath)
 {
-	QUrl fileUrl = QUrl::fromLocalFile(fileURLPath);
+	QUrl fileUrl = fileURLPath.startsWith("file:") ? QUrl(fileURLPath) : QUrl::fromLocalFile(fileURLPath);
 	if (!fileUrl.isLocalFile())
 	{
 		MessageForwarder::showWarning(tr("Open file"), tr("Cannot access file %1").arg(fileURLPath));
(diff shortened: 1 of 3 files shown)


Desktop/mainwindow.cpp
```diff
@@ -851,6 +851,32 @@ void MainWindow::showLogFolder() const
 	openFolderExternally(AppDirs::logDir());
 }
 
+void MainWindow::openURLFile(QString fileURLPath)
+{
+	QUrl fileUrl(fileURLPath);
```
shun2wang:
```suggestion
	QUrl fileUrl = QUrl::fromLocalFile(fileURLPath);
```

Desktop/mainwindow.cpp
```diff
@@ -851,6 +851,32 @@ void MainWindow::showLogFolder() const
...
+	QString filePath = fileUrl.toLocalFile();
+	QFileInfo fileInfo(filePath);
+
+	if (!fileInfo.exists())
+	{
+		MessageForwarder::showWarning(tr("Open file"), tr("File %1 is not found.").arg(filePath));
+		return;
+	}
+
+	if (!FileTypeBaseValidName(fileInfo.suffix().toStdString()))
```
shun2wang:
```suggestion
	if (!FileTypeBaseValidName(fileInfo.suffix().toLower().toStdString()))
```

shun2wang (changes requested):

Desktop/mainwindow.cpp
shun2wang: Which not work on Linux for me unless above change, it save to make file work as local file scheme and unicode file name I think. see: https://doc.qt.io/qt-6/qurl.html#fromLocalFile

Desktop/mainwindow.cpp
shun2wang: we only compare lower case file suffix to avoid not open `example.XLSX` file, actually uppercase suffixes are also valid file names.

Desktop/mainwindow.cpp
shun2wang: I mean On Linux path like `/home/shun/文档/测试中文abc.XLSX` the scheme was empty.

Desktop/mainwindow.cpp
JorisGoosen: depends on the platform, on linux this change would make it impossible to open `example.XLSX` ...

shun2wang: yeah, so lower case is robust more.

shun2wang: > depends on the platform, on linux this change would make it impossible to open `example.XLSX` ...

you mean now use `toLower` ? no , it could open because we just to determine whether it is supported instead of changing the suffix. and I tested it already.

JorisGoosen (approved):

PR #5958, opened 2025-07-03
Better handling of interaction terms

boutinb:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/3029

The error comes from an (ugly) old code, that duplicates the terms (variables) in the model of the VariablesList (in the InteractionModel). As it is duplicated, the type change is not notified to the duplicated terms.

Moreover to be able to know if some variables comes from a randomFactors or covariates VariablesList (the handling is then slightly different), the name of the variablesList was hardcoded checked. An extra property is added sourceWithoutDefaultInteraction and is set per default set to ["randomFactors", "covariates"].


JorisGoosen:
This actually seems to fix https://github.com/jasp-stats/jasp-test-release/issues/3031 as well

JorisGoosen (approved):

PR #5959, opened 2025-07-03
Wait for all controls to be created before setting them up

boutinb:
When controls are created dynamically, wait first for all controls to be created before calling their setup function: the setup links the controls together (with their sources), so it needs all controls to be created first.


QMLComponents/models/listmodelfactorsform.cpp
```diff
@@ -216,7 +216,7 @@ void ListModelFactorsForm::ensureNesting()
 		{
 			if (currentModel == _factors[i].listView->model())
 			{
-				if (i > 0)
+				if (i > 0 )
```
JorisGoosen:
```suggestion
				if (i > 0)
```

JorisGoosen (approved):

PR #5960, opened 2025-07-04
Set the add column + sign correctly

boutinb:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2906

I see that there is a horizontal line under the + sign, but I cannot see in the code where this line comes from. I just tried my best to make the rectangle big enough so that this extra line stays hidden, but sometimes with some zoom scale, you can still see it. Maybe @JorisGoosen knows better where this line comes from and can remove it.


JorisGoosen:
The line is drawn by DataSetView, but also the height of the plus shouldve been set by the datasetview.
Also just putting `+1.8` on a size in a GUI that is zoomed is not a very smart move...

Ill try and fix the issue I guess.

PR #5968, opened 2025-07-08
Fix infinite PLS SEM time

JorisGoosen:
Ok, so the loop is caused by modelReset causing rowCountChanged because of https://github.com/jasp-stats/jasp-desktop/pull/5350#issuecomment-1854088400

Just checking whether we already sent the rscript is actually enough now, ive tested it both adding columns manually and by synchronizing and the rscipt only gets sent if something relevant changes.

The analysis still runs a bunch of time in the beginning but eventually quiets down

### DIFF
--- a/Desktop/engine/enginesync.cpp
+++ b/Desktop/engine/enginesync.cpp
@@ -1163,13 +1163,13 @@ void EngineSync::enginesPrepareForData()
 
 	int64_t tryTill = Utils::currentMillis() + ENGINE_KILLTIME;
 
-	while(!allEnginesPaused(pauseOrKillThese) && tryTill >= Utils::currentMillis())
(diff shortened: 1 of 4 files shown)


Desktop/engine/enginesync.cpp
```diff
@@ -1163,13 +1163,13 @@ void EngineSync::enginesPrepareForData()
 
 	int64_t tryTill = Utils::currentMillis() + ENGINE_KILLTIME;
```
JorisGoosen:
```suggestion
	//int64_t tryTill = Utils::currentMillis() + ENGINE_KILLTIME;
```

JorisGoosen:
Hmm, maybe this needs some changes, apparently "" is a valid response.

boutinb (approved):

PR #5969, opened 2025-07-08
add store state etc option, default off

JorisGoosen:
implements https://github.com/jasp-stats/INTERNAL-jasp/issues/2915 Saving example file BFI Network goes from 2.3mb to 617kb

Ive also triggered nightly builds for testing.


JorisGoosen:
I think this can be merged pretty safely and if done so before the data library is updated we might get a smaller installer even.

JorisGoosen:
Contains a new commit from jaspColumEncoder btw. However, it does not need to be in jaspBase so no new bundles required.

JorisGoosen:
See the `dontAlwaysStoreStateEtc` nightlies

JorisGoosen:
shall we merge this?
@RensDofferhoff wanna review? Or no time?

JorisGoosen:
@boutinb wanna review this?

Desktop/analysis/analysis.cpp
```diff
@@ -987,7 +989,7 @@ void Analysis::setUpgradeMsgs(const Modules::UpgradeMsgs &msgs)
 bool Analysis::needsRefresh() const
 {
 	bool differentVersion = _moduleVersion != (_dynamicModule ? _dynamicModule->version() : AppInfo::version);
-	return _wasUpgraded || differentVersion;
+	return _wasUpgraded || _storedWithoutState || differentVersion;
```
boutinb, line 992: Maybe the needsRefresh should become a property, and we can remove the _wasUpgraded and _storedWithoutState properties.
It does not have to be a QML property since it is used in QML only in combination with hasVolativeNotes that should be also removed.

I can do this if you agree.

boutinb (approved):

PR #5970, opened 2025-07-09
Make sure doubles are converted to labels properly!

JorisGoosen:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2917

### DIFF
--- a/CommonData/column.cpp
+++ b/CommonData/column.cpp
@@ -1933,17 +1933,13 @@ void Column::upgradeDoublesToLabels()
 	
 	beginBatchedLabelsDB();
 	
-	for(double dbl : dbls)
-		if(!dblToIntsId.count(dbl))
-		{
-			std::string		dblStr	= Column::doubleToDisplayString(dbl);
-			Label		*	label	= labelByValue(dblStr);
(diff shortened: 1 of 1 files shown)


CommonData/column.cpp
```diff
@@ -1933,17 +1933,13 @@ void Column::upgradeDoublesToLabels()
 	
 	beginBatchedLabelsDB();
```
boutinb, line 1930: Why there is no endBatchedLabelsDB ?
Also line 1930 `doubleset dbls(_dbls.begin(), _dbls.end());` can be removed

JorisGoosen: Because this is done later in the batched update of all values.

See: https://github.com/jasp-stats/jasp-desktop/blob/65c34faaaa92f91db7870aa8b2c666406a82575c/CommonData/dataset.cpp#L75

JorisGoosen: > Also line 1930 doubleset dbls(_dbls.begin(), _dbls.end()); can be removed

Ok, I will make the suggestion then...
image

boutinb (approved):

PR #5972, opened 2025-07-14
feat: showing column descriptions also on variable list while hovered

shun2wang:
part of: https://github.com/jasp-stats/jasp-issues/issues/3524

It is a quick request from users they have many short name cols and want to know description quickly before drag it, minor but improves the user experience.

### DIFF
--- a/QMLComponents/components/JASP/Controls/VariablesList.qml
+++ b/QMLComponents/components/JASP/Controls/VariablesList.qml
@@ -559,6 +559,11 @@ VariablesListBase
 					hoverEnabled:	true
 					cursorShape:	itemRectangle.typeChangeable && mouseX < icon.width ? Qt.PointingHandCursor : Qt.OpenHandCursor
 
+					onEntered:
+					{
+						itemRectangle.formatToolTip()
+					}
+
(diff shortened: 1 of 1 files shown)


boutinb:
I'm not sure to understand: the description is already displayed in the tooltip of a variables in the VariablesList...
I don't see what the onEntered adds here.

shun2wang:
> I'm not sure to understand: the description is already displayed in the tooltip of a variables in the VariablesList... I don't see what the onEntered adds here.

No, it was just in showing in right assigned list but not in left variable list:
image

Now it will showing variable description in right list then user will saw which col means.

shun2wang:
Ah, sorry I know what happend, because the description was now updated immediately while input filed lose focus.It's really different things.

tomtomme:
though it would be cool if the tooltip would show the "long name" instead of the "short name" IF no "description is given

shun2wang:
> though it would be cool if the tooltip would show the "long name" instead of the "short name" IF no "description is given

Yes, the long name should be some thing like Lable for value. that was I means in the request.

but this PR not implement that.

PR #5986, opened 2025-07-31
Colorblind friendlier palettes

vincentott:
Context: https://github.com/jasp-stats/jaspGraphs/pull/118

I want users to switch to colorblind friendlier palettes.
This is why I would like to make Colorblind #3 the new default at the very top.

To ensure backwards compatibility, I kept Colorblind and Colorblind #1
Yet, they are marked as deprecated and moved to the very bottom.

Maybe this is not the optimal solution yet, so feel free to fix/build on my current proposal.

### DIFF
--- a/QMLComponents/components/JASP/Controls/ColorPalette.qml
+++ b/QMLComponents/components/JASP/Controls/ColorPalette.qml
@@ -27,15 +27,17 @@ DropDown
 	indexDefaultValue: 0
 	values:
 	[
-		{ label: qsTr("Colorblind"),		value: "colorblind"		},
-		{ label: qsTr("Colorblind #2"),		value: "colorblind2"	},
-		{ label: qsTr("Colorblind #3"),		value: "colorblind3"	},
-		{ label: qsTr("JASP"),				value: "jaspPalette"			},
-		{ label: qsTr("Viridis"),			value: "viridis"		},
-		{ label: qsTr("ggplot2"),			value: "ggplot2"		},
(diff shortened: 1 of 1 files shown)


QMLComponents/components/JASP/Controls/ColorPalette.qml
```diff
@@ -27,15 +27,17 @@ DropDown
...
-		{ label: qsTr("ggplot2"),			value: "ggplot2"		},
-		{ label: qsTr("Gray"),				value: "gray"			},
-		{ label: qsTr("Blue"),				value: "blue"			},
-		{ label: qsTr("Sports teams: NBA"),	value: "sportsTeamsNBA"	},
-		{ label: qsTr("Grand Budapest"),	value: "grandBudapest"	}
+		{ label: qsTr("Colorblind #3"),			   value: "colorblind3"	   },
+		{ label: qsTr("Colorblind #4"),			   value: "colorblind4"	   },
+		{ label: qsTr("Viridis"),				   value: "viridis"		   },
+		{ label: qsTr("Inferno"),				   value: "inferno"		   },
+		{ label: qsTr("JASP"),					   value: "jaspPalette"	   },
```
vandenman, line 34: I'm not sure if the naming makes sense here. Why not do Colorblind 1 through n and we mention elsewhere that this was different before version XXX? Also, please consistently use tabs for the indents instead of spaces.

vinniott:
Thanks @tomtomme for assigning me (writing from new github account).
I will try to address @vandenman's comment soon.

PR #5999, opened 2025-08-28
Recovery

JorisGoosen:
Implements [feature request of having autosave and a recovery mode](https://github.com/jasp-stats/INTERNAL-jasp/issues/2953)

How?
=====

It adds a preference to data preferences where (default enabled) autosave and its interval in minutes can be specified.
If enabled it will make an autosavefile if there are active unsaved changes, and if there already is an autosave it is "touched" so it will still look fresh. This file is stored in a folder next to "Logs" called "AutoSaves" and all files get a similar name like "autosave-*pid*-0. To allow the user to get some insight in what the file could be 2 extra fields are added to "analyses.json" in each jaspfile, with a name and some description. 

Recovering
====
Loading is then done through the added file open menu of "Recovery", where all recovery files are shown of the user, excluding the recoveryfile of the current jasp process (if any). This is ordered on last modified time and will, as a  side-effect, also delete autosaves older than a month.

image
image

If a recovery file is then saved the file is considered "recovered" and the autosave file is moved to the trash or, if that fails, removed.

### DIFF
--- a/Desktop/mainwindow.cpp
+++ b/Desktop/mainwindow.cpp
@@ -402,6 +402,8 @@ void MainWindow::makeConnections()
 	connect(_package,				&DataSetPackage::dataModeChanged,					_engineSync,			&EngineSync::dataModeChanged								);
 	connect(_package,				&DataSetPackage::dataModeChanged,					this,					&MainWindow::onDataModeChanged								);
 	connect(_package,				&DataSetPackage::askUserForExternalDataFile,		this,					&MainWindow::startDataEditorHandler							);
+	connect(_package,				&DataSetPackage::makeAnAutoSave,					this,					&MainWindow::saveTmpFileHandler								);
+	
 	connect(_package,				&DataSetPackage::runFilter,							_filterModel,			&FilterModel::sendGeneratedAndRFilter						);
 	connect(_package,				&DataSetPackage::showWarning,						_msgForwarder,			&MessageForwarder::showWarningQML,							Qt::QueuedConnection);
 	connect(_package,				&DataSetPackage::synchingExternallyChanged,			_fileMenu,				&FileMenu::dataAutoSynchronizationChanged					);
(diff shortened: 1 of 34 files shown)


Desktop/mainwindow.cpp
```diff
@@ -2053,6 +2083,15 @@ void MainWindow::saveJaspFileHandler()
 	dataSetIORequestHandler(saveEvent);
 }
 
+void MainWindow::saveTmpFileHandler()
+{
+	FileEvent * saveEvent = new FileEvent(this, FileEvent::FileSave);
```
boutinb, line 2088: I'd have added a FileEvent::FileSaveQuiet type, instead of reusing the FileSave and adding the `tmp` flag. This tmp flag is not really descriptive.

Desktop/utilities/settings.cpp
```diff
@@ -115,6 +115,8 @@ const Settings::Setting Settings::Values[] = {
 	{"useConfigurationFile",		true	},
 	{"startMaximized",				false	},
 	{"storeStateEtc",				false	},
+	{"autoSaveOn",					true	},
+	{"autoSaveInterval",			5*60	},
```
boutinb: I would say every minute per default instead.

boutinb (commented):
It works perfect!
Really nice feature.

I've just 2 comments, you may ignore.

Desktop/mainwindow.cpp
JorisGoosen: Then I would have had to check every place where FileEvent::FIleSave was being checked for. Now I just have this flag.

If you have a better name we can rename it

Desktop/utilities/settings.cpp
JorisGoosen:
```suggestion
	{"autoSaveInterval",			60	},
```

JorisGoosen:
So we can merge it ?

boutinb (approved):

PR #6000, opened 2025-08-29
Add auto transforms to computed columns window

Kucharssim:
Fix for Gitea SGC#15

A related PR that implements the transforms is here: https://github.com/jasp-stats/jaspBase/pull/176.

@JorisGoosen, is there a way to specify that an argument is optional? The R functions can be called just like `BoxCoxAuto(variable)`, but the computed columns window always expects that the user specifies all arguments.


Kucharssim:
Wow a pull request number 6000. Crazy.

JorisGoosen:
Ok, Ive added a "Boolean" support qml item to the filterconstructor to make that work (this wasnt supported).

However now I run into the problem that it isnt allowed to enter both "groupSize" and "predictor".
But, the filterconstructor functions assume that all arguments *must be* entered.

So it doesnt work right now.

Im simply going to add two dummy functions that pick one of either options.
Or add support for having optional arguments...

Edit: As I wouldve noticed if Id reread the actual short description of the PR haha.
It now works though

PR #6002, opened 2025-09-03
Testing Data Import JASP

JorisGoosen:
Adds a JASPTest executable which we can expand. 

Ive added a set of datafiles in Test/TestLibrary
All except the `jasp` ones are run by the test and compared to a `json` file right next to each.
If the json is missing it will generate it.


shun2wang:
should this work with GitHub workflows under headless mode?

JorisGoosen:
> should this work with GitHub workflows under headless mode?

I guess it could?

shun2wang:
> > should this work with GitHub workflows under headless mode?
> 
> I guess it could?

Great!

Which is the command argument actually? maybe I can add it after this is merged.

JorisGoosen:
The `BUILD_TESTS` property in cmake must be `ON` and then you should run `JASPTest`

JorisGoosen:
Ive found a way to improve this a bit, so I will.
Dont merge yet please

JorisGoosen:
This is better

Desktop/modules/ribbonmodel.cpp
```diff
@@ -141,6 +141,8 @@ void RibbonModel::addSpecialRibbonButtonsEarly()
 		new AnalysisEntry([&]() { emit setDataSynchronisation(false);	},					"stop-externaledit",			[](){ return fq(tr("Turn external data synchronisation off"));},		true, "")
 	});
 
+	if(QCoreApplication::applicationName() == "JASPTest")
+		_testButton			= new RibbonButton(this, "Test",					[&](){ return fq(tr("Test"));},						"summarize.svg",			false, [&](){ emit runTests();	},									[&](){return tr("Run tests");},													true);
```
boutinb, line 145: I don't see how this is used?

JorisGoosen: No this was for a possible GUI running on top of jaspTest, but I already put in the effort to add it and might use it later

JorisGoosen: (Like I had the gui running also but removed it again because it conflicted with the test executable main generated by QT)

JorisGoosen:
Why did you do that?

boutinb:
I have added a commit so that the testing resources are copied in the build folder.

JorisGoosen:
> I have added a commit so that the testing resources are copied in the build folder.

I can see that, what I would like to know is why

boutinb (approved):

PR #6020, opened 2025-09-16
Update jasp-module-workflow.md

doomlab:
updating the window view and options that don't cause JASP to crash when creating a new developer module

Just suggesting a change - working through the developer guidelines and this window has changed - I tried a couple combinations before this one finally worked without crashing. 

- am on mac os 15.6.1 (24G90)
- Apple M3 Max

### DIFF
--- a/Docs/development/jasp-module-workflow.md
+++ b/Docs/development/jasp-module-workflow.md
@@ -73,11 +73,9 @@ R CMD INSTALL . --preclean --no-multiarch --with-keep.source <module name>
 3. Navigate to 'Preferences'
 4. Navigate to 'Advanced'
 5. Place a checkmark before 'Developer mode'
-6. Tick 'Enable direct libpath mode'
-7. Choose the libpath where you precompiled your module (not sure?, execute `.libPaths()` in R)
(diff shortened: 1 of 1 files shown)


JorisGoosen:
Ah hey, thanks for the suggestion ^^

So we added a new way of installing development modules, where you use `renv` and its lockfiles to specify your exact dependencies. And then you can install your Rpackage/module within R to a particular custom libPaths (an renv handled library).
This is the system we will be moving towards in the future, because having JASP play nice with everyones development systems turns out to be quite painful.
Better to let R handle it.

However, we do still support installing devmodules through JASP, so I will add your screenshot too, with some extra text explaining the difference.

JorisGoosen:
So I couldnt check out your repo to push to this branch.

No matter, thanks for pointing out the inconsistency in the docs and uploading a screenshot.

Can you check out the new text I wrote in https://github.com/jasp-stats/jasp-desktop/pull/6023 and see if it is a bit clearer?
Thanks!

doomlab:
Thanks for the quick response! I checked it out.

PR #6024, opened 2025-09-18
Fix some loading related problems

JorisGoosen:
move the intialization of empty values up to avoid them showing up in the labelsview
fixes jasp-stats/INTERNAL-jasp#2955

Also loading a file made with a different locale can mess up the labels.
This is easily fixed though by simply not storing the double as a string.

Fairly sure this is all good but needs to be tested!

### DIFF
--- a/Desktop/data/datasetpackage.cpp
+++ b/Desktop/data/datasetpackage.cpp
@@ -1493,7 +1493,8 @@ void DataSetPackage::loadDataSet(std::function<void(float)> progressCallback)
 	_db->load();		
 	_db->upgradeDBFromVersion(_jaspVersion);
 	
-	bool do019Upgrade = _jaspVersion < "0.19";
+	bool 	do019Upgrade  = _jaspVersion < "0.19",
+			do0953Upgrade = _jaspVersion < "0.95.3";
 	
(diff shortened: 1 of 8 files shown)


Desktop/data/datasetpackage.cpp
```diff
@@ -1493,7 +1490,8 @@ void DataSetPackage::loadDataSet(std::function<void(float)> progressCallback)
 	_db->load();		
 	_db->upgradeDBFromVersion(_jaspVersion);
 	
-	bool do019Upgrade = _jaspVersion < "0.19";
+	bool 	do019Upgrade  = _jaspVersion < "0.19",
+			do0953Upgrade = _jaspVersion < "0.95.3";
```
boutinb: do0953Upgrade is not used

boutinb (commented):
I have tested with french alternative language, with different separators for thousands and decimals, and it works perfect.
Just one little comment in datasetpackage.cpp

Desktop/data/datasetpackage.cpp
boutinb:
```suggestion
	bool 	do019Upgrade  = _jaspVersion < "0.19";
```

PR #6025, opened 2025-09-19
Add possibility to open a JASP file with other data files and report results

### LINKED ISSUE #3260: [Feature Request]: Data synchronization with new csv
### Description

The "synchronization" button could use a dedicated sub-section named e.g. "switch data file" which then ignores possibly already existing CSVs and lets me search for a new one that I want to switch to

### Purpose

Switching the data file is kind of hard if I do not want to copy paste the new data e.g. from excel to jasp

### Use-case

copy pasting really big data is not very handy

1. open jasp file that already has a same name CSV alongside  
2. switch to data edit mode 
3. hit sync button 
4. I cannot choose another file because the CSV is found and opened. 
5. So I have to delete the CSV firt. 6. Now hitting "Synchronization" again lets me point to a different CSV

keep copy pasting

boutinb:
This adds the possibility from the command line to run analyses with other data files and export the results in PDF or HTML.
For this the command line supports extra options: `JASP  <JASP file> (<data file 1> <data file 2> ...) --inputDataDir <path to folder> --outputDir <path to folder> --exportPdf --dontExportResult  --keepJASPOpen`

- The JASP file is first loaded.
- It synchronizes with 1 or more data files: either the ones directly specified in the command line, or all data files found in the inputDataDir folder (and subfolder).
- For each data file, when all analyses are all refreshed, it exports the result in HTML format (or PDF format if `--exportPdf` is specified). The export file is generated in the same folder as the data file, or in the folder specified by `-outputDir`. This step can be discarded if `-dontExportResult` is specified.
- Afterwards JASP is closed per default, except if `--keepJASPOpen` is specified.

To realize this, the argument parsing of the main function has been rewritten and uses now a ParsedArguments object.
Also the handling of the FileEvent has been partly refactored: it has now a status property that states whether the event is `initialized`, `started`, `completed`. All the File Events follow now the same process, and are also deleted at the end (they were not deleted before). By making the process clearer, event chaining is now possible without risking that one handling of the first event does mangle with the handling of the second event. This was needed to start the synchronization of the data file only when the JASP file is completely opened.

I have also added the possibility to drag and drop a data file to a JASP instance, so that it synchronizes directly the analysis with the new data.


Desktop/data/fileevent.h
```diff
@@ -87,10 +92,11 @@ private slots:
 						_dataFilePath,
 						_last_error		= "Unknown error",
 						_message;
-	bool				_completed		= false,
+	bool				_started		= false,
```
JorisGoosen: an enum (via DECLARE_ENUM) describving the state of the fileevent is better

boutinb:
@FBartos You can review it

FBartos:
I will try to get it build so I can test it out. I will be operating in a very limited capacity till the end of July btw

boutinb:
I have added a --keepMissingColsWhenSyncing argument, so that if a column is missing during the synchronization, it is replaced by an empty column (with same name).
Also it checks now if the Form has an error: if a column has not anymore enough levels for example, the form gets an error, and the analysis is not run. Now this error is also displayed in the result.

FBartos:
I can reproduce two batch CLI forwarding issues on Windows at `7412d8123` (`Add keepMissingColsWhenSyncing argument and take care of error`).

1. `--dontExportResult` works in the single-file sync path, but not in the multi-file batch path.

Single-file control:

```powershell
JASPDesktop.exe template.jasp dra-baseline.csv --outputDir recheck-no-export-single --dontExportResult --hide --timeOut=1
```

Result: process finished and no `.html`/`.pdf` files were written.

Batch repro:

```powershell
JASPDesktop.exe template.jasp dra-baseline.csv dra-treatment-boost.csv --outputDir recheck-no-export-batch --dontExportResult --hide --timeOut=1
```

Result: two HTML files were still written:

```text
dra-baseline.html
dra-treatment-boost.html
```

The parent stdout showed the child processes were started without `--dontExportResult`:

```text
Starting subJASP with args: ... template.jasp ... dra-baseline.csv --outputDir ... --hide --timeOut=1
Starting subJASP with args: ... template.jasp ... dra-treatment-boost.csv --outputDir ... --hide --timeOut=1
```

2. `--keepMissingColsWhenSyncing` is also not forwarded to batch child processes.

Batch repro:

```powershell
JASPDesktop.exe template.jasp dra-baseline.csv dra-missing-column.csv --outputDir recheck-keep-missing-batch --keepMissingColsWhenSyncing --hide --timeOut=1
```

The child command lines again omitted the flag:

```text
Starting subJASP with args: ... template.jasp ... dra-baseline.csv --outputDir ... --hide --timeOut=1
Starting subJASP with args: ... template.jasp ... dra-missing-column.csv --outputDir ... --hide --timeOut=1
```

This looks like a small gap in `Desktop/main.cpp::syncDataFiles()`: it forwards `--outputDir`, `--exportPdf`, `--save`, `--hide`, and `--timeOut`, but not `--dontExportResult` or `--keepMissingColsWhenSyncing`. The single-file path in `Application::init()` does consume these options directly, which is why the behavior differs between single-file sync and batch sync.

FBartos:
Additional review findings beyond the earlier comment about `--dontExportResult` / `--keepMissingColsWhenSyncing` not being forwarded in batch mode:

1. Batch parent masks child failures. `syncDataFiles()` starts child JASP processes but never checks `waitForStarted()`, `exitStatus()`, or `exitCode()`, and the parent exits `0` after `--inputDataDir` processing. Repro: run recursive sync with at least one child that times out/crashes/fails to export; the parent only logs stderr and still exits successfully. Code: `Desktop/main.cpp:178-219`, `Desktop/main.cpp:307-310`.

2. Single child sync/export can continue after failure and still exit `0`. In `_open()`, `syncEvent->finalized` starts the export path without checking `syncEvent->isSuccessful()`, and both export/no-export finalizers emit `exitSignal(0)` unconditionally. Repro: use a valid `.jasp` template with a deliberately invalid data file, for example a fake `.xlsx` containing plain text; expected behavior is no successful export and a nonzero CLI result, but this path can still continue to export/exit success. Code: `Desktop/mainwindow.cpp:1142-1159`.

3. Database JSON CLI opening seems regressed. The first positional argument now goes through `checkFile()` while `mainFilePath` is empty; if it is not an existing/typed file, parsing errors before the database JSON branch can be reached. Repro: start JASP with a database JSON payload as the first positional argument, which the previous parser handled after the nonexistent-file check; current code treats it as an invalid file instead of setting `dbJson`. Code: `Desktop/parsedarguments.cpp:156-184`, `Desktop/parsedarguments.cpp:230-256`.

4. Canceling Computer open/save dialogs is treated like a failed file event. `browseOpen()` / `browseSave()` call `event->setComplete(false)` without `cancelled=true`, while `FileEvent` now auto-finalizes through `FileMenu`. Repro: open a Computer file dialog and press Cancel; this can go through failed open/save handling instead of silent cancellation. Likely fix: use `setComplete(false, "", true)` for intentional cancels, or avoid finalizing those as failed events. Code: `Desktop/widgets/filemenu/computer.cpp:61-69`, `Desktop/widgets/filemenu/computer.cpp:120-142`, `Desktop/data/fileevent.cpp:35-49`.

5. Hidden mode needs deployment support for `qminimal`. On Windows, `--hide` appends `-platform minimal`; without `platforms/qminimal.dll`, hidden/batch runs show `no Qt platform plugin could be initialized`. Repro: run the PR build with `--hide` before manually copying `qminimal.dll` into `build/platforms`.

Small hygiene item: `git diff --check upstream/development...github-desktop-boutinb/changeDataFile` reports trailing whitespace in `Desktop/mainwindow.cpp` and `Desktop/widgets/filemenu/recentfiles.h`, plus a blank line at EOF in `Desktop/parsedarguments.cpp`.

FBartos (changes requested):
See comments above, seems like a few more bugs are present

JorisGoosen:
Ill do a review soon

Desktop/data/importers/importer.cpp
```diff
@@ -316,14 +318,24 @@ void Importer::syncDataSet(const std::string &locator, std::function<void(int)>
 	
 	tasks.clear();
 	
-	DataSetPackage::pkg()->dataSet()->endBatchedToDB([&](float f){ progress(75 + f * 25); });
+	dataSet->endBatchedToDB([&](float f){ progress(75 + f * 25); });
 
 	for (Column * oldCol : oldColumns) //already checked for not being computed column at creation list
 	{
-		Log::log() << "Column deleted " << oldCol->name() << std::endl;
+		if (PreferencesModel::prefs()->keepMissingColsWhenSyncing())
```
JorisGoosen: Does this not mean that every dataset's extra columns (not present in other files) you open all get added to the dataset over and over? So suppose each dataset has 3 randomly named extra columns and you open 10 of such datasets in a row, then there are 30 extra columns right?

JorisGoosen: I need to add a test for this

boutinb:
The feature is now guarded by the PRO define

Desktop/components/JASP/Widgets/MainWindow.qml
```diff
@@ -89,25 +89,58 @@ Window
...
+						parent.state = "active"
+				}
+			}
+
+			states: [
+					State {
+						name: "active"
+						PropertyChanges {
+							warningRect {
+								color: jaspTheme.blueLighter
```
JorisGoosen, line 130: So we set the color for dropping a file via a state but the color for a fatalerror via binding? This is quite inconsistent.

Desktop/data/importers/importer.cpp
JorisGoosen: @boutinb ?

JorisGoosen:
Nice work overall — the FileEvent state machine, the FileMenu singleton routing, and the `ParsedArguments` extraction are real improvements over the old plumbing, and the CLI batch flow is a clean way to get Pro data-swapping without touching the core sync machinery. A few things should still be addressed before merge; several of my earlier review points stand as well. Everything below is checked against the current head.

## Blockers that I found

**1. Cancelled Open dialog resets the workspace (data loss)**
`Computer::browseOpen()`/`browseSave()` mark a cancelled dialog with `event->setComplete(false, "", true)` (computer.cpp). Because `FileEvent` now auto-connects `completed → FileMenu::finalizeFileEvent → MainWindow::fileEventRequestFinalize` in its constructor, a *dismissed* Open dialog now falls into the `FileOpen` failure branch and runs `_package->reset()` → `deleteWorkspace()` **unconditionally** (mainwindow.cpp ~1947). Previously, cancelled events never reached this handler, so the workspace survived. `isCancelled()` is checked only to suppress the warning — the reset itself is not guarded. Fix: early-return on `event->isCancelled()` in the failure branch.

**2. Sync is no longer runnable without the FileMenu singleton (regression of a core capability)**
Before this branch, `DataSetSyncer` → `AsyncLoader::onSyncRequired` created its own `FileEvent` and drove it through `io(event)` with no FileMenu involved — the per-dataset sync machinery worked against a bare `AsyncLoader` (that is exactly what `TestAll::testFileSyncerFullAsyncFlow` does, Tests/testall.cpp:1066). Now the `FileEvent` constructor hard-wires `started`/`completed` to `FileMenu::singleton()`, which is null in that context — the connect is refused, the event never starts and never finalizes, and the syncer's re-entrancy guard is never released. The tests will hang/fail, and any future headless/embedded use of the sync pipeline is bricked. This matters doubly because the whole point of this refactor was tidying the event pipeline — coupling the *data-sync* path (DataSetSyncer in CommonData → AsyncLoader) to a UI widget's lifetime undoes part of the layering. My strong preference: keep the sync path UI-independent — make the `starts()/finalized` targets injectable (or add a "no UI routing" constructor variant used by `onSyncRequired`), guard the connect against a null singleton with a loud log, and adjust the tests.

**3. The wait-for-analyses trigger is single-shot**
`waitForAllAnalysesFinishedBeforeStartingEvent()` permanently `disconnect()`s itself from `Analyses::analysisResultsChanged` after the first waiting event starts (mainwindow.cpp ~1227). A second queued event in the same process (`--keepJASPOpen` + a second sync/drop/export, or any repeated use) sets `_waitingEvent` and then **waits forever** — export never starts, and if `keepJASPOpen` is false the process hangs instead of exiting. Fix: use a `SingleShotConnection`, or re-connect each time `_waitingEvent` is assigned.

**4. `JASPTestParsedArgs` breaks non-Pro builds**
`--exportPdf`, `--outputDir`, `--inputDataDir`, `--dontExportResult`, `--keepJASPOpen`, `--keepMissingColsWhenSyncing` are all inside `#ifdef PRO` in `parsedarguments.cpp`, but the new tests exercise them unconditionally. In a community build (`PRO=FALSE` by default in CMakeLists.txt), those flags fall into the positional-argument branch → `checkFile` fails → **`exit(1)` kills the whole test binary mid-run**. Community CI for `testParsedArgs` will be red. Fix: guard those test cases with `#ifdef PRO`, or only register the test target under `PRO`.

**5. `FileMenu` singleton lifecycle is fragile**
`_singleton` is set once (assert-guarded) but never cleared in `~FileMenu`. It's also a hidden global dependency for every `FileEvent` created anywhere. Once blocker #2's explicit wiring is in, this becomes mostly a robustness matter: clear the pointer in the dtor (or drop the singleton in favor of explicit targets), and never silently accept a null target during `FileEvent` construction.

## My earlier review points that still stand

**6. `Importer::syncDataSet` column accumulation with `keepMissingColsWhenSyncing`** (my comment at importer.cpp:330)
Every column missing from the new file is kept-and-emptied on *every* sync, with nothing distinguishing "kept because missing" from "kept on a previous sync". Sync A→B keeps A's extras, sync B→C keeps B's *and* A's extras. With keepJASPOpen + repeated syncs this grows a monotonically increasing union of columns. This needs:
- a filter so columns are only kept once (e.g., snapshot the template schema on first sync, or only keep if the column is non-empty / newly introduced), and
- a test for exactly this scenario, per my follow-up. None exists yet (`Tests/` has no `keepMissingCols` coverage).

**7. MainWindow.qml color logic inconsistency** (my comment at MainWindow.qml:128)
The drop-highlight color is set via a `State`/`PropertyChanges`, while the fatal-error color is an inline ternary binding on the same `Rectangle` (`color: mainWindow.hadFatalError ? jaspTheme.red : "transparent"`). Please pick one mechanism for both (I'd suggest moving the fatal-error color into the state machine too). Also worth checking: a successful drop currently clears the fatal-error state (`parent.state = ""`).

**8. `FileEvent::EventStatus` should use `DECLARE_ENUM`** (my comment on fileevent.h)
Per my earlier comment — a bare enum has no `toString`/`fromString`/registration; the codebase pattern (`Common/enginedefinitions.h`, `Common/utilenums.h`) gives you logging and future QML exposure for free.

## Medium / should-fix

**9. `AsyncLoader::onSyncRequired` dereferences a possibly-null `dataSet`** (asyncloader.cpp ~71): the `!dataSet` guard logs "aborting" and then calls `emit syncCompleted(dataSet->id(), false)`. Capture the id before the guard.

**10. Untranslated string written into the results pane** (mainwindow.cpp:1235): `setErrorInResults("Validation error: " + …)` is user-visible in (and exported with) the results — wrap it in `tr("Validation error: ")`. The QML preference strings are all correctly `qsTr()`'d, so this is the only missing one.

## Minor / cleanup

- CLI help/diagnostics remain untranslated English — **accepted as-is** (consistent with the old `parseArguments` behavior; the parser also runs before `QApplication` exists, so `tr()` there would need deferral anyway).
- `syncDataFiles` exits 0 silently when there are no data files and `inputDataDir` doesn't match anything — fail loudly instead.
- First positional as an OSF/HTTP URL + trailing data files is silently mis-parsed (URL paths have no local `QFileInfo::exists()`, so later positionals overwrite `mainFilePath`).
- `--keepMissingColsWhenSyncing` help text typos: "afetr", "wlll".
- `FileMenu::startFileEvent` logs under the old `dataSetIOCompleted` label; duplicate `#include` blocks in fileevent.cpp.
- `FileMenu::setVisible(false)` now happens at *finalize* rather than request time — probably fine, but double-check the Sync-Data page UX while a sync is in flight.
- Exit codes >255 truncate in `syncDataFiles`.

boutinb:
All the issues have been tackled.

On #5 (FileMenu singleton)

Done. ~FileMenu now clears _singleton (guarded with == this so it can't null out another instance), so a FileEvent constructed after the menu is gone no longer routes into freed memory. The assert against a second FileMenu is gone in release builds, so there's a log next to it rather than silently re-pointing every future event's routing.

Combined with the UI-independence change, construction never silently accepts a null target either: it logs, and the data-syncer path opts out of FileMenu routing entirely instead of relying on it. I've left the singleton itself in place — replacing it with injected targets is a bigger refactor than this PR should carry.

On keepMissingColsWhenSyncing (#6)

The important part: your premise doesn't hold, and we have a test proving it. Worth stating plainly but with the evidence:

I went to add the filter and the regression test, and the test showed the accumulation doesn't actually work the way we thought. Before the keep/remove loop, syncDataSet recycles unmatched old columns into the columns the new file introduces (the changeNameColumns path) — so they're renamed, not kept alongside.

Concretely: dataset [a,b,c] synced against [p,q] gives [p,q,c] — a becomes p, b becomes q, and only the leftover c is kept as an empty column. I first wrote the test expecting the union (5 columns) and it failed with 3, which is how this came out.

So the column count is max(previous, new file) — it never shrinks within a session, but it doesn't grow into a union of everything either. Sync A→B→C doesn't accumulate A's and B's extras; each sync's new columns absorb the previous leftovers.

Given that, I don't think the template-schema filter earns its keep: it'd add state to DataSet to bound something that's already bounded. What I did instead is document the real behaviour in the --keepMissingColsWhenSyncing help text (leftovers persist for the session, which matters only with --keepJASPOpen, since every data file otherwise gets its own process) and add testSyncKeepMissingColumns covering both settings plus the [a,b,c] + [p,q] → [p,q,c] case. Happy to add the filter anyway if you still want the leftovers gone, but it'd be a behaviour change rather than a bug fix.

Two supporting facts you can deploy if he pushes back: the flag is CLI-only (not in Settings, not in QML — set once in Application::init), and batch mode spawns one subprocess per data file, so the documented workflow can't accumulate at all. Accumulation needs --keepJASPOpen plus manual repeated syncs.

JorisGoosen:
I dont really understand what "keepMissingColsWhenSyncing" even is supposed to do now?
If I see the name I think "Synch datafile, so all columns that now exist in my dataset get updated with the contents of the exactly same-name-having-cols in sourcefile. And any cols from synch not in current get added to current dataset. And all columns in the current dataset that ARENT in the source (so missing) are KEPT". In other words, source [a,b,c] and curdata [c,d,e] gets me [a,b,c,d,e].

But the way you explain it it sounds like what instead happens is:
> Concretely: dataset [a,b,c] synced against [p,q] gives [p,q,c] — a becomes p, b becomes q, and only the leftover c is kept as an empty column. I first wrote the test expecting the union (5 columns) and it failed with 3, which is how this came out.

Following that logic my example becomes: [a,b,c]... What is kept? What is missing?

boutinb:
> I dont really understand what "keepMissingColsWhenSyncing" even is supposed to do now? If I see the name I think "Synch datafile, so all columns that now exist in my dataset get updated with the contents of the exactly same-name-having-cols in sourcefile. And any cols from synch not in current get added to current dataset. And all columns in the current dataset that ARENT in the source (so missing) are KEPT". In other words, source [a,b,c] and curdata [c,d,e] gets me [a,b,c,d,e].
> 
> But the way you explain it it sounds like what instead happens is:
> 
> > Concretely: dataset [a,b,c] synced against [p,q] gives [p,q,c] — a becomes p, b becomes q, and only the leftover c is kept as an empty column. I first wrote the test expecting the union (5 columns) and it failed with 3, which is how this came out.
> 
> Following that logic my example becomes: [a,b,c]... What is kept? What is missing?

I have changed the code so that keepMissingColsWhenSyncing does really what it means.
The dataset will indeed grow after each synchronization if there are new columns and old columns disappear. But the problem is minimal when running multiple datafiles, since the JASP instance is restarted for each datafile.

JorisGoosen:
Can you please now explain here *what* the option is *supposed* to do?

boutinb:
> Can you please now explain here _what_ the option is _supposed_ to do?

If -keepMissingColsWhenSyncing is specified in the command line, synchronization will keep the columns that are in the current dataset, but are missing in the new dataset. If dataset has columns a & b, and is synchronized with a dataset with columns a & c, the result will be a dataset with columns a, b & c (a & c having the values of the new dataset, b having the values of the original dataset).
I have changed the command line help.

PR #6029, opened 2025-09-24
JASP Encryption Files 

RensDofferhoff:
JASP encryption files implemented using libsodium asym encryption.
A password is used to seed keypair generation.
 
Algo details: https://doc.libsodium.org/public-key_cryptography/authenticated_encryption
This stuff is top notch but the asymmetric case is  not quantum safe (yet,  its on the libsodium roadmap but realistically not an issue for looooong time).
symetric case is  quantum safe

Can be used to encrypt JASP files with a password.
Optional JASP submission checkbox so the JASP-Team can also decrypt the files in question.

How to use:

On the Save file select window, select Encrypted JASP file as type.
image

Enter Encryption details:
image

Chosen password hash function is a bit hefty to make weak password rainbow tables expensive.


RensDofferhoff:
wish to add one more thing

shun2wang:
Build test on Linux was failed.

boutinb:
Great feature!!
I have updated a bit the encryption dialog box (by using a Group, you get automatically alignment of the TextFields).
I found the feature a bit hidden when you save a JASP. It is (nearly) impossible to add a checkbox to a native dialog (default dialogs are native), so I understand that you chose to add a filter (even if it has the same jasp suffix), but I'm afraid that most of the users won't see this option. That's why I have changed the title of the dialog by 'Save (encrypted) JASP file". But maybe someone has a better idea to increase the attention of the user for this feature.

2 issues:

- If you do Save As with encryption, if you close the encryption dialog, it still saves the file.
- if you load an encrypted file, if you close the encryption dialog, it still tries to load the file, and you get an error that it could not load it: in this case it should not try to load it.

shun2wang:
>. But maybe someone has a better idea to increase the attention of the user for this feature.


We can add sub menu while cursor hover the "save as" button then push two options to the right, a "save as", and a "save as 
encrypted",and the second option always comes with an encryption dialog.

RensDofferhoff:
That's a good suggestion! I will implement this next week if Bruno agrees. But most people use ctrl+s I fear not the menu

Sent from Proton Mail Android

-------- Original Message --------

boutinb:
> That's a good suggestion! I will implement this next week if Bruno agrees. But most people use ctrl+s I fear not the menu 

I don't see exactly what you mean: there is now already 2 submenus (Computer and OSF). Do you want to add an extra submenu 'Computer with encryption' (I don't think that encryption can be added for OSF) ? Better is to add a checkbox 'Use encryption' in the Computer submenu. That would set de JASP encryption fiter per default in the dialog.
But as you said, this won't work with the CTRL-S shortcut. So I think to set in the title ''Save (encrypted) JASP file" is still needed.

shun2wang:
Ctrl+S shortcut is works for a saved file (whatever it is Encrypted or not) but should not a "save as...", in another case if user pressed that key from a created new file then it should just popup a save dialog ask save path. for example in Excel if you created a file from "mew file" button when you use shotcut to save it, application will ask you where to save it.

JorisGoosen:
> Great feature!! I have updated a bit the encryption dialog box (by using a Group, you get automatically alignment of the TextFields). I found the feature a bit hidden when you save a JASP. It is (nearly) impossible to add a checkbox to a native dialog (default dialogs are native), so I understand that you chose to add a filter (even if it has the same jasp suffix), but I'm afraid that most of the users won't see this option. That's why I have changed the title of the dialog by 'Save (encrypted) JASP file". But maybe someone has a better idea to increase the attention of the user for this feature.

We could just not use the native dialog for this particular dialog.

Is it then easier to add a checkbox perhaps?

JorisGoosen:
Works pretty smooth, however if you cancel the save dialog before entering the password and all the progressbar gets stuck. So I'll fix that.

JorisGoosen:
Alright, if looked through the code, it seems fine to me.
Ive made sure the progressbar doesnt get stuck if you fail to submit a password.

If @boutinb also thinks this is fine we could perhaps merge this?
I didnt test on windows or linux yet though.

QMLComponents/utilities/desktopcommunicator.cpp
```diff
@@ -34,4 +38,26 @@ bool DesktopCommunicator::engineSandbox()
...
+{
+#ifdef BUILDING_JASP
+	if(QThread::currentThread() == qApp->thread()) {
+		int x = 1;
+
+	}
+
+	queryCondition = false;
+	std::unique_lock<std::mutex> lock(queryLock);
+	emit queryEncryptionSettingsSignal();
```
JorisGoosen: This all couldve just been done with `Qt::BlockingQueuedConnection` but I cant be arsed to rewrite it for that now.

JorisGoosen (approved):

boutinb (approved):
I think we can merge this now, @JorisGoosen ?

PR #6030, opened 2025-09-24
Make database connection thread-safe

boutinb:
When working on the new feature to generate reports with feeding jasp files with other datafiles (https://github.com/jasp-stats/jasp-desktop/pull/6025), I came upon an issue that sometimes the synchronization hangs. 
This was due to the fact that after opening a JASP file, the Filter model generates some SQL statements, but the synchronization task was already started: this lead to concurrency issues with the database. Especially the connection object to SQLite is thread dependent, but `_transactionWriteDepth` and `_transactionReadDepth` are not. So one thread could increment _transactionWriteDepth, and another thread checking _transactionWriteDepth would think that a transaction was already started, and then would not execute a 'BEGIN EXCLUSIVE' statement.
So the solution is to make `_transactionWriteDepth` and `_transactionReadDepth` thread dependent.

Also the connection object is stored in a map, but a std map does not use mutex: so setting and querying the map must be done under a mutex to be sure that the manipulation of the map is thread safe.

This fix solves probably some hanging issues that some users report.

### DIFF
--- a/CommonData/databaseinterface.cpp
+++ b/CommonData/databaseinterface.cpp
@@ -12,6 +12,10 @@
 
 DatabaseInterface * DatabaseInterface::_singleton	= nullptr;
 
+thread_local 	int			_transactionWriteDepth	= 0,
+							_transactionReadDepth	= 0;
+
+
 //#define SIR_LOG_A_LOT
 
 const std::string DatabaseInterface::_dbConstructionSql =
@@ -2050,10 +2054,7 @@ sqlite3 * DatabaseInterface::_db()
 	if(_dbCreated && _dbCreator == id)
 		return _dbCreated;
 
-	if(!_dbs.count(id))
-		load();
-
-	return _dbs.at(id);
+	return load();
 }
 
 void DatabaseInterface::create()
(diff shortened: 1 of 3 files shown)


CommonData/databaseinterface.cpp
```diff
@@ -2132,40 +2133,41 @@ void DatabaseInterface::preloadInterfaceForThread()
...
-	assert(!_dbCreated || std::this_thread::get_id() != _dbCreator);
+    assert(!_dbCreated || std::this_thread::get_id() != _dbCreator);
 
     _loadMutex.lock();
     if(_dbs.count(std::this_thread::get_id()))
     {
+        sqlite3* connection = _dbs.at(std::this_thread::get_id());
         _loadMutex.unlock();
-        return;
+        return connection;
```
JorisGoosen: Why are you returning this?

CommonData/databaseinterface.h
```diff
@@ -197,13 +197,10 @@ class DatabaseInterface
 	void		_runStatementsRepeatedly(	const std::string & statements, std::function<bool(	std::function<void(sqlite3_stmt *stmt)> **	bindParameters, size_t row)> bindParameterFactory, std::function<void(size_t row, size_t repetition, sqlite3_stmt *stmt)> * processRow = nullptr, bool ignoreFails = false);
 
 	void		create();					///< Creates a new sqlite database in sessiondir and loads it
-	void		load();						///< Loads a sqlite database from sessiondir (after loading a jaspfile)
+	sqlite3*	load();						///< Loads a sqlite database from sessiondir (after loading a jaspfile)
```
JorisGoosen: I on purpose kept sqlite3 stuff out of the api because this is supposed to abstract away from that so we could change sqlite for some other sql some day.

Why did you add it?

boutinb: That's true, this is not necessary anymore (this is just a relic of old attempts). I'll change this

JorisGoosen: Ill do so, or not

JorisGoosen (approved):

CommonData/databaseinterface.cpp
```diff
@@ -2054,7 +2054,10 @@ sqlite3 * DatabaseInterface::_db()
 	if(_dbCreated && _dbCreator == id)
 		return _dbCreated;
 
-	return load();
+	if(!_dbs.count(id))
+		load();
+
+	return _dbs.at(id);
```
boutinb, line 2060: @RensDofferhoff and me thought also reading in a map is not really thread-safe: a map has no mutex, so if a thread add an element to the map at the same moment that count or at is called, it cannot guarantee the right value.
But as for the moment only one extra thread (the _loaderThread from MainWindow) is used, the map will have in fact only 1 value (the other connection is set in _dbCreated. So that's ok for now...

JorisGoosen: Which is why I put mutexes in the load and creation function...

JorisGoosen: And there are waaaaay more threads being used so im not sure why you think there is only one?

JorisGoosen: `class InitColumnTask : public QRunnable` look for that

RensDofferhoff: It doesn't matter you only guard against concurrent writes into the map with that mutex.
Reads can happen during load and creation.
count() and at() of std::map are not safe when the map is not const.

boutinb: But then reading in the map should also be enclosed by the mutex.

JorisGoosen: So we still have that potential problem then you mean?

RensDofferhoff: > But then reading in the map should also be enclosed by the mutex.

Yes.

JorisGoosen: well Ill add it to the PR im opening in a few minutes then!

PR #6031, opened 2025-09-24
Interactive graphs with plotly

### LINKED ISSUE #4077: [Feature Request]: Choice for plotly::ggplotly
### Description

Built interactive plots using plotly if the relevant button is clicked

### Purpose

Interactive data visualization would allow for easier data interrogation  

### Use-case

_No response_

_No response_

Mostly to Plot Builder but it's relevant for any plotting menu

A button that would provide the plotly version (maybe through ggplotly?) of the developed plot 

_No response_

### Additional context

_No response_

vandenman:
Requires https://github.com/jasp-stats/jaspBase/pull/178 and https://github.com/jasp-stats/jaspGraphs/pull/125

Probably best to discuss some of the implementation details in person or via video call or so, but in a nutshell:

- if possible, show plots as interactive plots by default (for now).
- switching between the static and interactive version of a plot is possible through the dropdown of a plot.
- resizing sort of works.

There are quite some rough edges though:

- For some plots the static version looks fine but the interactive version could use some tweaks (For example, more space between axis title and axis tick labels).
- The updating of the axes needs to be polished. Right now it can cause flickering when there are many plots.
- Passing the data now happens through the json directly (as it is json). However, the objects can be very large and make the logs hard if not impossible to read. Perhaps this should be saved as a separate object on disk and loaded in by js to keep the json messages from the desktop/ engine readable.
- I did not look at html/pdf export. The pdf should just use the static plots. But the html could, in theory, use the standalone feature from plotly. That way most interactivity is retained, although the file size can become quite large.

Nevertheless, I think this draft PR is far enough for:

1. An initial review.
2. A starting point for changes to the R analyses related to interactive figures. For example, the analyses in R need to decide what to show in the tooltip that is shown on hover (right now this is just whatever plotly does by default) and also need to check how the variables are named (right now this uses the internal aestethic names, which is what plotly does by default). 

Known issues:

- ggplot2 objects with polar coordinates do not work, see https://github.com/plotly/plotly.R/issues/878. This is a problem, because any pie chart (e.g., those in descriptives, but also the BF pizza) is currently implemented in polar coordinates. It's unclear what the best way is to fix this. plotly does support both polar coordinates, and has a separate 'pie' trace type. Both can do exactly what we need, but there is no automatic conversion from ggplot. It's not easy to see from the ggplot object that when we are creating a pie chart, because the ggplot object just contains a bar graph shown in polar coordinates. Possible solutions:
  - Fix the polar coordinate conversion.
  - Inside jaspGraphs, add special plotly code to everything that returns a pie chart, so that instead of calling ggplotly we use the custom plotly code (hacky).


shun2wang:
Great, this is what I have always wanted to do!👍
A suggestion: It seems that we can streamline some js dependencies. For example, for the jquery library, we can only keep *.min.js instead of .js and .map in the production environment, and only keep those redundant things in the development environment.

JorisGoosen:
Yes and maybe we can also depend on their respective source repositories as submodules or something, so we can stay up to date easier and with more easily traceable origins

shun2wang:
> Yes and maybe we can also depend on their respective source repositories as submodules or something, so we can stay up to date easier and with more easily traceable origins

Or simply copy them form somewhere while building.

vandenman:
>  It seems that we can streamline some js dependencies.

Sounds like a good idea! For what it's worth, I currently just copied these from a plotly not-standalone export. Basically, in R, export a figure as a not-standalone html which produces an html file and separate includes. The includes I copied over in this PR. Doing this generically is probably possible, although I'm not entirely sure how this will work when in the future people an install a newer or older version of a module, which uses a newer or older version of plotly.

Desktop/html/js/image.js
```diff
@@ -123,17 +180,166 @@ JASPWidgets.imagePrimitive = JASPWidgets.View.extend({
...
+			Plotly.newPlot(targetEl, payload.data, payload.layout)
+				.then(() => {
+					console.log("Plotly chart rendered successfully");
+					// Mark the element as having a valid Plotly chart
+					targetEl._plotlyInitialized = true;
+				})
+				.catch((err) => {
+					console.error("Plotly rendering failed:", err);
+					targetEl._plotlyInitialized = false;
+				});
```
vandenman, line 333: @JorisGoosen you need to adjust this part if you want to pass the plot differently.

JorisGoosen:
Ok, ive changed the code to use files as the basis of the plots so the results-json is nice and clean.
I couldnt find out how to switch back to normal plots though?

vandenman:
I'll take a look tomorrow, should be some boolean somewhere that's default should be flipped.

JorisGoosen:
I found the way to switch and everything seems to work.

JorisGoosen:
If I export to pdf it seems the interactive plot is exported, is that what we want?
In at least jaspTTests things look a little bit different.

Ill add a basic preference to show/hide interactive plots by default

JorisGoosen:
I love how smooth resizing the interactive plots feel btw ❤️ 
dit: resizing the normal plots behind them seems a bit broken though

shun2wang (commented):
I recommend we could just pack https://github.com/plotly/plotly.R/tree/master/inst/htmlwidgets to a htmlwidgets.zip file. and unzip it to cmake build dir and pack it into .qrc while building JASP. To make build clean and simple. Jquery is not necessary cause we already have it as library.

vandenman:
That sounds reasonable, but the plotly version used by R and the js need to be kept in sync somehow, so perhaps not just plotly@master but rather a specific tag or commit. I do remember running into issues with jquery somehow, either they or we (probably we) used an older version which caused some issue. Keeping this working while allowing for different versions of `plotly` (say when installing an older version of a module) is also something I haven't figured out yet.

shun2wang:
> That sounds reasonable, but the plotly version used by R and the js need to be kept in sync somehow, so perhaps not just plotly@master but rather a specific tag or commit. I do remember running into issues with jquery somehow, either they or we (probably we) used an older version which caused some issue. Keeping this working while allowing for different versions of `plotly` (say when installing an older version of a module) is also something I haven't figured out yet.


yes, I had tried that but cannot push to this pr then i made that comment above,it was because we already have jquery even that htmlwidgets doesn't include jquery, so plotly.r should never need it. 

some of files were not be used for example `plotly0/balabala`

vandenman:
@JorisGoosen with https://github.com/jasp-stats/jaspBase/pull/190 this should now also work when images are edited. It's pretty slow though because it only works after the ggplot images are rerendered. Technically that could be done async but right now this is all done serially. 

Also we (probably I) should indeed clean up all the unnecessary changes in this PR.

JorisGoosen:
image
the interactive plot doesnt get updated though?
The ggplot one is updated:
image

JorisGoosen:
Or is there some changes to jaspBase/jaspGraphs that im missing?

Edit: That was it, it works splendidly!

JorisGoosen (approved):

PR #6033, opened 2025-09-26
Allow read sqlite database of more than 2GB

### LINKED ISSUE #3725: [Crash]: Stopped running tests and  now every time I trie to open the file JASP crashes
### System Info

Hello, I was running tests (if recall well maybe that was T Tests or ANOVA) and at some point no matter how I moved variables around the analysis just wasn't happening and the report window said something like 'waiting to initiate the analysis'. After trying to close the test and run it again a few times, I decided just to save the last version (I also have saved the previous version), then close it and open it again but JASP started crashing whenever I tried to do it (tried to open the file directly, open it through a blank/new JASP, open it directly while having the blank JASP opened…). Then restarted my laptop but the problem is the same. It started crashing with 0.19.3 (Intel) but not improved after updating to 0.95.2 (Apple Silicon).
Would very much appreciate your help!

### JASP Module

T-Tests

### Is there a specific analysis that caused the crash?

if I remember correctly it was T Test (or ANOVA maybe)

### Log files

[JASP 2025-09-23 14_36_38 Desktop.log](https://github.com/user-attachments/files/22496291/JASP.2025-09-23.14_36_38.Desktop.log)
[JASP 2025-09-23 14_36_38 Engine 0.log](https://github.com/user-attachments/files/22496295/JASP.2025-09-23.14_36_38.Engine.0.log)
[JASP 2025-09-23 15_10_41 Desktop.log](https://github.com/user-attachments/files/22496292/JASP.2025-09-23.15_10_41.Desktop.log)
[JASP 2025-09-23 15_16_51 Desktop.log](https://github.com/user-attachments/files/22496293/JASP.2025-09-23.15_16_51.Desktop.log)
[JASP 2025-09-23 15_19_27 Desktop.log](https://github.com/user-attachments/files/22496294/JASP.2025-09-23.15_19_27.Desktop.log)

### Steps to Reproduce

Open JASP
Go to Open Recent Files
It starts loading and then crashes.

Same happens if I simply try to open the file from a Finder.

boutinb:
When reading the JASP file, it reads the number of bytes with an 'int' type, which means that if a file is bigger than 2GB, it fails. But the archive library we use, uses the 'long' type, and can handle much larger file.

Fixes https://github.com/jasp-stats/jasp-issues/issues/3725

### DIFF
--- a/CommonData/archivereader.cpp
+++ b/CommonData/archivereader.cpp
@@ -113,7 +113,7 @@ void ArchiveReader::writeEntryToTempFiles(std::function<void(float)> progressCal
     file.rdbuf()->pubsetbuf(streamBuff, sizeof(streamBuff)); //Set the buffer manually to make it much faster our issue https://github.com/jasp-stats/INTERNAL-jasp/issues/436 and solution from:  https://stackoverflow.com/a/15177770
 
     static char copyBuff[8192 * 4];
-    long		bytes		= 0;
+    int64_t		bytes		= 0;
     int         errorCode	= 0;
     float       tallyBytes  = 0;
     do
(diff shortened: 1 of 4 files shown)


JorisGoosen:
Maybe its better to put `uint64_t` ?

shun2wang:
Maybe `int64_t` is compatibility and better? Otherwise there may be differences on different platforms(msvc and gcc), see https://github.com/jasp-stats/jaspColumnEncoder/pull/11

JorisGoosen:
Filesize cant be negative can it?

So uint64_t

But a great idea to do this everywhere yeah...

shun2wang:
>But a great idea to do this everywhere yeah...

I think so.

JorisGoosen:
https://en.cppreference.com/w/cpp/filesystem/file_size.html

They use `uintmax_t` from stdint.h so maybe we should use that in both places

JorisGoosen:
also msvc uses a dword which is unsigned too.

JorisGoosen (approved):

boutinb:
I use a int64_t because the archive_read_data returns a signed long type (on mac at least)
A long type can go until 9,223,372,036,854,775,807, which  should be enough for the next years...

PR #6038, opened 2025-09-30
feat: drag to re-order parts of an analysis in results

### LINKED ISSUE #636: Ability to move and re-order parts of an analysis...
* Enhancement: Ability to re-order elements of an analysis
* Purpose: To be able to put the elements of the analysis e.g., descriptives, post-hocs, plots in a desired order
* Use-case: Setting up templates with the components of the analyses in the correct/desired order e.g., descriptives first.

Only that it is cumbersome to have to re-order the elements afterwards

Simple drag & drop in the settings or output window.

shun2wang:
Close: https://github.com/jasp-stats/INTERNAL-jasp/issues/2983
Close: https://github.com/jasp-stats/jasp-issues/issues/636

### DIFF
--- a/Desktop/html/css/jasp.css
+++ b/Desktop/html/css/jasp.css
@@ -68,8 +68,30 @@ h6, .h6-toolbar .jasp-menu {
 }
 
 .jasp-display-item-flat {
-	padding-left: 0;
+	padding-left: 0.6em;
 }
+
+.jasp-analysis > .jasp-display-item {
+	position: relative;
+}
+
+.jasp-analysis > .jasp-display-item:hover::before {
+	content: "";
+	position: absolute;
+	left: -7px;
+	top: 50%;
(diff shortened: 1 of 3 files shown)


shun2wang:
@tomtomme This is a feature not a bugfix so it would not in next hotfix release. thanks for your track.

boutinb:
This is quite interesting, but each time some options are changed, the order of the container is set back to the original one. Also the JASP file self does not store the new order, so when opening a JASP file, the order is lost. 
The order should be kept in the analyses.json, maybe not in  the `results`, since it is generated each time an option is changed, so probably a new entry is needed, like `userEdit`. Maybe the plot editing setting should be set here also...

JorisGoosen:
or the "userdata" where we also store the note stuff can also store a list with the changed order?

PR #6045, opened 2025-10-13
Options not set when loading analysis with TabViews

boutinb:
When testing the TabView test analysis in the jaspTestModule, when duplicating the analysis, the TabViews did not get the right values, and sometimes it even crashed.
The problem was that the controls in a Row Control are setup sometimes several times. The controls in the Row Control should be set up only once: if the Row Control is reused, only the values (and the context properties) should be reset.


JorisGoosen (approved):
Duplicating the test tabview analysis now seems to work perfectly

PR #6048, opened 2025-10-22
make sure to load the relevant properties from DESCRIPTION instead of Description.qml

JorisGoosen:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/1251

Will still need to update all the modules to avoid having ridiculous names like "A blabla module for JASP" as titles...
So setting it as draft until then.

### DIFF
--- a/Desktop/modules/description/description.h
+++ b/Desktop/modules/description/description.h
@@ -24,15 +24,15 @@ class AnalysisEntry;
 class Description : public QQuickItem
 {
 	Q_OBJECT
-	Q_PROPERTY(QString					name			READ name				WRITE setName				NOTIFY nameChanged				)
-	Q_PROPERTY(QString					title			READ title				WRITE setTitle				NOTIFY titleChanged				)
+	Q_PROPERTY(QString					name			READ name				WRITE setDummy						NOTIFY nameChanged				)
+	Q_PROPERTY(QString					title			READ title				WRITE setDummy				NOTIFY titleChanged				)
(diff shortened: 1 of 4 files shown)


Desktop/modules/description/description.h
```diff
@@ -112,8 +115,8 @@ public slots:
 	QUrl					_website;
 	Version					_version;
 	bool					_requiresDataDef	= true,
-							_hasWrappers		= false,
-							_preloadData		= false;
+							_hasWrappers		= true,
+							_preloadData		= true;
```
boutinb, line 119: Why did you change the default settings for hasWrappers and preloadData? Will you change these settings afterwards to all modules that did not set these properties explicitly?
Also I don't think that the default for hasWrappers should be true: if hasWrapper is true, JASP calls the R function with "Internal" as suffix. When making a new module, this will be confusing.
Maybe we should be more clever, and detect whether a wrapper exists by checking whether the Internal function of the analysis exists.

JorisGoosen: Yeah you are right about the hasWrappers

I have already (locally) changed all the modules though.
But I will simply add hasWrappers everywhere then.

It is an annoying property though. Because you are right, it really should be clear from the package.
Hmm

They should be exported in NAMESPACE right?
That should make it pretty easy to detect actually.

JorisGoosen: But I wont do that now as it would require quite a bit more testing than this pr needs.

JorisGoosen:
And I assume that for preloadData its pretty obvious why I changed the default.
Because we need all the modules to go to that setting for syntaxmode anyway.

PR #6049, opened 2025-10-22
Use sourceID and name in R Sources in the right way

boutinb:
During the development of the "Moderated Non-LInear Factor" analysis of the SEM module, @juliuspfadt came accross a strange bug: this analysis uses a R source, and it gives another value for the name of the R source and the sourceID.
The handling of R source could not cope with this. What happens is that in the meta part of the result, only the name of the R source was used. and only in the value part of the result the sourceID is given.
The sources are kept in the analysis with the _rSources property which is a map from the sourceID to its json value (the sourceID is what is used in QML to specify a R source).
The value of the R sources is given in the result, only if it is new or has another value.
But to detect whether a R source still exists, it searches in the _rSources the ones that do not exist anymore in the meta. As the _rSources uses the sourceID and the meta uses the source name, if they are not the same, the R source was then removed, and the controls depending of this R source in the QML form was removed.

In https://github.com/jasp-stats/jaspBase/pull/179, the jaspBase is changed so that the user does not have to specify a sourceID to define a R source (this is quite confusing), it takes per default the name of the source. Moreover the meta gets also the sourceID as property.
In this PR, the confusion in the code between sourceID and source name is fixed.

The "Test R source" analysis of jaspTestModule will be also changed to test different possibilities. But as jaspTestModule depends on jaspBase, the renv.lock must be changed to load the new version of jaspBase: so the changes in jaspBase must be first merged, before new tests on R sources can be added in jaspTestModule.

### DIFF
--- a/Desktop/analysis/analysis.cpp
+++ b/Desktop/analysis/analysis.cpp
@@ -1097,7 +1097,8 @@ void Analysis::checkForRSources()
 	}
 
 	//First check meta for qmlSources and collections
-	std::set<std::string> sourceIDs, isCollection;
+	QMap<std::string, std::string> sourceNamesMap; // Map source name to source ID
+	std::set<std::string> isCollection;
 
 	std::function<void(Json::Value & meta)> findNewSource = [&](Json::Value & meta) -> void
 	{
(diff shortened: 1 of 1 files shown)


boutinb:
A test is now added in R Source test analysis of the jaspTestModule: https://github.com/jasp-stats/jaspTestModule/pull/73

JorisGoosen:
This can be tested and reviewed right?

And it does not require any changes to any modules righ?

JorisGoosen:
(The code that is, I guess it would need to use the new jaspBase alright)

boutinb:
I've made the changes so that the Cochrane module works with the old and new jaspBase. The new SEM module of julius https://github.com/juliuspfadt/jaspSem/tree/newAnalysisMnlfa works now also with the old jaspBase because he has renamed the sourceID to be equal to the source name.
The jaspTestModule is already linked with the new jaspBase (if you load it as Development module since its bundle is not yet updated), and the R Source test analysis should prove now that the R Source can have another sourceID than the source name, and that if no sourceID is given, then the source name is used.
Afterwards, we can update all jaspBase in all modules (or will you do that automatically when you'll make a new 0.95.5 beta version?)

JorisGoosen:
I wasnt planning on updating the lockfiles, thats up to the module developers.

JorisGoosen (approved):
Im not 100% what was going on here or what is supposed to be going on but it seems to work...

A unittest for this would be nice too I think

PR #6051, opened 2025-10-24
Crash with autosave with analysis without data

boutinb:
If you start JASP, add an analysis without data, close the workspace, and add an analysis again without data, wait until the autosave starts: it crashes.

### DIFF
--- a/Desktop/data/datasetpackage.cpp
+++ b/Desktop/data/datasetpackage.cpp
@@ -46,12 +46,12 @@ DataSetPackage::DataSetPackage(QObject * parent) : QAbstractItemModel(parent)
 	if(_singleton) throw std::runtime_error("DataSetPackage can be constructed only once!");
 	_singleton = this;
 	//True init is done in setEngineSync!
-	
+
 	_db			= new DatabaseInterface(true);
 
 	_dataSet	= new DataSet(); //We create one here to make sure filter() etc can actually work
 	setDefaultWorkspaceEmptyValues();
-	
+
 	connect(this, &DataSetPackage::isModifiedChanged,					this, &DataSetPackage::windowTitleChanged);
 	connect(this, &DataSetPackage::loadedChanged,						this, &DataSetPackage::windowTitleChanged);
 	connect(this, &DataSetPackage::currentFileChanged,					this, &DataSetPackage::windowTitleChanged);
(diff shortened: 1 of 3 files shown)


Desktop/data/datasetpackage.cpp
```diff
@@ -2530,7 +2530,7 @@ void DataSetPackage::setAnalysesData(const Json::Value &analysesData)
 {
 	QString		previousASF					= analysesData.type() != Json::objectValue ? "" : tq(analysesData.get("autoSaveFileName", "").asString());
 				_analysesData				= analysesData;
-	QFileInfo	dataFile					( tq(_dataSet->dataFilePath()) ),
+	QFileInfo	dataFile					( tq(_dataSet ? _dataSet->dataFilePath() : "") ),
```
JorisGoosen: Shouldnt there always be an empty DataSet anyway?

JorisGoosen: Like, isnt there something else going wrong?
Does it not crash when you do a normal save?

boutinb: When JASP starts, the datasetpackage always make a DataSet object. If a (JASP) file is loaded, the current DataSet is deleted and a new one is made. If the workspace is closed, the DataSet is deleted. If afterwards an analysis is added that does not need a dataset, no DataSet object is created. 
Apparently, just in the case that a workspace is closed, and an analysis without dataset is added, then the DataSet object is null (and might be used). Maybe a DataSet object should be created in this case. Not sure about the consequences... So I thought , just check that the DataSet is not null (like in many other places in the datasetpackage code).
But maybe a better fix would be to create a DataSet object after the close workspace action: there should be always a DataSet object even if there is no dataset.

boutinb: And you are right: if you close the workspace, add a an analysis without data, and save manually, it crashes: no need to wait for the autosave.

JorisGoosen: > But maybe a better fix would be to create a DataSet object after the close workspace action: there should be always a DataSet object even if there is no dataset.

This is what I think the fix is

JorisGoosen: Because otherwise it would have crashed after saving a workspace without data the first time

boutinb: OK, then I change the code so that DataSetPackage always has a proper DataSet object

Desktop/data/datasetpackage.cpp
```diff
@@ -143,11 +143,11 @@ void DataSetPackage::reset(bool newDataSet)
 	Log::log() << "DataSetPackage::reset()" << std::endl;
 	_databaseIntervalSyncher.stop();
 	_delayedRefreshTimer.stop();
-	
+
 	beginLoadingData();
 
 	if(newDataSet)	createDataSet();
```
JorisGoosen, line 149: This is `if` is now a bit stupid isnt it?

JorisGoosen:
Im not sure why there need to be so many changes, the whitespace is now prettier sure, but it also is a lot of changes hiding the few actual changes.

The rename of deleteDataSet to resetDataSet is a bit weird, especially because it now also does createDataSet, despite that function still existing.

Instead of all of these changes I really only expected there to be a single `createDataSet` call somewhere...
I guess I will have a look and figure out which spot that is.

JorisGoosen:
So I made an alternative PR that I think does a bit less useless changing the code

PR #6053, opened 2025-10-28
Fix label load old jasp file

### LINKED ISSUE #3710: [Bug]: Nominal variable...loses values
### JASP Version

0.95.2

### Commit ID

_No response_

### JASP Module

Unrelated

Binomial Test

Windows 11

### Bug Description

I have toy dataset where I have nominal variable with 280 observations. The variable is binary. So I set it to be 1 or 0. I was using it to teach the binomial test. An initial run of the binomial test worked. But, then if I go change an attribute (e.g., Alt. Hypothesis), JASP seeminly loses one of the values so in the test it only lists 1's or 0s. The data is still the same and appears preserved. But, I lose one of the values. In this case the 0. I hav uploaded a JASP file that I saved for the class 2 years ago where it worked fine...but now I get the issue. See saved second file. 
I see it on Mac Intel and Windows 11. 
[Archive.zip](https://github.com/user-attachments/files/22414507/Archive.zip)

### Expected Behaviour

Values should not be dropped. 

### Steps to Reproduce

Take the original file and change an attribute (e.g., 2 sided test). 

### Log (if any)

_No response_

### More Debug Information

_No response_

### Final Checklist

JorisGoosen:
Several tweaks to JASP data loading
- Missing values should be hierarchical so workspace missing values should be used in addition to the ones from a column. See the stata test file, it doesnt contain specific `nan` missing value specifications for the columns, but it really ought to be a missing value as it is so defined anyway in the workspace. This change also allows one to disable and reenable custom missing values without losing them.
- Make sure labels imported from an old (<0.18) jasp file are sorted on load
- added a testcase for loading 0.16.4 and 0.18.3 jasp file (0.17.3 app didnt work... because apple said no)
- Small change to how titles are set to make loading an old jaspfile more like making a new one.
Fixes https://github.com/jasp-stats/jasp-issues/issues/3710
- a test has been added for verifying that titles adhere to namechanges correctly
- a test has been added for verifying that empty values are adhered to, both as workspace and non-workspace

Loading debug jasp file from 0.18.3 has some inconsistencies now with loading both older and newer files where:
- autoSort is disabled
- thus the labels are sorted differently
- the title is set to name regardless of the change made for the older files.


boutinb:
If I load the hypoTherapic jasp file from https://github.com/jasp-stats/jasp-issues/issues/3710, I get all labels, but if I refresh the analyses, I get an error: 
Screenshot 2025-10-29 at 12 57 27

JorisGoosen:
Ill have a look!

JorisGoosen:
(I guess the numeric things arent reset properly either? Sounds like a good testcase to add in any case)

JorisGoosen:
So labels werent being stored, and then when they werent they were getting bogus intsId (starting at -1)

Now it should work

boutinb (approved):

PR #6054, opened 2025-10-30
Save changed TextField when adding a new row

### LINKED ISSUE #3812: [Bug]: Hypothesis naming undo automatically when adding additional hypothesis in Binomial testing
### JASP Version

0.95.4

### Commit ID

_No response_

### JASP Module

Learn Bayes

Binomial Testing

Windows 11

### Bug Description

When I rename the hypothesis, the renaming is lost if I have not pressed enter. This happens if I am still in the type-in window, type the hypothesis name, then directly click to add a new hypothesis.

### Expected Behaviour

Added information should stay in there and not be lost

### Steps to Reproduce

1. go to module,
2. add hypothesis using green plus
3. rename hypothesis but keep cursor in the renaming window
4. click green button to add hypothesis again

### Log (if any)

_No response_

### More Debug Information

_No response_

### Final Checklist

boutinb:
Fixes https://github.com/jasp-stats/jasp-issues/issues/3812

In a ComponentsList with TextField's, if you change a value in a TextField and click 'Add row', then your change in the TextField is lost.
When a TextField loses the focus, the changes should be saved. But when adding a new row this action takes place first, and this sets the Row controls for each row (term) of the components list. Of course it reuses the existing Row controls, but it resets the values of each control. As the internal value of the TextField was not changed yet, it uses the old value and resets the display value of the TextField, which discards the changed value.

But when adding a new row, it should not set the Row controls of the previous rows: this fixes this issue.

### DIFF
--- a/QMLComponents/models/listmodel.cpp
+++ b/QMLComponents/models/listmodel.cpp
@@ -190,7 +190,7 @@ void ListModel::setRowComponent(QQmlComponent* rowComponent)
 	_rowComponent = rowComponent;
 }
 
-void ListModel::setUpRowControls()
+void ListModel::setUpRowControls(int startRow)
 {
 	if (_rowComponent == nullptr)
 		return;
@@ -198,14 +198,17 @@ void ListModel::setUpRowControls()
 	int row = 0;
 	for (const Term& term : terms())
 	{
-		if (!_rowControlsMap.contains(term.value()))
+		if (startRow <= row)
 		{
-			RowControls* rowControls = new RowControls(this, _rowComponent);
-			_rowControlsMap[term.value()] = rowControls;
-			rowControls->initValues(row, term, _rowControlsValues[term.value()]);
+			if (!_rowControlsMap.contains(term.value()))
+			{
+				RowControls* rowControls = new RowControls(this, _rowComponent);
+				_rowControlsMap[term.value()] = rowControls;
+				rowControls->initValues(row, term, _rowControlsValues[term.value()]);
+			}
+			else
+				_rowControlsMap[term.value()]->resetValues(row, term, _rowControlsValues[term.value()]);
 		}
-		else
-			_rowControlsMap[term.value()]->resetValues(row, term, _rowControlsValues[term.value()]);
 
 		row++;
 	}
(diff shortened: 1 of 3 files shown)


JorisGoosen:
Or maybe you should store the value in a property of JASPControl/BoundControl instead of in a qml property ?

JorisGoosen:
Or just make the "add row" button forceActiveFocus first?

boutinb:
> Or just make the "add row" button forceActiveFocus first?

I'm not sure if that would work, but that would be anyway quite hazardous, since it would depend still on how Qt implements this forceActiveFocus (would the add row signal be fired first, or the focus out signal?). 
I find my solution quite simple, and it implements something that should have be done anyway: the rows should not be reset when adding a new one.

JorisGoosen:
> I find my solution quite simple

Maybe, but Im having trouble figuring out how your rowcontrols actually work now... 
The implementation seems to keep changing all the time?

Like, I have no idea right now how exactly they are getting created and updated...

boutinb:
> Like, I have no idea right now how exactly they are getting created and updated...

Each time the terms of a model are updated the 'setUpRowControls' is called: this function takes care that for each term of the model gets a rowControls object (if a rowComponent is given). In this function, the rowControls object is created or reused (it's kept in the _rowControlsMap) for each term. The only thing I did now, is to tell that if a term is just added (at the end of the terms list), then do not reset the values of the rowControls of all previous terms.

JorisGoosen:
That does sound like a simple solution

QMLComponents/models/listmodel.cpp
```diff
@@ -190,22 +190,25 @@ void ListModel::setRowComponent(QQmlComponent* rowComponent)
...
+void ListModel::setUpRowControls(int startRow)
 {
 	if (_rowComponent == nullptr)
 		return;
 
 	int row = 0;
 	for (const Term& term : terms())
 	{
-		if (!_rowControlsMap.contains(term.value()))
+		if (startRow <= row)
```
JorisGoosen: They cant be updated in another order?

JorisGoosen: Like, if someone removes a term, then the rows above all move one spot down no?
Then they would get reset again instead of not?
Or is there no way to do that?

boutinb: OK, good point, I have added also the case when a row is removed.

JorisGoosen: And can they be reordered?

boutinb: The rows might have different kinds of manipulations: in many cases, the model will get a resetModel signal, the rowObjects will be reset, and that won't be a problem. The edge case here is when a rowObjects contain a TextField, that this TextField is modified, and its focus out signal (that should save its change) is in concurrence with another signal that might modify the rows. There is only the Add or Remove row button that could provoke this edge case.

JorisGoosen (approved):

JorisGoosen:
Ok so your fix works fine, i will try to make a unittest for this kind of thing

PR #6057, opened 2025-10-31
Add reference lines and point interface to ploteditor

JorisGoosen:
https://support.jasp-services.com/SCG-Chemicals/SCG_Chemicals_Public_Company_Limited/issues/16

@vandenman Ill open this as a draft to have it pinned somewhere.

image

is added to ploteditor, the lowest line is a dummy line where an edit results in a new line.

It sends some extra json in the plotediting as follows:
```
"editOptions" : 
		{
			"error" : "success",
			"references" : 
			[
				{
					"horizontal" : false,
					"point" : true,
					"text" : "A very  fancy label",
					"x" : 0.0,
					"y" : 0.0
				}
			],
			"resetPlot" : false,
			"xAxis" : 
			{
				etcetera....
```

### DIFF
--- a/Desktop/components/JASP/Widgets/MainPage.qml
+++ b/Desktop/components/JASP/Widgets/MainPage.qml
@@ -535,7 +535,7 @@ Item
 						"functionCall"	: functionCall
 					};
 
-					customMenu.toggle(resultsView, props, (optionsJSON['rXright'] + 10) * preferencesModel.uiScale, optionsJSON['rY'] * preferencesModel.uiScale);
+					customMenu.toggle(resultsView, props, (optionsJSON['rXright'] + 10 ) * preferencesModel.uiScale, optionsJSON['rY'] );
 
 					customMenu.scrollOri		= resultsView.scrollPosition;
 					customMenu.menuScroll.x		= Qt.binding(function() { return -1 * (resultsView.scrollPosition.x - customMenu.scrollOri.x) / resultsView.zoomFactor; });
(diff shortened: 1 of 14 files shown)


vandenman:
Nice! This is indeed everything I needed to make the R side work:

screenshot_referencelines_1

screenshot_referencelines_2

Code is at https://github.com/jasp-stats/jaspGraphs/tree/referenceLines

Not sure what your workflow is, but you can test this by building jaspDescriptives as a devmodule with renv and then installing jaspGraphs in that renv project with `renv::install("path_to_jaspGraphs_checkout")` or `renv::install("jasp-stats/jaspGraphs@referenceLines")`.

We can definitely still improve some stuff, but this is already a good start. Perhaps we should have a brief chat to determine what is in scope of this feature and what is not, because it's very easy for me to come up with 100+ tweaks to this (e.g., adding options for label placement, nudging, point color/ shape, etc.).

vandenman:
@JorisGoosen anything else we (I) need to do to get this over the finish line?

JorisGoosen:
This looks quite good!

We could maybe make a testversion for SGC first to see if they have some feedback on it?
I think for now we can just keep it like this, then when it is in usage people can suggest things based on the way they use it?

vandenman:
Is the GUI ready for that? I remember having to press enter in some situation to "finish" the component.

JorisGoosen (commented):
So the interface is not as smooth as possible, but I can work on that later.

PR #6065, opened 2025-11-11
Add a "computed" argument to creating columns from R

JorisGoosen:
Fix for jasp-stats/INTERNAL-jasp#3035
Now they are created with the computed column setting "analysis" which means they get removed when the analysis is removed.

This adds support for also creating computedcolumns straight from R
Requires https://github.com/jasp-stats/jaspBase/pull/180

There isnt really a good testcase right now, and it is such a hassle to get a jaspBase in there that I suggest the reviewer (@boutinb ) just changes the `computed=false` in jaspRcpp to `computed=true` to verify that the bug as reported by @koenderks is fixed.

Then we can merge both the jaspBase and jasp-desktop changes and then @koenderks can add the `computed=TRUE` flag to his computedcolumn creation code and then the issue should be solved.

Ive set this up in such a way that the original behaviour is kept even with an old jaspBase (cause it doesnt pass ` computed` top jaspRcpp it passes `-1` so by checking the `int computed` for begin 1 we can handle both the older and newer jaspBase versions without a problem.

### DIFF
--- a/CommonData/column.cpp
+++ b/CommonData/column.cpp
@@ -258,7 +258,7 @@ void Column::setCodeType(computedColumnType codeType)
 		_constructorJson["formulas"] = Json::arrayValue;
 	}
 
-	if(_codeType == computedColumnType::analysisNotComputed)
+	if(_codeType == computedColumnType::analysisNotComputed && codeType != computedColumnType::analysis)
 		_analysisId = -1;
 
 	_codeType	= codeType;
@@ -658,7 +658,7 @@ bool Column::setDescriptions(strstrmap labelToDescriptionMap)
 }
 
 
-bool Column::overwriteDataAndType(stringvec colData, columnType colType)
+bool Column::overwriteDataAndType(stringvec colData, columnType colType, bool computed)
 {
 	JASPTIMER_SCOPE(Column::overwriteDataAndType);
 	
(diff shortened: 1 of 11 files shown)


CommonData/column.cpp
```diff
@@ -709,6 +709,8 @@ bool Column::overwriteDataAndType(stringvec colData, columnType colType)
 	// In this case, the locale used is just UTF-8/C, so the locale chosen by the user should not be used to convert the values.
 	beginBatchedLabelsDB();
 	setValues(values, labels, 0, &changes, false);
+	if(computed && _codeType == computedColumnType::analysisNotComputed)
+		setCodeType(computedColumnType::analysis);
```
boutinb, line 713: And the other way around? Can `computed` be false and _codeType be `analysis`? Then the codeType should be set to `analysisNotComputed`

boutinb (approved):

CommonData/column.cpp
JorisGoosen: I agree with you in principle, but the scenario would be:
a) a qmlcomponent (ComputedColumnField) makes a computed column
b) the R code then decides it isnt computed after all and should be not-computed... Seems a bit weird no?

I guess the computed argument here is more of a "hint", like, hey, init it as a computed.

PR #6067, opened 2025-11-13
Remove jaspColumnEncoder package

boutinb:
The jaspColumnEncoder is not needed anymore, and all its files can be set directly in jasp-desktop.

After this PR, the https://github.com/jasp-stats/jaspBase/pull/182 must be merged, then the jaspColumnEncoder can be removed


JorisGoosen:
jaspColumnEncoder cannot be removed because maybe somebody wants to build an older version of jaspBase but we can archive it.

JorisGoosen (approved):

PR #6069, opened 2025-11-18
fix decimal reading from xlsx on non en_us locale systems (like german)

### LINKED ISSUE #3867: [Bug]: some decimals differ from source file
### JASP Version

0.95.4

### Commit ID

_No response_

### JASP Module

Other

_No response_

Windows 11

### Bug Description

Dear Jasp Team,
This is the first time that I have seen Jasp change values in the data, i.e. omitting decimals of some (but not all) values from the source file data. I have downloaded the newest version and tried finding an error with the source file but it keeps happening.
In the source file, the values are 0.9, 1.0 etc. (speed in meters per second). In the Jasp data view, some random values suddenly are 900, 1000 (now the participants walk 1000 meters per second). It happens in a second row as well, and maybe in some other rows too, but I cannot check them all. 

Image
Image
Image
Image

### Expected Behaviour

Jasp should keep all original data completely unchanged. 

### Steps to Reproduce

It happens everytime as soon as I open the original csv file with Jasp. 

[Jasp file for bug report.zip](https://github.com/user-attachments/files/23434242/Jasp.file.for.bug.report.zip)

### Log (if any)

_No response_

### More Debug Information

-------- Application Info --------
JASP Version: JASP 0.95.4
Build Branch: development
Build Date: Oct 13 2025 19:29:55 (Netherlands)
Last Commit: 65193e92adb17cf4054700175c9acffa82b36b37

-------- Basic Info --------
Operating System: Windows 11 Version 25H2
Product Version: 11
Kernel Type: winnt
Kernel Version: 10.0.26200
Architecture: x86_64
Install Path: C:/Program Files/WindowsApps/45842JASP.JASP_0.95.14.0_x64__wz6q9dg667qsj
Platform Name: windows
System Local: de_AT

-------- Extra Info --------
Current code page
Aktive Codepage: 850.
Aktive Codepage: 65001.

### Final Checklist

JorisGoosen:
fixes https://github.com/jasp-stats/jasp-issues/issues/3867

To test the issue run jasp with a run-environment setting of `LC_ALL=de_DE`
This will make reading decimals fail for the xlsx from the issue.
Then using this branch it will work fine.

A thing to keep in mind: the unittest JASPTest needs to be run in an LC_ALL=en_US locale.
I guess later we can also see about adding tests in different locales.

### DIFF
--- a/Desktop/data/importers/excel/excel.cpp
+++ b/Desktop/data/importers/excel/excel.cpp
@@ -18,7 +18,7 @@
 #include "excel.h"
 #include "utilities/qutils.h"
 #include <stringutils.h>
-
+#include <columnutils.h>
 #include <QFileInfo>
 #include <QDebug>
 
@@ -95,7 +95,7 @@ void Excel::getCellValue(uint32_t &row, uint16_t &col, std::string &cellValue)
 		cellValue = std::to_string(cell.value.int_value);
 		break;
 	case FREEXL_CELL_DOUBLE:
-		cellValue = std::to_string(cell.value.double_value);
+		cellValue = ColumnUtils::doubleToStringMaxPrec(cell.value.double_value, false);
 		break;
 	case FREEXL_CELL_NULL:
 	default:
--- a/Desktop/main.cpp
+++ b/Desktop/main.cpp
@@ -510,6 +510,9 @@ int main(int argc, char *argv[])
 
 
 			QLocale::setDefault(QLocale(QLocale::English)); // make decimal points == . in at least R? Anyway, this has been here forever, ill just leave it.
+			
+			char dsteng[] = "LC_ALL=en_US";
+			putenv(dsteng);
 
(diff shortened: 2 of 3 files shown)


Desktop/data/importers/excel/excel.cpp
```diff
@@ -95,7 +95,7 @@ void Excel::getCellValue(uint32_t &row, uint16_t &col, std::string &cellValue)
 		cellValue = std::to_string(cell.value.int_value);
 		break;
 	case FREEXL_CELL_DOUBLE:
-		cellValue = std::to_string(cell.value.double_value);
+		cellValue = ColumnUtils::doubleToStringMaxPrec(cell.value.double_value, false);
```
shun2wang, line 98: Ah, indeed, `std::to_string` does lose precision.

boutinb (approved):

PR #6078, opened 2025-12-10
Check that plot title is a real string

### LINKED ISSUE #3955: JASP crashed - solved
### System Info

* JASP version: 0.95.4
* OS name and version: Windows 11 Version 24H2
* Commit used: https://github.com/jasp-stats/jasp-desktop/commit/65193e92adb17cf4054700175c9acffa82b36b37

### JASP Module

Quality Control

### Is there a specific analysis that caused the crash?

Type 4 Linearity Study

### Log files

No log files are available. To get more information, please turn logging on. For this: open the file menu (the blue hamburger button left top), navigate to Advanced Preferences and check the 'Log to file' checkbox.
Debug information: 

### Steps to Reproduce

Right clicked on Percentage process variation graph > Edit image. Can reproduce crash each time step is taken.

boutinb:
Fixes https://github.com/jasp-stats/jasp-issues/issues/3955

### DIFF
--- a/Desktop/results/ploteditoraxismodel.cpp
+++ b/Desktop/results/ploteditoraxismodel.cpp
@@ -20,7 +20,9 @@ void AxisModel::setAxisData(const Json::Value & axis)
 
 	Json::Value	settings = axis.get(	"settings",		Json::objectValue);
 
-	setTitle(		tq(						settings.get(	"title",		""	).asString()));
+	Json::Value title = settings.get("title", "");
+
+	setTitle(		tq(						title.isString() ? title.asString() : ""));
 	setTitleType(	TitleTypeFromString(	settings.get(	"titleType",	""	).asString())); //Defaults are set in *TypeFromString
 	setType(		tq(						settings.get(	"type",			""	).asString()));
 	setBreaksType(	BreaksTypeFromString(	settings.get(	"breaksType",	""	).asString()));
--- a/Desktop/results/ploteditoraxismodel.cpp
+++ b/Desktop/results/ploteditoraxismodel.cpp
@@ -22,7 +22,7 @@ void AxisModel::setAxisData(const Json::Value & axis)
 
 	Json::Value title = settings.get("title", "");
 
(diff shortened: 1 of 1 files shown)


Desktop/results/ploteditoraxismodel.cpp
```diff
@@ -20,7 +20,9 @@ void AxisModel::setAxisData(const Json::Value & axis)
 
 	Json::Value	settings = axis.get(	"settings",		Json::objectValue);
 
-	setTitle(		tq(						settings.get(	"title",		""	).asString()));
+	Json::Value title = settings.get("title", "");
+
+	setTitle(		tq(						title.isString() ? title.asString() : ""));
```
JorisGoosen:
```suggestion
	setTitle(		tq(						title.isString() ? title.asString() : title.toStyledString()));
```

JorisGoosen: Maybe nicer?

boutinb: I don't think: `toStyledString` adds curly brackets `{}`

JorisGoosen: Yes I get that but then at least you see something...

The alternative is nothing, which drops information...

boutinb: I don't understand: the title must be an empty string.

JorisGoosen: Like, the fix is to avoid titles that are not a string to not crash jasp right?
So the title could be a number?
In case it is a number, doing 'toStyledString' might be slightly better than just dropping it?
Although I guess then it is even better to do:
```suggestion
	setTitle(		tq(						title.isString() ? title.asString() : title.isDouble() ? ColumnUtils::doubleToString(title.asDouble(), false) : title.isInt() ? std::to_string(title.asInt()) : ""));
```
No?

boutinb:
OK It takes now the case it is a number. I can merge this I think.

PR #6080, opened 2025-12-16
Improved documentation

PabRod:
- More consistent chapter naming
- Simplified texts
- Friendlier explanations
- Added diagrams
- Added figures
- Added chapters


Docs/development/r-analyses-guide.md
```diff
@@ -87,29 +99,39 @@ The main analysis function has the following arguments:
...
-    ready <- (length(options$variables) > 0)
+    ready <- (length(options$variables) > 0) # Only TRUE if 1 or more variables are provided
+
    ```
 
 </details>
 
 ## Step 3 - Reading the Dataset
-If your R function is called from JASP the dataset will not be provided directly to the analysis. Instead, you must ask for the dataset. This allows you to specify how variables must be read (e.g., as factor) and how missing values should be treated -- they can be ignored or excluded listwise. To read the dataset, call `.readDataSetToEnd()` with the following optional arguments
+After selecting a dataset in JASP, you still have to load it into your R analysis. Reading data is always a tricky matter. Computers are devoid of common sense, so very often we have to explicitly specify how variables must be interpreted (are they integers?, characters?, ...). We recommend you to use our auxiliary function `.readDataSetToEnd()`, with the following optional arguments.
```
JorisGoosen: It would be better to describe the `preloadData: true` way here.

JorisGoosen: For the rest this file looks really good now ^^

JorisGoosen (changes requested):
Changes look quite good mostly, but I think it would be better if the R analyses guide describes the new way of working with data, as it is both easy and we want to remove the other way when we can.

Because that makes syntaxmode a lot easier.

Docs/development/r-analyses-guide.md
PabRod: Check commit https://github.com/jasp-stats/jasp-desktop/pull/6080/commits/c1be13594b740d210edaf9dd3e6064d71d38cf3b

Docs/development/r-analyses-guide.md
```diff
@@ -122,54 +122,19 @@ We have to determine early in the analysis whether we are ready to compute the r
...
-  .binomReadData <- function(dataset, options) {
-    if (!is.null(dataset)) # If data has been already read...
-      return(dataset)      # ... just return it ...
-    else                   # ... otherwise, read it
-      return(.readDataSetToEnd(columns.as.factor = options$variables))
-  }
-  ```
-
-</details></p>
+Reading data is always a tricky matter. Computers are devoid of common sense, so very often we have to explicitly specify how variables must be interpreted (are they a scale?, a nominal?, an ordinal?, ...). The `Edit data` panel allows you to specify this kind of things.
```
JorisGoosen: We dont have to do this via edit data, you can define which type you want in each qml component.
The type in the data can stay as it is.

This used to be false, so im guessing that is where this text comes from

JorisGoosen:
```suggestion
Reading data is always a tricky matter. Computers are devoid of common sense, so very often we have to explicitly specify how variables must be interpreted (are they a scale?, a nominal?, an ordinal?, ...). Each column in the data has a type set on it, which is how it is shown in the dataviewer. However, for an analysis you might well want to get a different type, just for that analysis. Or even get a scale *and* a nominal version of the column.
The preloading of data in JASP actually takes that into account and adds every column selected in any qml component to a dataset, converted to exactly the type requested by the user.
Just like you can write `columnName.scale` or `columnName.nominal` in a JASP filter or computed column.
```

Docs/development/r-analyses-guide.md
```diff
@@ -122,54 +122,19 @@ We have to determine early in the analysis whether we are ready to compute the r
...
+Notice that the column titles in the `dataset` will look a bit jumbled, _e.g._, the first column is titled `JaspColumn_.1._Encoded`, the second column is titled `JaspColumn_.2._Encoded`. This is due to the encoding we perform on the column titles, which allows us to handle foreign characters. The values in `options$variables` are NOT encoded and therefore do not match the column names in the dataset. Obviously this will present difficulties if we try to subset data later during the computation phase. The way we solve this is by using `decodeColNames()` to decode column names and `encodeColNames` to encode column names. To exemplify this, the following would return `TRUE`:
```
JorisGoosen: Here a part suggestion (but see the question below it):
```suggestion
Notice that the column titles in the `dataset` will look a bit jumbled, _e.g._, the first column is titled `JaspColumn_.1._Encoded`, the second column is titled `JaspColumn_.2._Encoded`. This is due to the encoding we perform on the column titles and the selected type, which allows us to handle foreign characters and multiple copies of columns with a different type selected. The values in `options$variables` are encoded and often even include the desired columnType and can be used directly to get the desired data from the dataset.
```

vandenman: It is indeed encoded. Here is a suggested rewrite:
```suggestion
Notice that the column titles in the `dataset` will look a bit jumbled, _e.g._, the first column is titled `JaspColumn_.1._Encoded`, the second column is titled `JaspColumn_.2._Encoded`. This is due to the encoding we perform on the column titles, which allows us to handle foreign characters. The values in `options$variables` are also encoded and therefore you can use them directly to subset the data. Usually, the results from an analysis are automatically decoded (e.g., tables will not show `JaspColumn_.1._Encoded` but instead the actual column name in the dataset) and the fact that encoding happens can be completely ignored by the developers. In some scenarios, however, you may need to do the decoding manually. For example, if you use column names as axis tick labels in a plot, then the width of the string becomes relevant, as a longer string means that `ggplot2` reserves more space for the tick label. In this case, you should manually decode the tick labels (Note that for the more common scenario where column names are used as axis titles this is not necessary). To this end, we can use `decodeColNames()` to decode column names and `encodeColNames` to encode column names. To exemplify this, the following would return `TRUE`:
```

Sidenote, if each sentence is on its own line then view diffs and making suggestions becomes a little easier.

JorisGoosen:
Thanks Don!

PR #6082, opened 2025-12-16
Filter fixes - allow the user to filter out everything

JorisGoosen:
Allows for having no accepted filter row whatsoever.

This fixes some issues related with https://github.com/jasp-stats/jasp-issues/issues/3941
The other was that also somehow filtervalues were not always being saved to sqlite properly?

A problem here was that the `ColumnsModel` which is used for lots of things in JASP was a `QTransposeProxyModel`.
Which was used to pass on columnname, -type etc to variable views in qml *and* the columns list in the filter/computed columns. 
However, when you filter everything out the transpose model optimises all data calls away cause it knows the model is empty...

To avoid this ive turned the `ColumnsModel` into its own `QAbstractTableModel`.

All of this has as a side-effect that know you can make errors in your filter that make it empty and you can actually read your errormsgs.

### DIFF
--- a/Desktop/data/columnsmodel.cpp
+++ b/Desktop/data/columnsmodel.cpp
@@ -113,8 +113,6 @@ QString ColumnsModel::getColumnTransformedToolTip(const QString &name, columnTyp
 
 QVariant ColumnsModel::data(const QModelIndex &index, int role) const
 {
-	if(index.row() < 0 || index.row() >= rowCount()) return QVariant();
-
 	QString				colName		= QTransposeProxyModel::data(index, int(DataSetPackage::specialRoles::name)).toString();
 	columnType			colType		= static_cast<columnType>			(QTransposeProxyModel::data(index, int(DataSetPackage::specialRoles::columnType			)).toInt());
 	computedColumnType	codeType	= static_cast<computedColumnType>	(QTransposeProxyModel::data(index, int(DataSetPackage::specialRoles::computedColumnType	)).toInt());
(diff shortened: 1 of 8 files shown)


Desktop/data/columnsmodel.cpp
```diff
@@ -214,8 +226,8 @@ bool ColumnsModel::absorbInfo(VariableInfo::InfoType info, const QString &colNam
 		switch(info)
 		{
 		default:										return	false;
-		case VariableInfo::DataSetValue:				return	QTransposeProxyModel::setData(qValIndex, value, int(DataSetPackage::specialRoles::value));
-		case VariableInfo::DataSetValues:				return	QTransposeProxyModel::setData(qColIndex, value,	int(DataSetPackage::specialRoles::valuesStrList));
+		case VariableInfo::DataSetValue:				return	QAbstractTableModel::setData(qValIndex, value, int(DataSetPackage::specialRoles::value));
```
boutinb: The qValIndex is set as if it was still the transposed proxy model. So that won't work

boutinb (changes requested):
The Audit analyses do not work anymore

Desktop/data/columnsmodel.cpp
JorisGoosen: If you have the fix you can push it ^^

boutinb: OK I do it.

JorisGoosen: Merci!

boutinb: OK, I have tested some Audit analyses, it seems to work now.

boutinb (approved):

PR #6085, opened 2025-12-18
use collapse icon instead of "delete analysis" icon

JorisGoosen:
fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/3052


boutinb:
The 'collapse' button should not trigger a warning to save the computed column. I don't find that logic. And even if I click the 'Apply changes', I still get a '*' in the title that warns me that I have to save my workspace: so it does not seem to work.
I would say: the collapse button should not trigger any warning, and the red-cross icon can be added to remove the computed column.

JorisGoosen:
I know I know.

Im not going to add the red cross.
Im going to fix the menu popping up

JorisGoosen:
So, because its all very confusing apparently:
- the red cross never deleted anything
- there is a bug that makes the "apply computed column" thing popup even when it ISNT a computed column
- if you want to delete a column you use the data editing functionality

this PR just changes the icon (so far) to make it less confusing for people.
but it should also fix the bug with the apply menu, which I will do.

JorisGoosen:
Now it only asks if you want to apply changes to your computed column when it is one, and it has changes.

boutinb (approved):

boutinb:
This is al better, but if you open a Computed column and you don't do anything and close the Filter window, then you get still the pop-up question if we want to apply the changes.

JorisGoosen:
Ah yes, if done from the "Plus" button.. Im afraid that is rather hard to fix.

boutinb:
OK let's merge this. This is already an improvement

JorisGoosen:
Hehe I was just thinking of a full fix, ill open a new pr when I find it ;)

PR #6089, opened 2025-12-21
test: add build test workflow for Windows on GitHub workflow

shun2wang:


Tools/CMake/Config.cmake
```diff
@@ -140,28 +142,49 @@ if(WIN32)
...
-    set(VC_TOOLS_REDIST_PATH
-        "C:\\Program Files\\Microsoft Visual Studio\\2022\\Community\\VC\\Redist\\MSVC\\v143"
-    )
-    set(VC_VARS_PATH_NATIVE
-        "C:\\Program Files\\Microsoft Visual Studio\\2022\\Community\\VC\\Auxiliary\\Build"
-    )
-  elseif(MSVC_VERSION GREATER "1920")
+
+    # GitHub Actions uses Enterprise edition of Visual Studio
+    if(GITHUB_WORKFLOW_WIN)
```
JorisGoosen: Maybe having another variable to contain the `C:\\Program Files\\Microsoft Visual Studio\\2022\\Enterprise` part is practical?

shun2wang: Do you mean simplifying this? For example, using only an optional variable like `VS_TOOLSET_EDITION` and then only changing the path? It certainly aligns more with intuition..

JorisGoosen: Yes something like that.

Or a `VS_PATH` var where you just put either `C:\\Program Files\\Microsoft Visual Studio\\2022\\Enterprise\\VC\\` or `C:\\Program Files\\Microsoft Visual Studio\\2022\\Community\\VC\\` 

And then you do `set(VC_TOOLS_REDIST_PATH "${VS_PATH}Redist\\MSVC\\v143")`

shun2wang:
This is also good start (with docs) if new developer want to build JASP locally.

shun2wang:
Should we merge this? also seems user cannot download Microsoft Visual Studio 2022 **Community** version easily anymore.so configure should compatibility with 2026 version. @JorisGoosen

JorisGoosen (approved):

shun2wang:
> Should we merge this? also seems user cannot download Microsoft Visual Studio 2022 **Community** version easily anymore.so configure should compatibility with 2026 version. @JorisGoosen 

Just note: don't upgrade to VS 2026 now cause Qt6 is not ready for this.

PR #6094, opened 2026-01-09
Make file menu wide enough even in long translations

boutinb:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/3024 (at least for the File Menu)

The FileMenu.qml takes into account the implicitWidth of its loaded item to determine its width.


JorisGoosen (approved):
Works nicely!

PR #6106, opened 2026-01-22
Safer heartbeat check

boutinb:
If the PC is asleep, when waking up, the Engine process might awake before the Desktop: when checking the hearbeat file timetamp, it will  see a quite old one. It should not stop, but still wait 60 seconds for the Desktop to wake up.

Fixes https://github.com/jasp-stats/jasp-test-release/issues/3141

### DIFF
--- a/Common/enginedefinitions.h
+++ b/Common/enginedefinitions.h
@@ -32,6 +32,6 @@ struct unexpectedEngineReply  : public std::runtime_error
 #define ENGINE_BORED_SHUTDOWN (30 * 60)
 
 ///Engines need some time between closing and starting to avoid problems with shared memory
-#define ENGINE_COOLDOWN 1
+#define ENGINE_COOLDOWN 200
 
 #endif // ENGINEDEFINITIONS_H
--- a/CommonData/ipcchannel.cpp
+++ b/CommonData/ipcchannel.cpp
@@ -231,6 +231,7 @@ bool IPCChannel::jaspAlive()
 	
 	if(newTimestamp  != _lastHeartBeatTimestamp) {
 		_lastHeartBeatTimestamp = newTimestamp ;
+		_startHeartAttackTimestamp = 0;
 		return true;
 	}
 
@@ -242,9 +243,17 @@ bool IPCChannel::jaspAlive()
 	
 	if(Utils::currentSeconds() - _lastHeartBeatTimestamp > _maxHeartbeatDiffS)
 	{
-		Log::log() << "heartbeat time limit exceeded, last timestamp was from " << (Utils::currentSeconds() - _lastHeartBeatTimestamp) << " seconds ago. Heartbeat file is at '" << _jaspHeartBeatPath << "'."  << std::endl;
-		return false;
+		if (_startHeartAttackTimestamp == 0)
+			// Maybe the engine is just awaken, and the Desktop not yet. Measure the heart attack only from now.
+			_startHeartAttackTimestamp = Utils::currentSeconds();
+		else if (Utils::currentSeconds() - _startHeartAttackTimestamp > _maxHeartAttackDurationS)
+		{
+			Log::log() << "heartbeat time limit exceeded, last timestamp was from " << (Utils::currentSeconds() - _lastHeartBeatTimestamp) << " seconds ago, heart attack started " << (Utils::currentSeconds() - _startHeartAttackTimestamp) << " seconds ago. Heartbeat file is at '" << _jaspHeartBeatPath << "'."  << std::endl;
+			return false;
+		}
 	}
+	else
+		_startHeartAttackTimestamp = 0;
 
 	return true;
 }
--- a/CommonData/ipcchannel.h
+++ b/CommonData/ipcchannel.h
@@ -79,10 +79,12 @@ private:
 	static bool										heartbeat(std::string path, unsigned int delayMs);
 	static std::thread								_heartbeatThread;
 
-	int64_t											_lastHeartBeatTimestamp = 0;
+	int64_t											_lastHeartBeatTimestamp = 0,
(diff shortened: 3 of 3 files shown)


JorisGoosen:
the heartbeatcheck was meant to make an engine stop when waking from sleep...
You sure this is a fix?

boutinb:
> the heartbeatcheck was meant to make an engine stop when waking from sleep... You sure this is a fix?

Why should an engine stops after sleeping?
I think it should stop only if the JASP Desktop is really dead. And anyway, the code is written in such a way that the Engine may or may not stay alive whether it reads the heartbeat file timestamp before or after the Desktop updates it: the behaviour depends on such a concurrency issue, this seems to me quite wrong. So I think my code is better, apart if there is a good reason to kill anyway an Engine after sleeping.

But this is true: it does not explain the crash...

So I have tried to understand further the crash. What strikes me in the log files of the issue, it that the Engine #1 is restarted twice in the same millisecond:

```
22:32:44.614: Engine # 1 finished while in state 'idle' normally and with exitCode 0!
22:32:44.614: EngineRepresentation::handleEngineCrash():
**Engine # 1** process *does not exist (anymore)* and it's state is *idle*.


###########################################################################################
#######         Engine #1 (re)started at 2026-01-22 22:36:45
###########################################################################################

(...)

22:32:44.615: restartEngineAfterCrash(1)


###########################################################################################
#######         Engine #1 (re)started at 2026-01-22 22:36:45
###########################################################################################
```

If I look carefully at the code, I see that the 2nd ` Engine #1 (re)started` is due to the normal handleEngineCrash (called by the processFinished signal), because the `restartEngineAfterCrash` log is written just before calling `startSlaveProcess` that writes this `Engine #1 (re)started`
The first ` Engine #1 (re)started` is called apparently by the `createNewEngine` in the `EngineSync::process()`

Starting twice the same Engine in the same millisecond may probably explain the crash, and is anyway not good.

I see that  `EngineRepresentation::requestEngineRestartAfterCrash` is connected with `EngineSync::restartEngineAfterCrash` with Qt::QueuedConnection. Why? If restartEnginesAfetrCrash is called immediately, then there is no risk that `EngineSync::process()` will call `createNewEngine` in the meantime.

JorisGoosen:
> But this is true: it does not explain the crash...
> So I have tried to understand further the crash. What strikes me in the log files of the issue, it that the Engine #1 is restarted twice in the same millisecond:


Cant we just set the following in enginedefinitions.h to something higher:
```
///Engines need some time between closing and starting to avoid problems with shared memory
#define ENGINE_COOLDOWN 1
```
Likes 200ms instead of 1?
Probably solves the problem without having to change fundamental stuff about how it all works?
It used to be 50

JorisGoosen:
> Why should an engine stops after sleeping? 

The whole point was that I (and I thought we) thought that engines trying to do all sort of things right after a computer comes out of sleep is not good. So this heartbeat achieves that, because an engine that manages to do stuff before the desktop even beats its heart gracefully shuts down.
And then de desktop can start a new engine as needed.
 
 
> I think it should stop only if the JASP Desktop is really dead.

I think it should *also* stop if desktop is reall dead.

> And anyway, the code is written in such a way that the Engine may or may not stay alive whether it reads the heartbeat file timestamp before or after the Desktop updates it: the behaviour depends on such a concurrency issue, this seems to me quite wrong. 

Which concurrency issue?

> So I think my code is better, apart if there is a good reason to kill anyway an Engine after sleeping.

Well, maybe "better" in the sense that it now does another thing, that im not sure we need for anything.

boutinb:
> Which concurrency issue?

The engine will not kill itself if the Desktop is awaken before the Engine and touches the heartbeat file. So whether the engine is killed, depends on how quick the Desktop and the Engine processes will awake.

I still do not understand why the Engine should be killed if it has slept. What is the problem? If an engine was busy to run an analysis before being asleep, it could still go on with running the analysis when it is awaken. If it puts stuff in the IPCChannel while the Desktop process is still asleep, as this IPCChannel is just a shared memory, it should not be a problem. What is the use-case where it is a real problem?

By not killing the Engine when it is not necessary (that is when the Desktop is not really dead), this avoids restarting the Engine, which causes (indirectly) the crash. So this PR fixes the issue.

Increasing the ENGINE_COOLDOWN might help, but I'm not at all sure. It depends whether the _engineStopTimes get the right timestamps...
I think really that removing the Qt::QueuedConnection in
`connect(engine, &EngineRepresentation::requestEngineRestartAfterCrash, this, &EngineSync::restartEngineAfterCrash, Qt::QueuedConnection);`
should avoid the crash, because the crash handling will first restart the engine, and the Desktop won't add a new Engine with the same channel.

JorisGoosen:
https://github.com/jasp-stats/jasp-desktop/pull/5001

JorisGoosen:
Its probably this part:
>On MacOS crucial memory mapped files of other instances are deleted upon start up.
This PR should fix the unnecessary removal of the temp files which can lead to crashing engine and entering eternal resume-crash loop.

>This PR builds in safety margin by extending time limit to 24 hours and fastening the heartbeat to guard against possible hidden bugs, likely related to post sleep and app-nap interference with timers on MacOS.

So Id suggest just trying changing that one single number to a higher value because the crash is related to the engine getting restarted very quickly repeatedly right after awaking. So avoiding that by using the already builtin define we avoid that, hence fix the problem... Also requires less testing and other stuff.

boutinb:
OK I have set the ENGINE_COOLDOWN to 200ms. Hopefully this solves this issue. I have tried to reproduce it by stopping the engine every 2 seconds, but did not succeed to get any crash. So just pray that this is the right solution!

JorisGoosen:
:crossed_fingers:

PR #6108, opened 2026-01-24
bugfix: make the tooltip background color style consistently

shun2wang:
Before: （e.g. the split view handle on Linux)
image

After:

image

### DIFF
--- a/Desktop/components/JASP/Widgets/MainWindow.qml
+++ b/Desktop/components/JASP/Widgets/MainWindow.qml
@@ -290,7 +290,6 @@ Window
 	{
 		//These properties override those for ALL attached ToolTips in the application
 		//ToolTip.toolTip shouldn't be changed anywhere else otherwise we get hard to debug behaviour
-		ToolTip.toolTip.background:		Rectangle { color: jaspTheme.tooltipBackgroundColor }
 		ToolTip.toolTip.contentItem:	Text
 		{
 			font:			jaspTheme.font
--- a/Desktop/components/JASP/Widgets/AnalysisFormExpander.qml
+++ b/Desktop/components/JASP/Widgets/AnalysisFormExpander.qml
@@ -134,7 +134,6 @@ DropArea
 			timeout:		jaspTheme.toolTipTimeout
 			delay:			jaspTheme.toolTipDelay
 			font:			jaspTheme.font
-			background:		Rectangle { color:	jaspTheme.tooltipBackgroundColor }
 			visible:		mouseArea.containsMouse && !analysesModel.moving && analysesModel.rowCount() > 1
 			y:				mouseArea.mouseY
 			x:				mouseArea.mouseX + 5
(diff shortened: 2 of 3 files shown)


Desktop/components/JASP/Widgets/MainWindow.qml
```diff
@@ -290,7 +290,6 @@ Window
 	{
 		//These properties override those for ALL attached ToolTips in the application
 		//ToolTip.toolTip shouldn't be changed anywhere else otherwise we get hard to debug behaviour
-		ToolTip.toolTip.background:		Rectangle { color: jaspTheme.tooltipBackgroundColor }
 		ToolTip.toolTip.contentItem:	Text
```
boutinb, line 293: _toolTipOverrideItem is never used, and does not seem tp set the Tootip properties either. It should be removed (after the release)

boutinb (approved):
We need a proper custom Tootip. It is now redefined each it is used. 
I've made an issue for this: https://github.com/jasp-stats/INTERNAL-jasp/issues/3081

JorisGoosen:
Kind of a pointless PR...
Please revert, and if you want to temporarily set the background color to white just set that in `Theme.qml` and `DarkTheme.qml` instead.

shun2wang:
> Kind of a pointless PR... Please revert, and if you want to temporarily set the background color to white just set that in `Theme.qml` and `DarkTheme.qml` instead.

I don't think so, not only me have this problem:https://github.com/jasp-stats/INTERNAL-jasp/issues/3076   that might because Qt Quick2 tooltip on Linux/flatpak has different behavior(black tooltip on Linux and white on Windows). that not a user behavior. on the other hand, yellow background color is not consistently with other tooltips.

boutinb:
I will reset the tooltip background, and set it property to all Tooltips. Then set a proper color for dark and light theme.

shun2wang:
But why^^ If we think the default tooltip background color style is not suitable for JASP, then I believe we definitely need a customizable and reusable component.

JorisGoosen:
> > Kind of a pointless PR... Please revert, and if you want to temporarily set the background color to white just set that in `Theme.qml` and `DarkTheme.qml` instead.
> 
> I don't think so, not only me have this problem:[jasp-stats/INTERNAL-jasp#3076](https://github.com/jasp-stats/INTERNAL-jasp/issues/3076) that might because Qt Quick2 tooltip on Linux/flatpak has different behavior(black tooltip on Linux and white on Windows). that not a user behavior. on the other hand, yellow background color is not consistently with other tooltips.

Hence the "if you want to temporarily set the background color to white just set that in `Theme.qml` and `DarkTheme.qml` instead"

JorisGoosen:
> But why^^ If we think the default tooltip background color style is not suitable for JASP, then I believe we definitely need a customizable and reusable component.

Thats what I tried through the mainwindow override thing, which used to work at a time.
However, ToolTips in QML are done via the attached ToolTip. properties. Which is why I fixed it as described in https://github.com/jasp-stats/INTERNAL-jasp/issues/3081#issuecomment-3804879778

Once that PR is merged we can tweak the color away from white again.

JorisGoosen:
Ah so it turns out that all the stuff that I fixed also stopped working :')

I guess Qt is serious about the "use a style to theme tooltip" after all

boutinb:
I have fixed this by using a style: https://github.com/jasp-stats/jasp-desktop/pull/6113

PR #6112, opened 2026-01-27
On uiScaleChanged also zoom modules app

sverhoeven:
Refs https://github.com/jasp-stats/INTERNAL-jasp/issues/3060

Todo
- [x] use either setZoomFactor or runJavascript function, see https://doc.qt.io/qt-6/qwebenginepage.html#zoomFactor-prop
  to act on PreferencesModelBase::uiScaleChanged notification

Also updates qt version in devcontainer.


boutinb:
Hi @sverhoeven, I'm not sure what is the goal of this PR, because zooming works with the Module Library.

sverhoeven:
The parent issue was closed because I made the text bigger.

If I use the nightly flatpak zoom in/out with CTRL+ / CTRL - then it makes modules library column wider but the text size of the module library stays the same while text size outside it grows/shrinks.
Same for the Windows nightly.

So I still want make zoom work.

JorisGoosen:
And you can't just use something like:
```
window.setZoom			= function (zoom)			{ document.body.style.zoom = "" + Math.floor(zoom * 100) + "%";	}
```
As we do in the results?

sverhoeven:
I am using zoomFactor because that was being used at https://github.com/jasp-stats/jasp-desktop/blob/32bc32431a0004e158af1d51154ca99c21f050b3/Desktop/components/JASP/Widgets/HelpWindow.qml#L39
and seemed the more simple/maintainable approach

boutinb (approved):

PR #6117, opened 2026-01-28
Reset delay and timeout each time the ToolTip is used.

boutinb:
Also reset the ToolTip in TextField (when value is wider than its field width), that was unfortunately removed in the previous PR.

### DIFF
--- a/Desktop/components/JASP/Style/ToolTip.qml
+++ b/Desktop/components/JASP/Style/ToolTip.qml
@@ -23,6 +23,11 @@ QtC.ToolTip {
 		color: jaspTheme.textEnabled
 	}
 
-	timeout:	jaspTheme.toolTipTimeout
 	delay:		jaspTheme.toolTipDelay
+	timeout:	jaspTheme.toolTipTimeout
+
+	// Each time the ToolTip is used, QML resets the delay and timeout to the defualt values (delay: 0 and timeout: -1)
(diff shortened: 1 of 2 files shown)


JorisGoosen:
Ah, I was indeed wondering whether that was really on purpose but forgot to doublecheck

Desktop/components/JASP/Style/ToolTip.qml
```diff
@@ -23,6 +23,11 @@ QtC.ToolTip {
 		color: jaspTheme.textEnabled
 	}
 
-	timeout:	jaspTheme.toolTipTimeout
 	delay:		jaspTheme.toolTipDelay
+	timeout:	jaspTheme.toolTipTimeout
+
+	// Each time the ToolTip is used, QML resets the delay and timeout to the defualt values (delay: 0 and timeout: -1)
+	onDelayChanged: if (delay === 0) delay = jaspTheme.toolTipDelay
+	onTimeoutChanged: if (timeout === -1) timeout = jaspTheme.toolTipTimeout
```
JorisGoosen, line 31:
```suggestion
	onDelayChanged: if (delay === 0) 			delay = Qt.Binding(function(){ return jaspTheme.toolTipDelay; }
	onTimeoutChanged: if (timeout === -1) 	timeout = Qt.Binding(function(){ return jaspTheme.toolTipTimeout; }
```

Maybe this is better?
Although they never really change do they?

JorisGoosen (approved):

PR #6119, opened 2026-01-30
Set MouseArea to the right zone in a control

boutinb:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2916

The tooltip is activated by a MouseArea for each control. This MouseArea used to fill the whole control area, but it should not be always the case:

-  for CheckBox: only the inner control CheckBox should have this MouseArea, not its sub groups
-  for RadioButtonGroup: only the RadioButtonGroup lable should get this MouseArea, each RadioButton should have its own MouseArea
- for TextField, a special MouseArea is set in case the content of the TextField is wider than its own width: this special MouseArea is used to activate a tooltip to show the value of the TextField. In this case the MouseArea should cover only the label of the TextField.
- for DropDown: an extra MouseArea is set for each value of the DropDown, so that if there is an info for a value, this info can be dsplayed in a tooltip.

To allow these cases, a MouseAreaZone property is added, that can be adjusted for each situation.

### DIFF
--- a/QMLComponents/components/JASP/Controls/CheckBox.qml
+++ b/QMLComponents/components/JASP/Controls/CheckBox.qml
@@ -33,6 +33,7 @@ CheckBoxBase
 	focusIndicator:		focusIndicator
 	childControlsArea:	childControlsArea
 	innerControl:		control
+	mouseAreaZone:		control
 	title:				text
 
 	default property alias	content:				childControlsArea.content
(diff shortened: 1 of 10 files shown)


boutinb:
I have added extra tests in Tooltip analysis of jaspTestModule

JorisGoosen (approved):

PR #6120, opened 2026-01-30
Do not reset all encoding when new columns are added

boutinb:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/2991

When an analysis adds a computed column, the column encoding is reset, so that some variables may get a new encoding. In the SEM analyses, the variables are also encoded in the options, and this encoding should be preserved when a new column is created.
Just adding the column in the encoding was not enough, since the column is first added with an unknown type, and afterwards its type is set. So to make the encoding more robust, it removes first the variables that does not exist or have a another type, and encode only the new variables (or variables with new type).

### DIFF
--- a/Desktop/data/datasetpackage.cpp
+++ b/Desktop/data/datasetpackage.cpp
@@ -2672,6 +2672,8 @@ void DataSetPackage::checkDataSetForUpdates()
 
 	if(_dataSet->checkForUpdates(&changedCols, &missingCols, &newCols, &rowCountChanged))
 	{
+		ColumnEncoder::setCurrentColumnNames(	getColumnTypesMap());
+
 		Log::log()	<< "Updates found for DataSet " << _dataSet->id() 
 					<< "| missing cols: '" << tq(missingCols).join(",")
 					<< "' | changed cols: '" << tq(changedCols).join(",")
--- a/QMLComponents/controls/textareabase.cpp
+++ b/QMLComponents/controls/textareabase.cpp
@@ -42,13 +42,15 @@ void TextAreaBase::setUpModel()
 	{
 		_model = new ListModelTermsAvailable(this);
 		_model->setNeedsSource(_textType == TextType::TextTypeCSem || _textType == JASPControl::TextType::TextTypeMetaSem);
+		connect(VariableInfo::info(),	&VariableInfo::dataSetChanged,		this,	&TextAreaBase::checkSyntaxHandler);
 
 		if (_textType == TextType::TextTypeLavaan)
 		{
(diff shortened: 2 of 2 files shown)


JorisGoosen (changes requested):
I would rather appreciate keeping the old function too.
So we can start afresh when we want to or we use the incremental update function that you've now written.

Common/columnencoder.cpp
```diff
@@ -158,24 +158,59 @@ void ColumnEncoder::setCurrentColumnTypePerName(const colTypeMap &theMap)
...
+		if (!namesWithTypes.count(nameType.first) || namesWithTypes.at(nameType.first) != nameType.second)
+			namesToRemove[nameType.first] = nameType.second;
+
+	for (const auto & nameType : namesWithTypes)
+		if (!_dataSetTypes.count(nameType.first) || _dataSetTypes.at(nameType.first) != nameType.second)
+			namesToAdd[nameType.first] = nameType.second;
 
 	_dataSetTypes = namesWithTypes;
+
+	auto encodedNameGenerator	= [this]() { return _encodePrefix + std::to_string(_runningCounter++) + _encodePostfix; }; //Slightly weird (but R-syntactically valid) name to avoid collisions with user stuff.
```
JorisGoosen: does runningcounter ever get reset?
Or it just goes up forever?

boutinb:
OK, my first attempt did not work completely: when changing names of columns, or deleting/adding new columns, the SEM syntax generates sometimes errors.

What I noticed, is that when the Engine creates a column, the Desktop noticed it, but does not update the column encoder. So I have added this, and add a signal so that the TextArea containing the SEM syntax must update its option.

JorisGoosen:
So column encoder needs no changes?
And I should review again?

boutinb:
> So column encoder needs no changes? And I should review again?

No, column encoder does not need to change. My first thought was to change the Column Encoder so that it keeps the encoding of columns, even after deleting or adding columns. But as the same encoding must be used by the Desktop and the Engine, it was too difficult to ensure that the encoding keeps synchronized between the Engine and the Desktop (especially when several columns are added by the Engine).
The current implementation of the Column Encoder ensures that each time the columns are changed, the whole encoding is reset, and as the encoding used ordered maps, the encoding is then automatically the same for the Desktop and the Engine.
So my second thought was just to update the options sent by the SEM TextArea (that contains encoding columns), each time the dataset changes. Here I saw that the Desktop did not reset the columns encoder, when a column was added (or removed) by the Engine (at least not immediately), so I just added this code, and now it seems to work.
So yes, you should review again.

JorisGoosen (approved):

PR #6126, opened 2026-02-04
Some fixes and tweaks for modules and the online module library

JorisGoosen:
- Adds a "isDevMod" role to ribbonmodel to make sure only development modules show a "refresh" icon
- Sort the modules as the get loaded on startup (in the correct order) to have newer version and/or user-installed modules have preference over the others.
- Make sure to store the analyses json before reloading the dynamic modules to recreate the analyses at the end. This probably also applies upgrades and should fix https://github.com/jasp-stats/jasp-test-release/issues/3128

I open it now as a draft to remove the edits I made that weren't helpful and because some other things probably need to be retested (like language changing etc and devmodule installs)

### DIFF
--- a/Desktop/analysis/analyses.cpp
+++ b/Desktop/analysis/analyses.cpp
@@ -267,7 +267,15 @@ Json::Value Analyses::asJson() const
 
 void Analyses::saveAnalysesJsonForReload()
 {
-	_tempSave = asJson();	
+	_tempSave = asJson();
+	
+	destroyAllForms();
+	
+	for(auto & idAnalysis : _analysisMap)
+		delete idAnalysis.second;
+	
+	_analysisMap.clear();
+	_orderedIds.clear();
 }
 
 void Analyses::reloadSavedAnalysesJson()
@@ -278,6 +286,8 @@ void Analyses::reloadSavedAnalysesJson()
 	bool errorFound;
 	std::stringstream errors;
 	loadAnalysesFromJaspFileJson(_tempSave["analyses"], _tempSave["meta"], errorFound, errors, RibbonModel::singleton());
+	
+	applyToAll([](Analysis * a){ a->setBeingTranslated(true); });
 }
 
 
@@ -365,11 +375,20 @@ void Analyses::rescanAnalysisEntriesOfDynamicModule(Modules::DynamicModule * mod
 			else
 			{
 				Analysis * a = keyval.second;
+				
+				//Json::Value prevBound = Json::nullValue;
+				//if(a->form())
+				//	prevBound = a->asJSON();
 
 				//This function is called once after 300 ms upon translation due to delayedUpdate in description
 				//and causes the set analysis options to be cleared without this additional flag check set at the start of translations
 				if(a->readyToCreateForm() && !a->beingTranslated())
+				{
 					a->createForm();
+					
+					//if(!prevBound.isNull())
+					//	a->form()->bindTo(prevBound);
+				}
 				a->setBeingTranslated(false);
 			}
 		}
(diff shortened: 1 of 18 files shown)


Desktop/modules/installedmodules.cpp
```diff
@@ -72,7 +72,7 @@ std::vector<InstalledModules::ModuleInfo> InstalledModules::getModules() {
 	auto modulesAll = getAllAvailableModules();
 	std::map<std::string, InstalledModules::ModuleInfo> modules;
 	for(const auto& module : modulesAll) { //remove duplicates take highest version
-		if(modules.find(module.name) == modules.end() || modules[module.name].version >= module.version) {
+		if(modules.find(module.name) == modules.end() || modules[module.name] < module) {
```
JorisGoosen, line 76: So this needs to >= for the module library itself and < for the ribbonmodel, so I should make this into an argument or something

JorisGoosen:
This now properly fixes https://github.com/jasp-stats/jasp-test-release/issues/3128
And it now also upgrades options for development modules (only tested with renv style module install, but I see no reason why it would be different for the other method)

JorisGoosen:
It is a bit hard to test the proper upgrade of the module library now though cause the module library right now does not actually allow me to update anything at the moment.

Desktop/modules/installedmodules.cpp
JorisGoosen: So this is not really true, but the problem is more likely to lie at the serverside.

Desktop/analysis/analyses.cpp
```diff
@@ -64,50 +64,19 @@ Analysis* Analyses::createFromJaspFileEntry(Json::Value analysisData, RibbonMode
...
 	bool						wasUpgraded		= Upgrader::upgrader()->upgradeAnalysisData(analysisData, msgs);
 	Json::Value				&	optionsJson		= analysisData["options"];
-	Modules::AnalysisEntry	*	analysisEntry	= nullptr;
-
-	if(analysisData.get("dynamicModule", Json::nullValue).isNull()) //This used to be the code to load our "standard modules", seeing as how those are gone and the've become dynamic as well it should figure out somehow which one is meant
+	std::string					title			= analysisData.get("title", "").asString();
+	Modules::AnalysisEntry	*	analysisEntry	= Modules::DynamicModules::dynMods()->retrieveCorrespondingAnalysisEntry(analysisData["dynamicModule"]);
+	Analysis				*	analysis		= create(analysisData, analysisEntry, id, status, false, title, analysisData["dynamicModule"]["moduleVersion"].asString(), &optionsJson);
+	
+	if(msgs.count(Modules::analysisLog))
```
boutinb, line 75: It seems that this code does not take old JASP files into account anymore, where the version was directly set in analysisData["version"]

QMLComponents/analysisbase.h
```diff
@@ -28,7 +28,7 @@ class AnalysisBase : public QObject
 	virtual				bool				isDuplicate()												const	{ return false;				}
 	virtual				bool				wasUpgraded()												const	{ return false;				}
 	virtual				bool				needsRefresh()												const	{ return false;				}
-	virtual				const std::string & module()													const	{ return emptyString;		}
+	virtual				const std::string   module()													const	{ return emptyString;		}
```
boutinb, line 31: Why?

QMLComponents/analysisform.cpp
```diff
@@ -672,9 +672,9 @@ void AnalysisForm::setAnalysisUp()
 
 	_setUpControls();
 
-	Json::Value defaultOptions = _analysis->orgBoundValues();
-	_analysis->clearOptions();
-	bindTo(defaultOptions);
+	Json::Value defaultOptions	= _analysis->orgBoundValues(),
+				prevOptions		= _analysis->boundValues();
+	bindTo(prevOptions.isArray() && prevOptions.size() > 0 ? prevOptions : defaultOptions);
```
boutinb: The boundValues is an json object, not an array.
But the orgBoundValues is just confusing, and is in fact not needed (anymore). I'll add a commit to remove it.

PR #6127, opened 2026-02-05
FormulaFields from JASP file do not get always the right value

boutinb:
Bug found when testing jaspTextModule "Test Input Field" analysis. It is reproduceable with any analysis using FormulaField (as Frequencies Binomial Test, Test Value field).

- Set an expression like 1/3 (different than the default)
- Duplicate -> default value is set instead of 1/3
- Or Save the file, and open it -> default value is set instead of 1/3

The expression 1/3 is well stored in JASP file (in the option json), but when reading it, the `_value` property was not set.

It should run a R script in the engine to solve the expression. But this does not work since the analysis is not yet stored in the analyses object (and signals are not yet set between this analysis and analyses).
But as the real value is used only to check the constraints (min or max value for example), and as the problem occurs during the initialization of the analysis (using value from a JASP file, or from default values), we can expect that the constraints are fulfilled, and so we don't need to know the real value.

### DIFF
--- a/Desktop/modules/installedmodules.cpp
+++ b/Desktop/modules/installedmodules.cpp
@@ -6,6 +6,7 @@
 #include <fstream>
 #include "gui/preferencesmodel.h"
 #include "log.h"
+#include "resultstesting/compareresults.h"
 
 
 const std::string InstalledModules::settingsPath = "modules-settings.json";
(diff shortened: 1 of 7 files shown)


JorisGoosen:
Will you also add a nice unittest?

JorisGoosen:
Works really nicely in any case, but I would really appreciate a unittest to avoid regression en la future.

boutinb:
I have added a unit test for FormulaField.
I cannot test the evaluation of the R formula, since the unit test framework does not start an engine, but I could mimic the loading of a JASP file, and in this way the unit test tests automatically the bug.

JorisGoosen (commented):
❤️

JorisGoosen (approved):

PR #6139, opened 2026-02-19
Check known issues per module version

boutinb:
Now we have a module library, the known issue check should be done per module version.

The syntax of the KnownIssues.json file is changed (from the 0.96.0 version):
```
	"0.96":
	{
		"jaspTTests":
		{
			"version": "0.95.5",
			"analyses":
			{
				"TTestIndependentSamples":
				{
					"info": "Bad error with normality Test. Please upgrade the T-Tests module",
					"options":
					[
						"normalityTest"
					]
				}
			}
		}
	}
```

To test it, just update the loaded KnownIssues.json in your AppData (for Mac: ~/Library/Application Support/JASP/JASP/KnownIssues.json).

There was also a bug I fixed: the KnownIssues.json is downloaded once per day from our server, but the json file is read only when just downloaded. If you start afterwards JASP, the local json file was not read, and no known issues were displayed. Now, it reads always the local KnownIssues json file each time JASP is started.

This is a temporary fix. I think it's better that the module self keeps their known issues (in their master branch).
Also instead of a json file, this should be a QML file, so that we can translate the warning message.

### DIFF
--- a/Desktop/analysis/analysis.cpp
+++ b/Desktop/analysis/analysis.cpp
@@ -31,7 +31,7 @@
 #include "modules/description/description.h"
 #include "gui/jaspConfiguration/jaspconfiguration.h"
 
-Analysis::Analysis(size_t id, Modules::AnalysisEntry * analysisEntry, std::string title, std::string moduleVersion, Json::Value *data) :
+Analysis::Analysis(size_t id, Modules::AnalysisEntry * analysisEntry, const std::string & title, const Version & moduleVersion, const Json::Value & data) :
 	  AnalysisBase(Analyses::analyses(), moduleVersion),
 		_id(				id),
 		_name(			analysisEntry->function()),
@@ -41,11 +41,13 @@ Analysis::Analysis(size_t id, Modules::AnalysisEntry * analysisEntry, std::strin
 		_moduleData(		analysisEntry),
 		_dynamicModule(	_moduleData->dynamicModule())
 {
+	// If the moduleVersion parameter is given, this is the version this analysis was stored with (in a JASP file).
+	// This version might be not the same as the current module version: in this case, the analysis will have to be refreshed.
 	if(_moduleVersion.isEmpty() && _dynamicModule)
 		_moduleVersion = _dynamicModule->version();
 
 	if (data)
-		setBoundValues(*data); //Same story as other constructor
+		setBoundValues(data); //Same story as other constructor
 
 	_codedReferenceToAnalysisEntry	= analysisEntry->codedReference(); //We need to store this to be able to find the right analysisEntry after reloading the entries of a dynamic module (destroys analysisEntries). Or replacing the entry if a different version of the module gets loaded of course.
 	_helpFile						= dynamicModule()->helpFolderPath() + tq(analysisEntry->function());
@@ -479,6 +481,7 @@ void Analysis::setStatus(Analysis::Status status)
 		
 		_wasUpgraded		= false;
 		_storedWithoutState	= false;
+		// The analysis has been run, so its version is the same as the module version.
 		_moduleVersion		= _dynamicModule ?  _dynamicModule->version() : AppInfo::version;
 
 		if(neededRefresh != needsRefresh())
--- a/Desktop/analysis/analysis.cpp
+++ b/Desktop/analysis/analysis.cpp
@@ -31,8 +31,8 @@
 #include "modules/description/description.h"
 #include "gui/jaspConfiguration/jaspconfiguration.h"
 
-Analysis::Analysis(size_t id, Modules::AnalysisEntry * analysisEntry, const std::string & title, const Version & moduleVersion, const Json::Value & data) :
-	  AnalysisBase(Analyses::analyses(), moduleVersion),
+Analysis::Analysis(size_t id, Modules::AnalysisEntry * analysisEntry, const std::string & title, const Version & optionsVersion, const Json::Value & options) :
+	  AnalysisBase(Analyses::analyses()),
 		_id(				id),
 		_name(			analysisEntry->function()),
 		_qml(			analysisEntry->qml().empty() ? _name : analysisEntry->qml()),
@@ -41,13 +41,13 @@ Analysis::Analysis(size_t id, Modules::AnalysisEntry * analysisEntry, const std:
 		_moduleData(		analysisEntry),
 		_dynamicModule(	_moduleData->dynamicModule())
 {
-	// If the moduleVersion parameter is given, this is the version this analysis was stored with (in a JASP file).
+	// If the optionsVersion parameter is given, this is the version this analysis was stored with (in a JASP file).
 	// This version might be not the same as the current module version: in this case, the analysis will have to be refreshed.
-	if(_moduleVersion.isEmpty() && _dynamicModule)
-		_moduleVersion = _dynamicModule->version();
+	if(!optionsVersion.isEmpty() && _dynamicModule)
+		_optionsFromDifferentVersion = _dynamicModule->version() != optionsVersion;
 
-	if (data)
-		setBoundValues(data); //Same story as other constructor
+	if (options)
+		setBoundValues(options); //Same story as other constructor
 
 	_codedReferenceToAnalysisEntry	= analysisEntry->codedReference(); //We need to store this to be able to find the right analysisEntry after reloading the entries of a dynamic module (destroys analysisEntries). Or replacing the entry if a different version of the module gets loaded of course.
 	_helpFile						= dynamicModule()->helpFolderPath() + tq(analysisEntry->function());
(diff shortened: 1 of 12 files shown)


boutinb:
One issue to discuss: if a JASP file is loaded with an analysis having a module version older than the current module version, if there exist some known issues with this older version, it currently does not show the known issues, because the analysis is already upgraded to the the current version. 
There is now no warning message if an analysis comes from an older version and no upgrade was needed. But I can see in the code that the temporary files are deleted in this case. Maybe would it be informative to warn the user that the analysis comes from an older module version, and it would be wise to refresh the analysis.

Desktop/analysis/analysis.cpp
```diff
@@ -990,8 +992,7 @@ void Analysis::setUpgradeMsgs(const Modules::UpgradeMsgs &msgs)
 
 bool Analysis::needsRefresh() const
 {
-	bool differentVersion = _moduleVersion != (_dynamicModule ? _dynamicModule->version() : AppInfo::version);
-	return _wasUpgraded || _storedWithoutState || differentVersion;
+	return _wasUpgraded || _storedWithoutState || _optionsFromDifferentVersion;
```
JorisGoosen, line 995: _wasUpgraded == _optionsFromDifferentVersion right?
Why add an extra property?
Or is _wasUpgraded "An upgrade was applied" and  _optionsFromDifferentVersion "options from a different version without having an upgrade applied" ?

Because I think you couldve just set `_wasUpgraded` to `true`

Desktop/modules/dynamicmodule.cpp
```diff
@@ -856,7 +856,7 @@ Json::Value DynamicModule::asJsonForJaspFile(const std::string & analysisFunctio
 	Json::Value json(Json::objectValue);
 
 	json["moduleName"]			= name();
-	json["moduleVersion"]		= version();
+	json["moduleVersion"]		= version().asString();
```
JorisGoosen, line 859: Why?
The default is already to turn it into a string?

Desktop/modules/dynamicmodule.h
```diff
@@ -107,8 +107,7 @@ class DynamicModule : public QObject
 	QString				titleQ()			const { return QString::fromStdString(title());			}
 	bool				requiresData()		const { return AnalysisEntry::requiresDataEntries(_menuEntries); }
 	std::string			author()			const { return _author;									}
-	std::string			version()			const { return _version;								}
-	QString				versionQ()			const { return QString::fromStdString(_version);		}
```
JorisGoosen, line 111: Ah ok, I see

QMLComponents/knownissues.cpp
```diff
@@ -20,10 +20,14 @@ KnownIssues::KnownIssues(QObject * parent) : QObject(parent)
 {
 	assert(!_knownIssues);
 	_knownIssues = this;
+	loadKnownJson();
 }
 
 void KnownIssues::loadLocalJson(const std::string & filePath, bool saveIt)
 {
+	if (!QFileInfo::exists(tq(filePath)))
```
JorisGoosen, line 28: This is already checked before calling `loadLocalJson` ?
Anyway, I guess it can't hurt to have an extra check...

QMLComponents/knownissues.cpp
```diff
@@ -64,19 +68,27 @@ void KnownIssues::loadJson(const Json::Value & json, bool saveIt)
...
 		if(!json.isObject()) throw std::runtime_error("expected issues json to be an object");
 
-		const std::string version = AppInfo::version.asString();
-
-		if(json.isMember(version))
-			for(		const std::string & module		: json[version]						.getMemberNames())
-				for(	const std::string & analysis	: json[version][module]				.getMemberNames())
+		const std::string jaspVersion = AppInfo::version.asString();
+
+		if(json.isMember(jaspVersion))
```
JorisGoosen, line 73: This implies that we would have to add known issues for every single module with issues for every single version of JASP we release...

I think it would be better to just put as a "JASP Version" something like "ModuleLibrary" instead of a version number.
And then every JASP from now on just puts the module specific errors in there.
And then we could still have a specific per JASP version error?

On the other hand, I guess this is temporary anyway, so we could just leave it like this. Its not like we are constantly updating known issues

QMLComponents/knownissues.cpp
```diff
@@ -64,19 +68,27 @@ void KnownIssues::loadJson(const Json::Value & json, bool saveIt)
...
+		{
+			const Json::Value & modulesWithKnownIssues = json[jaspVersion];
```
JorisGoosen, line 75: Really loving the fact that you felt the need to rename everything...

QMLComponents/knownissues.cpp
```diff
@@ -138,34 +150,37 @@ void KnownIssues::addIssue(const std::string & module, const std::string & analy
 	default:				Log::log() << "KnownIssues::addIssue got unexpected type for \"options\", so ignoring it." << std::endl;
 	}
 
-	_issues[module][analysis].push_back(newIssue);
+	_issues[std::make_pair(module, version)][analysis].push_back(newIssue);
```
JorisGoosen: Why not just put an extra map in there :'(

QMLComponents/knownissues.h
```diff
@@ -18,12 +19,12 @@ class KnownIssues : public QObject
...
-		std::string	info;
-		stringset	options;
+		std::string		info;
+		stringset		options;
 	};
 
-	typedef std::map<std::string, std::vector<issue>>	issuesPerAnalysis;
-	typedef std::map<std::string, issuesPerAnalysis>	issuesPerModule;
+	typedef std::map<std::string, std::vector<issue>>						issuesPerAnalysis;
+	typedef std::map<std::pair<std::string, Version>, issuesPerAnalysis>	issuesPerModule;
```
JorisGoosen:
```suggestion
	typedef std::map<std::string, std::map<Version, issuesPerAnalysis>>	issuesPerModule;
```

This seems to make a bit more sensen to me... Why use a pair?

QMLComponents/knownissues.cpp
```diff
@@ -20,10 +20,14 @@ KnownIssues::KnownIssues(QObject * parent) : QObject(parent)
 {
 	assert(!_knownIssues);
 	_knownIssues = this;
+	loadKnownJson();
```
JorisGoosen, line 23: The check whether the file already exists couldve just been here, but whatever

JorisGoosen (changes requested):
I think you should just use a map instead of having a pair as an indexer.

I also think the "_optionsFromDifferentVersion" is superfluous.

Desktop/analysis/analysis.cpp
boutinb: The wasUpgraded() function is used in the code, and I did not want to take the risk to change the behavior by setting _wasUpgrade to true each time an old JASP file is loaded.

JorisGoosen: Safety first :D

But if you look for where it gets used you'll see that this is exactly what it is meant for though ^^
So you can safely use it for that.

QMLComponents/knownissues.cpp
boutinb: Yes I was hesitating here. Some issues will be independent of the JASP version, other issues might be solved by an update of QML Controls (so depending of the JASP version).
But as noticed, this is a temporary fix. KnownIssues are module specific, so seems more logical to be set in the GitHub repo of the module, and even better in a QML format so that we can translate it.

JorisGoosen: But if you are pulling the KnownIssues.qml  from a repo directly you arent getting the translations...
Because the point of this is to tell users about errors in JASP directly that wont really work.

I think for the future we could just have the module library tell JASP that there is a problem. And anyway, we can just push updates for the broken modules :)

JorisGoosen: Well, not push, but make available for pull

Desktop/analysis/analysis.cpp
boutinb: Hmm, if I read this PR https://github.com/jasp-stats/jasp-desktop/pull/4544 with all the comments, I'm not so sure.
Also here https://github.com/jasp-stats/jasp-desktop/blob/62b061009a8a125d448de1699829e33c7b960d05/QMLComponents/boundcontrols/boundcontrolrlangtextarea.cpp#L197 for example: does this have really no effect if wasUpgraded is true each time an old JASP file is loaded?
I know I'm not proud to have such a code which makes me shaky when changing it...

QMLComponents/knownissues.cpp
boutinb: > But if you are pulling the KnownIssues.qml from a repo directly you arent getting the translations...

Oh yes! Stupid remark from me!

Desktop/analysis/analysis.cpp
JorisGoosen: Fair enough, keep it like this then :)

JorisGoosen (approved):
Ca marche ^^

PR #6141, opened 2026-02-24
fix crash during import inconsistent row count under msvc

shun2wang:
These line will trigger assertion of `std::vector::operator[]` under **MSVC debug build**, because of different row count.

```cpp
			_doubleTroubleBinder(	stmt,	i++, col->dbls()[rowOutside]);
			sqlite3_bind_int(		stmt,	i++, col->ints()[rowOutside]);
```

Test case: importing this minitab file in JASP: https://support.minitab.com/en-us/datasets/media/generated-content/datasets/ToothpasteSales.MWX


JorisGoosen:
But isnt there then something wrong with the minitab code determining how many rows there are?

shun2wang:
OK, see: https://github.com/jasp-stats/jasp-desktop/pull/6142

PR #6142, opened 2026-02-25
enh: also read column description from minitab and fix crash under msvc

shun2wang:
- fix crash during import inconsistent row count under msvc
- also read column description from minitab


boutinb (approved):

Desktop/data/importers/minitab/mwx.cpp
```diff
@@ -86,6 +90,11 @@ void Minitab::getColumns(std::vector<MwxImportColumn *> &columns, ImportDataSet
 				impCol->addValue(val.isNull() ? "" : ColumnUtils::doubleToStringMaxPrec(val.asDouble(), false));
 		}
 
+		while (impCol->size() < _numRows)
+		{
+				impCol->addValue("");  //filling to keep col length consistently
```
JorisGoosen, line 95: So minitab just has random columns with less entries?

shun2wang: yes.

PR #6146, opened 2026-02-26
Allow link in Module Library

boutinb:
In the Module Library, a module may have a link to a home page (cf Audit module).
This link does not work because the WebEngineView does not allow this per default.


shun2wang (approved):

RensDofferhoff (approved):
Great!

PR #6148, opened 2026-02-27
In computed column, do not show a CheckBox for a boolean parameter when no boolean is given

boutinb:
Fixes https://github.com/jasp-stats/jasp-test-release/issues/3186


RensDofferhoff (approved):
seems fine

PR #6156, opened 2026-03-06
Make standalone MSIX default installer + Core Modules Only + bundle update action

### LINKED ISSUE #3390: [Feature Request]: MSIX directly downloadable from https://jasp-stats.org/download/
### Description

Make MSIX downloadable without Microsoft account

### Purpose

To lower the bar of using the MSIX over the problematic MSI installer

### Use-case

MSI does not work and one does not want to use MS-Store

problematic MSI installation

Unrelated

MSIX download button on top of the MSI download button

MSIX nightlies are arleady available directly, but that page is kind of hidden

### Additional context

_No response_

### LINKED ISSUE #4175: [Bug]: Jasp 0.96.0 - Nested manifest folders
### JASP Version

0.96.0

### Commit ID

_No response_

### JASP Module

Unrelated

_No response_

Windows 11

### Bug Description

Upon installing JASP 0.96.0 on Windows 11 Enterprise 25H2 PCs, we have noticed a possible error with the way JASP writes data to the directory below C:\Program Files\JASP\Modules\manifests\MANIFE~1

It creates multiple MANIFE~1 folders nested within each other, and creates multiple copies of the .json files within each folder.

Image

 

### Expected Behaviour

We would presume it would only create one copy of MANIFE~1 and the .json files ?

### Steps to Reproduce

Installed JASP on a Windows 11 Enterprise PC.

### Log (if any)

n/a

### More Debug Information

n/a

### Final Checklist

RensDofferhoff:
Only build with core modules
update remote bundles every day
remove msi and move to MSIX only future
MSI creates junction in appdata like MSIX to get rid of future installer troubles
faster junction creation by not booting up R

Solves not being able to open Log files from advanced menu.
Solves some settings not sticking when updating via MSIX

Helps Admins enforce Policies


QMLComponents/utilities/appdirs.cpp
```diff
@@ -115,15 +115,15 @@ QString AppDirs::documents()
...
+    QDir res(AppDirs::documents());
 	res.mkdir(name);
 	res.cd(name);
 	return res.absolutePath();
 }
 
 QString AppDirs::logDir()	
 {
-	QString path = appData();
+    QString path = sandboxedDocuments();
```
RensDofferhoff, line 126: Cant reference appdata folder to external program like file browser. 
The windows msix environment translates this to contained folder and not actual Settings folder Qt thinks it is writing to. 
So windows explorer opens non existing path.
Throw it in the sandbox folder we have to create anyhow

Desktop/main.cpp
```diff
@@ -456,6 +456,8 @@ int main(int argc, char *argv[])
 		QCoreApplication::setOrganizationDomain("jasp-stats.org");
 		QCoreApplication::setApplicationName("JASP");
 	}
+	QSettings::setDefaultFormat(QSettings::IniFormat);
```
RensDofferhoff, line 433: Writing into registry is cursed so lets just write an ini file like civilized people

RensDofferhoff:
In future MSI will just throw junction in appdata the same as MSIX

Desktop/main.cpp
JorisGoosen: This effectively resets everyones settings everywhere right?
Although I guess switching from msi to msix will do that anyway

JorisGoosen (commented):
Code looks great, also real nice to have windows policies queried in there.
Let me just make some testbuilds to see how it looks in real life

JorisGoosen:
Actually, Ill just try https://static.jasp-stats.org/Nightlies/WindowsMSIX/JASP-nightly-msixSwitch-902dd047762f8f2d151c4e6d6d33a672110c15f7.msix

shun2wang:
For individual users, the difference between .msix and .msi is minimal and poses no significant obstacle. The main impact lies in enterprise-level batch silent deployments, such as support for Windows Group Policy installation methods, which may depend on the preferences of enterprise IT personnel. I suggest we prepare some instructions and solutions for batch deployments on enterprise platforms; I anticipate we'll be asked about these things by enterprise users someday.

JorisGoosen:
That sounds like a great idea!

Do you know anything about such batch deployments?
I can read up of course, but I've never done much windows admin stuff.

JorisGoosen:
The MISX works fine, but we should maybe work on the errormessage we get when you load a file with a missing module: 
image
We could at least open the module menu and show the module or something

shun2wang:
Here are some comparison: https://www.advancedinstaller.com/msix-limitations.html

It is actually very dependent on the OS version used by enterprise end users.

JorisGoosen:
You think it's a bad idea to drop msi then?

So far companies we heard from did use the msi, but previously the msix was rather annoying.
Now that we have a key I think its not so bad.

shun2wang:
To be honest, I'm a little worried because enterprise IT engineers don't seem to all like "embracing new technologies." Instead, they prioritize stability and convenience. My concern is that during the solution selection phase before make a decision, they might give lower scores to technologies due to bundled packages (for example, mistakenly assuming that .msix relies on the Microsoft Store, a feature most enterprises have disabled).

In fact, once they decide to use .msix, large-scale deployment isn't difficult; JASP can provide PowerShell scripts for deployment, etc.

Sorry, in short, I think the current approach should be coexistence. We can hide or weaken the .msi installation option and primarily promote .msix, so that we won't be passive when encountering enterprise users who prefer .msi.

EJWagenmakers:
I agree with Shun, let's not close any doors unless we have to

JorisGoosen:
I dont think we have to, it looks to me that the changes to remove msi are not required for the other stuff to work.
I can see about splitting that off

RensDofferhoff:
We can still make MSIs.  A new toolchain has been developed to do so from ZIP. The current installer is a hot mess and building it takes lots of time. Also its WIX version is ancient and that will cause issues in the future. 
The new MSI is build with newest WIX version

RensDofferhoff:
Admin features are being incorporated in the new MSI tool

Desktop/junction_tool/main.cpp
```diff
@@ -0,0 +1,279 @@
...
+    DWORD bytesReturned;
+    return DeviceIoControl(hFile, FSCTL_SET_REPARSE_POINT, reparseData, 
+                           reparseData->ReparseDataLength + 8, NULL, 0, &bytesReturned, NULL);
+}
+
+// --- APP MODES ---
+void RunScan(const fs::path& baseDir, bool deleteAfterScan) {
+    std::ofstream outFile("junctions_map.txt");
+    if (!outFile) {
+        std::cerr << "[ERROR] Could not create output file.\n";
```
shun2wang, line 120: Should these errors will be recoded in .logs or just print them?

RensDofferhoff: These are shown during the build process

JorisGoosen (approved):
Looks fine to me, maybe the merge conflicts can be fixed?
After that Im ok with merging this

PR #6158, opened 2026-03-09
Use JASP.Module in JASP R Syntax

boutinb:
JASP Syntax has functions that generate wrappers. For this it has to read the Description.qml file. But this needs the JASP.Module module. Until now the Description.qml was parsed manually, which is was quite unhandy.

To be able to use the JASP.Modules module, all the JASP Modules files are moved to the QMLComponent folder. Only the DynamicModules class stays in Desktop. The initialization of JASP.Modules is done in QmlUtils, so that JASP Syntax can call this.

I have also removed the Upgrades.json file, that was needed only for really old JASP files (<0.17.0): I think it's time to remove this, and the extra code needed to handle such old files.

The jaspSyntax has been already updated.

To test jaspSyntax with this new code, JASP needs to be compiled with the static Qt libraries, and by adding these 2 variables in configure:
```
JASP_SOURCE_DIR="<path to your JASP source files>"
JASP_BUILD_DIR="<path to your static Qt build>"
```
Then install JASPSyntax with `install.packages("<path to JASPSyntax module source>", repos = NULL, type = "source")`

But once this PR is merged, I will also change the libSyntaxInterface.dylib and dyntaxInterface.dll files in the static server (and the Hash code in the configure file of JASPSyntax), so that you can install directly JASPSyntax via `remotes::install_github("jasp-stats/jaspSyntax")`

To test if the wrappers are generated, use generateModuleWrappers or generateAnalysisWrapper like this: `jaspSyntax::generateModuleWrappers("<Path to a JASP module sources, like jaspTTests>")`: this will generate all wrappers of jaspTTests.

The goal is that these wrappers will be automatically generated with a GitHub action. Also roxygen2::roxygenise() should be called automatically to generate the rd files in the `man` folder: all the generated help files will be then available.


JorisGoosen:
Can you then change also the minimum version of jasp jaspimporter will load?

In the old importer

boutinb:
`jaspSyntax::parseDescription(<module path>)` has been also added.

JorisGoosen (approved):

PR #6163, opened 2026-03-18
chore: don't reconfigure dependencies every time

shun2wang:
On Windows cmake will reconfigure JAGS and freexl every time because '_deps' dir was removed after run conan command.The downside of doing this is that when switching branches locally or running configuration commands, everything has to be downloaded and rebuilt, increasing configuration time.

I'm not sure the impact was but we set that 10 months ago . maybe @RensDofferhoff know?

### DIFF
--- a/Tools/CMake/Conan.cmake
+++ b/Tools/CMake/Conan.cmake
@@ -56,7 +56,6 @@ if(USE_CONAN)
     else()
       message(CHECK_FAIL "build freexl failed")
     endif()
-    file(REMOVE_RECURSE ${CMAKE_BINARY_DIR}/_deps/)
     
     execute_process(
       COMMAND_ECHO STDOUT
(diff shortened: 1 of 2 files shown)


JorisGoosen:
Maybe we should add it to some sort of clean command though?

shun2wang:
CMake has its own build clean commands, doesn't it?

 cmake --build --target clean
 cmake --build --clean-first

for manually in Qt creator we can click the menu buttons of clean.

JorisGoosen:
Yes, but they don't clean `_deps`...

JorisGoosen:
Alright, now when you run clean you can actually rebuild JASP afterwards and everything is fresh then.

JorisGoosen:
I think Rens added the remove because of https://github.com/jasp-stats/jasp-desktop/pull/5909#issuecomment-2929043186
We now have everything working via conan right?

shun2wang:
> We now have everything working via conan right?

no, but librdata and readstat (Windows only) are not from conan.

JorisGoosen:
So I guess I should test this some more on windows

JorisGoosen (approved):

PR #6164, opened 2026-03-18
Look at the modules requested before trying to load all analyses

JorisGoosen:
Send out a signal that is caught in ModulesMenu.qml to show and select the missing modules That last part is still missing however

Belongs with https://github.com/jasp-stats/INTERNAL-jasp/issues/3159 to try and improve the handling of missing modules

### DIFF
--- a/Desktop/analysis/analyses.cpp
+++ b/Desktop/analysis/analyses.cpp
@@ -410,7 +410,7 @@ void Analyses::setAnalysesUserData(Json::Value userData)
 
 void Analyses::loadAnalysesFromDatasetPackage(bool & errorFound, stringstream & errorMsg, RibbonModel * ribbonModel)
 {
-	if (DataSetPackage::pkg()->hasAnalyses())
+	if (DataSetPackage::pkg()->hasAnalysesData())
 	{
 		Json::Value analysesData = DataSetPackage::pkg()->analysesData();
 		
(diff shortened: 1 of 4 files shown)


JorisGoosen:
@sverhoeven is there some easy way now to do a search via `js` ?

To use the new search you fixed in https://github.com/jasp-stats/INTERNAL-jasp/issues/3159

sverhoeven:
> @sverhoeven is there some easy way now to do a search via `js` ?
> 
> To use the new search you fixed in [jasp-stats/INTERNAL-jasp#3159](https://github.com/jasp-stats/INTERNAL-jasp/issues/3159)

I assume you already know which modules where used in the analysis. So you could fetch https://module-library.jasp-stats.org/index.json to get a list of all available modules and filter on their id for example jaspBain.

JorisGoosen:
Well, but I just want to show the user the module library and the analyses pre-searched for them.
Then they can choose what to install, or update.

This is already a lot more friendly then just dropping the analyses and showing a lot of warnings. Which is what we do now. And autoinstalling stuff seems a bit impractical.

sverhoeven:
Ah I see, you can add set the search field with

```js
function selectModules(moduleIds) {
  const searchParam = moduleIds.map(id => `id:${id}`).join(' OR ');
  const url = new URL(window.location);
  url.searchParams.set('s', searchParam);
  window.history.replaceState({}, '', url);
  window.location.reload(); // needed otherwise React nuqs does not pick it up
}
selectModules(['jaspBain', 'jaspAnova']);
```

After that the module library would look like
image




If you do not like the reload then I can create a custom event in the module-app.

sverhoeven:
Using an empty array will clear the search field.

tomtomme:
Why is auto-installing impractical? Would be gread IMHO - like firefox also updates itself silently in the background. Of course this should be "opt out" or "opt-in" via the options or via the questions JASPs asks (to check for updates) after the first time of install.  Even beta-auto-updates could be opt in.

PR #6173, opened 2026-04-02
fix: suppress UI Automation events during form creation to prevent slowdown

FBartos:
## Bug

Addresses [INTERNAL-jasp#3136](https://github.com/jasp-stats/INTERNAL-jasp/issues/3136) — **On some Windows 11 machines, duplicating or opening multiple analyses causes JASP to freeze for 10–30+ seconds.**

Affected ~15% of Windows workshop participants. The 3rd analysis took ~19 seconds to open. Each subsequent analysis was progressively slower.

## Root Cause

**Power Automate Desktop** — pre-installed on Windows 11 — runs a background UI Automation server (`PAD.AutomationServer.exe`) that intercepts every text-change accessibility event from all applications. Each `setText()` on a QML TextField triggered a cross-process UIA notification through the Windows kernel, adding **~65ms overhead per call**. With 35+ text fields per analysis form and the cost scaling with total QML objects, form creation ground to a halt.

**How we found it** (diagnostic journey):
1. Performance instrumentation traced the bottleneck to `setProperty("displayValue", ...)` in `TextInputBase::setDisplayValue()` — a single Qt property write taking 65–400ms
2. Ruled out: Qt render loop, accessibility env var, GPU backend (D3D11/OpenGL/software), DPI scaling, QML alias overhead — all had zero effect
3. Key breakthrough: running JASP as **Administrator** made it instant (UIPI blocks UIA events from elevated processes to non-elevated UIA clients)
4. CPU profiling (ETW/WPR) revealed **71% of CPU time in kernel** (`ntoskrnl.exe`) and `PAD.AutomationServer.exe` consuming massive kernel time via `UIAutomationCore.dll`
5. Killing PAD processes confirmed: instant form creation without PAD, 19-second freezes with PAD

## Fix

**1 file changed, 14 lines added** (10 of which are the comment).

Installs a no-op `QAccessible` update handler for the duration of `Analysis::createForm()`, suppressing all accessibility notifications during bulk QML object construction. The previous handler is restored immediately after, so accessibility works normally during regular user interaction.

```cpp
auto previousHandler = QAccessible::installUpdateHandler([](QAccessibleEvent*) {});
AnalysisBase::createForm(parentItem);
QAccessible::installUpdateHandler(previousHandler);
```

- Uses Qt's **public API** (`QAccessible::installUpdateHandler`)
- **No platform-specific code** (`#ifdef`) — works cross-platform
- **No form lifecycle changes** — forms are not destroyed/recreated
- **No functionality reduction** — all controls, validations, and features remain
- **No side effects on unaffected machines** — suppression window is ~200ms during which the form isn't visible anyway

## Performance Results

With Power Automate Desktop running:

| Analysis # | Before fix | After fix |
|---|---|---|
| 1st | 280ms | 208ms |
| 2nd | 3,249ms | **173ms** |
| 3rd | 18,837ms | **179ms** |
| 4th | — | **168ms** |

## What to Test

1. **Basic**: open + duplicate ClassicalMetaAnalysis (or any analysis) 3+ times — should be consistently fast
2. **Options preservation**: set options, duplicate, verify options are correct in the duplicate
3. **Screen reader compatibility**: verify JASP remains accessible with a screen reader after form creation completes
4. **Cross-platform**: verify no regression on macOS and Linux

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Desktop/analysis/analysis.cpp
+++ b/Desktop/analysis/analysis.cpp
@@ -32,6 +32,7 @@
 #include "modules/description/description.h"
 #include "gui/jaspConfiguration/jaspconfiguration.h"
 #include <QAccessible>
+#include <QScopeGuard>
 
 Analysis::Analysis(size_t id, Modules::AnalysisEntry * analysisEntry, const std::string & title, const Version & optionsVersion, const Json::Value & options) :
 	  AnalysisBase(Analyses::analyses()),
@@ -370,10 +371,9 @@ void Analysis::createForm(QQuickItem* parentItem)
 	// these notifications. The handler is restored immediately afterwards so that
 	// accessibility works normally during regular user interaction.
 	auto previousAccessibilityUpdateHandler = QAccessible::installUpdateHandler([](QAccessibleEvent*) {});
+	auto restoreAccessibilityHandler = qScopeGuard([&]() { QAccessible::installUpdateHandler(previousAccessibilityUpdateHandler); });
 
 	AnalysisBase::createForm(parentItem);
-
-	QAccessible::installUpdateHandler(previousAccessibilityUpdateHandler);
 
 	if (_analysisForm)
 	{
--- a/Desktop/analysis/analysis.cpp
+++ b/Desktop/analysis/analysis.cpp
@@ -31,6 +31,7 @@
 #include "utilities/messageforwarder.h"
 #include "modules/description/description.h"
 #include "gui/jaspConfiguration/jaspconfiguration.h"
+#include <QAccessible>
 
 Analysis::Analysis(size_t id, Modules::AnalysisEntry * analysisEntry, const std::string & title, const Version & optionsVersion, const Json::Value & options) :
 	  AnalysisBase(Analyses::analyses()),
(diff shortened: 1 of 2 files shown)


Desktop/analysis/analysis.cpp
```diff
@@ -359,6 +361,18 @@ Analysis::Status Analysis::parseStatus(std::string name)
...
+	// Suppress accessibility (UI Automation) events during bulk form creation.
+	// On Windows 11, background UIA clients such as Power Automate Desktop intercept
+	// every text-change event via the OS accessibility infrastructure. Each setText()
+	// on a QML TextField triggers a cross-process UIA notification through the kernel
+	// (~65 ms overhead per call). With 35+ text fields per analysis form this causes
+	// multi-second freezes that grow with the number of open analyses.
+	// Installing a no-op update handler for the duration of form creation suppresses
+	// these notifications. The handler is restored immediately afterwards so that
+	// accessibility works normally during regular user interaction.
+	auto previousAccessibilityUpdateHandler = QAccessible::installUpdateHandler([](QAccessibleEvent*) {});
```
JorisGoosen, line 373: I do not see `QAccessible::installUpdateHandler` in https://doc.qt.io/qt-6/qaccessible-members.html
Perhaps its called different in Qt 6.11?
Also not in Qt 6.10.

Ill try to build it though, maybe the online docs are confusing.

JorisGoosen: Alright, it's just an undocumented function. It is there.

Now to see whether we still have accessibility features in JASP.

JorisGoosen: Ill build a nightly for @FBartos to test.

FBartos: The API is present in Qt 6.10 — it's declared in qaccessible_base.h:402 (included via <QAccessible>), not in the   qaccessible.h that the online docs are generated from. The online member listing seems to have a gap here. It    
  compiles and works correctly on Qt 6.10.2, and the build on this PR confirms it. The function has existed since Qt
   5 and is still in Qt 6's public headers alongside the related installFactory() and installRootObjectHandler().

JorisGoosen (approved):

PR #6176, opened 2026-04-06
fix showing long password in tooltip and add eye button

shun2wang:
This fixed a long password showing in tooltip while inputing, that was not privacy safety:

image

And add a eye icon to show/hide the password for user inputing.


JorisGoosen (approved):
Very nice!

PR #6181, opened 2026-04-08
Add functions in syntaxbridge so that jaspSyntax can read JASP file

boutinb:
jaspSyntax should be able to read JASP files in order to load its database, and also to get the options of an analysis in this JASP file.

When this PR is merged, I will then update the jaspSyntax so that it can use these functions.

### DIFF
--- a/QMLComponents/datasetprovider.cpp
+++ b/QMLComponents/datasetprovider.cpp
@@ -57,7 +57,7 @@ void DataSetProvider::resetDataSet()
 		_dataSet->dbDelete();
 		delete _dataSet;
 	}
-	
+
 	_dataSet = new DataSet();
 }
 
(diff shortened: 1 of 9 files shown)


JorisGoosen:
There is a small merge conflict 😇

boutinb:
> There is a small merge conflict 😇

Solved

QMLComponents/datasetprovider.cpp
```diff
@@ -113,6 +113,20 @@ void DataSetProvider::loadDataSet(const std::map<std::string, stringvec > & data
...
 
+void DataSetProvider::loadDatabase(const Version & jaspVersion)
+{
+	delete _dataSet;
+
+	_db->close();
+	_db->load();
+	_db->upgradeDBFromVersion(jaspVersion);
+
+	_dataSet = new DataSet(0);
```
JorisGoosen:
```suggestion
	_dataSet = new DataSet(0); // Setting 0 for "do nothing" because otherwise we can't pass on jaspVersion
```

JorisGoosen (approved):

PR #6198, opened 2026-04-28
Add a CSV preview window

### LINKED ISSUE #3608: [Bug]: values from csv-files with semicolon separator (where commas are used for separation of fractional part of numbers) are uploaded incorrectly
### JASP Version

0.95.0

### Commit ID

_No response_

### JASP Module

Unrelated

opening file

Windows 10

### Bug Description

When I open the csv-file with semicolon separator (where commas are used for separation of fractional part of numbers) the file is opened but some values are read incorrectely. For example, the value "0,435" became "435" instead of "0.435".
The same file was read correctly in previous version of JASP (see screenshot).

The file in example contains 6 values, which must be read as:

variable Valence:
0.435
1.286
0.118

variable Arousal:
2.87
4
2.059

But they are read as:

variable Valence:
435
1,286
118

variable Arousal:
2.87
4
2,059

That is the values in JASP 0.95.0 became wrong.

Image

[!problem_Open_csv_file.csv](https://github.com/user-attachments/files/21815709/problem_Open_csv_file.csv)

### Expected Behaviour

The values in example file should be read as:

variable Valence:
0.435
1.286
0.118

variable Arousal:
2.87
4
2.059

### Steps to Reproduce

Open -> Computer -> file from attachement

### Log (if any)

_No response_

### More Debug Information

_No response_

### Final Checklist

JorisGoosen:
Fixes jasp-stats/jasp-issues#3608
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/3176
image
You can also change the locale in it, but the screenshot is old.


shun2wang:
Maybe move Windows-BOM things https://github.com/jasp-stats/jasp-desktop/blob/7363a9f091a8193f8894795d83c92f0d7a28411c/Desktop/components/JASP/Widgets/FileMenu/PrefsData.qml#L276 in this preview.

JorisGoosen:
> Maybe move Windows-BOM things
> 
> https://github.com/jasp-stats/jasp-desktop/blob/7363a9f091a8193f8894795d83c92f0d7a28411c/Desktop/components/JASP/Widgets/FileMenu/PrefsData.qml#L276
> in this preview.

I think this might be something for another time. It would complicate the flow a bit. (As it needs to use the bom before it sends the rawData to the previewmodel)

boutinb:
Claude Overview (my remarks in italic)                                                                                                                                                  
                                                                                                                                                              
  This PR adds a modal dialog that lets users preview and choose the CSV delimiter before loading. It introduces CsvPreviewModel (C++/QML), wires it through DesktopCommunicator with the same blocking condition-variable pattern used for encryption dialogs, and persists the chosen delimiter to the JASP SQLite database so it survives sync. The language-preferences panel is also refactored out of PrefsUI.qml into a reusable PrefsLanguage component.                 

  ---
  Bugs / Correctness

  Critical

  - deEuropeaniseForImport removed from columnutils.cpp — This function converted European decimal notation (e.g. 1.234,56 → 1234.56) during CSV import.      
  Removing it silently breaks import of European-locale CSV files. The advanced language panel in the dialog doesn't substitute for this automatic fallback:
  users who don't open the advanced panel will get silent data corruption on European CSVs. This needs to be restored.

_I tend to agree with Claude: of course the user should take care of having the right locale to read the csv file, but users are lazy or do not pay attention of these locale issues. And I don't see why a string like "20,3" should not be converted to a decimal number anyway: what would be the problem?_
                                        
  - wait(120) in asyncloaderthread.cpp — Changed from wait() (indefinitely) to wait(120) (120 ms). On a slow machine or a large file, the loader thread won't finish in 120 ms and the destructor will continue while the thread is still running, causing a crash or data corruption. Should be reverted to wait() or at minimum use several seconds (e.g., wait(30000)).

_Here apparently Claude did not see that this wait is only used in the destructor of the Mainwindow. So fortunately we are still sometimes more clever than him!_  

  - readLineRaw() silently truncates at buffer boundary — If a line spans multiple read buffers (long header rows, wide CSVs), the function returns "" instead of reading more data. The preview will then stop at that line and show fewer rows than expected, or stop entirely. The comment in the code acknowledges this. This should be fixed to properly handle multi-buffer lines, or at minimum emit a warning.

_I can reproduce this. I'll ask him to fix it_

  - catch(...) fallback uses , not the auto-detected delimiter — In csvimporter.cpp:                                                                          
  try {                                                                                                                                                       
      delimiter = DesktopCommunicator::singleton()->askCsvDelimiter(...);                                                                                     
  } catch (...) {                                                                                                                                             
      delimiter = ',';   // ← ignores the auto-detected delimiter                                                                                             
  }                                                                                                                                                           
  -  If an exception occurs, delimiter should fall back to the value it had before the try, not hardcoded ,.

_Definitely a good point. I'll change this_

  - Tests will fail for non-numeric values — CsvPreviewModel::data() wraps non-numeric strings in double quotes ("\"" + val + "\"") but the tests assert raw strings like QString("Col1") and QString("Val1"). These assertions will fail. Either the quote-wrapping should not apply to header display, or the tests need to be updated to match the actual output.                                                                                                              
 
_The double quote for column names are quite annoying. I'll remove them._
                                                                                                                                                              
  Moderate        

  - Preview data values wrapped in "..." is misleading — Displaying "Val1" in the preview when the imported value is Val1 misrepresents what will actually load. Consider not wrapping strings in the preview, or using visual styling (e.g. colour) to distinguish string vs. numeric cells instead.

_Maybe we could use italic for strings instead of adding these double-quotes_

  - firstRowsPlease() reset was incomplete (now fixed in this branch) — The original PR reset 4 of 6 buffer/stream state fields, leaving _utf8BufferEndPos, _rawBufferEndPos, _filePosition and the stream position in a partially-read state. For BOM-encoded files (common on Windows/Excel) this caused BOM bytes to appear as part of the first column name token, preventing the if (colName == "") check from firing, which is why V1/V2 auto-naming broke. The fix (calling
  readRaw(), determineEncoding(), readUtf8(), determineDelimiters() after reset) has been applied in this branch.                                             

_Yes, this is in my last commit_

  ---
  Code Quality
                                                                                                                                                              
  - Entire csv.cpp reformatted tabs → spaces — This creates ~700 lines of noise in the diff that have no functional effect. It makes it very hard to see the real changes and should be reverted. If a formatting change is wanted, it should be a separate commit.

_I'll reformat the file_
                                                      
  - Copyright header modified incorrectly — csv.cpp and csvimporter.cpp changed "free software: you can redistribute" to "free to redistribute". This alters the standard GPL license header text, which is legally incorrect and should be reverted.

_Funny change @JorisGoosen !_
                                                            
  - Debug no-ops left in production code — Both askCsvDelimiter and queryEncryptionSettings contain:
 ```
 if(QThread::currentThread() == qApp->thread()) {                                                                                                            
      int x = 1;  // ← does nothing, debugging leftover
  }     
```                                                                                                                                                      
 This should be removed or replaced with an assertion or Log::log().

_OK, I'll remove it_

  - Typo in log message — "Loading CSV Previeww" (double w) in mainwindow.cpp.                                                                                
  - Include guard mismatch — csvpreviewmodel.h uses #ifndef CsvPreviewModel_H but project convention is ALL_CAPS_H (e.g. CSVPREVIEWMODEL_H).                  
  - public slots: indentation — Extra indentation before the public slots: label in csvpreviewmodel.h.

_OK, I'll change all that_
                                                     
  - columnCount() is O(n) per call — It iterates all rows every time QML requests the column count. Cache _maxCols and update it in updateInternalStructure().

_No cache please!_

  - selectDelimiter(char) is dead code — Defined on CsvPreviewModel but never called (QML sets delimiter directly via property). Remove it or make it actually used.

_I'll remove it_
                                                                                                                                                   
  - Duplicate TESTLIBRARY_DIR definition in CMakeLists.txt — target_compile_definitions(JASPTest PUBLIC TESTLIBRARY_DIR=...) is added twice (lines 2849–2851  
  and 2853–2857 in the diff).

_I'll remove it_
                                                                                                                              
  - <fstream> and <sstream> included in csvimporter.cpp without being used. Remove.

_I'll remove it_
                                                                                                                                                   
  ---             
  QML                                                                                                                                                         
                                                                                                                                                              
  - Hardcoded height: 50 * jaspTheme.uiScale — The delimiter toolbar height is fixed. Using implicitHeight from the RowLayout inside would be more robust.

_I'll ask Claude to do this_

  - Delimiter list includes '.' — Period is a common decimal separator. Including it as a delimiter option will confuse users who try to use it on European-format files where . is the thousands separator. 

_Yes, but maybe other countries might use it_

  - No visual highlight for the active delimiter — The current delimiter button is enabled: false (greyed out) to indicate selection. This is unintuitive. The active delimiter should instead be highlighted (e.g. different background/border), and all buttons should remain enabled: true.

_That's quite a good point. I'll ask him to make it better_
                            
  - DropDown.qml csvPreviewWindow reference — The maxHeight calculation references csvPreviewWindow by id. This only works if the DropDown is inside a parent
  that has this id in scope. This is a fragile coupling that will break if the QML structure changes.                                                         

_I know, but this PR is big enough now_

  ---                                                                                                                                                         
  Tests           
       
  - testcsvpreviewmodel.cpp missing newline at EOF — Minor but will cause compiler warnings.
  - No test for V1/V2 auto-naming — There's no test that an empty column name in the CSV becomes "V1" in the preview.

_The debug.csv test it_
            
  - No test for BOM handling — The BOM fix is not covered.
  - No test for delimiter auto-detection — The csv.delimiter() API is not exercised in tests.

_I'll ask Claude to make these tests_
                   
  - preparePreview signature takes const char* but the internal rawData is QString — The slot signature in csvpreviewmodel.h takes const QString&, so calling with rawData.toStdString().c_str() in the test creates an unnecessary conversion. Minor.                                                                    
                                                                                                                                                              
  ---                                                                                                                                                         
  Summary         
         
  The core idea is solid and well-architected — the DesktopCommunicator blocking pattern is a clean fit. The main blockers before merge are: the
  deEuropeaniseForImport regression (data corruption for European locales), the wait(120) crash risk, the broken tests, and the noisy formatting diff. The    
  readLineRaw() buffer-boundary truncation is also worth addressing before ship.

JorisGoosen:
> I tend to agree with Claude: of course the user should take care of having the right locale to read the csv file, but users are lazy or do not pay attention of these locale issues. And I don't see why a string like "20,3" should not be converted to a decimal number anyway: what would be the problem?

This would be the problem: https://github.com/jasp-stats/INTERNAL-jasp/issues/3176

JorisGoosen:
So were are the fixes that claude made during your conversation?

PR #6212, opened 2026-05-08
Repair JAGS model

boutinb:
Applying a JAGS model triggers an error.
https://github.com/jasp-stats/jasp-desktop/pull/6170 causes the problem. It sets the bound value directly as if it is the model string. But the bound value of JAGS or R TextArea are complex: it's an object with the model string, but also with the columns used, the model with encoded columns, etc.
So use rather the checkSyntax function instead, that takes care that the right bound value is made.


boutinb:
Claude agrees

PR #6220, opened 2026-05-12
enh: use zoomFactor to scale results page and disable js solutions for now

shun2wang:
Fix: https://github.com/jasp-stats/INTERNAL-jasp/issues/3214

To use js/css zoom the page is not the same behavior with zoom by browser(zoomFactor).

### DIFF
--- a/Desktop/components/JASP/Widgets/MainPage.qml
+++ b/Desktop/components/JASP/Widgets/MainPage.qml
@@ -318,6 +318,8 @@ Item
 
 			url:					resultsJsInterface.resultsPageUrl
 
+			zoomFactor:				resultsJsInterface.zoom
+
 			onContextMenuRequested: (request) => request.accepted = true
 
 			backgroundColor:		jaspTheme.uiBackground
--- a/Desktop/html/js/main.js
+++ b/Desktop/html/js/main.js
@@ -39,7 +39,8 @@ $(document).ready(function () {
 
 	analysesGlobal 			= analyses
 
-	window.setZoom			= function (zoom)			{ document.body.style.zoom = "" + Math.floor(zoom * 100) + "%";	}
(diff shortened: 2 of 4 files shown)


JorisGoosen:
This pr messes up the custommenu: 
image
Here we see the "Data Split" menu after zooming around and out a bit.

JorisGoosen:
~~But I think I can fix it fairly easily~~

So @boutinb maybe you can have a look? I think you changed custom menu last time around no?

JorisGoosen:
It does nicely fix the bug after all.

boutinb (approved):

shun2wang:
> This pr messes up the custommenu: image Here we see the "Data Split" menu after zooming around and out a bit.

hummm. good find, indeed a problem.This is a bit tricky; we might have to switch all custommenu to the JavaScript side.

JorisGoosen:
> we might have to switch all custommenu to the JavaScript side.

Impossible, we also use it for the rest of the gui.
Bu we should probably just make an issue for this and assign it to @boutinb although maybe it is already fixed by removing/adding zoomFactor around where it gets created in MainPage.qml

shun2wang:
> Impossible, we also use it for the rest of the gui.

Sorry, I mean custommenu in result page, then pass signal/data from webchannel to ensure synchronized interaction.

> @boutinb although maybe it is already fixed by removing/adding zoomFactor around where it gets created in MainPage.qml

Great!

JorisGoosen:
> Sorry, I mean custommenu in result page, then pass signal/data from webchannel to ensure synchronized interaction.

We do ^^ 
Its just that the calculations are counting on slightly different numbers. Its just that @boutinb just rewrote the reposition logic of the customMenu so I figured he might know a quick fix (which might be the thing I mentioned)

PR #6221, opened 2026-05-14
Expose SyntaxInterface bridge APIs for jaspSyntax

FBartos:
﻿## Summary
- expose status-returning SyntaxInterface APIs for `.jasp` option/dataset load, bridge lifecycle reset, dataset callbacks, and column decoding
- make `.jasp` dataset loading file-backed and status-observable while preserving the existing void compatibility exports
- reuse the shared DataBridge/DataSetProvider path for the native bridge and update CMake wiring for SyntaxInterface consumers
- update the `jaspBase` submodule pointer to the runtime support commit used by the bridge

## Why
`jaspSyntax` bundles/links against SyntaxInterface, so the Desktop source needs to expose the same ABI and native behavior that `jaspSyntax` and `jaspTools` now rely on for local replay. This PR keeps those semantics in Desktop/SyntaxInterface instead of reimplementing them in `jaspTools`.

## Base branch
This targets `development` because `jasp-stats/jasp-desktop` does not currently expose `master` or `main`; `development` is the upstream default branch.

## Related PRs
- jaspBase: https://github.com/jasp-stats/jaspBase/pull/204 is the submodule runtime support commit referenced by this PR.
- jaspSyntax: https://github.com/jasp-stats/jaspSyntax/pull/7 consumes these SyntaxInterface APIs from the R bridge package.
- jaspTools: https://github.com/jasp-stats/jaspTools/pull/79 is the top-level local analysis replay migration that uses the bridge.
- jaspDescriptives: https://github.com/jasp-stats/jaspDescriptives/pull/500 removes the real-module replay QML blocker.

## Verification
- rebased onto current `upstream/development`
- `git diff --check HEAD~1..HEAD`
- native compile not run in this pass

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -17,13 +17,14 @@ jobs:
     runs-on: ubuntu-latest
     env:
        GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
+       JASP_R_VERSION: '4.5.3'
   
     steps:
       - uses: actions/checkout@v3
       
       - uses: r-lib/actions/setup-r@v2
         with:
-          r-version: 'release'
+          r-version: ${{ env.JASP_R_VERSION }}
       - name: Get Installed R Version
         run: echo "R_VERSION=$(R --version | head -n 1 | awk '{print $3}')" >> $GITHUB_ENV
       
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -12,6 +12,7 @@ jobs:
     runs-on: windows-latest
     env:
        GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
+       JASP_R_VERSION: '4.5.3'
        
     name: build test on Windows
         
@@ -19,7 +20,7 @@ jobs:
 
       - uses: r-lib/actions/setup-r@v2
         with:
-          r-version: 'release'
+          r-version: ${{ env.JASP_R_VERSION }}
           windows-path-include-rtools: false
           windows-path-include-mingw: true
         
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -207,6 +207,8 @@ else()
     USES_TERMINAL
     COMMENT "------ Configuring and Building the libR-Interface")
 
+  add_dependencies(SyntaxInterface R-Interface)
+
 endif()
 
 if(NOT USE_QT_STATIC_LIBS)
--- a/CommonData/CMakeLists.txt
+++ b/CommonData/CMakeLists.txt
@@ -60,6 +60,10 @@ target_link_libraries(
 
 )
 
+if(MSVC)
+  target_compile_options(CommonData PRIVATE /EHsc)
+endif()
+
 target_compile_definitions(
   CommonData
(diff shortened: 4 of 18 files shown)


FBartos:
Senior follow-up pass pushed in `4eca8fe4e`.

This fixes the native issues found while validating the jaspTools -> jaspSyntax replay bridge against real module archives:
- `QGuiApplication` now receives process-lifetime argument storage instead of pointers derived from local temporaries. This fixes the hard crash seen when loading a direct data frame and then parsing variable-list QML.
- `.jasp` archive loading now goes through `provider->loadDatabase(jaspVersion)`. That preserves the archive version during DB load so older typed-column archives, such as `jaspMixedModels/examples/Larks and Owls.jasp`, are read with the correct pre-upgrade schema instead of trying to select non-existent `Column_1` fields.

Verification run locally:
- Rebuilt `SyntaxInterface` successfully with Ninja.
- Direct in-process `jaspSyntax` read of `Larks and Owls.jasp` returns `260 x 9` with expected names.
- Public installed `jaspSyntax::readDatasetFromJaspFile()` subprocess read also returns `260 x 9`.
- End-to-end `jaspTools::runAnalysis(MixedModelsGLMM, dataset, opts)` completes with `status = complete` on the patched stack.

Remaining note: native teardown still emits noisy QML/stack-imbalance/QThreadStorage warnings in some focused tests. They are not blocking replay now, but they are real lifecycle debt worth tracking separately from this bridge contract.

Related PRs for the full bridge:
- jaspTools orchestration layer: https://github.com/jasp-stats/jaspTools/pull/79
- jaspSyntax bridge API layer: https://github.com/jasp-stats/jaspSyntax/pull/7

FBartos:
Follow-up on the noisy focused-test output:

- The `GridLayout` QML warning was not a module layout issue. In the headless/minimal SyntaxInterface path the layout check can run during cleanup/re-entry with a non-positive or non-finite width, while `implicitWidth` is still populated. I guarded `_checkColumns()` for non-positive/non-finite dimensions.
- The `QThreadStorage` message was a real lifecycle issue. `syntaxBridgeCleanup()` is a reusable state reset and should stay that way, but the bridge also needed a terminal shutdown path that deletes the `QQmlEngine`, owned `QGuiApplication`, encoder, and data-provider state. Added `syntaxBridgeShutdown()` for package unload/session exit.
- The stack-imbalance warning was not a MixedModels warning. It came from SyntaxInterface parse-only startup bootstrapping more of the R bridge than it needed. Desktop now keeps `jaspBase` lazy for the uncommon QML-R-script path, and `rbridge_init(..., insideJasp=false)` skips the Engine-only friendly R helper bundle while still exposing the native dataset/callback bridge.

Validation with a fresh `jaspSyntax` install using the rebuilt Desktop bridge:

- `ninja -C C:/JASP-Packages/jasp-desktop/build SyntaxInterface`
- `test-module-options.R`: 49 pass
- `test-dataset-helpers.R`: 85 pass
- full `jaspSyntax/tests/testthat`: 256 pass, 1 skip
- real `jaspMixedModels` QML parse for `MixedModelsGLMM`: 76 options returned

The warning scans for those runs were clean for `GridLayout`, `qml:`, `QThreadStorage`, `QObject`, `Timers cannot`, `stack imbalance`, eager `jaspBase` init, and friendly-R-helper loading.

boutinb (approved):

PR #6227, opened 2026-05-19
Load labels when reading sav files

### LINKED ISSUE #4283: [Bug]: Labelled datasets loaded pre-0.97, but no longer loading
### JASP Version

0.97.0

### Commit ID

_No response_

### JASP Module

Unrelated

_No response_

macOS Silicon

### Bug Description

Labelled .Rdata and .sav files used to load labels in JASP when opening, no longer the case in 0.97. "Use labels" is off by default, even after enabling it and reloading data, new instance is started and labels don't get loaded.

### Expected Behaviour

Same as above, labelled variables should load properly as they did before.

### Steps to Reproduce

[Labelled_data.sav.zip](https://github.com/user-attachments/files/27951399/Labelled_data.sav.zip)
Please load this labelled dataset in 0.96 vs 0.97. In 0.97, ordinal variable labels not loading, even if "Use labels" is set to ON in last instance.

### Log (if any)

_No response_

### More Debug Information

_No response_

### Final Checklist

boutinb:
Fixes https://github.com/jasp-stats/jasp-issues/issues/4283


shun2wang:
Has the unit test covered this situation? If not,it's better to add one.

shun2wang (approved):
Looks good for me, tested with .sav files with labels.

boutinb:
I have fixed the unit tests for loading different datafiles, and add a special unit test for labels with sav file.

PR #6229, opened 2026-05-20
Improve user experience for plot editing

vandenman:
Implement a "debounce timer" to avoid that requests "pile up" and the engine is still busy rendering request 1 while in the meantime 4 other requests arrived (which would be rendered in order before).

Also adds a nice overlay so users know when the plot is being rerendered.

### DIFF
--- a/Desktop/components/JASP/Widgets/PlotEditor/PlotEditor.qml
+++ b/Desktop/components/JASP/Widgets/PlotEditor/PlotEditor.qml
@@ -386,6 +386,48 @@ Popup
 							{
 								src:			plotImg
 							}
+
+							// Overlay shown while the engine is re-rendering
+							Rectangle
+							{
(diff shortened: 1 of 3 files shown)


Desktop/components/JASP/Widgets/PlotEditor/PlotEditor.qml
```diff
@@ -386,6 +386,48 @@ Popup
...
+											to:			360
+											duration:	1200
+											loops:		Animation.Infinite
+											running:	plotEditorModel.updating
+										}
+									}
+
+									JASPC.Text
+									{
+										text:				qsTr("Rerendering\u2026")
```
JorisGoosen: Maybe its nicer to simply put the actual unicode character here?

JorisGoosen (approved):
Works very smooth now

PR #6230, opened 2026-05-21
Label filter does not work with JASP files from Data Library

### LINKED ISSUE #4293: Filters broken in 0.97 for categorial data of older JASP files
filtering of nominals and ordinals string data of older jasp files is broken:
- data library => mixed models => larks and owls data 
- filter gender

problem disappears if I click and unclick this:

Image

Problem:

Image

another one:

Image

boutinb:
Fixes https://github.com/jasp-stats/jasp-issues/issues/4293

The JASP files from the Data Library were apparently saved in a beta version of JASP, because the bug is not reproducible with JASP made in an officilal version of 0.97 or 0.96.
What happens is that the vector of Labels  _labels in Column, contains the labels but also the empty values. In the JASP files of the Data Library, an empty value label is at the start of the _labels vector. This is wrong, but it is also wrong to fetch a label from its index in the Label window, since in this window only non empty labels are displayed. So by fetching the right Label object, the issue is fixed, and makes JASP a bit robuster.


JorisGoosen:
Is there now a unittest added for this too?

JorisGoosen:
Uhm, shouldnt we just fix the data library files? If this bug cannot occur with files created with 0.95.4 and 0.96?  Or is there some specific sort of file that was broken in this way that now can be loaded properly? Because then there should be at least 1 or 2 unittest jasspfiles for this!

Desktop/data/datasetpackage.cpp
```diff
@@ -566,22 +566,23 @@ QVariant DataSetPackage::data(const QModelIndex &index, int role) const
...
-		case int(specialRoles::description):					return index.row() >= labels.size() ? "" : tq(labels[index.row()]->description());
-		case int(specialRoles::filter):							return index.row() >= labels.size() || labels[index.row()]->filterAllows();
+		case int(specialRoles::description):
+		case int(specialRoles::filter):
+		{
+			Label * label = dynamic_cast<Label*>(indexPointerToNode(index));
+			if (!label)	return QVariant();
+			return (role == int(specialRoles::description)) ? QVariant(tq(label->description())) : QVariant(label->filterAllows());
+		}
 		case int(specialRoles::value):							return tq(column->labelByIndexNonEmpty(index.row())->originalValueAsString());
```
JorisGoosen, line 586: Did you see the function `column->labelByIndexNonEmpty(index.row())` ? Why not just use that...

JorisGoosen:
It does fix the problem at least

JorisGoosen:
`getColumnFilterAllows` why did you change it?

Desktop/data/datasetpackage.cpp
boutinb: dynamic_cast<Label*>(indexPointerToNode(index)) is what is used by all setLabel functions in datasetpackage, that's why I used it. 
I see that labelByIndexNonEmpty is used also, but if possible, I would replace all labelByIndexNonEmpty by indexPointerToNode function: we could then remove the _labelByNonEmptyIndex mapping in column, which would be a good thing.

boutinb:
> Uhm, shouldnt we just fix the data library files? If this bug cannot occur with files created with 0.95.4 and 0.96? Or is there some specific sort of file that was broken in this way that now can be loaded properly? Because then there should be at least 1 or 2 unittest jasspfiles for this!

I could not reproduce the problem with JASP files made in 0.96 or 0.95.4, but it's difficult to be sure that this is not reproducible. I suspect that the problem in the JASP files of the Data Library comes from a JASP beta version, but cannot be sure either. There is good chance that some JASP users may have such a JASP file.
But what is clear for me is that the code is not robust when handling with labels: it should not use the row given by the Label Editor to fetch the label in _labels.
I can add a JASP file from the Data Library as test case. I'll do this.

boutinb:
> `getColumnFilterAllows` why did you change it?

Because the generateLabelFilter function in labelFilterGenerator was using this code:
```
boolvec				filterAllows	= pkg->getColumnFilterAllows(col);
stringvec			labels		= pkg->getColumnLevelsAsStrVec(col);
```
That was supposing that both functions give the same ordered vector for the labels, which was not true since  getColumnFilterAllows gave also the empty values labels. So I thought in place of expecting that 2 functions gave the same vector, just use 1 that gives all necessary information, especially because getColumnFilterAllows is used only in this case.

JorisGoosen:
> > `getColumnFilterAllows` why did you change it?
> 
> Because the generateLabelFilter function in labelFilterGenerator was using this code:
> 
> ```
> boolvec				filterAllows	= pkg->getColumnFilterAllows(col);
> stringvec			labels		= pkg->getColumnLevelsAsStrVec(col);
> ```
> 
> That was supposing that both functions give the same ordered vector for the labels, which was not true since getColumnFilterAllows gave also the empty values labels. So I thought in place of expecting that 2 functions gave the same vector, just use 1 that gives all necessary information, especially because getColumnFilterAllows is used only in this case.

If its used only in this case you could just make it return the same vector as getColumnLevelsAsStrVec no?

JorisGoosen:
Seems easier

boutinb:
> Seems easier

But less error prone.
I have added a unit test with the Directed Reading Activities JASP file from the Data Library. The test fails before this fix, and succeeds afterwards.
If you don't have more remarks, let's merge this.

JorisGoosen (approved):

Desktop/data/datasetpackage.cpp
JorisGoosen: That function will be removed so please dont do that...

PR #6234, opened 2026-05-26
Fixes the encoding and data-preloading of the repeated measures cells

JorisGoosen:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/3192

### DIFF
--- a/Common/columnencoder.cpp
+++ b/Common/columnencoder.cpp
@@ -470,7 +470,7 @@ std::string ColumnEncoder::encodeRScript(std::string text, const std::map<std::s
 		for(size_t bracePos = pos; bracePos < text.size() && endIsFree && keepGoing; bracePos++)
 			if(text[bracePos] == '(')
 				endIsFree = false;
-			else if(text[bracePos] != '\t' && text[bracePos] != ' ')
+			else if(text[bracePos] != '\t' && text[bracePos] != ' ' && text[bracePos] != '\n')
 				keepGoing = false; //Aka something else than whitespace or a brace and that means that we can replace it!
 		return endIsFree;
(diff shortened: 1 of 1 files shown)


JorisGoosen:
I made a local anova bundle with preloadData on and that now works so I think repeated measures is now fixed.
Also checked with Johnnys branch of jaspAnova

QMLComponents/boundcontrols/boundcontrolmeasurescells.cpp
```diff
@@ -63,6 +63,13 @@ Json::Value BoundControlMeasuresCells::createJson() const
 	return result;
 }
 
+Json::Value BoundControlMeasuresCells::createMeta() const
+{
+	Json::Value meta(BoundControlBase::createMeta()); //Work from parent to get everything it defines
+		
+	return meta;
+}
+
```
boutinb: I don't understand this: why the implicit call to BoundControlBase::createMeta() would not give the same result?

boutinb (commented):
Should jaspBase be in this PR?

JorisGoosen:
> Should jaspBase be in this PR?

No I guess not.

QMLComponents/boundcontrols/boundcontrolmeasurescells.cpp
JorisGoosen: Uhm, now that I look at the code again, im also not sure how it was working? Or how this fixes it... Uhm, let me double check whats going on here

JorisGoosen:
Better like this?

boutinb (approved):

PR #6235, opened 2026-05-26
Update jaspBase for decoded result objects

FBartos:
fixes: https://github.com/jasp-stats/INTERNAL-jasp/issues/3220

﻿## Summary
- updates `Engine/jaspBase` to `8c4f861`, which decodes the R-facing result copy returned by `$toRObject()` and exposes `decodeJaspResultState()` for stored figure state
- fixes direct `jaspSyntax` replay leakage of encoded bridge column names in returned table metadata, footnotes, stored ggplot labels, and result-state figure objects
- keeps encoded names available to modules while analyses run, so model fitting and dataset lookup continue to use the backend/runtime names

## Why
The `jaspTools` JSON path already runs result/state decoding after an analysis. Direct `jaspSyntax` usage, for example `jaspSyntax::loadDataSet(myData)` followed by `jaspMixedModels::MixedModelsLMM(...); res$toRObject()`, bypasses that `jaspTools` post-processing and receives raw `jaspBase` wrapper objects.

For MixedModels this meant the model ran correctly, but `res$toRObject()` could still expose internal names such as `JaspColumn_2_Encoded` in ANOVA footnotes and ggplot labels. The owning fix is in `jaspBase`, and this PR moves Desktop to that commit.

This also removes the remaining result-state gray area from `jaspTools`: state figure decoding now has a public `jaspBase::decodeJaspResultState()` API, so `jaspTools` can delegate instead of calling `jaspBase:::decodeplot()` directly.

## Dependencies
- jaspBase: https://github.com/jasp-stats/jaspBase/pull/204
- jaspTools consumer update: https://github.com/jasp-stats/jaspTools/pull/79
- related jaspSyntax result-text decoder: https://github.com/jasp-stats/jaspSyntax/pull/8

## Verification
- `Rscript -e "pkgload::load_all('C:/JASP-Packages/jasp-desktop/Engine/jaspBase', quiet = TRUE); testthat::test_dir('tests/testthat', reporter = 'summary')"`
- Direct R 4.6 replay of the `lme4::cake`/`ABC` MixedModelsLMM example through `jaspSyntax::loadDataSet()` and `res$toRObject()`; ANOVA footnote decoded to `recipe`, plot labels decoded to `ABC` and `temperature`, and the returned object had zero scalar `jaspColumn`/`JaspColumn_` hits.
- `git diff --check`

### DIFF
--- a/Common/tempfiles.cpp
+++ b/Common/tempfiles.cpp
@@ -55,7 +55,7 @@ void TempFiles::createSessionDir()
 
 	std::filesystem::path sessionPath = _sessionDirName;
 	
-	std::cout<< "'" << sessionPath.string() << "' about to be (removed and re)created." << std::endl;
+	Log::log() << "'" << sessionPath.string() << "' about to be (removed and re)created." << std::endl;
 
 	std::filesystem::remove_all(sessionPath, error);
 	std::filesystem::create_directories(sessionPath, error);
--- a/QMLComponents/datasetprovider.cpp
+++ b/QMLComponents/datasetprovider.cpp
@@ -91,7 +91,7 @@ QVariant DataSetProvider::data(const QModelIndex & index, int role) const
 
 void DataSetProvider::loadDataSet(const std::map<std::string, stringvec > & dataSet, int threshold, bool orderLabelsByValue)
 {
-
+	beginResetModel();
 	_dataSet->beginBatchedToDB();
 
 	int rowCount = 0;
@@ -119,7 +119,7 @@ void DataSetProvider::loadDataSet(const std::map<std::string, stringvec > & data
 	_dataSet->endBatchedToDB([](float f) {});
 
 	ColumnEncoder::columnEncoder()->setCurrentNames(_dataSet->getColumnTypesMap());
-
+	endResetModel();
 }
 
 void DataSetProvider::closeDatabase()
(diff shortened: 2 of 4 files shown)


FBartos:
Clarifying status because this PR is easy to confuse with the newer bridge work.

This is the older aggregate branch from the original decoded-result-object effort. It mixes a desktop `Engine/jaspBase` gitlink update with several SyntaxInterface/QML/runtime fixes that were explored along the way. It should not be treated as the active desktop PR for the current jaspTools -> jaspSyntax bridge architecture.

The active desktop-side bridge PR is now:
- https://github.com/jasp-stats/jasp-desktop/pull/6249

That newer PR keeps the desktop work focused on the native `ColumnEncoder` / SyntaxInterface bridge and intentionally does not carry the `Engine/jaspBase` submodule update. The jaspBase package work is tracked separately in:
- https://github.com/jasp-stats/jaspBase/pull/204

So my recommendation is: do not merge this PR as-is. If any specific non-decoder fix here is still needed, we should extract/cherry-pick it into a small focused PR instead.

boutinb:
This PR was a duplicate https://github.com/jasp-stats/jasp-desktop/pull/6249.
I merged the small extra changes in https://github.com/jasp-stats/jasp-desktop/pull/6261

PR #6233, opened 2026-05-26
add fileSystemWriteVirtualization on msix

shun2wang:
maybe a solution for https://github.com/jasp-stats/INTERNAL-jasp/issues/3217 and https://github.com/jasp-stats/jasp-issues/issues/4292

Refer: https://learn.microsoft.com/en-us/uwp/schemas/appxpackage/uapmanifestschema/element-virtualization-filesystemwritevirtualization

Could some one make a MS Store beta version and test it with me?

How to actually test it: You need to build two versions, one before the other. Based on the first version, create logs and configurations, then upgrade to the second version and see if the previous settings still exist.

PR #6236, opened 2026-05-27
fix files URI scheme in html image source

shun2wang:
This helps alleviate some issues with pasting images in certain office software(e.g. WPS Office).

I intend to keep this PR separate from https://github.com/jasp-stats/jasp-desktop/pull/6233, because they are different things.
see: https://en.wikipedia.org/wiki/File_URI_scheme, `file:///path`  here 3rd `/` means empty hostname, is robust solution mayebe.


RensDofferhoff (approved):

RensDofferhoff:
good catch that is more correct

PR #6239, opened 2026-05-27
fix full result copy on Windows MSIX

RensDofferhoff:
Keeps settings virtual.
Put tmp clipboard  in sanbox folder for external apps

### DIFF
--- a/Common/tempfiles.cpp
+++ b/Common/tempfiles.cpp
@@ -32,7 +32,6 @@ const long 				outOfDateDelta = 24 * 3600;
 long					TempFiles::_sessionId		= 0;
 std::string				TempFiles::_sessionDirName	= "";
 std::string				TempFiles::_statusFileName	= "";
-std::string				TempFiles::_clipboard		= "";
 int						TempFiles::_nextFileId		= 0;
 int						TempFiles::_nextTmpFolderId	= 0;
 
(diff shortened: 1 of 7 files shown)


RensDofferhoff:
MSIX is now building

shun2wang:
Hi, maybe should rebase this PR on [6349636](https://github.com/jasp-stats/jasp-desktop/commit/6349636677416636d4031adc46b3c5542d58a69f) when build test version.This also allows us to test that fix.

RensDofferhoff:
good point

RensDofferhoff:
https://static.jasp-stats.org/Nightlies/WindowsMSIX/JASP-nightly-fixCopy-928d3f5e79b0ce4c7c24661d891165f18dd21bc7.msix

shun2wang:
Did you signed the package? cause it says:`A certificate chain processed, but terminated in a root certificate which is not trusted by the trust provider.`

If you add a OID then should install from unsigned pkg. https://learn.microsoft.com/hr-hr/windows/msix/package/unsigned-package#create-an-unsigned-package

OK, If I install the JASP digital certificate as a trusted root certificate, then it can be installed.

shun2wang (commented):
Clipboard now works good, but settings still not be synced between two versions(https://github.com/jasp-stats/INTERNAL-jasp/issues/3217).

For example `conf.toml` of configuration files does not in sandbox directories, and if I check/uncheck `sandbox engine` option the setting was not synced in other sotre version installation.

RensDofferhoff:
current behaviour syncs the settings from msi registry once if it doesnt have the option set yet.
Not an issue people will no be running multiple versions.

shun2wang:
> current behaviour syncs the settings from msi registry once if it doesnt have the option set yet. Not an issue people will no be running multiple versions.

Not .msi registry, but I mean the same .msix after upgrade.because I saw now setting files still not in sandbox dir.

RensDofferhoff:
the next upgrade should properly retain settings on upgrade.
But if you uninstall  it the settings are lost (cleans up nicely).

PR #6244, opened 2026-05-28
change advice when something is wrong so EJ gets less email

JorisGoosen:
Changes ", if that doesn't help, you could try to contact the module's maintainer: info@jasp-stats.org" 

To:   ", if that doesn't help, you could try our github issuetracker: https://github.com/jasp-stats/jasp-issues/issues."

### DIFF
--- a/Desktop/modules/dynamicmodules.cpp
+++ b/Desktop/modules/dynamicmodules.cpp
@@ -450,7 +450,7 @@ Modules::AnalysisEntry* DynamicModules::retrieveCorrespondingAnalysisEntry(const
 	throw Modules::ModuleException(moduleName,
 		"Module is not available, to load this JASP file properly you will need to install it first and then retry.\n"
 		"If you do not have this module you can try the module's website: \""  + jsonFromJaspFile.get("moduleWebsite", "jasp-stats.org").asString()	 +  "\" or"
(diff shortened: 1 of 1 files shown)


Desktop/modules/dynamicmodules.cpp
```diff
@@ -450,7 +450,7 @@ Modules::AnalysisEntry* DynamicModules::retrieveCorrespondingAnalysisEntry(const
 	throw Modules::ModuleException(moduleName,
 		"Module is not available, to load this JASP file properly you will need to install it first and then retry.\n"
 		"If you do not have this module you can try the module's website: \""  + jsonFromJaspFile.get("moduleWebsite", "jasp-stats.org").asString()	 +  "\" or"
-		", if that doesn't help, you could try to contact the module's maintainer: \"" + jsonFromJaspFile.get("moduleMaintainer", "the JASP team").asString() + "\"."
+		", if that doesn't help, you could try our github issuetracker: https://github.com/jasp-stats/jasp-issues/issues."
```
shun2wang, line 453: this should be marked as translatable strings I think.

JorisGoosen: Yes why not indeed. Although it should also link in with https://github.com/jasp-stats/jasp-desktop/pull/6164

PR #6249, opened 2026-06-01
Expose native column encoder context

FBartos:
## Summary

This PR exposes the Desktop-owned encoder context needed by the jaspTools -> jaspSyntax bridge:

- replaces the exported decoder snapshot/mapping API with a serializable `ColumnEncoder` context
- keeps the context/schema/scoped restore helper in `Common/columnencodercontext.*`, next to `ColumnEncoder`
- captures dataset column names/types and extra QML option encodings from the native source state
- decodes text by temporarily restoring native `ColumnEncoder` state and then calling the normal decoder path
- removes the native `decodeAllWithMapping()` compatibility path so R cannot silently decode from a stale map
- keeps native bridge failures structured at the C ABI boundary

## Why

Encoding/decoding should remain an internal Desktop/SyntaxInterface concern. The R packages now pass around an opaque encoder context, and Desktop remains the only implementation of token replacement. That prevents incorrect replay when the live dataset has changed, and it avoids duplicating encoding rules in jaspSyntax, jaspBase, or jaspTools.

## Related PRs

- jaspSyntax R bridge API: https://github.com/jasp-stats/jaspSyntax/pull/8
- jaspBase result materialization: https://github.com/jasp-stats/jaspBase/pull/204
- jaspTools replay orchestration: https://github.com/jasp-stats/jaspTools/pull/79

## Validation

- rebuilt `SyntaxInterface` with `ninja -C build -j1 SyntaxInterface` through the local MSVC toolchain
- verified the jaspSyntax/SyntaxInterface C ABI exports with `tools/check-syntaxinterface-symbols.sh`
- ran focused jaspSyntax bridge tests against the rebuilt Desktop target: `test-desktop-jasp-contract.R`, `test-dataset-helpers.R`
- ran focused jaspBase tests: `test-result-object-decoding.R`, `test-runWrappedAnalysis.R`
- ran focused jaspTools bridge lifecycle tests: `test-jaspSyntax-lifecycle.R`
- ran R parse checks for jaspSyntax, jaspBase, and jaspTools
- ran `git diff --check`

Note: this PR should be reviewed together with the linked package PRs because the contract is intentionally split by ownership: Desktop owns token replacement, jaspSyntax exposes the bridge, jaspBase decodes JASP-owned result surfaces, and jaspTools carries the captured context through replay.

### DIFF
--- a/Common/columnencoder.h
+++ b/Common/columnencoder.h
@@ -63,6 +63,7 @@ public:
 	
 	static	colVec				columnNames();
 	static	colVec				columnNamesEncoded();
+	static	const char*			extraOptionsPrefix()										{ return "JaspExtraOptions_"; }
 
 			bool				shouldEncode(const std::string & in);
 			bool				shouldDecode(const std::string & in);
--- a/Common/columnencoder.h
+++ b/Common/columnencoder.h
@@ -63,12 +63,11 @@ public:
 	
 	static	colVec				columnNames();
 	static	colVec				columnNamesEncoded();
-	static	colMap				decodingMapSnapshot();
-	static	std::string			decodeAllWithMapping(const std::string & text, const colMap & decodingMap);
 
 			bool				shouldEncode(const std::string & in);
 			bool				shouldDecode(const std::string & in);
 			void				setCurrentNames(const colTypeMap & names);
+			const colTypeMap&	currentNames() const { return _dataSetTypes; }
 			void				updateColumnTypesOnly(const colTypeMap & names);
 			void				setCurrentNames(const std::vector<std::string> & names, bool generateTypesEncoding=true);	///< Do not use! Deprecated
 			void				setCurrentColumnTypePerName(const colTypeMap & theMap);									///< Do not use! Deprecated
--- a/Common/columnencoder.h
+++ b/Common/columnencoder.h
@@ -63,6 +63,8 @@ public:
 	
 	static	colVec				columnNames();
 	static	colVec				columnNamesEncoded();
+	static	colMap				decodingMapSnapshot();
+	static	std::string			decodeAllWithMapping(const std::string & text, const colMap & decodingMap);
 
 			bool				shouldEncode(const std::string & in);
 			bool				shouldDecode(const std::string & in);
@@ -87,7 +89,7 @@ public:
 	static	std::string			encodeAll(const std::string & text) { return replaceAll(text, encodingMap(), originalNames()); }
 
 			///Replace all occurences of encoded columnNames in a string by their decoded versions, regardless of word boundaries or parentheses.
-	static	std::string			decodeAll(const std::string & text) { return replaceAll(text, decodingMap(), encodedNames());  }
+	static	std::string			decodeAll(const std::string & text) { columnEncoder(); return replaceAll(text, decodingMap(), encodedNames());  }
 
 			///Replace all occurences of columnNames in a string by their encoded versions in all json-names and string-values, regardless of word boundaries or parentheses.
 	static	void				encodeJson(Json::Value & json, bool replaceNames = false, bool replaceStrict = false);
--- a/Common/columnencoder.h
+++ b/Common/columnencoder.h
@@ -63,7 +63,6 @@ public:
 	
 	static	colVec				columnNames();
 	static	colVec				columnNamesEncoded();
-	static	const char*			extraOptionsPrefix()										{ return "JaspExtraOptions_"; }
 
 			bool				shouldEncode(const std::string & in);
 			bool				shouldDecode(const std::string & in);
--- a/Common/log.cpp
+++ b/Common/log.cpp
@@ -15,7 +15,19 @@
 
 std::ofstream Log::_logFile;// = bofstream();
 
-std::ostream* Log::_nullStream = &std::cout;
+namespace
+{
+	class NullLogBuffer : public std::streambuf
+	{
+	protected:
+		int overflow(int c) override { return traits_type::not_eof(c); }
+	};
+
+	NullLogBuffer	nullLogBuffer;
+	std::ostream	nullLogStream(&nullLogBuffer);
+}
+
+std::ostream* Log::_nullStream = &nullLogStream;
 
 std::string Log::logFileNameBase	= "";
 
--- a/Common/log.cpp
+++ b/Common/log.cpp
@@ -15,21 +15,7 @@
 
 std::ofstream Log::_logFile;// = bofstream();
 
-namespace
-{
-	// Used before Log::init() receives an application-owned null stream. The
-	// default logType::null destination must discard output, not write to cout.
-	class NullLogBuffer : public std::streambuf
-	{
-	protected:
-		int overflow(int c) override { return traits_type::not_eof(c); }
-	};
-
-	NullLogBuffer	nullLogBuffer;
-	std::ostream	nullLogStream(&nullLogBuffer);
-}
-
-std::ostream* Log::_nullStream = &nullLogStream;
+std::ostream* Log::_nullStream = &std::cout;
 
 std::string Log::logFileNameBase	= "";
 
--- a/Common/log.cpp
+++ b/Common/log.cpp
@@ -17,6 +17,8 @@ std::ofstream Log::_logFile;// = bofstream();
 
 namespace
 {
+	// Used before Log::init() receives an application-owned null stream. The
+	// default logType::null destination must discard output, not write to cout.
 	class NullLogBuffer : public std::streambuf
 	{
 	protected:
--- a/SyntaxInterface/syntaxbridge.cpp
+++ b/SyntaxInterface/syntaxbridge.cpp
@@ -43,9 +43,11 @@
 #include "modules/dynamicmodule.h"
 #include "archivereader.h"
 #include "databaseinterface.h"
+#include "columnencoder.h"
 
 #include <string>
 #include <vector>
+#include <stdexcept>
 
 #include <QtPlugin>
 #ifdef USE_QT_STATIC_LIBS
@@ -150,6 +152,98 @@ static const char* statusError(Json::Value status, const std::string & error)
 	status["error"] = error;
 	Log::log() << error << std::endl;
 	return statusResult(status);
+}
+
+static Json::Value columnDecoderSnapshotJson()
+{
+	Json::Value snapshot(Json::objectValue);
+	Json::Value columns(Json::arrayValue);
+	const ColumnEncoder::colMap decodingMap = ColumnEncoder::decodingMapSnapshot();
+
+	for(const auto & keyVal : decodingMap)
+	{
+		Json::Value column(Json::objectValue);
+		column["encoded"] = keyVal.first;
+		column["decoded"] = keyVal.second;
+		columns.append(column);
+	}
+
+	snapshot["version"] = 1;
+	snapshot["columns"] = columns;
+	return snapshot;
+}
+
+struct ColumnDecoderSnapshot
+{
+	ColumnEncoder::colMap	decodingMap;
+	bool					supplied = false;
+};
+
+static ColumnDecoderSnapshot columnDecoderMapFromSnapshot(const Json::Value & snapshot)
+{
+	ColumnDecoderSnapshot snapshotState;
+	snapshotState.supplied = true;
+	const Json::Value & columns = snapshot["columns"];
+	if(!columns.isArray())
+		return snapshotState;
+
+	for(const Json::Value & column : columns)
+		if(column.isObject() && column["encoded"].isString() && column["decoded"].isString())
+			snapshotState.decodingMap[column["encoded"].asString()] = column["decoded"].asString();
+
+	return snapshotState;
+}
+
+static ColumnDecoderSnapshot columnDecoderMapFromSnapshotString(const char * snapshotJson)
+{
+	if(!snapshotJson || std::string(snapshotJson).empty())
+		return ColumnDecoderSnapshot();
+
+	Json::Value snapshot;
+	Json::Reader reader;
+	if(!reader.parse(snapshotJson, snapshot))
+		throw std::runtime_error("Could not parse column decoder snapshot JSON.");
+
+	return columnDecoderMapFromSnapshot(snapshot);
+}
+
+static Json::Value parseStringArrayJson(const char * valuesJson)
+{
+	if(!valuesJson)
+		throw std::runtime_error("Cannot decode column text from a null JSON payload.");
+
+	Json::Value values;
+	Json::Reader reader;
+	if(!reader.parse(valuesJson, values))
+		throw std::runtime_error("Could not parse column text JSON payload.");
+	if(!values.isArray())
+		throw std::runtime_error("Column text JSON payload must be an array.");
+
+	return values;
+}
+
+static Json::Value decodeColumnTextJson(const Json::Value & values, const ColumnDecoderSnapshot & snapshot)
+{
+	Json::Value decodedValues(Json::arrayValue);
+
+	for(const Json::Value & value : values)
+	{
+		if(value.isNull())
+		{
+			decodedValues.append(Json::Value());
+		}
+		else if(value.isString())
+		{
+			const std::string text = value.asString();
+			decodedValues.append(snapshot.supplied ? ColumnEncoder::decodeAllWithMapping(text, snapshot.decodingMap) : ColumnEncoder::decodeAll(text));
+		}
+		else
+		{
+			throw std::runtime_error("Column text JSON payload must contain only strings or null values.");
+		}
+	}
+
+	return decodedValues;
 }
 
 static Json::Value analysisOptionsStatus(const char * filePath, int analysisNr)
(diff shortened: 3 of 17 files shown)


vandenman:
@JorisGoosen do you have time to review this? You're more familiar with the decoding stuff than I am. The basic idea is that we want to export its contents as json, so that we can reinstantiate the decoder in R in jaspSyntax. This is necessary when you save results to disk, restart R, and then load the results again. However, the round trip should be perfect and easy to maintain in the future (so you should understand what's going on here).

Common/log.cpp
```diff
@@ -15,7 +15,19 @@
...
+	{
+	protected:
+		int overflow(int c) override { return traits_type::not_eof(c); }
+	};
+
+	NullLogBuffer	nullLogBuffer;
+	std::ostream	nullLogStream(&nullLogBuffer);
+}
+
+std::ostream* Log::_nullStream = &nullLogStream;
```
JorisGoosen: Maybe this is a good idea but I will need to know why this was done?

JorisGoosen (changes requested):
So this PR has some good ideas, but its mostly adding a lot of hassle on top of columnencoder. 

Instead of adding all of these functions it would be a lot easier to simply store the output of `DataSet::getColumnTypesMap()` somewhere in your R datastructure (so you can reload it on demand). Then when you need to decode stuff you just use ColumnEncoder as usual after setting those columnTypes in `setCurrentColumnNames`. 

Then there is almost no extra code to maintain and everything happens in the same way.

Or is there something Im missing that makes this obligatory?

JorisGoosen:
jaspBase shouldnt be updated.

The extra columnencoder related classes should be either in the file WITH columnencoder or in a separate class next to columnencoder. but not with syntaxbridge.

FBartos:
@JorisGoosen addressed in `ed7c12e`.

- Removed the `Engine/jaspBase` gitlink from this Desktop PR. The jaspBase work stays isolated in https://github.com/jasp-stats/jaspBase/pull/204.
- Moved the encoder context/schema/scoped restore code out of `SyntaxInterface` into `Common/columnencodercontext.*`, next to `ColumnEncoder`.
- `SyntaxBridge` now only handles the exported C ABI: it captures `DataSet::getColumnTypesMap()` plus the extra-option encoder state, and decoding restores those maps into `ColumnEncoder` before calling the normal `ColumnEncoder::decodeAll()` path.

So the package side stores/passes an opaque context, and Desktop remains the only place implementing token replacement. I also re-ran the native build plus the focused jaspSyntax, jaspBase, and jaspTools bridge tests listed in the PR body.

Common/log.cpp
FBartos: This is needed so `logType::null` actually discards output in the static jaspSyntax bridge. Before this change `_nullStream` pointed at `std::cout`, so `syntaxBridgeSetVerbose(false)` could still let native bridge logging reach stdout before `Log::init()` installed an application-owned null stream. The local streambuf is a true sink; verbose mode still switches the bridge back to stdout. I added a short code comment for that rationale.

JorisGoosen: We could just call the following in the syntaxbridge constructor:
```
Log::logFileNameBase = (AppDirs::logDir() + "JASP "  + getSortableTimestamp()).toStdString();
Log::init(&nullstream);
Log::setLogFileName(Log::logFileNameBase + " Desktop.log");
Log::setLoggingToFile(_preferences->logToFile());
```
Like MainWindow::initLog ?

SyntaxInterface/syntaxbridge.cpp
```diff
@@ -152,6 +154,98 @@ static const char* statusError(Json::Value status, const std::string & error)
...
+	Json::Reader reader;
+	if(!reader.parse(valuesJson, values))
+		throw std::runtime_error("Could not parse column text JSON payload.");
+	if(!values.isArray())
+		throw std::runtime_error("Column text JSON payload must be an array.");
+
+	return values;
+}
+
+static Json::Value decodeColumnTextJson(const Json::Value & values, const ColumnDecoderSnapshot & snapshot)
```
JorisGoosen: This whole function looks like it should just call `setCurrentColumnNames` first with the stuff from the snapshot. And than `ColumnEncoder::decodeJson`. Not really sure why we need to have this whole function.

SyntaxInterface/syntaxbridge.cpp
```diff
@@ -655,6 +758,42 @@ const char* STDCALL syntaxBridgeGetVariableNames()
...
+
+const char* STDCALL syntaxBridgeColumnDecoderSnapshot()
+{
+	static std::string result;
+
+	result = columnDecoderSnapshotJson().toStyledString();
+	return result.c_str();
+}
+
+const char* STDCALL syntaxBridgeDecodeColumnText(const char* valuesJson, const char* decoderSnapshotJson)
```
JorisGoosen: Why isnt the decoder snapshot json not simply set from R? And the standard ColumnEncoder functions used? WHy all these extra functions?

Common/log.cpp
FBartos: That is exactly the issue this was trying to fix: when the static bridge is used from R/jaspSyntax, `MainWindow::initLog()` is never called, so `logType::null` still pointed at `std::cout` and JASP/native messages leaked into R sessions even with verbose disabled.

I agree this is bridge initialization concern rather than part of the decoder API. Before I patch it: do you want the SyntaxInterface bridge to only install a local null stream via `Log::init(&nullStream)` and then toggle `Log::setDefaultDestination()` / `Log::setWhere()` from `syntaxBridgeSetVerbose()`, or do you want the bridge to do the fuller `MainWindow::initLog()`-style log-file setup as well?

My preference would be the small bridge-local null-stream initialization, without Desktop preferences/log-file setup, because the R/headless bridge should not depend on `PreferencesModel` or Desktop-owned logging policy.

SyntaxInterface/syntaxbridge.cpp
FBartos: Agreed, this should be simplified. I can change the native path to parse the incoming JSON, temporarily install the captured context into `ColumnEncoder`, call the standard `ColumnEncoder::decodeJson()`, and restore the previous state before returning.

One detail to confirm: should this bridge call decode arbitrary JSON with `replaceNames = true`, or keep the current R contract as a character-vector JSON payload and still use `decodeJson()` internally? The former is more directly the standard ColumnEncoder API; the latter keeps the R API smaller and avoids exposing more surface than jaspSyntax currently needs.

SyntaxInterface/syntaxbridge.cpp
FBartos: I think I understand the intended simplification: R should not carry mappings or use bridge-specific replacement logic; it should pass the captured native context back, native code should install that context into `ColumnEncoder`, and then call standard `ColumnEncoder` decode functions.

The point I want to clarify before patching is whether you prefer a persistent R-facing set-current-ColumnEncoder-context API, or a scoped decode call that installs the snapshot, calls `ColumnEncoder::decodeJson()`, and restores the previous native state before returning.

I prefer the scoped form because result replay often happens after the live dataset has changed, and persistent mutation from R is easier to misuse. But it would still remove the custom decode machinery and use the normal `ColumnEncoder` path. Is that aligned with what you intended?

Related detail: should the same context continue to carry the extra QML-option encoder state (`JaspExtraOptions_...`) and restore that alongside the dataset column encoder?

Common/log.cpp
JorisGoosen: You could use a `nullstream` like the one in MainWindow::initLog. And then Log::init(&nullstream). Much simpler than the code in this PR. And ofc in the syntaxbridge.

SyntaxInterface/syntaxbridge.cpp
JorisGoosen: Yes indeed, R should not carry mappings or bridge specific logic.

A scoped decode would require a bunch of extra interfacing functions right? 
Why not simply make sure to set the right context from R? or is that where the scope would be?

FBartos: No extra public bridge functions should be needed. My intended scoped version is a single native decode call: R passes the captured context JSON and payload JSON, SyntaxInterface installs that context into `ColumnEncoder` for that call, calls `ColumnEncoder::decodeJson()`, and restores the previous context before returning.

So the scope is inside the native decode call, not an R-side sequence of set/decode/reset calls. That keeps the R API from leaving global native decoder state changed after replay, but still uses the standard `ColumnEncoder` path. I will implement it that way unless you prefer the persistent set-current-context API.

JorisGoosen: > SyntaxInterface installs that context into ColumnEncoder for that call, calls ColumnEncoder::decodeJson(), and restores the previous context before returning.

How?

FBartos: Mechanically:

1. Parse the context JSON into the same `ColumnEncoder::colTypeMap` that came from `DataSet::getColumnTypesMap()` plus the extra-options encoder map.
2. Save the currently active maps:
   - `ColumnEncoder::columnEncoder()->currentNames()`
   - `gl_extraEncodings->currentNames()`
3. Install the captured maps:
   - `ColumnEncoder::setCurrentColumnNames(contextColumns)`
   - `gl_extraEncodings->setCurrentNames(contextExtra)`
4. Parse the payload JSON and call the standard native decoder:
   - `ColumnEncoder::decodeJson(payload, replaceNames)`
5. Restore the previously saved maps in an RAII guard/destructor before returning.

So the only custom code is context JSON <-> `colTypeMap` plus the scoped restore guard. The actual replacement would be `ColumnEncoder::decodeJson()`, not a separate bridge-specific decoding loop.

SyntaxInterface/syntaxbridge.cpp
JorisGoosen: If it is the standard API for ColumnEncoder it is already accessible from R via jaspBase right? So that should make "The R API smaller" than the latter? Or am I missing something?

FBartos: You are not missing much; my phrasing about "R API smaller" was imprecise.

`jaspBase::decodeColNames()` / `decodeName()` are module-facing convenience helpers. They eventually resolve the decoder functions available in the current JASP runtime, but they are not a standalone way for jaspSyntax to install a saved Desktop `ColumnEncoder` context after the live dataset has changed. The dependency direction should also stay jaspBase -> jaspSyntax/native bridge for replay, not jaspSyntax -> jaspBase.

So I think the native bridge still needs one minimal context-aware decode entry point. But I agree it should be generic and standard-ColumnEncoder-shaped: context JSON + payload JSON in, install context, call `ColumnEncoder::decodeJson()`, restore, return decoded JSON. Then the exported R helper can remain a thin convenience wrapper for character vectors if that is all current callers need.

SyntaxInterface/syntaxbridge.cpp
JorisGoosen: Ok sounds good.

SyntaxInterface/syntaxbridge.cpp
JorisGoosen: Ok well im curious to see it.

Common/log.cpp
FBartos: Done in c21e37f98. I reverted the `Common/log.cpp` change entirely, so the global logging defaults outside SyntaxInterface are unchanged.

SyntaxInterface now owns a local null stream and calls `Log::init(&nullstream)` from `configureBridgeLogging()` before bridge status/error paths can log. `syntaxBridgeSetVerbose()` only toggles between `cout` and `null`; no Desktop preference/file logging is needed for this standalone bridge path. I also smoke-tested the jaspSyntax status-only error path with `verbose = FALSE`; it returns the structured error with zero captured native output.

SyntaxInterface/syntaxbridge.cpp
FBartos: Done in c21e37f98. The bridge-specific manual JSON decoding loop is gone.

`decodeColumnJson()` now parses the payload, installs the snapshot context with `ScopedColumnEncoderContext`, calls the standard `ColumnEncoder::decodeJson(payload, replaceNames)`, and returns the decoded payload. The payload is no longer restricted to a string array, so the normal ColumnEncoder traversal remains the single implementation for decoded JSON objects.

SyntaxInterface/syntaxbridge.cpp
FBartos: Implemented in c21e37f98 with the scope inside the single native bridge call.

R/jaspSyntax still passes only the encoded payload plus the opaque encoder context. SyntaxInterface parses that context, `ScopedColumnEncoderContext` saves the current `ColumnEncoder` state, installs the snapshot names/types on `ColumnEncoder` and the extra encoder, calls the standard `ColumnEncoder::decodeJson()`, then restores the previous native state before returning. So R does not carry mappings and there are no separate set/unset calls exposed across the ABI.

SyntaxInterface/syntaxbridge.cpp
FBartos: The standard decoder implementation is still the one we use, but replay needs to decode against the context captured from the source file/analysis, not whatever dataset context happens to be live when R materializes the result.

That is why I kept this as a single SyntaxInterface bridge operation instead of routing state management through jaspBase: the only cross-ABI API is `decode(payload, context)`, and native code handles install/decode/restore internally. jaspSyntax remains a thin caller and jaspBase does not need to know about encoder mappings.

FBartos:
@JorisGoosen I pushed c21e37f98 with the changes discussed in the threads.

Summary:
- reverted the broad `Common/log.cpp` null-stream change; SyntaxInterface now owns the bridge-local null stream and initializes logging there
- removed the bridge-specific manual decoder loop; the replay path now installs the captured encoder context, calls `ColumnEncoder::decodeJson()`, and restores the previous native state before returning
- kept R/jaspSyntax as a thin `decode(payload, context)` caller, with no exposed set/unset mapping API and no jaspBase mapping responsibility

I rechecked the three review threads after the push; they are now outdated and have implementation replies. Ready for another look.

Common/columnencoder.h
```diff
@@ -63,10 +63,12 @@ private:						ColumnEncoder() { invalidateAll(); }
 	
 	static	colVec				columnNames();
 	static	colVec				columnNamesEncoded();
+	static	const char*			extraOptionsPrefix()										{ return "JaspExtraOptions_"; }
```
JorisGoosen: This is now a double definition. Very ugly. Should be moved to DataBridge instead. That way it covers both engine and syntaxbridge

FBartos: Done in 3c0575ce6. I moved this out of `ColumnEncoder` and made `DataBridge` the single owner of the extra-options encoder.

Concretely:
- removed `ColumnEncoder::extraOptionsPrefix()`
- added a bridge-owned extra encoder in `DataBridge`
- removed `Engine::_extraEncodings`
- removed the separate SyntaxInterface `gl_extraEncodings`
- changed `rbridge_init()` to derive the extra encoder from the active `DataBridge`, and `rbridge_setDataBridge()` now refreshes that borrowed pointer whenever SyntaxInterface recreates its DataBridge

So Engine and SyntaxInterface now share the same ownership model through DataBridge, and SyntaxBridge only borrows the encoder for context capture/decode.

FBartos:
@JorisGoosen I pushed 3c0575ce6 to address the remaining DataBridge ownership comment.

The extra-options encoder is now owned by `DataBridge`; Engine and SyntaxInterface no longer allocate their own copies, `ColumnEncoder::extraOptionsPrefix()` is gone, and `rbridge_init()`/`rbridge_setDataBridge()` borrow the encoder from the active DataBridge.

Validation after the push:
- rebuilt `SyntaxInterface`
- verified SyntaxInterface exported declarations against `jaspSyntax`
- ran focused jaspSyntax tests: `test-desktop-jasp-contract.R`, `test-dataset-helpers.R`
- ran focused jaspTools lifecycle tests: `test-jaspSyntax-lifecycle.R`
- ran focused jaspBase tests: `test-result-object-decoding.R`, `test-runWrappedAnalysis.R`
- ran `git diff --check`

The previous review threads are now outdated with implementation replies. Ready for another look.

SyntaxInterface/syntaxbridge.cpp
```diff
@@ -85,6 +88,31 @@ static std::string							gl_param_resultFont				=
...
+namespace
+{
+	class SyntaxBridgeNullBuffer : public std::streambuf
+	{
+	protected:
+		int overflow(int c) override { return traits_type::not_eof(c); }
+	};
+
+	SyntaxBridgeNullBuffer	gl_nullLogBuffer;
+	std::ostream			gl_nullLogStream(&gl_nullLogBuffer);
```
JorisGoosen: Why not just use the boost null stream we had before?

Or, if this is somehow better, why dont we replace the boost nullstream with something like this?

FBartos: Done in 647fde916. I switched SyntaxInterface to the same `boost::iostreams::null_sink` pattern used by Engine/MainWindow and removed the local std-only null stream implementation.

One build detail: SyntaxInterface now enables `/EHsc` on MSVC. Without that, Boost leaves `boost::throw_exception` unresolved for the null stream path in this target. After that change, `SyntaxInterface` rebuilds cleanly.

Validation:
- rebuilt `SyntaxInterface`
- verified SyntaxInterface exports against `jaspSyntax`
- ran focused jaspSyntax bridge contract test
- smoke-tested `verbose = FALSE` status-error path: zero captured native output

SyntaxInterface/syntaxbridge.cpp
```diff
@@ -681,6 +774,8 @@ void sendMessage(const char * msg)
 
 bool init(bool dbInMemory)
 {
+	configureBridgeLogging(gl_verbose);
```
JorisGoosen, line 765: I see `configureBridgeLogging` here. Which is the logical place! Good. But why is it at all those other places?

FBartos: Agreed. Those extra calls were defensive leftovers from the earlier logging-leak fix for pre-init status/error paths, but they make the ownership less clear.

I pushed 1766f05e3 to simplify this:
- `configureBridgeLogging()` is now only called from `init()` as the central setup point
- `syntaxBridgeSetVerbose()` only reconfigures if logging was already initialized, so changing verbosity after init still works
- pre-init/status-only failures now just return structured status errors instead of logging through `Log::log()` before initialization

Validation:
- rebuilt `SyntaxInterface`
- verified SyntaxInterface exports against jaspSyntax
- ran the focused jaspSyntax desktop bridge contract test
- smoke-tested the pre-init `verbose = FALSE` error path; it still captures zero native output

JorisGoosen:
It compiles and the code looks good now!

Im just missing some unittest(s) in Tests/

JorisGoosen (approved):

PR #6250, opened 2026-06-01
Fix MacOS Link issue

RensDofferhoff:
When you update macOS new linker is not content with the double definitions

### DIFF
--- a/CommonData/CMakeLists.txt
+++ b/CommonData/CMakeLists.txt
@@ -59,7 +59,9 @@ target_link_libraries(
 	$<$<NOT:$<PLATFORM_ID:Windows>>:R-Interface>
 
 )
-
+if(MSVC)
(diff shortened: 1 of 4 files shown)


CommonData/CMakeLists.txt
```diff
@@ -53,17 +53,13 @@ target_link_libraries(
     CommonData
     PUBLIC
     LibArchive::LibArchive
-#    Common
     SQLite::SQLite3
+	PRIVATE
 	$<$<PLATFORM_ID:Windows>:${_LIB_R_INTERFACE_STATIC_DLL_A}>
 	$<$<NOT:$<PLATFORM_ID:Windows>>:R-Interface>
 
 )
```
boutinb:
```suggestion
if(MSVC)
  target_compile_options(CommonData PRIVATE /EHsc)
endif()
```

boutinb (approved):

PR #6252, opened 2026-06-02
Try to get back the "good things" of 0.16.3 formatting

JorisGoosen:
This adds a unittest for the formatting. And fixes formatting

- Add Desktop/html/js/utils-test.js with 168 test assertions covering all formatters: formatFixed, formatPrecision, formatPrecisionWRFD, formatMoney,  formatNumber, handleNoZeroLead, toExponential, formatCellforLaTeX, and all  formatColumn paths (string, significance, decimals, percentage, monetary,  p-value, mixed, edge cases)
- Add utils-test.js to index-jasp.html so tests load in WebEngine
- Add test_utils_comprehensive() to tst_webengine.qml, calls runAllUtilTests()  via runJavaScript and asserts zero failures
- Fix minLSD computation in utils.js line 279: when dp is NaN (sf-only format  without dp token), Math.max(-NaN, minLSD) evaluated to NaN, causing  formatFixed to produce empty/invalid output. Now guards with  isFinite(dp) ? -dp : -Infinity
- sf:4;dp:3 now uses formatPrecision(sf) for non-fixDecimals normal-range  values, restoring sig-fig display (eg 0.001230 instead of 0.001)
- Force exponential notation when abs(value) >= 10^sf for !fixDecimals,  matching v0.16.3's toPrecision() behavior (eg 15606 -> 1.561×10^4)
- Remove dpP threshold override; < p now shows the actual configured  threshold instead of substituting the precision floor (eg < .010 not < ~.001)

### DIFF
--- a/Desktop/html/js/utils.js
+++ b/Desktop/html/js/utils.js
@@ -342,7 +342,7 @@ function formatColumn(column, type, format, alignNumbers, combine, modelFootnote
 					}
 					else if (content == 0)
 					{
-						formatted["content"] = fixDecimals ? formatFixed(content, dp, noZeroLead, thousands) : (isFinite(dp) ? formatFixed(content, dp, noZeroLead, thousands) : formatPrecision(content, sf, false, thousands))
+						formatted["content"] = (fixDecimals || isFinite(dp)) ? formatFixed(content, dp, noZeroLead, thousands) : formatPrecision(content, sf, false, thousands)
 					}
 					else if (Math.abs(content) >= upperLimit || Math.abs(content) < Math.pow(10, -dp) || (!fixDecimals && Math.abs(content) >= Math.pow(10, sf)))
 					{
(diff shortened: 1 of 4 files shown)


JorisGoosen:
Now we get 
image

Desktop/html/js/utils.js
```diff
@@ -341,18 +342,19 @@ function formatColumn(column, type, format, alignNumbers, combine, modelFootnote
 					}
 					else if (content == 0)
 					{
-						formatted["content"] = isFinite(dp) ? formatFixed(content, dp, noZeroLead, thousands) : formatPrecisionWithRespectForFixedDecimals(content, sf, 0, noZeroLead, thousands)
+						formatted["content"] = fixDecimals ? formatFixed(content, dp, noZeroLead, thousands) : (isFinite(dp) ? formatFixed(content, dp, noZeroLead, thousands) : formatPrecision(content, sf, false, thousands))
```
boutinb:
```suggestion
						formatted["content"] = (fixDecimals || isFinite(dp)) ? formatFixed(content, dp, noZeroLead, thousands) : formatPrecision(content, sf, false, thousands)
```

PR #6253, opened 2026-06-03
Cast variable should keep their type when used as source

boutinb:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/3227
The code to recast the variable (if the VariablesList allows it), was run each time a variable was set to a new VariablesList, but this should be done only when the user set manually a variable to another Variables List, not when the variables come from a source.

This issue comes after a fix for https://github.com/jasp-stats/INTERNAL-jasp/issues/3182


JorisGoosen (approved):
Looks good and seems to fix the issue

PR #6258, opened 2026-06-09
Manipulating value of labels may cause a crash in debug mode.

boutinb:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/3236
A unit test is added in testdebugdata.

First fix was to add rememberCurrentOrigValDisplay when _labelMapUpdates is called.

But a second fix is necessary:
```
 The real bug: _labelByValDis keys are built via origValDisplay() which calls processLabel(label, value). For a numeric label whose display equals its value (e.g., _label="6",
  origVal=6), processLabel("6","6") returns "" — so the map key is ("6","").

  But when labelDisplayChanged is later called with previousDisplay="6" (the raw _label), it was constructing oldValDis = ("6","6") — which doesn't exist in the map, failing the
  assert at line 1806.

  The same mismatch existed in labelValDisplayChanged and labelsRemoveByIntsId. Three fixes applied:

  1. labelDisplayChanged:1799 — wrap previousDisplay in Label::processLabel(previousDisplay, origValStr)
  2. labelValDisplayChanged:1822 — wrap previousDisplay in Label::processLabel(previousDisplay, oldOrigValS)
  3. labelsRemoveByIntsId:1159 — replace std::make_pair(label->originalValueAsString(), label->label()) with label->origValDisplay(), which already applies processLabel internally

```

### DIFF
--- a/CommonData/column.cpp
+++ b/CommonData/column.cpp
@@ -1624,6 +1624,8 @@ void Column::labelValueChanged(Label *label, const Json::Value & previousOrigina
 {
 	assert(_hasLabels);
 	
+	label->rememberCurrentOrigValDisplay();
+	
 	auto prevOrigV	= Label::originalValueAsString(this, previousOriginal);
 	auto oldValDis	= std::make_pair(prevOrigV, label->label());
 	bool merged		= _labelByValDis.count(label->origValDisplay()) != 0;
--- a/CommonData/column.cpp
+++ b/CommonData/column.cpp
@@ -1156,7 +1156,7 @@ void Column::labelsRemoveByIntsId(std::set<int> valuesToRemove, bool updateOrder
 				{
 					_labelByIntsIdMap.erase(label->intsId());
 					
-					auto valDis = std::make_pair(label->originalValueAsString(), label->label());
+					auto valDis = label->origValDisplay();
 					if(_labelByValDis.count(valDis) && _labelByValDis.at(valDis) == label)
 						_labelByValDis.erase(valDis);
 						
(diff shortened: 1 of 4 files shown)


CommonData/column.cpp
shun2wang:
```suggestion
	_labelMapUpdates(label, previousDisplay, oldOrigValS);
```

I can still reproduce it in the steps of https://github.com/jasp-stats/INTERNAL-jasp/issues/3236 says, unless I mad this change.

JorisGoosen: That looks like something is still going wrong with the flow there... Ill also do some research on this today

JorisGoosen: The suggestion works, thanks shun

JorisGoosen:
Looks like this at least fixes the original crash. Im looking into this a bit more to see if I can add some more tests for this

Desktop/utilities/settings.cpp
```diff
@@ -144,6 +144,7 @@ const Settings::Setting Settings::Values[] = {
 	{"showInteractiveDefault",		true	},
 	{"autoSaveOn",					true	},
 	{"autoSaveInterval",			5*60	},
+	{"mergeDuplicateLabelsNoAsk",	true	},
```
JorisGoosen:
```suggestion
	{"mergeDuplicateLabelsNoAsk",	false	},
```

JorisGoosen:
I need to test these changes but my worksystem is now in a state of limbo because im recom-piling webengine a bunch, but its crashing now. The dialog really neeeds to be tested and checked still so dont merge

PR #6265, opened 2026-06-19
Add Variables and Label editor help

boutinb:
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/3243


shun2wang (commented):
could we move these markdown to qml info to make it translatable?

boutinb:
> could we move these markdown to qml info to make it translatable?

Quite a good idea.
We could make a QML item for the help icon with an info property where the markdown is set. Then it will be automatically available in the jaspDesktop component of our WebLate project.
I'll do that, it's quite easy.

boutinb:
The change became bigger than I thought. I have made another PR for this: https://github.com/jasp-stats/jasp-desktop/pull/6272
So I close this one.

PR #6268, opened 2026-06-23
Speed up old file issue checks

JorisGoosen:
Instead of doing a lot of dbcalls to check a single thing (possibly repeatedly) and within a nested loop split it off and make sure to collect all labels at once

fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/3259


JorisGoosen:
So it will still take a little while to load EJs file, but this is a 1 time thing, just for the conversion to the current format.

boutinb (approved):

PR #6271, opened 2026-06-23
Add OverrideCommon functionality for customizing Common modules

JorisGoosen:
# OverrideCommon Feature

## Overview

The `OverrideCommon` feature allows you to customize which modules appear as "Common" (visible on the ribbon) and which appear as "Extra" (in the "Add analyses" menu).

## How It Works

When `OverrideCommon` is specified, the system:
1. Combines the Common and Extra module lists from `modules-settings.json`
2. Moves modules listed in `OverrideCommon` to the Common group (in the specified order)
3. Moves all other modules (including originally Common modules NOT in OverrideCommon) to the Extra group

## Configuration Source

`OverrideCommon` can be specified in the TOML configuration file:

### TOML Configuration File (config.toml)
```toml
OverrideCommon = [
    "jaspDescriptives",
    "jaspTTests",
    "jaspRegression"
]
```

### Example 1: Subset of Original Common Modules
```toml
OverrideCommon = [
    "jaspDescriptives",
    "jaspTTests",
    "jaspRegression"
]
```
**Result:**
- **Common**: jaspDescriptives, jaspTTests, jaspRegression (in that order)
- **Extra**: All other modules (including jaspAnova, jaspFrequencies, jaspFactor, etc.)

### Example 2: Reordering Common Modules
```toml
OverrideCommon = [
    "jaspFrequencies",
    "jaspFactor",
    "jaspDescriptives"
]
```
**Result:**
- **Common**: jaspFrequencies, jaspFactor, jaspDescriptives (custom order)
- **Extra**: All other modules

### Example 3: Combining with EnabledModules
```toml
EnabledModules = [
    "jaspDescriptives",
    "jaspTTests",
    "jaspRegression"
]
OverrideCommon = [
    "jaspDescriptives",
    "jaspRegression"
]
```
**Result:**
- **Common**: jaspDescriptives, jaspRegression (on ribbon, enabled)
- **Extra**: jaspTTests (enabled but not on ribbon), all other modules (disabled)

## Usage

To use this feature:

1. Edit your `config.toml` file (usually in JASP's configuration directory)
2. Add or uncomment the `OverrideCommon` array
3. List the modules you want to be Common, in the order they should appear
4. Restart JASP to see the changes

## Default Behavior

If `OverrideCommon` is not specified (empty array or missing), JASP uses the default behavior:
- Common modules are read from `modules-settings.json` "common" field
- Extra modules are read from `modules-settings.json` "extra" field

Implements https://support.jasp-services.com/JASP-Services/Company/issues/40


Desktop/modules/ribbonbutton.h
```diff
@@ -66,10 +66,10 @@ class RibbonButton : public QObject
 	std::string					title()														const			{ return _titleF ? _titleF() : _title;						}
 	QString						titleQ()													const			{ return QString::fromStdString(title());					}
 	QString						iconSource()												const			{ return _iconSource;										}
-	bool						enabled()													const			{ return _enabled;											}
+	bool						enabled()													const			{ return _isCommonModule || _enabled;						}
```
boutinb, line 69: Why? It means that the user cannot disable a common module anymore.
I don't see the use of this. And if it is a wanted feature, then the possible to disable it, should be also disabled: now if you click in the modules list on a common module, then it stays enabled

Desktop/modules/ribbonmodelfiltered.cpp
```diff
@@ -16,13 +16,17 @@ void RibbonModelFiltered::setRibbonModel(RibbonModel * ribbonModel)
 	setSourceModel(_ribbonModel);
 
 	connect(ribbonModel, &RibbonModel::highlightedModuleIndexChanged, this, &RibbonModelFiltered::highlightedModuleIndexChanged);
+	
+	connect(ribbonModel, &QAbstractListModel::modelReset, this, &QSortFilterProxyModel::invalidate);
```
boutinb, line 20: Why do we need to invalidate the sortFilterProxyModel when the ribbonModel is reset? I don't see the logic here

Desktop/mainwindow.cpp
```diff
@@ -2405,6 +2405,13 @@ void MainWindow::loadModulesFromUserConfiguration(configState state)
 		auto button = _ribbonModel->ribbonButtonModel(moduleName.toStdString());
 		_ribbonModel->setModuleEnabled(_ribbonModel->ribbonButtonModelIndex(button), true);
 	}
+
+	// Apply OverrideCommon to refresh Common/Extra
+	const QStringList* overrideCommon = _jaspConfiguration->getOverrideCommon();
+	if(overrideCommon && !overrideCommon->isEmpty())
+	{
+		DynamicModules::dynMods()->refreshCommonModules(*overrideCommon);
+	}
```
boutinb, line 2816: If I understand, this code will be called if the configuration is remote, and might be not ready before the ribbon is displayed.
It seems to me a bit dangerous: the user can block (maybe unintentonally) the access to the remote configuration, and then has access to JASP in a unwanted way from the company.
I would wait for the return of remote configuration with a timeout, and maybe in case of an error, use the local configuration.

boutinb (commented):
I have 3 comments

Desktop/modules/ribbonbutton.h
JorisGoosen: you never could do that...

Desktop/modules/ribbonmodelfiltered.cpp
JorisGoosen: I was having issues seeing changes appear in the ribbon. So I figured ill just FORCE invalidation on the internal filtermodel. Maybe it is is not needed.

Desktop/mainwindow.cpp
JorisGoosen: Ok, but that is more of a general thing about how the configs work right? Not really about this PR

PR #6272, opened 2026-06-26
Make help translatable

boutinb:
This PR replaces https://github.com/jasp-stats/jasp-desktop/pull/6265
Fixes https://github.com/jasp-stats/INTERNAL-jasp/issues/3243

What started with a small feature, became quite a big change.
I made all the help files in JASP Desktop translatable.
For this I used a mechanism that reads all the MD files, set them as translatable strings, and add for each of them a property in the AllHelp class, so that it can be used directly in QML
I have asked my AI to translate the jaspDesktop po files in Dutch, French, German and Chinese. Maybe @tomtomme and @shun2wang can check these translations. If they are good enough, we could use this method to translate to all other languages.
One drawback of translating the whole MD file, is that for each small change in the MD file. the whole translation is lost, but if we can use AI to translate, then there is no problem anymore.


tomtomme:
Interesting. 

Some questions: 
a) So the weblate translation is not possible / needed anymore for help files?
b) This certainly is welcome for languages that are common and thus well translatable via AI. But I guess rare languages like Tibetan will be a problem?
c) Also what about all the other strings that are currently translated via weblate - could and should those also be ported over to this complete AI-translation?
d) Will there be a possibility for manual intervention if a translation-AI causes a huge mistake?
e) Is this one of the files I should review: 
https://github.com/boutinb/jasp-desktop/blob/f5a6ac3fcb0cac64743423f2c8a40e64b37bf6f2/Desktop/po/jaspDesktop-de.po
f) Which AI did you use?

I am also adding @JoKeyser to the conversation, maybe he has also time to review and ideas for this one

boutinb:
> Interesting.
> 
> Some questions: a) So the weblate translation is not possible / needed anymore for help files? b) This certainly is welcome for languages that are common and thus well translatable via AI. But I guess rare languages like Tibetan will be a problem? c) Also what about all the other strings that are currently translated via weblate - could and should those also be ported over to this complete AI-translation? d) Will there be a possibility for manual intervention if a translation-AI causes a huge mistake? e) Is this one of the files I should review: https://github.com/boutinb/jasp-desktop/blob/f5a6ac3fcb0cac64743423f2c8a40e64b37bf6f2/Desktop/po/jaspDesktop-de.po f) Which AI did you use?
> 
> I am also adding @JoKeyser to the conversation, maybe he has also time to review and ideas for this one

All the strings (with the help files) will be in WebLate, and all translations made in WebLate will be used in JASP.
But as the AI's are now really performant, and can take much better the context into account to translate, we should use them for translation now. 
The french and dutch translations are as far as I can see really good. But I'm not s statistician, so it would be good if you and Shun check the translations in jaspDesktop-de.po and jaspDesktop-zh_Hans.po.

I have used Claude / Opus for this.
I think we could use an AI to fill the non-translated strings, and let the translators check and correct them in WebLate.

boutinb:
> One drawback of translating the whole MD file, is that for each small change in the MD file. the whole translation is lost, but if we can use AI to translate, then there is no problem anymore.

Here an AI can be prompted to be clever enough, to see what is the change, and translate only the change. In this way, if a human translator made a correction, this correction will still hold.

tomtomme:
This is awesome. I have reviewed some stuff in jaspDesktop-de.po - and so far all is good!

tomtomme (approved):
Looks good to me. On reading some of the translations I recognized some strange stuff, but this was caused by the original english text. I made some changes here to remedy this: https://github.com/boutinb/jasp-desktop/pull/8

shun2wang:
The AI ​​translation GUI interface guidance is generally accurate, but I'm unsure how accurate the Markdown in the analysis module will be. However, I agree with your statement below:

>I think we could use an AI to fill the non-translated strings, and let the translators check and correct them in WebLate.

shun2wang (approved):

JoKeyser:
Hi @boutinb, hi @tomtomme , this is really helpful and promising.
 
I had a brief look, and an LLM with more context clearly beats the Automatic Suggestions from Weblate.
However, it's also still far from perfect in my opinion, and it will be crucial (at least to me) that the human translators can effortlessly override the automatic translations.
So please don't let it auto-change stuff that already has been human-translated; perhaps pointing out inconsistencies would be nice, though.

Just as an example that has lots of sharp edges, see the translation of "scale" in line 8956 of `jaspDesktop-de.po`
```txt
EN: #~ "* **Scale (Ruler icon):** Continuous numerical data (e.g., age, weight, test scores).\n"
DE: 
"* **Skala (Lineal-Symbol):** Kontinuierliche numerische Daten (z. B. Alter, Gewicht, Testergebnisse).\n"
```

Specifically, the LLM translation of EN "Scale" into DE "Skala" is predictable, but doesn't make sense, as discussed in https://github.com/jasp-stats/jasp-issues/issues/231#issuecomment-2189109372.

There are several other cases like this, so again, overriding will be just as important as this is helping.

tomtomme:
@JoKeyser 
Should we rename scale to continous or metric? I never got why its called scale in EN. Is it a synonym for metric really?

JoKeyser:
@tomtomme:

> Should we rename scale to continous or metric? I never got why its called scale in EN. Is it a synonym for metric really?

You mean in the EN original? From last time I checked, I think "continuous" would be the better term, unless the concept gets split further into "interval" vs "ratio" scale...
Overall, renaming from "scale" to "continuous" seems like a fairly simple improvement.

tomtomme:
Okay, I will propose this in a separate issue.

tomtomme:
@JoKeyser, see https://github.com/jasp-stats/jasp-issues/issues/4391 for the translation issue

shun2wang:
@RensDofferhoff @boutinb Why reverted https://github.com/jasp-stats/jasp-desktop/pull/6272

boutinb:
> @RensDofferhoff @boutinb Why reverted #6272

I did not pay attention enough about what the AI bots did, but there were a lot of duplicated strings, malformed markdown strings, some translations were overridden. I tried to tell the bots to repair these problems. but they made other mistakes. So I preferred to cancel this PR, and to start again, by controlling better what the bots made. Unfortunately it could not reach the 0.98.0 version. But probably we will have a new version for September, and that will be in this version.

boutinb:
See PR https://github.com/jasp-stats/jasp-desktop/pull/6278

PR #6275, opened 2026-06-29
JASP AI Beta

RensDofferhoff:
Here's a polished GitHub PR summary:

---

### Overview

This PR introduces the **JASP AI Beta**, an AI-powered assistant that lets users interact with JASP through natural language. Users can ask questions, create and run statistical analyses, inspect results, and generate reports — all from a chat interface inside JASP.

The assistant communicates with any OpenAI-compatible API (OpenAI, Anthropic, Mistral, OpenRouter, local models, etc.) and uses **tool calling** (function calling / native tool use) to interact with the JASP  directly.

---

#### AI Chat Window
- A dedicated `ChatWindow.qml` window with a WebEngineView hosting the `deep-chat` web component
- Control bar with conversation reset, Markdown export, and live token counting
- Intro message on first open that explains JASP's capabilities (table results, plot results, options, etc.)
- Avatar support for both AI and user, provided by the persona system

#### AI Bridge (`AiBridge`)
- Manages HTTP SSE streaming to AI providers
- Orchestrates tool-call loops: the AI can call JASP RPC methods and feed results back into the conversation
- Conversation history management with token estimation and context window tracking
- Handles concurrency (busy guard, intro-stream kill on new message, deferred clear-chat)
- On AI provider/model/persona config changes, chat is cleared to avoid stale responses

#### Agent State Tracker
- Tracks workspace changes (analysis added/removed, options edited, data changed) via signal hooks into `Analyses` and `DataSetPackage`
- On every RPC response, attaches a `_stateUpdate` delta so the AI knows what changed since it last observed state
- On user divergence during mutation methods, returns error `-32001` with a full workspace snapshot

#### JASP RPC Framework
- JSON-RPC 2.0 dispatcher (`JaspRpcDispatcher`) with schema-based param validation
- OpenRPC 1.2.6 spec loaded from `Resources/JASP_RPC.json` — defines all available RPC methods
- HTTP server (`JaspRpcServer`) on `127.0.0.1:48164/rpc` for external clients (MCP, custom integrations)
- Methods: `analysis_create`, `analysis_run`, `get_analyses_state`, `analyses_list`, `analysis_results`, `data_info`, `modules_list`, `analysis_createAnnotation`, `analysis_context`, `write_report`
- `rpc_discover` for self-describing method discovery
- State divergence handling: mutation methods (`x-failOnStateDiverged`) fail if the user has edited the workspace since the AI last observed it

#### AI Configuration
- **Preferences → AI tab** (`PrefsAI.qml`): full provider/model configuration UI
- Shipped provider definitions (`Resources/defaultProviders.json`): 7 providers with ~30 models
- Custom provider support for ad-hoc API endpoints
- Per-model overrides: model name, extra params, system prompt postfix, chat limit, `useCompleteSchema` flag
- Encrypted API key storage via `SecretStore` (libsodium crypto_secretbox with machine-derived master key)
- **Test Connection** button to verify endpoint/key/model before use

#### Persona System
- Shipped personas (`Resources/defaultPersonas.json`): Alfred, Evelyn, Socrates
- Custom personas with user-configurable avatars, system prompts, and capability toggles
- `PersonaSchemeHandler`: serves persona images via custom `jaspPersona://` URL scheme
- Persona avatars in the chat window update theme-aware (dark/light)

#### Analysis Annotation & Reports
- **Annotate Analysis** button in the analysis form header — sends the current analysis results to the AI and asks for a plain-English summary
- Reports: AI can compose multi-analysis documents with Markdown interleaved with embedded analysis results
- Editable report mode where the user can toggle the report QML form between read-only and editable

#### Form Control Introspection (`optionMeta`)
- `AnalysisForm::optionMeta()` generates structured metadata about every control on a form (kinds: checkbox/combo/variables/number/string, choices, min/max, allowed column types, single/multi)
- Used to build the LLM's system prompt with exact control schemas so the AI emits valid option JSON

---

### Architecture

```
ChatWindow.qml → WebEngineView → chat.html → deep-chat
                                                    ↓
chat-bridge.js ←→ AiBridge (C++) ←→ OpenAI-compatible API (SSE)
                    ↓
            JaspRpcDispatcher ←→ JASP  (Analyses, Data, Modules)
                    ↓
            AgentStateTracker (workspace change detection)
```

7 Mermaid flow diagrams in `Docs/development/aiBridge/` document the full architecture.

---

### Notable Supporting Changes

- **SecretStore**: encrypted credential storage using libsodium (XSalsa20-Poly1305) with BLAKE2b key derivation from machine + user identity. Replaces plaintext storage
- **`ComponentsListBase`** (Bruno Boutin): fixes selection loss on model reset; guards add/remove/key-change when populated from a source model
- **`SourceItem`** (Bruno Boutin): refactors native model reading from single-role to multi-role (display/value/info/type), enabling type-aware term population
- **`JASPDoubleValidator`**: optional decimal-place relaxation (`relaxDecimals`) for AI form input mode
- **Build system**: added `Qt6::HttpServer` dependency; version bumped to 0.98.0
- **Translation files**: removed Farsi and Macedonian translations (unmaintained); updated BO and ID translations

---

### Configuration

Users need to:
1. Enable AI in **Preferences → AI**
2. Select a provider (OpenAI, Anthropic, Mistral, OpenRouter, etc.)
3. Enter an API key (encrypted at rest)
4. Optionally customize model, endpoint, extra params, chat limits, and persona

---

### Testing

- Verified with OpenAI (GPT-5-mini), Anthropic (Claude Sonnet and haiku via OpenRouter), Mistral, and DeepSeek
- Tested with and without `useCompleteSchema` (function calling vs. native tool use)
- Manual testing: create analyses, set options, inspect results, write reports, annotation flow
- Server tested with curl for JSON-RPC protocol compliance

---

### For Reviewers

The highest-risk changes that need careful review:

1. **`QMLComponents/controls/sourceitem.cpp`** — refactors core source-model reading from single-role to multi-role. Every analysis form's variable list depends on this
2. **`QMLComponents/analysisform.cpp`** — new `_controlOptionMeta()` method (~300 lines) that introspects form controls. Complex but only called by the AI path, not normal analysis workflow
3. **`Desktop/ai/aiBridge.cpp`** — orchestrates SSE streaming and tool-call loops. Well-documented; the two-path `currentSignals` architecture is described in flow diagrams
4. **`Desktop/ai/agentstatetracker.cpp`** — workspace change detection. Verify signal connections and dirty-flag semantics


Resources/defaultPersonas.json
```diff
@@ -0,0 +1,23 @@
...
+		"personaPrompt": "Your name is Alfred and you are a generic statistical advisor. When advising on analyses, first use the information available in JASP, including the data set if available, variable names, measurement levels, current analyses, and output. Consider the user's research question, variable types, study design, sample size, missing data, assumptions, and the distinction between exploratory and confirmatory analysis. Recommend analyses that are appropriate and available in JASP. Explain the reasoning behind your recommendations, including relevant assumptions, limitations, effect sizes, confidence or credible intervals, diagnostics, and possible alternatives.\n\nIf the user's request is unclear or important information is missing, ask concise follow-up questions before giving firm advice or conducting an analysis. When possible, provide a provisional recommendation and clearly state the assumptions on which it depends.\n\nWhen selecting options for an analysis, prefer options that improve interpretation, diagnostics, assumption checking, effect-size estimation, uncertainty quantification, and reporting. Avoid enabling unnecessary options merely for completeness.",
```
shun2wang, line 5: Could we translate these persona prompt? or we can request the language output by the AI ​​based on the JASP interface language?

RensDofferhoff: We should not translate them as they are part of systemprompts for the AI.
Adding a note  of the interface language could be smart but the be honest the AI picks up your language perfectly after the first user msg.

Only the introduction would be english by default.
@EJWagenmakers what do you think ? should we put the intro message in the interface language?
Not everyone of them would be supported (well) depending on model. Perhaps leave it for the beta.

boutinb:
### 1. `buildAnalysesSnapshot` in `get_analyses_state` doesn't clear data dirty flags

**File**: `Desktop/analysis/analyses.cpp`, `get_analyses_state` handler

The handler calls `AgentStateTracker::notifyAnalysisObserved()` for each requested analysis ID, but never calls `notifyDataObserved()`. If column changes occurred, `_dataState.dirty` stays `true`. The next RPC response will attach a full `_stateUpdate`, which does include data — so the agent eventually learns about the changes, just one round-trip late.

**Fix**: Either add `AgentStateTracker::notifyDataObserved()` in the handler, or document that data changes surface via `_stateUpdate` on the next read call.

### 2. `ChatWindow.qml` — `onClosing` fires `clearChat()` which may outlive the window

**File**: `Desktop/components/JASP/Widgets/ChatWindow.qml`

```qml
onClosing: {
    aiBridge.clearChat();
}
```

`clearChat()` calls `stopStream()`, `clearConversation()`, `emit onClearChat()`, and `sendIntroMessage()` — the last of which makes a network request. If the reply arrives after `AiBridge` is destroyed, it crashes. Window close is not the right time to kick off an intro stream.

**Fix**: Replace with just `clearConversation()` (no network), or set a flag to suppress the intro after close.

### 3. `AgentStateTracker::markAnalysisRemoved` wipes all dirty flags for that analysis

**File**: `Desktop/ai/agentstatetracker.cpp`

```cpp
void AgentStateTracker::markAnalysisRemoved(size_t analysisId) {
    auto & state = _analysisStates[analysisId];
    state.dirtyFlags.clear();
    state.dirtyFlags.insert(AnalysisChange::Removed);
}
```

If an analysis was marked as `Added` by the user and is then `Removed` in the same observation window, the `Added` flag is lost. The agent only sees `Removed`. This is arguably correct end-state semantics (the analysis is gone), but if the agent was planning to mutate that analysis, it won't know it was the user who created it. Not a crash, but worth confirming this is the intended behavior.

### 4. `SecretStore::kMasterKeySeed` is in the header

**File**: `Desktop/utilities/secretstore.h`

The 32-byte seed is `static constexpr` in the header. Moving it to the `.cpp` file keeps it out of every translation unit that includes the header. Minor, but easy to fix.

RensDofferhoff:
@boutinb 4. is not a problem forgot its just the seed not and the key is derived from machine ID + seed

RensDofferhoff:
those 4 points have been addressed where necessary

boutinb (approved):

PR #6255, opened 2026-06-06
bugfix: fix parsering and preview complex csv files

shun2wang:
Fixed: https://github.com/jasp-stats/jasp-issues/issues/4396

Original csv(please copy and paste in a text file and name to .csv):
```csv
ID,Names,Gender,Age,Add.,Tel.,Text
1,One,M,32,ABBA,1234567891,"test""Quote"""
2,Two,M,21,,1234567890,"test：
newlines"
3,Three,F,18,,,"this,is,text"
```
✗Before preview behavior:
image

✓ Now preview:
image

✗Before Imported behavior:
image

✓ Now imported:
image

And added a unit test for it.

PR #6291, opened 2026-07-21
Fix csv preview

JorisGoosen:
So @shun2wang I was working on stuff in this branch. It uses the fixes of you but supposedly merged everything into a single class. 

But as I wrote this with the help of an LLM something slipped in and I need to fix that still. And it seems I missed your last commit from https://github.com/jasp-stats/jasp-desktop/pull/6255 what was that adding exactly?


shun2wang:
My last commit fixed importing this csv,

```csv
ID,Names,Gender,Age,Add.,Tel.,Text
1,One,M,32,ABBA,1234567891,"test""Quote"""
2,Two,M,21,,1234567890,"test：
newlines"
3,Three,F,18,,,"this,is,text"
4,Four,M,25,Address,1234567892,Comment,ExtraCol1,ExtraCol2,ExtraCol3
```

and you can [cherry pick](https://git-scm.com/docs/git-cherry-pick) to your branch.

JorisGoosen:
So, this all seems to now work, and be unified. 

However, trimming is removed, which now triggers testfailures in the csv/tsv unittest. Because I put in a " " in one of the columns, that always used to be trimmed. But now it converts a scalar column into nominal because " " is not a double... 

Im considering that it might be good to do a trim at the end of the import. But then again, if someone took the trouble to put a space somewhere, I guess they really want the space? 
So maybe its just my debugfile that is the problem.

Desktop/data/importers/csv/csv.h
```diff
@@ -49,12 +50,15 @@ class CSV
 
 	Status status();
 	char delimiter() { return _delim; }
```
shun2wang, line 52:
```suggestion
```

shun2wang: Seems unnecessary?

Desktop/data/importers/csv/csvparser.h
```diff
@@ -0,0 +1,123 @@
...
+
+
+
+	/// Get number of rows parsed
+	/// @return Row count
+	size_t getRowCount() const;
+
+	/// Check if parser has any pending data
+	/// @return true if there's unextracted data
+	bool hasPendingData() const;
```
shun2wang, line 79: We defined them `getRowCount` and `hasPendingData`  but never be used. maybe keep them here so can be used in the future.

JorisGoosen:
@boutinb reviews?

boutinb (approved):

PR #6297, opened 2026-08-04
i18n: Arabic as complete and add Tibetan language

shun2wang:


shun2wang:
see also: https://github.com/jasp-stats/jasp-actions/pull/79

PR #6301, opened 2026-08-17
 Bump JASP to 1.0.0, R to 4.6.1, update dependencies 

RensDofferhoff:


RensDofferhoff:
builds are running

Modules/Rcpp_RInside.lock
```diff
@@ -1,10 +1,10 @@
 {
   "R": {
-    "Version": "4.3.3",
+    "Version": "4.6.1",
     "Repositories": [
       {
         "Name": "CRAN",
-        "URL": "https://packagemanager.posit.co/cran/2026-03-31"
+        "URL": "https://p3m.dev/cran/2026-08-15"
```
JorisGoosen, line 7: is cran really so bad?

RensDofferhoff: yes binaries are lost after update making it break for all new developers while current ones dont notice due to caching.

I added cran as second repo.

PR #6305, opened 2026-08-26
Ribbon config initial

RensDofferhoff:
## What

Makes the ribbon fully user-configurable in two ways:

1. **Every module — including the seven core/common ones — can be (de)selected** in the `+` modules menu.
2. **The order of modules on the ribbon is user-configurable** via ↑/↓ arrows in the modules menu.

## Design decisions

- **Deselecting never destroys work.** Hiding a module no longer calls `DynamicModules::unloadModule()`, so live analyses stay in the results. Unloading remains reserved for genuine uninstall/replace flows.
- **"Common" means default-on, nothing more.** The `isCommon` flag (from `modules-settings.json`, optionally seeded by enterprise TOML `OverrideCommon`) only provides the initial set of enabled modules; users may change everything afterwards.
- **The stored selection is authoritative** once it exists. A one-time migration (`modulesSelectionMigrated`) appends the core modules to selections stored before this change, so upgrading users don't lose their core modules.
- **`OverrideCommon` seeds, doesn't enforce.** It seeds the initial module order once (when no order is stored yet); afterwards the user's stored order wins.

## How it works

- Enabled-set persists in `MODULES_REMEMBERED` (existing setting, now `'`|`'`-joined and parsed with `SkipEmptyParts`); order persists in the new `MODULES_ORDER`.
- Order is applied at the end of `RibbonModel::loadModules()`. Special buttons (Data, New-data, separator, R-console) stay anchored; disabled modules keep their position, so re-enabling returns a module to its spot. Modules not mentioned in the stored order (newly installed, first run) append at the end.
- Reordering goes through `RibbonModelUncommon::moveModule(filtered, filtered)` → `RibbonModel::moveModule(source, source)` with proper `beginMoveRows`/`endMoveRows`, so both the menu and the ribbon update live.

## Fixed along the way

- Ribbon separators (the divider next to New/Edit Data) disappeared: separators inherited `remember = true` and were disabled by the new selection pass. They're now excluded from the selection.
- Special buttons (`Data`, `Data-Resize`, …) were being written into `MODULES_REMEMBERED` every time their enabled state changed; only selection participants (modules + R-console) are stored now.
- Empty-string artifact from splitting an unset `'`|`'`-joined setting.

## How to test

1. Fresh config → core modules on ribbon, all modules listed with checkboxes in `+` menu.
2. Deselect a core module with a live analysis of it → analysis stays, button disappears; restart → persists (with remember-pref on).
3. Move modules with ↑/↓ → ribbon order changes live; restart → order persists.
4. Disable a middle module, restart, re-enable → returns to its position.
5. Deselect everything → only special buttons remain, no crash.
6. ALT-navigation and keyboard Left/Right still traverse the ribbon in visual order.

## Deferred

Drag & drop reordering (menu and/or ribbon) — arrows cover the functionality accessibly; ribbon-side DnD specifically fights `reuseItems`, `interactive: false`, the wheel MouseArea and ALT/keyboard navigation, so it needs dedicated testing time.


RensDofferhoff:
@juliuspfadt  see https://static.jasp-stats.org/Nightlies/ for builds

its not quite what you wanted but its a start

juliuspfadt (changes requested):
I like the functionality. I think it is a very good step in the right direction. 
Naively, though, I would like to drag and reorder the modules. I think that is the intuitive solution. 
If not possible in the ribbon itself, then maybe it would be possible in the modules column on the right? I am not sure.

RensDofferhoff:
bruno got it working!

boutinb (approved):

juliuspfadt:
I didnt even have a chance to look at it again 😳

PR #6313, opened 2026-09-08
Test analyses load and fuzztest analyses

JorisGoosen:
Implements a test where all modules+analyses are loaded via rpc/mpc at least once to see if they load at least the default.
This pointed the bot to a bug in filtereddataentry, which it then fixed.

I still need to review all of this as well.

Should implement https://github.com/jasp-stats/INTERNAL-jasp/issues/3323


Resources/JASP_RPC.json
```diff
@@ -78,6 +78,114 @@
 				}
 			}
 		},
+		{
+			"name": "analysis_getOptions",
```
JorisGoosen: See https://github.com/jasp-stats/jasp-desktop/pull/6313/changes#diff-20fa801ee8984476e60d05df55b4127cafe1529ff9e94291b1cc609e1ab7c867L183

FBartos:
what am I supposed to review here?

PR #6318, opened 2026-09-10
The Windows started via a batch should not have the focus.

boutinb:


FBartos:
I don't think it worked. Also now I see this pop up message with every analysis:

image

FBartos:
Prepared [d6469897e](https://github.com/FBartos/jasp-desktop/commit/d6469897e8fb72aea2b89c28805a0a9aec108c9b), based directly on this PR's current head, to address the remaining batch problems found in Windows testing.

Opening the saved template still followed the interactive linked-data synchronization path. If the template's original data file had changed, JASP could show **Datafile changed** and attempt another import while the batch was supposed to use its explicitly supplied data file. The commit skips that automatic synchronization during batching and keeps unattended workers hidden on the normal Qt platform, which supports rendering the exported results.

The batch also treated a successful export as a successful analysis. A report could contain an analysis error, and `--exportType=No` could exit before the analyses finished. Workers now wait for completion, report import/analysis/export failures, and reject data containing only missing values. The existing batch loop continues to the next input and prints the reason beside `Failed`, then lists errors and warnings with final counts. Raw Qt/Chromium diagnostics remain separate because they also occur for successful inputs and should not inflate the JASP warning count.

The change retains the existing directory traversal and process structure, adds no CLI flags, and limits shared dialog handling to JASP warning messages. Help text and focused tests are included; existing parser tests were updated to use `exportType` because their old boolean fields no longer exist.

Validation on Windows, Qt 6.10.2, Release Pro build:

- Four CSV inputs produced three correct HTML exports and one expected failure for 99 rows of empty fields. No worker window became visible or took focus.
- A zero-byte file and an analysis error both failed without preventing a following valid input from completing, including with `--exportType=No`.
- Direct single-file execution returned failure for an analysis error while exporting its error report.
- `testBatchResult`, `testParsedArgs`, and `git diff --check` passed. macOS and Linux runtime behavior has not been tested locally.

GitHub denied my account permission to push to `boutinb:tdk`, so the commit is published on `FBartos:codex/pr-6318-batch-fixes`. It can be applied from this PR's `tdk` branch with:

```sh
git fetch https://github.com/FBartos/jasp-desktop.git codex/pr-6318-batch-fixes
git cherry-pick d6469897e8fb72aea2b89c28805a0a9aec108c9b
```

PR #6322, opened 2026-09-15
Add possibility to select multiple folders in Batch

boutinb:
Now you can select several files and folders, and then unselect (or remove) the data files you want to use for the batch.
Also small UI changes.


JorisGoosen (approved):
It looks much nicer now!
I dont have modules installed currently so its hard to test but I will test it with the next build.

PR #6326, opened 2026-09-17
Trim the unquoted padding of a csv field again 

boutinb:
Fixes the JASP unit tests.

JASP has always read the whitespace around a csv field as padding rather than as
data: a cell holding nothing but spaces is a missing value, and " 1 " is the
number 1. The old hand-rolled tokenizer did that with a trim(token) at each of
its exit points, and https://github.com/jasp-stats/jasp-desktop/pull/6291 dropped those when it replaced the tokenizer with the
RFC 4180 CSVParser state machine. Since then a cell of one space imports as the
literal " ", which becomes a level of its own and pushes a scale or ordinal
column over to nominal because " " is not a number. This is what makes
testDataImport fail on Data Import Test CSV.csv and .tsv: the stored json is
still right, the importer is not.

Whitespace inside quotes is content the user asked for, though, so rather than
trimming everything back the parser now remembers where the quoted part of a
field begins and ends and only eats the padding around it:

  ` a `    -> `a`          `" a "`  -> ` a `
  ` "a" `  -> `a`          `"  "`   -> `  `
  `   `    -> ``           `"a,b"`  -> `a,b`

An unterminated quote at the end of the file leaves the end unset, and the rest
of the field then counts as quoted, so nothing of it is eaten.

CsvPreviewModel shares CSVParser, so the preview and the import agree again. The
two stored jsons match unchanged, and testCsvParserTrimsOnlyUnquotedPadding pins
the table above on the parser itself. Also drops the using-declaration for
boost's trim that csv.cpp kept after https://github.com/jasp-stats/jasp-desktop/pull/6291 removed its last use.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>


JorisGoosen:
Wouldnt it be simpler to just only trim non-quoted strings?

JorisGoosen (approved):

PR #6327, opened 2026-09-17
bugfix: fix to JASP new version checker

shun2wang:
The old mechanism:
Latest 1.0.0.0 → 1 * 1,000,000 = 1,000,000
Old 0.96.0.0 → 0 + 96 * 100,000 = 9,600,000
Then Old verison will always larger than latest version! When you build locally, you will find that the "Download new version" button becomes enabled, which is incorrect.

After:
Class `Version` already has full exicographical comparison, lets just compare them.

Safely release `_networkReply`, This is more robust in the event of network failure see, https://github.com/jasp-stats/jasp-issues/issues/1986


shun2wang:
However, the remote version file "http://static.jasp-stats.org/JASP-Version.txt" must remains locked at  0.98.1.0 and should never changed it to `1.0.0`; otherwise, users on versions prior to 0.98.1 would never receive notifications about new versions. 

Once this fix is ​​applied, for version 1.0 and later, we can set up an another remote file  at"http://static.jasp-stats.org/JASP-Version_1.txt" and change the client-side index to request version check from this url, thereby enabling the new mechanism for versions following `1.0.0`

